PR #37, opened 2023-04-03
feat: make libxmtp npm consumable

michaelx11:
## Overview

Attempt to make libxmtp consumable as a repo dependency within an external package.json file. To do this, we rename the package to `xmtpv3_wasm` from `libxmtp` to avoid some ambiguity. This package contains the WASM bindings for xmtpv3, but there could be another npm package with node-bindgen bindings for example.

## Test Plan

Doing `npm install xmtp/libxmtp#michaelx_make_libxmtp_npm_consumable` on the xmtp-js#michaelx_voodoo_client branch at least lets things build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "./package.json": "./bindings/wasm/package.json"
   },
   "files": [
-    "dist"
+    "bindings/wasm/dist"
   ],
   "scripts": {
     "prepare": "cd bindings/wasm && npm install && npm run build",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "libxmtp",
+  "name": "xmtpv3",
   "version": "0.0.1",
   "author": "jazzz <jazz@xmtp.com",
   "license": "MIT",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -0,0 +1,52 @@
...
+    },
+    "./libxmtp.wasm": "./bindings/wasm/dist/libxmtp_bg.wasm",
+    "./package.json": "./package.json"
+  },
+  "files": [
+    "./bindings/wasm/dist"
+  ],
+  "scripts": {
+    "prepare": "cd bindings/wasm && npm install && npm run build",
+    "build": "wasm-pack build -t web --out-dir src/pkg bindings/wasm/crate && rm -rf bindings/wasm/dist/ && rollup -c && cp -r bindings/wasm/dist dist",
```
neekolas, line 33: To reduce confusion, you should be able to remove the rest of these scripts

neekolas (approved):
If it works, good enough for me. This is all throwaway, as eventually we are going to have to start publishing this thing to NPM. This just lets us move faster without have to think about package versioning.

michaelx11:
The final test for me is to get the Github actions to pass on our corresponding `xmtp-js` branch: https://github.com/xmtp/xmtp-js/pull/322

PR #39, opened 2023-04-03
Encrypt and Decrypt support

neekolas:
## Summary

Trying to teach myself some Rust here. Attempted to add Encrypt and Decrypt support. This also required me to change the output type from `createInboundSession` and `createOutboundSession` to include the `sessionId` in the payload. I went for the `Box<JsValue>` type and am returning the two values as a tuple for simplicity, but I'm not sure what the best pattern is here.

## Notes
- I initially wrote the last test with Alice decrypting her own message. That didn't work. Is that not possible with the way we have Olm configured? The error I got was `Bad MAC`.
- This is _my first ever Rust PR_. Have at me. Be ruthless. I have no idea what I'm doing.

### DIFF
--- a/bindings/wasm/crate/src/lib.rs
+++ b/bindings/wasm/crate/src/lib.rs
@@ -29,7 +29,7 @@ pub fn create_outbound_session(
     sending_handle: &str,
     receiving_handle: &str,
     message: &str,
-) -> Result<String, JsValue> {
+) -> Result<Box<[JsValue]>, JsValue> {
     // Look up both handles in INSTANCE_MAP
     let instances = INSTANCE_MAP.lock().unwrap();
     let mut sending_instance = instances
@@ -53,8 +53,13 @@ pub fn create_outbound_session(
         .get(receiving_handle)
         .unwrap()
         .set(receiving_instance);
+
     match result {
-        Ok((_, ciphertext_json)) => Ok(ciphertext_json),
+        Ok((session_id, ciphertext_json)) => Ok(vec![
+            JsValue::from_str(&session_id),
+            JsValue::from_str(&ciphertext_json),
+        ]
+        .into_boxed_slice()),
         Err(e) => Err(JsValue::from_str(&e.to_string())),
     }
 }
@@ -64,7 +69,7 @@ pub fn create_inbound_session(
     sending_handle: &str,
     receiving_handle: &str,
     message: &str,
-) -> Result<String, JsValue> {
+) -> Result<Box<[JsValue]>, JsValue> {
     // Look up both handles in INSTANCE_MAP
     let instances = INSTANCE_MAP.lock().unwrap();
     let sending_instance = instances
@@ -90,7 +95,52 @@ pub fn create_inbound_session(
         .set(receiving_instance);
 
     match result {
-        Ok((_, plaintext)) => Ok(plaintext),
+        Ok((session_id, ciphertext_json)) => Ok(vec![
+            JsValue::from_str(&session_id),
+            JsValue::from_str(&ciphertext_json),
+        ]
+        .into_boxed_slice()),
+        Err(e) => Err(JsValue::from_str(&e.to_string())),
+    }
+}
+
+#[wasm_bindgen]
+pub fn encrypt_message(
+    sending_handle: &str,
+    session_id: &str,
+    message: &str,
+) -> Result<String, JsValue> {
+    let instances = INSTANCE_MAP.lock().unwrap();
+    let mut instance = instances
+        .get(sending_handle)
+        .ok_or("sending_handle not found")?
+        .take();
+
+    let result = instance.encrypt_message_serialized(session_id, message);
+    // Do we actually need to do this?
+    instances.get(sending_handle).unwrap().set(instance);
+
+    match result {
+        Ok(ciphertext_json) => Ok(ciphertext_json),
+        Err(e) => Err(JsValue::from_str(&e.to_string())),
+    }
+}
+
+#[wasm_bindgen]
+pub fn decrypt_message(
+    handle: &str,
+    session_id: &str,
+    ciphertext: &str,
+) -> Result<String, JsValue> {
+    let instances = INSTANCE_MAP.lock().unwrap();
+    let mut instance = instances.get(handle).ok_or("handle not found")?.take();
+
+    let result = instance.decrypt_message_serialized(session_id, ciphertext);
+    // Do we actually need to do this?
+    instances.get(handle).unwrap().set(instance);
+
+    match result {
+        Ok(plaintext) => Ok(plaintext),
         Err(e) => Err(JsValue::from_str(&e.to_string())),
     }
 }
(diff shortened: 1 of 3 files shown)


bindings/wasm/crate/src/lib.rs
```diff
@@ -90,7 +95,52 @@ pub fn create_inbound_session(
...
+pub fn decrypt_message(
+    handle: &str,
+    session_id: &str,
+    ciphertext: &str,
+) -> Result<String, JsValue> {
+    let instances = INSTANCE_MAP.lock().unwrap();
+    let mut instance = instances.get(handle).ok_or("handle not found")?.take();
+
+    let result = instance.decrypt_message_serialized(session_id, ciphertext);
+    // Do we actually need to do this?
```
michaelx11, line 139: ditto above: don't need to follow my mistakes, `get_mut` on a `Cell` should work here I think

bindings/wasm/crate/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub fn create_outbound_session(
     sending_handle: &str,
     receiving_handle: &str,
     message: &str,
-) -> Result<String, JsValue> {
```
jazzz, line 32: I'm not sure I follow the need for the Box here.

My understanding was that JsValues through WASM_BINDGEN get copied between the memory space?

 Heap Allocating here seems unnecessary at a high level.

bindings/wasm/tests/wasmpkg.test.ts
```diff
@@ -12,19 +12,48 @@ it("can instantiate", async () => {
...
+  await bob.createInboundSession(alice, payload);
+  expect(typeof sessionId).toBe("string");
+
+  const msg = "hello there";
+  const encrypted = await alice.encryptMessage(sessionId, msg);
+  expect(typeof encrypted).toBe("string");
+
+  // Alice can't decrypt her own message. Does work for Bob though
+  // const decrypted = await alice.decryptMessage(sessionId, encrypted);
+  const decrypted = await bob.decryptMessage(sessionId, encrypted);
```
michaelx11, line 57: beautiful

michaelx11 (approved):
LGTM - I think we're about evenly matched already in Rust so I found nothing to be harsh about

bindings/wasm/crate/src/lib.rs
```diff
@@ -53,8 +53,13 @@ pub fn create_outbound_session(
         .get(receiving_handle)
         .unwrap()
         .set(receiving_instance);
+
     match result {
-        Ok((_, ciphertext_json)) => Ok(ciphertext_json),
+        Ok((session_id, ciphertext_json)) => Ok(vec![
+            JsValue::from_str(&session_id),
+            JsValue::from_str(&ciphertext_json),
+        ]
```
jazzz, line 55: Using the Vec here changes the memory structure. Given the size cannot be known at compile time. 

Imagine using a tuple would be easier to use and would avoid the need for boxing of the values.

bindings/wasm/crate/src/lib.rs
```diff
@@ -90,7 +95,52 @@ pub fn create_inbound_session(
...
+) -> Result<String, JsValue> {
+    let instances = INSTANCE_MAP.lock().unwrap();
+    let mut instance = instances
+        .get(sending_handle)
+        .ok_or("sending_handle not found")?
+        .take();
+
+    let result = instance.encrypt_message_serialized(session_id, message);
+    // Do we actually need to do this?
+    instances.get(sending_handle).unwrap().set(instance);
```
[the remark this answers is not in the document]
michaelx11: In fact, now that we're using `.public_account`, I think we can do away with this pattern in those existing functions too. I might tackle that in a separate PR.

bindings/wasm/crate/src/lib.rs
```diff
@@ -90,7 +95,52 @@ pub fn create_inbound_session(
...
+
+    match result {
+        Ok(ciphertext_json) => Ok(ciphertext_json),
+        Err(e) => Err(JsValue::from_str(&e.to_string())),
+    }
```
jazzz, line 111: This seems fine to me, however there are more idomatic ways to use a Match statement.

1. Pretty sure that RHS `Ok()` is not needed. Values are implicitly `Ok`
2. result is an 'Result' which you are trying to map the error. You can use `result.map_error(|e| JsValue::from_str(&e.to_string()))`

bindings/wasm/crate/src/lib.rs
jazzz: `.unwrap` will panic on error. Becareful with these outside of prototyping code.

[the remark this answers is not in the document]
neekolas: The type checker doesn't seem to like that, but it recommends that I replace the cell with a `RefCell`. RefCell then has a method `borrow_mut`, which if I use will succeed.

Does have some scary language about concurrent usage leading to a panic, but I think that's fine for now given that all our code is synchronous.

bindings/wasm/crate/src/lib.rs
```diff
@@ -90,7 +95,52 @@ pub fn create_inbound_session(
...
+}
+
+#[wasm_bindgen]
+pub fn encrypt_message(
+    sending_handle: &str,
+    session_id: &str,
+    message: &str,
+) -> Result<String, JsValue> {
```
jazzz, line 99: These types look both to be Strings, only the error gets converted to a wasmbind:JsValue. Shouldn't they be the same type?

bindings/wasm/crate/src/lib.rs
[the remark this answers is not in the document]
neekolas: I've committed that change [here](https://github.com/xmtp/libxmtp/pull/39/commits/4781a67d83b24b1f0002a2f6f26614427c32829a)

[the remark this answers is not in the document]
michaelx11: Sounds good, anything works for now as long as borrow/type checker likes it. We can take a holistic look at the map/cell strategy in a subsequent PR

bindings/wasm/crate/src/lib.rs
```diff
@@ -90,7 +95,52 @@ pub fn create_inbound_session(
...
+    let instances = INSTANCE_MAP.lock().unwrap();
+    let mut instance = instances
+        .get(sending_handle)
+        .ok_or("sending_handle not found")?
```
jazzz, line 103: I like this conversion from an Option to Result. 

You could have used a match, but this is much cleaner

jazzz (approved):

bindings/wasm/crate/src/lib.rs
neekolas: Unfortunately, `wasm-bindgen` doesn't appear to support tuples as return values. https://github.com/rustwasm/wasm-bindgen/issues/122

bindings/wasm/crate/src/lib.rs
```diff
@@ -20,7 +20,7 @@ lazy_static! {
 pub fn new_voodoo_instance() -> String {
     let mut instances = INSTANCE_MAP.lock().unwrap();
     let handle = (instances.len() as u64).to_string();
-    instances.insert(handle.clone(), Cell::new(VoodooInstance::new()));
+    instances.insert(handle.clone(), RefCell::new(VoodooInstance::new()));
```
jazzz, line 23: RefCell seems odd here: Though WASM is odd so it might be the best way. Naively I would have thought we could handle borrows at compile time.

PR #40, opened 2023-04-04
feat: methods to allow insecure export and import of vodozemac identies

michaelx11:
## Overview

This PR adds methods to xmtpv3 and xmtpv3_wasm to allow us to export a JSON dump of the "public identity" (which is currently equivalent to the whole identity). We pair this with a PR in xmtp-js which publishes the contact bundle to a `voodoo-contact` topic, fetches it and uses it for tests.

## Changes

These changes are slightly painful, but necessary until we have a real public/private distinction. For now, our strategy is to allow `xmtpv3_wasm` to take JSON, save the account, and return a handle. This handle can be used to start conversations.

So the basic flow (encapsulated by our Javascript code in xmtpv3) is:
```
const myHandle = new_voodoo_instance()
const otherExport = other.get_public_account_json()
const otherHandle = add_or_get_public_account_from_json(otherExport)
create_outbound_session(myHandle, OtherHandle)
```

## Test Plan

- new roundtrip test with JSON export/import

### DIFF
--- a/bindings/wasm/crate/src/lib.rs
+++ b/bindings/wasm/crate/src/lib.rs
@@ -31,7 +31,9 @@ pub fn create_outbound_session(
     message: &str,
 ) -> Result<Box<[JsValue]>, JsValue> {
     // Look up both handles in INSTANCE_MAP
-    let instances = INSTANCE_MAP.lock().unwrap();
+    let instances = INSTANCE_MAP.lock().map_err(|e| {
+        JsValue::from_str(&format!("Error getting instance map lock: {}", e))
+    })?;
     let mut sending_instance = instances
         .get(sending_handle)
         .ok_or("sending_handle not found")?
@@ -65,7 +67,9 @@ pub fn create_inbound_session(
     message: &str,
 ) -> Result<Box<[JsValue]>, JsValue> {
     // Look up both handles in INSTANCE_MAP
-    let instances = INSTANCE_MAP.lock().unwrap();
+    let instances = INSTANCE_MAP.lock().map_err(|e| {
+        JsValue::from_str(&format!("Error getting instance map lock: {}", e))
+    })?;
     let sending_instance = instances
         .get(sending_handle)
         .ok_or("sending_handle not found")?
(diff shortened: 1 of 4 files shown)


bindings/wasm/crate/src/lib.rs
```diff
@@ -131,6 +131,47 @@ pub fn decrypt_message(
     }
 }
 
+// This function is the most barebones (incorrect) implementation of obtaining
+// a "contact bundle" for a given VoodooIdentity by handle. The output string
+// is a JSON object which can subsequently be imported into another VoodooInstance
+#[wasm_bindgen]
+pub fn get_public_account_json(handle: &str) -> Result<String, JsValue> {
+    let instances = INSTANCE_MAP.lock().unwrap();
```
jazzz, line 139: Instead of `unwrap` you can use `?`. lock returns a LockResult which should convert to result.

bindings/wasm/src/xmtpv3.ts
```diff
@@ -134,6 +140,15 @@ export class XMTPWasm {
...
   }
 
+  getPublicAccountJSON(handle: string): string {
+    return get_public_account_json(handle);
+  }
+
+  addOrGetPublicAccountFromJSON(json: string): VoodooInstance {
+    const handle = add_or_get_public_account_from_json(json);
+    return new VoodooInstance(this, handle);
+  }
```
jazzz, line 150: I'd expect the return types to match given the names of the functions

jazzz (approved):

bindings/wasm/src/xmtpv3.ts
michaelx11: Like it should be returning a handle? It's more ergonomic if it returns a VoodooInstance but I can rename it

bindings/wasm/crate/src/lib.rs
michaelx11: yup!

bindings/wasm/src/xmtpv3.ts
michaelx11: I added a comment to explain that it returns a VoodooInstance for now, until we have a corresponding public type

PR #41, opened 2023-04-04
Return JsValue instead of string and remove match statements

neekolas:
## Summary

Address some of Jazz's feedback from https://github.com/xmtp/libxmtp/pull/39


michaelx11 (approved):
Nice!

jazzz (approved):

PR #42, opened 2023-04-05
Jazzz/add reg

jazzz:
Adds Registration mechanic

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -253,6 +253,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
+name = "base64"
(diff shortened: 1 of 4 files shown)


jazzz:
Currently hung up on a Dependency conflict between Vodezomac and ethers

michaelx11:
Exciting! I'll take a quick look at the dependency issue too later

michaelx11:
If we're okay with downgrading to `ethers=0.6.0`, it'll accept zeroize 1.3.0 which is what vodozemac 0.3.0 wants. However, `ethers=0.6.0` drags in `k256=0.9.6` which pulls in `sha2 v0.9.9` and finally`sha2-asm v0.6.2` which is not wasm-unknown-unknown compatible.

Diff here: https://github.com/xmtp/libxmtp/compare/jazzz/add_reg...michaelx-add_reg_version_finagle?expand=1

PR #43, opened 2023-04-05
fix: update ffi and node-bindgen bindings

michaelx11:
## Overview

Neither binding type is actively used but keeping them running is a good test. It'll help us avoid adding dependencies that break another platform silently.

## Changes

- For ffi, just include `::manager` to reflect renaming within `xmtpv3`
- For node-bindgen, I had to rename `app.js` to `app.cjs` to use `require`

## Test Plan

- build the Android app with new ffi bindings
- run `./run.sh` in node-bindgen


neekolas (approved):

michaelx11:
Gonna follow this up with adding back the .github/actions I wiped out when we started this branch

michaelx11:
Tested Android app, works nicely still!

PR #46, opened 2023-04-07
feat: incredibly bare-bones in-memory message store by wrapping olm::Session

michaelx11:
## Overview

This is a quick PR to start the idea of storing messages in WASM land. To do this, we wrap the `olm::Session` with our own struct. This struct should be a serializable type at some point.

## Test Plan

Added unit tests to check `my_messages` and `their_messages`. Rest of tests should continue to pass.

### DIFF
--- a/crates/xmtpv3/src/session.rs
+++ b/crates/xmtpv3/src/session.rs
@@ -1,0 +1,34 @@
+use anyhow::Result;
+use vodozemac::olm::{OlmMessage, Session};
+
+// Wrapper struct for Olm sessions with additional metadata
+pub struct VoodooSession {
+    pub session: Session,
+    // NOTE: this does not save the first plaintext encoded in the
+    // outbound session
+    // TODO: instead of storing pure plaintext, we should store higher level structs
+    pub my_messages: Vec<String>,
+    pub their_messages: Vec<String>,
+}
+
+impl VoodooSession {
+    pub fn new(olm_session: Session) -> Self {
+        Self {
+            session: olm_session,
+            my_messages: Vec::new(),
+            their_messages: Vec::new(),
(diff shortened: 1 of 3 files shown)


crates/xmtpv3/src/session.rs
```diff
@@ -0,0 +1,34 @@
+use anyhow::Result;
+use vodozemac::olm::{OlmMessage, Session};
+
+// Wrapper struct for Olm sessions with additional metadata
+pub struct VoodooSession {
+    pub session: Session,
+    // NOTE: this does not save the first plaintext encoded in the
+    // outbound session
+    // TODO: instead of storing pure plaintext, we should store higher level structs
```
jazzz, line 9: 👍

crates/xmtpv3/src/session.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub my_messages: Vec<String>,
+    pub their_messages: Vec<String>,
+}
+
+impl VoodooSession {
+    pub fn new(olm_session: Session) -> Self {
+        Self {
+            session: olm_session,
+            my_messages: Vec::new(),
+            their_messages: Vec::new(),
```
jazzz, line 19: How might this look like in the final version? Would there still be a need to store plaintexts on the Session or is this to facilitate the prototype only?

jazzz (approved):

crates/xmtpv3/src/session.rs
michaelx11: Great question: I'm open to however we want to structure this. We're going to be exporting both session state and message history from Rust land at some point right? (or not - asking to gut-check my assumptions)

PR #47, opened 2023-04-10
Add public contact bundle type

richardhuaaa:
In the long-term the client should be able to query the server for a contact (either via wallet address, or via device identity key), and receive one or more contact bundles (public device identity key, public one time key). This PR adds the contact bundle type. Some simplifications for the prototype:

- The contact bundle is served directly from the client rather than the server
- The 'one time key' in the bundle is the fallback key, and it never changes

Verified existing tests pass.

I could propagate these changes into js, but I think ultimately we want to fetch these bundles directly from the server. 

I think I can prototype publishing and fetching bundles as part of multi-device support next. My thought is that we upload a large bundle of keys (PreKeyTopUpBundle) for ourself (e.g. 1000 one time keys and a fallback key), and we retrieve a small bundle (ContactBundle) for others (device identity key, one time key). The server is responsible for constructing the contact bundle from the prekey topup bundle.

### DIFF
--- a/crates/xmtpv3/src/account.rs
+++ b/crates/xmtpv3/src/account.rs
@@ -1,5 +1,8 @@
 use anyhow::Result;
-use vodozemac::{olm::{Account, AccountPickle}, Curve25519PublicKey};
+use vodozemac::{
+    olm::{Account, AccountPickle},
+    Curve25519PublicKey,
+};
 
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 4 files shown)


neekolas:
> The server is responsible for constructing the contact bundle from the prekey topup bundle.

I'd like to understand what is being proposed here. Any change where the server is returning something other than a byte-perfect representation of what was uploaded comes with complications as we decentralize the network.

Few questions:
1. Are we expecting the server to maintain some sort of state of which one time keys have already been used?
2. If not, what logic would be used to select the one time key?
3. What would be signed in this scheme? How do we verify that the server-constructed bundles were not tampered with?

michaelx11 (approved):
LGTM - for the next step we might need to rework the bindings crate for `add_or_get_public_account_from_json`. Instead of storing other's public info as a new `VoodooInstance` (my crappy code) we can store a separate map of public contacts!

jazzz:
>(either via wallet address, or via device identity key) 

Does this mean that two versions are submitted the network? One indexed by Address and one by IdentityKey? Or are Contact Bundles powered by a different service?

PR #50, opened 2023-04-14
feat: Build grpc API with swift-bridge and package with Swift

michaelx11:
## Overview

Top-level issue [link](https://github.com/xmtp/xmtp-rust-swift/pull/1/files#diff-7a17750f3b7a3f3aa4fcece43436a8f84200a7a7095328f23ea4fad3ccf4b3baR11)

This PR goes hand-in-hand with an update to our Swift package [repo](https://github.com/xmtp/xmtp-rust-swift/pull/1/files#diff-7a17750f3b7a3f3aa4fcece43436a8f84200a7a7095328f23ea4fad3ccf4b3baR11) to hide all of this nonsense behind a single API defined in `ApiService.swift` in that repo.

## Why is this PR so big?

Things to ignore: 1) generated Rust proto code 2) generated Rust-Swift binding code 3) I also deleted `corecrypto` from the crate cause we didn't need it

## What changed in the Rust<>Swift flow?

xmtp-ios should just be consuming a run-of-the-mill Swift package with a super simple Swift-only API.

How we did this was:
1) Use [swift-bridge](https://chinedufn.github.io/swift-bridge/) to generate nice `C` headers and Swift files from our code
2) Push the `C` headers into our `include` directory to get bundled into the XMTPRustSwift.xcframework
3) Behind the scenes, push the Swift code into `xmtp-rust-swift` after adding an "import XMTPRustSwift" line to each file
4) Finally within `xmtp-rust-swift` (our Swift package), we use the XMTPRustSwift.xcframework, wrap it with ApiService and deliver a nice clean Swift package interface

Right now the shim strategy is to convert protos to JSON and throw them across the Rust<>Swift boundary. We can optimize this as necessary but number one priority is to get it working.

## Next Steps

Need to implement "subscribe" and make sure our Rust gRPC can target TLS.

## Test Plan

Bring up a local xmtp-go-node with gRPC port mapped to 15555 (or change port in the test) and run unit tests.


neekolas:
Copying all our protos into this package seems less than ideal. Maybe a project for me next week is to avoid the need for that

crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -0,0 +1,99 @@
...
+    } else {
+        None
+    };
+
+    // Do the query and get a Tonic response that we need to process
+    let response = query(host, topic, paging_info)
+        .await
+        .map_err(|e| format!("{}", e))?;
+    // Response is a v1::QueryResponse protobuf message, which we need to serialize to JSON
+    let json = serde_json::to_string(&response).map_err(|e| format!("{}", e))?;
```
neekolas, line 48: How do the proto fields that are of type `bytes` make it through this JSON serialization?

michaelx11: Great question, in my experience with a different Rust proto library they become base64-encoded.

michaelx11:
> Copying all our protos into this package seems less than ideal. Maybe a project for me next week is to avoid the need for that

Oh yeah I should make them a submodule now actually. Let me fix that

crates/xmtp-networking/src/grpc_api_helper.rs
michaelx11: I lied, looks like they become byte arrays: "{\"content_topic\":\"\",\"timestamp_ns\":0,\"message\":[1,2,3,4]}"

neekolas:
I'm hoping to get around the nightmare that is our proto repo by using Buf Schema Registry. Just pushed up [the first package](https://buf.build/xmtp/proto).

Submodule is fine for now, but I'm happy to take on buf-ifying this thing on Monday as a test before I buf-ify the mobile SDKs.

michaelx11:
@neekolas this commit addresses the base64 issue in a very targeted and brittle way. As long as we only serialize the MessageApi protos to JSON this should be fine, but gotta keep an eye on this for sure.

https://github.com/xmtp/libxmtp/pull/50/commits/263a4d28be490f791feee5bc19c7835d3d6f3ce8

https://github.com/tokio-rs/prost/issues/75

crates/xmtp-networking/src/grpc_api_helper.rs
neekolas: OK. That's great

neekolas:
> @neekolas this commit addresses the base64 issue in a very targeted and brittle way. As long as we only serialize the MessageApi protos to JSON this should be fine, but gotta keep an eye on this for sure.
> 
> [263a4d2](https://github.com/xmtp/libxmtp/commit/263a4d28be490f791feee5bc19c7835d3d6f3ce8)
> 
> [tokio-rs/prost#75](https://github.com/tokio-rs/prost/issues/75)

For a prototype, that all seems fine. I suspect we're going to want to pass these back and forth as structs to optimize perf at some point.

crates/xmtp-networking/src/grpc_api_helper.rs
michaelx11: I think that's bad, Swift expects base64 from my test. I added in a hack until Prost becomes proto3 json compliant

michaelx11:
Gonna add one quick fix to `bindings/xmtp_rust_swift`, since the bindings crate is in the Cargo.toml all artifacts go into a top-level `target` directory. I'll change the target-dir in the `Makefile`.

crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -0,0 +1,144 @@
...
+use tonic::{metadata::MetadataValue, transport::Channel, Request};
+
+// Do a barebones unpaginated Query gRPC request
+// With optional PagingInfo
+pub async fn query(
+    host: String,
+    topic: String,
+    paging_info: Option<v1::PagingInfo>,
+) -> Result<v1::QueryResponse, tonic::Status> {
+    let mut client = v1::message_api_client::MessageApiClient::connect(host)
```
neekolas, line 12: At some point we are going to want to optimize this to re-use a long-lived GRPC connection, given how nice and performant it is that GRPC can multiplex a single HTTP/2 connection.

But that's a problem for another day.

crates/xmtp-proto/src/gen/xmtp.keystore_api.v1.rs
```diff
@@ -0,0 +1,904 @@
+// @generated
```
neekolas, line 1: I could have sworn I `.gitignore`d these files. They don't really need to be checked in, given that they are recreated on every build.

neekolas (approved):

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "crates/xmtp-networking/proto/googleapis"]
```
neekolas, line 1: You can remove this `submodule` with my changes

.gitignore
```diff
@@ -124,3 +124,7 @@ dist
 # Swap files for vim
 *.swp
 *.swo
+
+
+# Generated code
+crates/xmtp_proto/src/gen
```
neekolas, line 130: Whoops. I changed this folder from `xmtp_proto` to `xmtp-proto` but didn't update the .gitignore

michaelx11:
FYI: I removed `bindings/xmtp_rust_crate` from the Cargo workspace so its release profile in its own Cargo.toml would be respected. This was why we were getting: `warning: profiles for the non root package will be ignored, specify profiles at the workspace root:` warnings

PR #52, opened 2023-04-16
Simplify proto gen

neekolas:
## Summary

Trying a different approach to generating protobuf code.

Advantages are:
- No submodules or copying proto files
- Simpler `build.rs` config with no special overrides needed for serialization.
- Uses buf schema registry for our protos

## Notes

Some things to watch out for:
- Includes a generated `.serde.rs` file for all the protos. One side effect of this is that zero'd proto fields are no longer present in JSON output. I had to update one of your tests to accomodate the change.
- I changed the grpc port for the tests to :5556, which may break your local setup
- There may be subtle differences in serialization I'm not aware of. The only thing I checked is that `bytes` fields are serialized as base64 strings, which this generated Serde setup seems to do by default. But there may be other less desirable changes that I haven't noticed yet.
- I have no idea why I can only make this work as a separate crate. I tried a bunch of ways to make it work inside `xmtp-networking`, but my knowledge of Rust modules is just not there to debug the issue. If you can make it work, that's probably better.


michaelx11:
Wow awesome I'll take a look!

michaelx11 (approved):
This is an amazing improvement. I was hunting for the reference to our protos, but it's in build.rs pointing at https://buf.build/xmtp/proto right?

Having a separate crate for protos is probably a better idea too!

crates/xmtp-proto/src/gen/mod.rs
```diff
@@ -0,0 +1,25 @@
+// @generated
+pub mod xmtp {
```
michaelx11, line 2: I feel like a moron for having to handwrite this file myself in the original branch. I felt like a genius at the time for figuring out how to deal with our package names though 😅

neekolas: Oh man, I never would have figured that out on my own.

michaelx11:
Feel free to merge whenever!

PR #53, opened 2023-04-17
feat: rustls and full grpc api support

michaelx11:
## Overview

This implements TLS for gRPC api connections, albeit in a slightly ugly way. My main inspiration was the rust_tls [example](https://github.com/hyperium/tonic/blob/master/examples/src/tls_rustls/client.rs) in tonic repo. This code has been tested with `xmtp-ios` and passes all tests if used as an approximate drop-in replacement for gRPC.

I want to DRY things up, especially with client creation but I'm wrangling with Rust lifetimes and complicated return types. For now, I'll stick to this (works and is tested) and we can refactor iteratively.

## Changes

- Rework publish/query/subscribe to initialize either TLS enabled or non-TLS clients, depending on the host string passed in.
- Change `publish` to take a Vec of JSON-encoded envelope protos. This ends up being more ergonomic than trying to construct then JSON encode a list of Envelopes in Swift

## Tests
- Unit tests (if local node is up)
- E2E tests by running xmtp-ios (requires additional code in xmtp-rust-swift repo and [xmtp-ios](https://github.com/xmtp/xmtp-ios/tree/4052e23) to utilize the Rust gRPC api service

### DIFF
--- a/bindings/xmtp_rust_swift/.gitignore
+++ b/bindings/xmtp_rust_swift/.gitignore
@@ -15,8 +15,8 @@ bundle.zip
 .build
 
 # Ignore swift files in include/
-./include/Generated/*.swift
-./include/Generated/xmtp_rust_swift/*.swift
+/include/Generated/*.swift
+/include/Generated/xmtp_rust_swift/*.swift
 
 # Ignore Generated, gets rebuilt on cargo build
-./Generated
+/Generated
--- a/bindings/xmtp_rust_swift/.gitignore
+++ b/bindings/xmtp_rust_swift/.gitignore
@@ -13,4 +13,10 @@ bundle.zip
 /target
 /Cargo.lock
 .build
+
+# Ignore swift files in include/
+./include/Generated/*.swift
+./include/Generated/xmtp_rust_swift/*.swift
+
+# Ignore Generated, gets rebuilt on cargo build
 ./Generated
--- a/crates/xmtp-networking/src/grpc_api_helper.rs
+++ b/crates/xmtp-networking/src/grpc_api_helper.rs
@@ -53,17 +53,7 @@ pub async fn query(
     topic: String,
     paging_info: Option<v1::PagingInfo>,
 ) -> Result<v1::QueryResponse, tonic::Status> {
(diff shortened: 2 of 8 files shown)


bindings/xmtp_rust_swift/.gitignore
```diff
@@ -13,4 +13,10 @@ bundle.zip
 /target
 /Cargo.lock
 .build
-./Generated
\ No newline at end of file
+
+# Ignore swift files in include/
+/include/Generated/*.swift
+/include/Generated/xmtp_rust_swift/*.swift
```
neekolas, line 19: We can't ignore the `.h` files as well?

crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -53,44 +127,76 @@ pub async fn query_serialized(
...
+                let tls = tls.clone();
+
+                hyper_rustls::HttpsConnectorBuilder::new()
+                    .with_tls_config(tls)
+                    .https_or_http()
+                    .enable_http2()
+                    .wrap_connector(s)
+            })
+            .service(http);
+        let client = hyper::Client::builder().build(connector);
```
neekolas, line 163: I see what you mean about the need to DRY this up

bindings/xmtp_rust_swift/.gitignore
michaelx11: The .h files actually need to be added to `include` and exported along with the XCFramework. Not sure if this is the ideal way but works and avoids having us require the app developer to add separate headers as source in their XCode project. From high level it makes sense to package the headers with the static library.

crates/xmtp-networking/src/grpc_api_helper.rs
michaelx11: yeah :/ my naive attempt at just taking this code and pasting it into a function resulted in some indecipherable Rust errors (at least with my knowledge at this point)

michaelx11:
@neekolas I'm going to close this PR in favor of your refactor for single-client instantiation

PR #55, opened 2023-04-18
feat: initial crate for libxmtp-core

richardhuaaa:
This is just an initial crate for our xmtpv3 work, building off @michaelx11 and @jazzz's prior work on rust bindings. I'm still piecing together how crates/workspaces/bindings work in rust-land so go easy on me!

Tested via `cargo test` in root directory, and `npm run test` in `bindings/wasm`

Some Q's here:
1. Any reason we're using vitest over jest (which xmtp-js uses)? Would prefer standardization across repos if possible.
2. I imagine some of the existing prototyping code can be removed, but I don't have the confidence to do it. Can @michaelx11 or @jazzz do this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,8 @@ members = [
   "crates/xmtp-keystore",
   "crates/corecrypto",
   "crates/xmtp-networking",
+  "crates/libxmtp-core",
+  "bindings/wasm/crate",
 ]
 
 # Exclude since
--- a/bindings/wasm/crate/Cargo.toml
+++ b/bindings/wasm/crate/Cargo.toml
@@ -8,10 +8,11 @@ edition = "2018"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-wasm-bindgen = "0.2"
 js-sys = "0.3"
-xmtp-keystore = { path = "../../../crates/xmtp-keystore" }
+libxmtp-core = { path = "../../../crates/libxmtp-core" }
 once_cell = "1.17.1"
+wasm-bindgen = "0.2"
+xmtp-keystore = { path = "../../../crates/xmtp-keystore" }
 
 
 [dev-dependencies]
--- a/crates/libxmtp-core/Cargo.toml
+++ b/crates/libxmtp-core/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "libxmtp-core"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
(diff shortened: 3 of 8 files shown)


Cargo.toml
```diff
@@ -5,6 +5,8 @@ members = [
   "crates/xmtp-keystore",
   "crates/corecrypto",
   "crates/xmtp-networking",
+  "crates/libxmtp-core",
```
michaelx11, line 8: Is it worth tagging with `v3`? Or is libxmtp-core's scope going to go beyond? I do think once we pull in vodozemac we might hit conflicts that require excluding it from the Cargo workspace, but we can cross that bridge when we get there.

bindings/wasm/crate/Cargo.toml
```diff
@@ -8,10 +8,11 @@ edition = "2018"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-wasm-bindgen = "0.2"
 js-sys = "0.3"
-xmtp-keystore = { path = "../../../crates/xmtp-keystore" }
+libxmtp-core = { path = "../../../crates/libxmtp-core" }
 once_cell = "1.17.1"
+wasm-bindgen = "0.2"
+xmtp-keystore = { path = "../../../crates/xmtp-keystore" }
```
michaelx11, line 15: We can just nuke all the `xmtp-keystore` code in `bindings/wasm` at this point and start fresh. That can be done in a follow-up PR

crates/libxmtp-core/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "libxmtp-core"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
michaelx11, line 8: Happy to merge PR as is, but if you're feeling adventurous you can add `vodozemac` here to see how the dependency situation plays out. I could be wrong about conflicts!

michaelx11 (approved):
Overall it's a reasonable starting crate structure. I do think the overall crate names need a rework, but I usually prefer as specific as possible. If this crate is for v3, we can just call it xmtp-v3 IMO.

Not a particularly strong opinion but why sow more confusion than I already have with my names haha

Cargo.toml
michaelx11: It's a good question, I'm not sure if consumers will even see this crate directly. They might consume an intermediary crate (like a bindings/wasm crate). In that sense we have a lot of flexibility with naming to make things clearer. `libxmtp-core` is totally fine since it'll hopefully be very clear what code lives in it (V3 for now + some state management goodies maybe)

PR #58, opened 2023-04-18
Streaming support

neekolas:
## Summary

- Adds `Subscription` struct for managing the GRPC subscription, storing messages, querying for those messages, and ending the subscription.
- Wraps in a binding that returns a `subscription_id`. The `subscription_id` can be polled for later to get the most recent messages.

## To do
- Add binding to end a subscription. They just go on forever right now
- I'd like to remove the ResponseJson type and just return real structs. All this serializing is not going to be very maintainable, and makes testing harder.

## Notes

In order for the tests to pass, you currently have to use a local version of `xmtp-node-go` that has authn disabled. That way you don't need a valid auth token to publish. Will automate the running of that later, but I included the docker-compose file in this repo.


crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -133,6 +134,24 @@ pub async fn subscribe_once_serialized(
...
+    println!("Starting subscribe process");
+    let mut client = v1::message_api_client::MessageApiClient::connect(host)
+        .await
+        .map_err(|e| tonic::Status::new(tonic::Code::Internal, format!("{}", e)))?;
+    let request = v1::SubscribeRequest {
+        content_topics: topics,
+        ..Default::default()
+    };
+    println!("Sending request: {:?}", request);
+    let stream = client.subscribe(request).await.unwrap().into_inner();
```
neekolas, line 150: This is where things get stuck

neekolas:
Moved this PR description to a comment for posterity:

Everything was going great here...until I actually tried to test the code. client.subscribe(...) hangs forever instead of returning a Streaming<Envelope> as expected.

Halp!

Things I've Tested
I used a local version of tonic and traced down where it hangs. It makes it [this far](https://github.com/hyperium/tonic/blob/master/tonic/src/client/grpc.rs#L309), and then gets stuck.
This is unusual. My code looks a whole lot like the [client streaming example](https://github.com/hyperium/tonic/blob/master/examples/src/streaming/client.rs#L28) in the Tonic repo, which I've run alongside the server-streaming example and does work as expected. The stream is returned immediately and then you wait for messages on the stream.
This issue has nothing to do with my Subscription code, since that code never gets run. Also, the same issue is reproducible on subscribe_once, which doesn't use the Subscription struct at all.
I've confirmed from the node logs that the subscription request does make it to the node and no errors are logged
I haven't found anyone else with the same Tonic issue as me. If I can come up with a reproducible test case in their repo I'll create one
Theories
Something weird is happening with Tokio and deadlocks. No idea why this wouldn't be happening with the query endpoint, which works great. But possible.
Our nodes are doing something non-standard with GRPC, and Tonic is uniquely sensitive about the issue. Maybe something to do with trailers?
Anyone have any ideas?

bindings/xmtp_rust_swift/src/subscriptions.rs
```diff
@@ -0,0 +1,30 @@
+use lazy_static::lazy_static;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+use xmtp_networking::grpc_api_helper::Subscription;
+use xmtp_proto::xmtp::message_api::v1::Envelope;
+
+// Define the global HashMap for storing Subscriptions
+lazy_static! {
+    static ref SUBSCRIPTIONS: Mutex<HashMap<String, Arc<Subscription>>> =
```
neekolas, line 9: There must be an easier way to get access to mutable state, but the ChatGPT endorsed solution seems to work.

bindings/xmtp_rust_swift/src/lib.rs
```diff
@@ -50,18 +56,100 @@ async fn publish(host: String, token: String, json_envelopes: String) -> ffi::Re
...
+            ADDRESS.to_string(),
+            "test-query".to_string(),
             "".to_string(),
         ));
         assert_eq!(result.error, "");
-        println!("Got result: {}", result.json);
+    }
+
+    #[tokio::test]
+    async fn test_subscribe() {
```
michaelx11, line 121: nice!

dev/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+services:
+  waku-node:
+    image: xmtp/node-go:latest
+    platform: linux/amd64
+    environment:
+      - GOWAKU-NODEKEY=8a30dcb604b0b53627a5adc054dbf434b446628d4bd1eccc681d223f0550ce67
+    command:
+      - --ws
+      - --store
+      - --message-db-connection-string=postgres://postgres:xmtp@db:5432/postgres?sslmode=disable
```
michaelx11, line 10: 👍

bindings/xmtp_rust_swift/src/lib.rs
```diff
@@ -50,18 +56,100 @@ async fn publish(host: String, token: String, json_envelopes: String) -> ffi::Re
...
+            }
+        }
+        Err(e) => ffi::ResponseJson {
+            error: e.to_string(),
+            json: "".to_string(),
+        },
+    }
+}
+
+async fn poll_subscription(subscription_id: String) -> ffi::ResponseJson {
```
michaelx11, line 85: Is this still `async`?

crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -98,47 +100,78 @@ pub async fn publish_serialized(
...
-    Ok(json)
+    return Ok(Subscription::start(stream).await);
 }
 
 // Return the json serialization of an Envelope with bytes
-pub fn test_envelope() -> String {
+pub fn test_envelope(topic: String) -> String {
+    let time_since_epoch = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
     let envelope = v1::Envelope {
+        timestamp_ns: time_since_epoch.as_nanos() as u64,
```
michaelx11, line 120: thanks for this

crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -98,47 +100,78 @@ pub async fn publish_serialized(
...
+    close_tx: Option<oneshot::Sender<()>>,
+}
+
+impl Subscription {
+    pub async fn start(stream: Streaming<Envelope>) -> Self {
+        let pending = Arc::new(Mutex::new(Vec::new()));
+        let pending_clone = pending.clone();
+        let (close_tx, close_rx) = oneshot::channel::<()>();
+        tokio::spawn(async move {
+            let mut stream = Box::pin(stream);
```
michaelx11, line 139: dumb question for my understanding: what does the `pin` do here?

michaelx11 (approved):
LGTM

bindings/xmtp_rust_swift/src/lib.rs
neekolas: Good call. It's sync now. Updated.

crates/xmtp-networking/src/grpc_api_helper.rs
neekolas: The compiler made me do it. I [read this](https://rust-lang.github.io/async-book/04_pinning/01_chapter.html) and still can't say I understand when I should and shouldn't be pinning things.

PR #59, opened 2023-04-19
feat: client class that takes in persistence callbacks

richardhuaaa:
Some thoughts here:

1. In terms of structuring the library, I'm going to suggest that we loosely follow the JS SDK with `Client`, `Conversation`, etc classes - that encourages us to abstract as much as is feasible in Rust and provides an easy template to follow.

2. I think the actual definitions of these callbacks can be implemented in the bindings, or passed down from the app layer. I made them public for now so the tests can access them, but they should be private.

### DIFF
--- a/bindings/wasm/crate/src/lib.rs
+++ b/bindings/wasm/crate/src/lib.rs
@@ -2,5 +2,5 @@ use wasm_bindgen::prelude::*;
 
 #[wasm_bindgen]
 pub fn add_two_numbers(left: usize, right: usize) -> usize {
-    libxmtp_core::add(left, right)
+    left + right
 }


crates/libxmtp-core/src/client.rs
```diff
@@ -0,0 +1,31 @@
...
+        write_to_persistence_fn: WriteToPersistenceFn,
+        read_from_persistence_fn: ReadFromPersistenceFn,
+    ) -> Client {
+        Client {
+            write_to_persistence_fn,
+            read_from_persistence_fn,
+        }
+    }
+
+    pub fn add(left: usize, right: usize) -> usize {
```
michaelx11, line 20: Do we still need this?

michaelx11 (commented):
Looks reasonable, is the next step to implement example persistence for a single binding?

crates/libxmtp-core/src/lib.rs
```diff
@@ -1,14 +1,41 @@
...
-        let result = add(2, 2);
+        let result = Client::add(2, 2);
         assert_eq!(result, 4);
     }
+
+    #[test]
+    fn can_pass_persistence_methods() {
+        let persisted_bytes = Rc::new(RefCell::new(Vec::new()));
+        let write_to_persist_fn = {
+            let persisted_bytes = persisted_bytes.clone();
```
michaelx11, line 19: no need for `.clone()` here

michaelx11 (commented):
After playing around with removing the `Box` references (possible with lifetimes) I think rather than callbacks we really want a `Persistence` trait and then the Client will take a reference to some struct that implements it. Similar to your `BackupClient` in xmtp-js or the Persistence for Keystores in xmtp-js.

Makes it easier to test also by passing in a possible `InMemoryPersistence` struct

michaelx11 (approved):
Looks much nicer! Tests in wasm are failing because the `add` method was deleted upon my request.  We can delete it for now.

PR #61, opened 2023-04-19
Create a real client

neekolas:
## Summary

- Creates a stateful client struct in `grpc_api_helpers` that re-uses a single connection for all API calls
- Removes any use of JSON in the API boundaries and replaces with a mix of [transparent](https://chinedufn.github.io/swift-bridge/bridge-module/transparent-types/structs/index.html) and [opaque](https://chinedufn.github.io/swift-bridge/bridge-module/opaque-types/index.html) structs for request and response types.
- Wraps and exposes the Client and Subscription types to Swift so they can be used as regular classes

## Notes
I've looked through the most popular repos that have `swift-bridge` as a dependency and most are using passed structs and opaque types. That gives me hope that the memory management of shared data is manageable, although I suspect we are going to run into issues.

## Up next
- Integrate this into our Swift library and test end to end in `xmtp-ios`
- Integrate with @jazzz and @michaelx11's TLS work to allow for TLS clients


bindings/xmtp_rust_swift/Package.swift
```diff
@@ -16,5 +16,6 @@ let package = Package(
             .binaryTarget(
                 name: "XMTPRustSwift",
                 path: "XMTPRustSwift.xcframework"
+            ),
```
michaelx11, line 19: Oh wow, I forgot that this Package.swift existed 🤦‍♂️ good catch

bindings/xmtp_rust_swift/src/types.rs
```diff
@@ -0,0 +1,122 @@
+use xmtp_proto::xmtp::message_api::v1::{
+    cursor as cursor_proto, Cursor, Envelope, IndexCursor, PagingInfo,
+    QueryResponse as ProtoQueryResponse, SortDirection,
+};
+
+impl From<crate::ffi::SortDirection> for SortDirection {
```
neekolas, line 6: This was super annoying, but Copilot was able to write most of these on the first shot.

crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -1,143 +1,95 @@
...
-        .await
-        .map_err(|e| format!("{}", e))?;
-    // Response is a v1::QueryResponse protobuf message, which we need to serialize to JSON
-    let json = serde_json::to_string(&response).map_err(|e| format!("{}", e))?;
-    Ok(json)
-}
+impl Client {
+    pub async fn create(host: String) -> Result<Self, tonic::Status> {
+        let host = host.to_string();
+        let channel = Channel::from_shared(host)
```
michaelx11, line 17: This doesn't create a TLS client yet right?

bindings/xmtp_rust_swift/src/lib.rs
```diff
@@ -1,155 +1,219 @@
...
-        Ok(json) => ffi::ResponseJson {
-            error: "".to_string(),
-            json,
-        },
-        Err(e) => ffi::ResponseJson {
-            error: e,
-            json: "".to_string(),
-        },
+    extern "Rust" {
+        type QueryResponse;
```
michaelx11, line 41: These types are such a huge improvement

michaelx11 (commented):
Looking really good! It doesn't have the TLS stuff integrated yet right?

crates/xmtp-networking/src/grpc_api_helper.rs
neekolas: Not yet

michaelx11 (approved):
LGTM - I'm going to rewrite xmtp-rust-swift to use the new bindings and test e2e

neekolas:
I'm gonna merge this as-is and take a stab at TLS

neekolas:
> I'm realizing the ApiService wrapper within xmtp-rust-swift may not make sense for these new bindings. Instead, xmtp-ios should interface directly with the bindings.

Ya. I think that's the long-term goal. Not sure how it would work with the package steps, type conversions, etc. But I hope we can get there.

michaelx11:
@neekolas FYI some stuff that I found out from trying to build in Swift land. We can look into it in a subsequent PR. I think they summarize as:

1) Generated code doesn't seem to like Strings as errors because the Swift representation of a Rust Result expects something that conforms to `any Error`. I wonder if a Rust `std::error` would fit the bill.
```
           let wrapper = Unmanaged<CbWrapper$RustClient$publish>.fromOpaque(cbWrapperPtr!).takeRetainedValue()
            if rustFnRetVal.is_ok {
                wrapper.cb(.success(()(ptr: rustFnRetVal.ok_or_err!))) // <--- error here, doesn't convert void return correctly
            } else {
                wrapper.cb(.failure(RustString(ptr: rustFnRetVal.ok_or_err!))) // <--- error here, expects any Error instead of RustString
            }
```
2) Envelope isn't Vectorizable? Not sure about what's going on with this one
3) Seems like `            self: &mut RustClient,` in publish isn't converted correctly?

Screenshot 2023-04-19 at 12 36 07 PM

michaelx11:
Re: 1) from above, looks like we might have to annotate in Swift: https://github.com/chinedufn/swift-bridge/issues/150

neekolas:
@michaelx11 oof. Should have gone through the Swift imports earlier. 

Let's try the Rust errors. Worst case scenario we can make an error type. Or really worst case scenario instead of using Result we return objects with both results and errors. But Result conversion is nice because it just throws in Swift. 

I'll take a look at how to make Envelope vectorizable.

michaelx11:
I'm debugging now too. I'm wondering if I'm fundamentally misunderstanding `swift-bridge`. Do we even need to bundle the generated `xmtp_rust_swift.swift` if we're consuming from Swift or is the XCFramework enough?

bindings/xmtp_rust_swift/src/types.rs
jazzz: 🤖 🙏

PR #62, opened 2023-04-20
TLS Support

neekolas:
## Summary

- Adds `is_secure` option to client creation and uses @jazzz and @michaelx11's code to implement TLS
- Adds a new TLS test against the dev network

## Notes

I do not have the Rust generics chops to figure out how to merge the `MessageApiClient` types into one, so I settled on an `enum` for both the `Client` and the `Conn`.

Ideally these would be a single value, and given that `MessageApiClient` is generic itself this feels plausible. But this will do for now.

### DIFF
--- a/crates/xmtp-networking/src/grpc_api_helper.rs
+++ b/crates/xmtp-networking/src/grpc_api_helper.rs
@@ -1,9 +1,11 @@
+use http_body::combinators::UnsyncBoxBody;
 use hyper::{client::HttpConnector, Uri};
 use hyper_rustls::HttpsConnector;
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
 use tokio::sync::oneshot;
 use tokio_rustls::rustls::{ClientConfig, OwnedTrustAnchor, RootCertStore};
+use tonic::Status;
 use tonic::{metadata::MetadataValue, transport::Channel, Request, Streaming};
 use xmtp_proto::xmtp::message_api::v1::{
     message_api_client::MessageApiClient, Envelope, PagingInfo, PublishRequest, PublishResponse,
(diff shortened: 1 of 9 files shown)


crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -35,23 +117,46 @@ impl Client {
...
     }
 
     pub async fn subscribe(&mut self, topics: Vec<String>) -> Result<Subscription, tonic::Status> {
         let request = SubscribeRequest {
             content_topics: topics,
         };
-        let stream = self.client.subscribe(request).await.unwrap().into_inner();
+        let stream = match &self.client {
+            InnerApiClient::Plain(c) => c
+                .clone()
```
neekolas, line 148: I'm doing a lot more cloning than I would like, but I get compiler errors without it.

crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -35,23 +117,46 @@ impl Client {
...
-                Ok(req)
-            });
-
-        return client
-            .publish(PublishRequest { envelopes })
-            .await
-            .map(|r| r.into_inner());
+        match &mut self.conn {
+            InnerApiConnection::Plain(c) => {
+                return MessageApiClient::with_interceptor(c, move |mut req: Request<()>| {
```
michaelx11, line 122: You already figured out the interceptor approach, but I struggled with `with_interceptor` here. It didn't want to take the client that I was passing in. I had a realization that we can modify headers on the tonic request directly:
```
        let mut tonic_request = Request::new(request);
        // Insert the auth token into the request
        tonic_request.metadata_mut().insert("authorization", token);
        non_tls_client.publish(tonic_request).await
```

The longer story is the MessageApiClient has two components, a connector (TCP connection) and a part that handles the Body, in this case a Request. Rather than muck with an interceptor to modify the body, we can just modify it ourselves. Took me embarrassingly long to figure that out.

michaelx11 (approved):
LGTM - I'm impressed you finagled the TLS client to work with `with_interceptor`

crates/xmtp-networking/src/grpc_api_helper.rs
neekolas: https://github.com/xmtp/libxmtp/pull/62/commits/d9c994de9cd6746d467341bd769dd2acf159871a

michaelx11: nice!

PR #63, opened 2023-04-20
Run all tests on PRs

neekolas:
## Summary

- Updates the Github workflows to run _all_ the tests in CI. Takes advantage of our workspace `Cargo.toml` to get most of our tests in one shot, and then separately runs the `xmtp_rust_swift` tests which have an unconnected `Cargo.toml`
- Simplifies lint job

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,20 +11,14 @@ jobs:
       - uses: actions/checkout@v1
       - run: rustup component add clippy
       - uses: actions-rs/clippy-check@v1
-        name: xmtp-keystore
+        name: root
         with:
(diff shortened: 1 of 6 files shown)


michaelx11 (approved):
Ship it! I think xmtp-rust-swift might fail cause I have hardcoded local node requirements or something. We can nuke those tests or change them up for now.

neekolas:
> I think xmtp-rust-swift might fail cause I have hardcoded local node requirements or something. We can nuke those tests or change them up for now.

I've also hardcoded local node requirements. We'll figure out a better configuration management setup. But for now, I just spin up a local node using the `docker-compose.yml`

PR #65, opened 2023-04-20
feat: added hashing and k256 utilities for iOS to consume

michaelx11:
## Overview

This adds the functionality that powered https://github.com/xmtp/xmtp-ios/releases/tag/0.1.3-beta0 but without hooking it up to the `bindings/xmtp-rust-swift` crate just yet. This will give us some time to finalize the gRPC side and make sure it plays nicely with Swift.

## Changes

- hashes and k256_helpe Rust files
- also included include/Generated/.swift files in the .gitignore to avoid churn, those files shouldn't be there anyways

## Test Plan

- All unit tests that were also included in the 0.1.3-beta0 prelease Cocoapod, so functionality should be the same


crates/corecrypto/src/hashes.rs
```diff
@@ -0,0 +1,21 @@
...
+use sha3::Keccak256;
+
+pub fn sha256(data: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha256::new();
+    hasher.update(data);
+    let result = hasher.finalize();
+    let mut digest = [0u8; 32];
+    digest.copy_from_slice(&result);
+    digest
+}
```
nakajima, line 12: We actually don't need this exposed in swift because the built-in CryptoKit stuff provides a `SHA256` function, unless there's something special about how we're doing it here.

nakajima (approved):
Can't super speak to the rust stuff but left one small note. Looks good to me though!

crates/corecrypto/src/hashes.rs
michaelx11: Nope, nothing special in this sha256! I can remove the function.

crates/corecrypto/src/lib.rs
```diff
@@ -204,4 +216,135 @@ mod tests {
...
+            0x9d, 0x61, 0xb1, 0xde,
+        ];
+        let public_key = k256_helper::get_public_key(secret.as_slice()).unwrap();
+        // Assert 65 bytes, first is 0x04
+        assert_eq!(public_key.len(), 65);
+        assert_eq!(public_key[0], 0x04);
+    }
+
+    #[test]
+    fn test_public_key_from_private() {
```
neekolas, line 260: Appreciate these thorough tests with real external data

crates/corecrypto/src/k256_helper.rs
```diff
@@ -0,0 +1,101 @@
...
+    message: &[u8],
+    signature: &[u8],
+) -> Result<Vec<u8>, String> {
+    // Recovery id is the last byte of the signature, split signature into 64:1
+    let (signature, recovery_id) = signature.split_at(64);
+    let recovery_id = recovery_id[0];
+    let recovery_id = RecoveryId::try_from(recovery_id).map_err(|e| e.to_string())?;
+    let signature = Signature::try_from(signature).map_err(|e| e.to_string())?;
+    // Create a pre-digested FixedOutput digest
+    let digest = Keccak256::new().chain(message);
```
neekolas, line 87: Just out of curiosity, what's the difference between this and the `keccak256` function in `hashes.rs`?

neekolas (approved):

crates/corecrypto/src/k256_helper.rs
michaelx11: Good question! I struggled with this for a while but the recovery method signature cannot take a &[u8] slice. It specifically wants marker traits to denote which Hash function was used. This prevented me from implementing a generic `32 byte scalar + signature => curve math and recovered key` method like iOS wanted originally.


```
error[E0277]: the trait bound `[u8; 32]: FixedOutput` is not satisfied
   --> crates/corecrypto/src/k256_helper.rs:89:59
    |
89  |     let recovered_key = VerifyingKey::recover_from_digest(digest, &signature, recovery_id)
    |                         --------------------------------- ^^^^^^ the trait `FixedOutput` is not implemented for `[u8; 32]`
    |                         |
    |                         required by a bound introduced by this call
    |
    = help: the following other types implement trait `FixedOutput`:
              CoreWrapper<T>
              SimpleHmac<D>
    = note: required for `[u8; 32]` to implement `Digest`
note: required by a bound in `ecdsa::recovery::<impl ecdsa::VerifyingKey<C>>::recover_from_digest`
   --> /Users/michaelx/.cargo/registry/src/github.com-1ecc6299db9ec823/ecdsa-0.15.1/src/recovery.rs:242:12
    |
242 |         D: Digest,
    |            ^^^^^^ required by this bound in `ecdsa::recovery::<impl ecdsa::VerifyingKey<C>>::recover_from_digest`

error[E0277]: the trait bound `[u8; 32]: HashMarker` is not satisfied
   --> crates/corecrypto/src/k256_helper.rs:89:59
    |
89  |     let recovered_key = VerifyingKey::recover_from_digest(digest, &signature, recovery_id)
    |                         --------------------------------- ^^^^^^ the trait `HashMarker` is not implemented for `[u8; 32]`
    |                         |
    |                         required by a bound introduced by this call
    |
    = help: the following other types implement trait `HashMarker`:
              CShake128Core
              CShake256Core
              CoreWrapper<T>
              CtVariableCoreWrapper<T, OutSize, O>
              Keccak224Core
              Keccak256Core
              Keccak256FullCore
              Keccak384Core
            and 12 others
    = note: required for `[u8; 32]` to implement `Digest`
```

bindings/xmtp_rust_swift/.gitignore
```diff
@@ -13,4 +13,8 @@ bundle.zip
 /target
 /Cargo.lock
 .build
-./Generated
\ No newline at end of file
+/Generated
+
+# Ignore any swift files in include/Generated
+/include/Generated/SwiftBridgeCore.swift
```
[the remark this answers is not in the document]
michaelx11: Yup, that's definitely a more robust solution

PR #67, opened 2023-04-20
Throw errors on reading closed subscriptions

neekolas:
## Summary

- Running `get_messages` on a closed subscription will now throw an error.
- Subscriptions can be closed by either connection failure or manually closing the stream

## Notes

- Currently we make no distinction between forcibly closed Subscriptions (`close_stream`) and closure due to connection error. We also do not do retries on connection error. I kinda think retries should be the responsibility of the caller and we should leave this level simple and unopinionated on retry logic.

### DIFF
--- a/crates/xmtp-networking/src/grpc_api_helper.rs
+++ b/crates/xmtp-networking/src/grpc_api_helper.rs
@@ -2,6 +2,7 @@ use http_body::combinators::UnsyncBoxBody;
 use hyper::{client::HttpConnector, Uri};
 use hyper_rustls::HttpsConnector;
 use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use tokio::sync::oneshot;
 use tokio_rustls::rustls::{ClientConfig, OwnedTrustAnchor, RootCertStore};
@@ -177,6 +178,7 @@ impl Client {
 pub struct Subscription {
     pending: Arc<Mutex<Vec<Envelope>>>,
     close_sender: Option<oneshot::Sender<()>>,
+    closed: Arc<AtomicBool>,
 }
 
 impl Subscription {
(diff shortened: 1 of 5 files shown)


crates/xmtp-networking/src/grpc_api_helper.rs
```diff
@@ -177,13 +178,16 @@ impl Client {
...
     close_sender: Option<oneshot::Sender<()>>,
+    closed: Arc<AtomicBool>,
 }
 
 impl Subscription {
     pub async fn start(stream: Streaming<Envelope>) -> Self {
         let pending = Arc::new(Mutex::new(Vec::new()));
         let pending_clone = pending.clone();
         let (close_sender, close_receiver) = oneshot::channel::<()>();
+        let closed = Arc::new(AtomicBool::new(false));
```
michaelx11, line 189: Took a bit of time to read up on AtomicBool. From what I can tell, Arc<AtomicBool> + clone is the right approach. You have my support :)

crates/xmtp-networking/src/lib.rs
```diff
@@ -70,16 +71,20 @@ mod tests {
                 .unwrap();
 
             // Sleep to give the response time to come back
-            std::thread::sleep(std::time::Duration::from_millis(100));
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
michaelx11, line 74: How did this differ from the std::thread::sleep?

michaelx11 (approved):
LGTM - one question for my understanding

crates/xmtp-networking/src/lib.rs
neekolas: I realized in testing that this was also blocking the reading of the stream (everything goes to sleep), whereas this will yield and let other async processes continue working.

michaelx11: Gotcha!

PR #68, opened 2023-04-21
feat: add crypto bindings, fix and test e2e with Swift

michaelx11:
## Overview

This adds the crypto functionality from #65 and exposes it via new bindings. I also changed the bindings to handle errors mentioned here: https://github.com/xmtp/libxmtp/pull/61#issuecomment-1515280324

1) Updated `make push-to-swift` process to inject `extension` code in the generated Swift so `Result<T, String>` in Rust gets transformed into descriptive errors when thrown on the Swift side
2) Had to replace the `()` return type with String. This one was quite weird, bool or UInt32 wouldn't work either - it needed to be a type that had a constructor generated on the Swift side
3) Made Envelope an opaque type with getters, otherwise it can't be Vectorizable.

## API Changes for Swift Package / Cocoapod
- `ApiService` and `CoreCrypto` are going away. There is no more intermediary Swift code. Consumers will call directly into XMTPRust.
- No more JSON serialization across the binding boundary (thanks to @neekolas's hard work)
- Long-lived client instance [example](https://github.com/xmtp/xmtp-ios/blob/michaelx-test_no_json_bindings/Sources/XMTP/Client.swift#L77)

The end result should be more performant and easier to maintain (no Swift code needs to be hand maintained).

## Test Plan

Consumed the XMTPRustSwift.xcframework in xmtp-rust-swift and then in xmtp-ios and checked all tests passed except `subscribe` - I haven't implemented the necessary AsyncThrowableStream wrapper code in `ApiClient` yet.

NOTE: I've added sha256 back @nakajima it just felt wrong without it.


nakajima (approved):
> NOTE: I've added sha256 back @nakajima it just felt wrong without it.

live your truth!

neekolas:
Thank you for cleaning up my mess with the Swift bindings

bindings/xmtp_rust_swift/Generated/xmtp_rust_swift/xmtp_rust_swift.swift
```diff
@@ -1,3 +1,6 @@
+public func create_envelope<GenericIntoRustString: IntoRustString>(_ topic: GenericIntoRustString, _ sender_time_ns: UInt64, _ payload: RustVec<UInt8>) -> Envelope {
```
neekolas, line 1: Think you might need to merge main into your branch to get the .gitignore changes

bindings/xmtp_rust_swift/copy_xcframework_to_swift.sh
```diff
@@ -15,13 +15,37 @@ cp -R "XMTPRustSwift.xcframework" "$REPOPATH"
...
 # Here's the ultra-hack, we need to inject "import XMTPRustSwift" into the top of the Swift files
 # before copying them over so they'll get the headers we moved to "include/Generated".
+#
+# Moreover, we inject the following lines to allow RustStrings to be interpreted as NSErrors and automatically
+# displayed in the debugger:
+# https://github.com/chinedufn/swift-bridge/issues/150
+#
+# extension RustString: @unchecked Sendable {}
+#
+# extension RustString: LocalizedError {
```
neekolas, line 29: 🤯 

This is going on my list of things we should make `swift-bridge` issues for. Seems to be actively maintained, but just not a lot of users.

neekolas (approved):
Looks great to me

bindings/xmtp_rust_swift/copy_xcframework_to_swift.sh
michaelx11: I feel like there has to be a better way or I'm misconfiguring something, but this beats their suggestion of just empty `NSError` extension which satisfies the Swift compiler but provides no useful output on error.

PR #72, opened 2023-04-24
Round-trip in-memory persistence through WASM

richardhuaaa:
# Description
Routing the in-memory persistence through WASM first, before implementing a LocalStorage binding. Add `?w=1` to the URL to ignore whitespace changes.

We define a `Client` JS class that wraps the equivalent Rust object, intending for the interface to resemble that used by the JS SDK. We use a similar pattern to the Voodoo prototype here, where we hold Rust objects in memory in Rust-land and refer to them by ID's in JS-land. There could be some argument for moving the wrapper class directly into the JS SDK rather than implementing it here in the wasm binding library.

Dynamic traits such as the Persistence trait in this PR are not thread-safe by default. I am assuming that the entry-point into our WASM library is single-threaded, so I added the `Sync` and `Send` annotations to the `Persistence` trait to disable the compiler error about thread safety - feedback definitely welcome here.

# Test plan
`cargo test` in `xmtpv3/`
`npm run test` in `xmtpv3/bindings/wasm`

### DIFF
--- a/xmtpv3/bindings/wasm/crate/src/lib.rs
+++ b/xmtpv3/bindings/wasm/crate/src/lib.rs
@@ -1,7 +1,7 @@
-use std::{sync::Mutex};
+use std::sync::Mutex;
 
+use libxmtp_core::client::Client;
 use libxmtp_core::persistence::InMemoryPersistence;
-use libxmtp_core::{client::Client};
 use wasm_bindgen::prelude::*;
 
 static CLIENT_LIST: Mutex<Vec<Client<InMemoryPersistence>>> = Mutex::new(Vec::new());
@@ -14,14 +14,21 @@ pub fn client_create() -> usize {
 }
 
 #[wasm_bindgen]
-pub fn client_write_to_persistence(client_id: usize, key: String, value: &[u8]) -> Result<(), String> {
+pub fn client_write_to_persistence(
+    client_id: usize,
+    key: String,
+    value: &[u8],
+) -> Result<(), String> {
     let mut clients = CLIENT_LIST.lock().unwrap();
     let client = clients.get_mut(client_id).expect("Client not found");
     client.write_to_persistence(key, value)
 }
 
 #[wasm_bindgen]
-pub fn client_read_from_persistence(client_id: usize, key: String) -> Result<Option<Vec<u8>>, String> {
+pub fn client_read_from_persistence(
+    client_id: usize,
+    key: String,
+) -> Result<Option<Vec<u8>>, String> {
     let mut clients = CLIENT_LIST.lock().unwrap();
     let client = clients.get_mut(client_id).expect("Client not found");
     client.read_from_persistence(key)
--- a/xmtpv3/crates/libxmtp-core/src/client.rs
+++ b/xmtpv3/crates/libxmtp-core/src/client.rs
@@ -8,7 +8,7 @@ where
 }
 
 impl<P: Persistence> Client<P> {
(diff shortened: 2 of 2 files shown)


michaelx11 (commented):
I had to do some research/thinking but our management of Persistence should be simplified. IMO we should only vary rarely go to heap (e.g. `Box`). Here's an example: https://github.com/xmtp/libxmtp/pull/81

The memory hierarchy here is still simple (Client instance "owns" Persistence implementation struct), we can save Heap logic for when we need it.

xmtpv3/bindings/wasm/crate/src/lib.rs
```diff
@@ -1,6 +1,28 @@
...
+    clients.len() - 1
+}
+
+#[wasm_bindgen]
+pub fn client_write_to_persistence(client_id: usize, key: String, value: &[u8]) -> Result<(), String> {
+    let mut clients = CLIENT_LIST.lock().unwrap();
+    let client = clients.get_mut(client_id).expect("Client not found");
+    client.write_to_persistence(key, value)
+}
+
```
jazzz, line 22: This file doesn't look like it was formatted using rustfmt. Might want to add it to your editor to minimize headaches

xmtpv3/crates/libxmtp-core/src/client.rs
```diff
@@ -1,13 +1,13 @@
 use crate::persistence::Persistence;
 
-pub struct Client<P: Persistence> {
-    persistence: P,
+pub struct Client {
+    persistence: Box<dyn Persistence>,
```
jazzz, line 4: For my own knowledge, whats the driver over the runtime determined type here rather the caller choosing it at compile time via generic?

xmtpv3/crates/libxmtp-core/src/client.rs
```diff
@@ -1,13 +1,13 @@
...
 }
 
-impl<P: Persistence> Client<P> {
+impl Client {
     pub fn new(
-        persistence: P,
-    ) -> Client<P> {
+        persistence: Box<dyn Persistence>,
+    ) -> Client {
```
jazzz, line 10: nit: You can use `Self` in the return type rather than explicitly returning client.

jazzz (approved):
👍

xmtpv3/crates/libxmtp-core/src/client.rs
michaelx11: I think we should hold off on merging until we sanity-check the Box usage here. I have a PR (linked above) that just transfers the ownership of the underlying Persistence-implementing instance to the client. For this specific usage, I believe the ownership approach is simpler and avoids us having that `Send + Sync` race condition.

Thoughts @jazzz ?

michaelx11: I had to write up some code to see if I'm crazy, and at least in the local_storage case we don't (unit test doesn't pass but example browser HTML test works): https://github.com/xmtp/libxmtp/compare/michaelx-local_storage_example?expand=1#diff-f3a6a7b2860bea48f5b4267aa7a8139b8bc48126dcb45424abd91d830cd41f90R26

For the iOS/Android cases, at least Mozilla Uniffi (one possible binding generator) does account for `Send + Sync` already in their callback interface: https://mozilla.github.io/uniffi-rs/udl/callback_interfaces.html#1-define-a-rust-trait because it can be useful. I can foresee us using a global Mutex to manage SQLite

PR #81, opened 2023-04-24
example: wasm persistence traitbound ownership

michaelx11:
## Overview

Example rework for #72 

Our strategy for managing Persistence structs right now involves passing static lifetime (live as long as the program) references via (dyn X) around our program.

This is what caused the need for `Send + Sync`

An alternative approach is to just make the Client "own" the specific Persistence object. We basically already do this, because we're initializing a new InMemoryPersistence for each client. However, right now we're putting that memory on the Heap (so it lives statically-ish) and then throwing references around.

By using trait bounds, we can say the `Client` takes a specific Persistence type and then just pass ownership of that Persistence concrete object directly to the Client. This tells Rust that "the persistence's object's memory is now handled by Client" which is really the most simple/correct variation.

## PR steps

This is just a draft example, feel free to close after looking at it for your own reference @richardhuaaa

### DIFF
--- a/xmtpv3/crates/libxmtp-core/src/client.rs
+++ b/xmtpv3/crates/libxmtp-core/src/client.rs
@@ -1,18 +1,15 @@
 use crate::persistence::Persistence;
 
 pub struct Client<P>
-where P: Persistence
+where
+    P: Persistence,
 {
     persistence: P,
 }
 
 impl<P: Persistence> Client<P> {
-    pub fn new(
-        persistence: P,
-    ) -> Client<P> {
-        Client {
-            persistence,
-        }
+    pub fn new(persistence: P) -> Client<P> {
+        Client { persistence }
     }
 
     pub fn write_to_persistence(&mut self, s: String, b: &[u8]) -> Result<(), String> {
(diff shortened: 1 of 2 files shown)


xmtpv3/bindings/wasm/crate/src/lib.rs
```diff
@@ -4,12 +4,12 @@ use libxmtp_core::persistence::InMemoryPersistence;
...
 use wasm_bindgen::prelude::*;
 
-static CLIENT_LIST: Mutex<Vec<Client>> = Mutex::new(Vec::new());
+static CLIENT_LIST: Mutex<Vec<Client<InMemoryPersistence>>> = Mutex::new(Vec::new());
 
 #[wasm_bindgen]
 pub fn client_create() -> usize {
     let mut clients = CLIENT_LIST.lock().unwrap();
-    clients.push(Client::new(Box::new(InMemoryPersistence::new())));
+    clients.push(Client::new(InMemoryPersistence::new()));
```
jazzz, line 12: Definitely like getting the Persistence object off the heap

PR #82, opened 2023-04-24
feat: can reference any branch and have protos build correctly

michaelx11:
## Overview

Issue: https://github.com/orgs/xmtp/projects/8/views/4?pane=issue&itemId=26452558

This provides an example for being able to reference any xmtp/proto branch we want to target. Since `xmtp-proto` is shared, whatever branch we choose needs to be kept up to date with `xmtp/proto#main`.

## Test Plan

- Confirm that a new test field was added correctly
- Ensure that `cargo build` passes in `xmtp-proto` and in `xmtp-networking` which consumes xmtp-proto

## Open Question

What do you want to call our working branch? Should we utilize `beta` again? Edit: no, we're using `xmtpv3` on xmtp/proto

### DIFF
--- a/crates/xmtp-proto/build.rs
+++ b/crates/xmtp-proto/build.rs
@@ -29,7 +29,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let status = Command::new("buf")
         .arg("generate")
-        .arg("buf.build/xmtp/proto")
+        .arg("https://github.com/xmtp/proto.git#branch=michaelx-rust_v3_protos,subdir=proto")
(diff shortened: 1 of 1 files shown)


michaelx11:
Correction: it seems like the branch may need https://github.com/xmtp/proto/blob/main/proto/buf.yaml - so it has to be based off of recent `main`

jazzz (approved):
🤷

michaelx11:
FYI using `xmtpv3` as the branch: https://github.com/xmtp/proto/tree/xmtpv3

PR #87, opened 2023-04-25
Project Restructure

jazzz:
This PR sanitizes the LibXMTP repo, to follow a more traditional style. 

The only significant change is that modules no longer use Vodozemac and now point to a Fork which has been updated to avoid package conflicts.

Swift bindings have been moved, but I am unable to test them. I tried my best but might need some love.

Guidlines::

```
├── Cargo.toml        <---- Virtual Workspace
├── bindings_swift    <---- Bindings are flattened and given a common naming scheme
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── bindings_wasm     <---- Use snake_case for packages, (package names are auto converted to snake_case in code anyways)
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── examples          <---- Examples at the workspace root level
│   ├── Cargo.toml
│   └── chat.rs
├── utils             <---- Isn't published as the name is not prefixed with 'xmtp_'
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── xmtp              <---- Core Client Lives here
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── xmtp_crypto       <---- Package names match the folder names
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── xmtp_networking
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
└── target
```

PR #88, opened 2023-04-26
feat: generate contact bundle and populate proto

michaelx11:
## Overview

This can wait until the project restructure PR #87 , but provides an initial example of using new protos with vmac. This uses @richardhuaaa 's contact bundle proto structure from the MudPup sprint.

Notably - `tonic` does not build for `wasm`, so I needed to gate `tonic` behind a feature "grpc" inside xmtp-proto. xmtp-networking then relies on that feature. FYI @neekolas so that doesn't catch you by surprise.
## Test Plan

Unit tests pass


xmtpv3/crates/libxmtp-core/src/vmac.rs
```diff
@@ -0,0 +1,90 @@
...
+) -> VmacUnsignedPublicKey {
+    VmacUnsignedPublicKey {
+        created_ns: 0,
+        union: Some(Union::Curve25519(VodozemacCurve25519 {
+            bytes: key.to_bytes().to_vec(),
+        })),
+    }
+}
+
+fn account_linked_key_from_curve25519_public_key(key: Curve25519PublicKey) -> VmacAccountLinkedKey {
```
neekolas, line 27: Wouldn't this be more idiomatic as a `impl from<Curve25519PublicKey> for VmacAccountLinkedKey {`

michaelx11: I might be misinterpreting, but Rust's orphan rules don't allow me to write an Impl where both the trait and type live in different crates. Now that I think about it we maybe can work around it with either a Wrapper or a fake Generic type?

https://stackoverflow.com/q/75765502

michaelx11: This is what the compiler spits at me:

```
error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate
  --> crates/libxmtp-core/src/vmac.rs:27:1
   |
27 | impl From<Curve25519PublicKey> for VmacUnsignedPublicKey {
   | ^^^^^-------------------------^^^^^---------------------
   | |    |                             |
   | |    |                             `VmacUnsignedPublicKey` is not defined in the current crate
   | |    `Curve25519PublicKey` is not defined in the current crate
   | impl doesn't use only types from inside the current crate
   |
   = note: define and implement a trait or new type instead
```

neekolas: I also totally skimmed past your comment above the first function here

michaelx11: Very understandable it's a tiny comment punting a large problem haha

neekolas: Ya. A type wrapper would work, but I'm not sure of the downstream consequences of actually using the wrapped type.

PR #89, opened 2023-04-26
feat: contact bundle with ProtoWrapper From trait

michaelx11:
## Overview

Issue: https://github.com/xmtp/proto/issues/81

Rebuild of #88 - this time with `ProtoWrapper` which allows us to implement `From` traits for conversion in the `vmac_protos.rs` module (happy to rename)

~~**NOTE**: I think the wasm bindings situation is currently borked. `bindings_js` and `bindings_wasm` are one cohesive operation and don't seem to build while separated (`npm test` in bindings_js complains about lack of Cargo.toml). I'll look into this next.~~

GRPC crate feature notes, copied from #88 
```
Notably - tonic does not build for wasm, so I needed to gate tonic behind a feature "grpc" inside xmtp-proto. xmtp-networking then relies on that feature. FYI @neekolas so that doesn't catch you by surprise.
```

## Tests

- Unit tests in `xmtp` core crate pass
- Fix and run bindings_js tests by adding `buf` to the Github action


xmtp/src/vmac_traits.rs
```diff
@@ -0,0 +1,60 @@
+use xmtp_proto::xmtp::v3::message_contents::vmac_unsigned_public_key::{
```
jazzz, line 1: I find this file naming interesting as there appears to be no traits in it, only Impls. 😄

xmtp/src/vmac.rs
```diff
@@ -0,0 +1,54 @@
...
+    };
+    VmacContactBundle {
+        identity_key: Some(identity_key),
+        prekey: Some(fallback_key),
+    }
+}
+
+// Generate an outbound session (Olm Prekey Message) given a VmacContactBundle
+pub fn generate_outbound_session(bundle: VmacContactBundle) -> Vec<u8> {
+    let account = Account::new();
```
jazzz, line 36: How is this function intended to function? 

Is an account generated for every session?

xmtp/src/lib.rs
```diff
@@ -24,4 +29,15 @@ mod tests {
...
+    fn test_can_generate_test_contact_bundle_and_session() {
+        let bundle = generate_test_contact_bundle();
+        assert!(bundle.identity_key.is_some());
+        assert!(bundle.prekey.is_some());
+
+        // Generate an outbound session (Olm Prekey Message) given a VmacContactBundle
+        let session = generate_outbound_session(bundle);
+        assert!(!session.is_empty());
+    }
 }
```
jazzz, line 43: These tests could be placed in the same file as the functioned being tested. 

E.g. https://github.com/xmtp/libxmtp/blob/main/xmtp/src/builder.rs#L44

xmtp/src/vmac_traits.rs
michaelx11: I can rename to proto_helpers or move code back to vmac.rs? I have no strong opinions here

xmtp/src/vmac.rs
michaelx11: I should prefix with `test_` - meant almost as a demo that the proto code isn't just baloney

michaelx11: actually moved all of the functions directly into the new `mod test` block

xmtp/src/lib.rs
michaelx11: did it!

michaelx11:
FYI @neekolas - this provides a paradigm (one of many, probably better ones) that allows us to convert between Vodozemac types and protos. Feel free to open a PR against xmtp/proto#xmtpv3 with a completely redesigned Account proto message!

jazzz (approved):

michaelx11:
@jazzz - the wrapper situation could definitely use your thoughts. I couldn't implement `From` directly between Vodozemac types and protos, because both live outside the crate (Rust orphan rules). I thought of this wrapper approach but there's almost certainly a better way.

michaelx11:
Will ship for now to get the bindings fix (need to install buf) and the `grpc` feature flag situation in main. Very happy to revisit or completely rework the proto wrapper logic, for now it acts as a working placeholder for further protobuf-based development.

Issue #73, opened 2023-04-24
Generate and persist account keys


richardhuaaa:
We can generate/store the vmac keys as-is.
If there are already keys on disk, we should read those into memory as part of initialization.

Wallet signing the identity key and possibly the fallback key is left as a separate task for simplicity.

neekolas:
I'm marking this as closed, and we can create a new ticket for migrating to proto. But that doesn't block anything for now.

PR #90, opened 2023-04-26
add builder pattern for Client

jazzz:
# Problem

Adding new parameters to the `Client` struct results in having to update all existing instantiations (tests and external apps) - Making development/future usability painful. 

# Solution
Implement the builder pattern for Client. This allows developers an mechanism to use a sensible default implementation, while also customizing the object as needed. 

Advantages:
- Insulates developers from changes to the client Struct
- Allows customization of the client object, as opposed to only implementing Default for Client 
- Allows specific implementations for test clients etc.

### DIFF
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -41,18 +41,18 @@ where
     }
 }
 
-impl ClientBuilder<InMemoryPersistence> {
-    pub fn new_test() -> Self {
-        Self::new().persistence(InMemoryPersistence::new())
-    }
-}
-
 #[cfg(test)]
 mod tests {
 
     use crate::{client::Network, persistence::InMemoryPersistence};
 
     use super::ClientBuilder;
+
+    impl ClientBuilder<InMemoryPersistence> {
+        pub fn new_test() -> Self {
+            Self::new().persistence(InMemoryPersistence::new())
+        }
+    }
 
     #[test]
     fn builder_test() {
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -1,6 +1,6 @@
 use crate::{
     client::{Client, Network},
-    persistence::{InMemoryPersistence, Persistence},
+    persistence::Persistence,
 };
 
 #[derive(Default)]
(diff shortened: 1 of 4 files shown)


xmtp/src/builder.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    pub fn persistence(mut self, persistence: P) -> Self {
+        self.persistence = Some(persistence);
+        self
+    }
+
+    pub fn build(self) -> Client<P> {
+        Client {
+            network: self.network,
+            persistence: self.persistence.expect("A persistence engine must be set"),
```
jazzz, line 39: Theres an open question of whether the persistence engine should initialized in New, which allows for Compile time check, or if it should be lazily initialized on build which allows more flexibility. 

went with flexibility for now

jazzz: Adding Default trait to Persistence would allow the builder to instantiate on build and avoid a panic, but seems heavy weight at this time.

michaelx11:
~~Changes look like the right direction, but might break something in bindings_wasm (pointed out by linter on PR)~~ (your commit just now fixed this). I think the bindings_wasm/bindings_js relationship seems broken currently unless I'm missing something. `npm test` in bindings_js complains about missing Cargo.toml.

xmtp/src/builder.rs
michaelx11: Default for persistence would just be choosing an engine like `InMemoryPersistence` right? It makes sense to just panic to me. The consumer should be forced to acknowledge the persistence method they want to use.

jazzz: The Default trait would provide a function called default which accepts no params. Adding this to the persistence trait would force every struct who implements Persistence to have a default initialization. 

For InMem likely makes sense, however for more complicated Persistence implementation a default might not make sense (a hypothetical HttpPersistence would minimally need a url to be useful). 

Ultimately it would be great to throw a compile time error if a value is missing, but I Figure it doesn't need to be decided right meow...

neekolas (approved):
I've also been thinking about this, since I'm going to be attaching the Account to the client for https://github.com/xmtp/libxmtp/issues/73. 

Makes sense to me

michaelx11:
FYI #93 should fix any `bindings_js/bindings_wasm` wonkiness

PR #91, opened 2023-04-26
fix: allow multiple getter calls on QueryResponse without transferring ownership

michaelx11:
## Overview

When we try to call both `.envelopes()` and `.paging_info()` on a Rust QueryResponse from the Swift side, we get a "memory already freed" exception. The semantics of the getters was transferring ownership on first call.

This change makes it so the getters return a clone (not ideal) - maybe we can look into returning a slice or a Smart Pointer-like object in a future optimization.

## Test Plan

- ios side no longer crashes when calling `.paging_info()` after a `.envelopes()` call

### DIFF
--- a/bindings_swift/src/types.rs
+++ b/bindings_swift/src/types.rs
@@ -55,7 +55,7 @@ impl From<PagingInfo> for crate::ffi::PagingInfo {
             },
             None => None,
         };
-        
+
         crate::ffi::PagingInfo {
             limit: paging_info.limit,
             direction: crate::ffi::SortDirection::from(
@@ -93,12 +93,13 @@ pub struct QueryResponse {
 }
 
 impl QueryResponse {
-    pub fn envelopes(self) -> Vec<crate::Envelope> {
-        self._envelopes
+    pub fn envelopes(&self) -> Vec<crate::Envelope> {
+        self._envelopes.clone()
     }
 
-    pub fn paging_info(self) -> Option<crate::ffi::PagingInfo> {
-        self._paging_info
+    pub fn paging_info(&self) -> Option<crate::ffi::PagingInfo> {
+        self._paging_info.as_ref()?;
+        self._paging_info.clone()
     }
 }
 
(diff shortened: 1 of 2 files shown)


neekolas (approved):
Nice fix. And it's good that we are finding these allocation errors and have a path to fix more if needed.

michaelx11:
Yeah I want to really hammer on this integration before we ship it to partners e.g. do a whole slew of automated testing.

bindings_swift/src/types.rs
```diff
@@ -93,12 +93,13 @@ pub struct QueryResponse {
 }
 
 impl QueryResponse {
-    pub fn envelopes(self) -> Vec<crate::Envelope> {
-        self._envelopes
+    pub fn envelopes(&self) -> Vec<crate::Envelope> {
+        self._envelopes.clone()
```
jazzz, line 97: I'm curious about the intended ownership here. 

Who owns the QueryResponse? is passing a ref to the Vec not an option?

Want to make sure were only cloning when needed

michaelx11: It's a good question, AFAICT the issue here boils down to how `swift-bridge` emulates ownership on the Swift side.

See this code: https://github.com/xmtp/xmtp-rust-swift/blob/main/Sources/XMTPRust/xmtp_rust_swift.swift#L406 which looks like:
```
extension QueryResponse {
    public func envelopes() -> RustVec<Envelope> {
        RustVec(ptr: __swift_bridge__$QueryResponse$envelopes({isOwned = false; return ptr;}()))
    }

    public func paging_info() -> Optional<PagingInfo> {
        __swift_bridge__$QueryResponse$paging_info({isOwned = false; return ptr;}()).intoSwiftRepr()
    }
}
```

It appears that on either call, ownership is "given up" via an explicitly tracked variable which tells Swift it's okay to free the memory on `deinit`. I don't pretend to fully understand why the `deinit` is invoked while the `response` is still within scope on the Swift side.

PR #92, opened 2023-04-26
Dedupe test actions

jazzz:
The project used to have two separate workspaces (Pre: #87) 

Now that the entire workspace tests are run, it's not needed to run the core crate independently

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,9 +34,3 @@ jobs:
         with:
           command: test
           args: --manifest-path bindings_swift/Cargo.toml
-
-      - name: Run cargo test on xmtpv3
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --manifest-path xmtp/Cargo.toml


michaelx11:
Looks good, do we need to run tests on `bindings_wasm` or any other excluded members separately? We can just replace `xmtp` with those excluded members

jazzz:
@michaelx11 bindings_wasm is part of the workspace so those tests are executed at the worspace level. 

The only excluded members are _swift and _js. bindings_swift are ran separately, and js doesn't have a crate. So from what I can see all tests are being run

ref: https://github.com/xmtp/libxmtp/blob/main/Cargo.toml

michaelx11:
Got it, we used to have an e2e test for JS that should run `npm test` inside `bindings_js` (which is excluded from the workspace and not covered). I can look into getting that back up separately. RE: #93

michaelx11:
I'll basically be adding: https://github.com/xmtp/libxmtp/blob/vodozemac_dev/.github/workflows/wasm_npm_test.yaml back in

PR #95, opened 2023-04-26
Get or create account as part of initialization

neekolas:
## Summary

- First pass at getting or creating a VmacAccount as part of the client building process. 
- Makes `Account` serializable, by wrapping in a struct and using the pickled value to JSON serialize
- Includes the account on the `Client` struct

## Ticket
https://github.com/xmtp/libxmtp/issues/73

## Notes
- I haven't included any state on whether the key has been uploaded or not. Going to handle that in a subsequent PR, but it should be pretty straightforward to modify the serialize method to include more data as needed.
- I am currently using JSON as the serialization format. We may ultimately want protos even for some of these data types that don't go to the network, just because it enforces some type safety and is faster to read/write. But we can handle that separately if we even care.
- We are going to want to prefix the storage key for the account with a wallet address, so that one persistence can handle multiple accounts. But we currently don't know that address, so I am defaulting to `unknown`

### DIFF
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -55,7 +55,7 @@ where
                 let account = VmacAccount::generate();
                 let data = serde_json::to_string(&account).map_err(|e| format!("{}", e))?;
 
-                persistence.write(key.clone(), data.as_bytes())?;
+                persistence.write(key, data.as_bytes())?;
 
                 self.account = Some(account)
             }
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -108,27 +108,27 @@ mod tests {
             .is_empty())
     }
 
-    #[test]
-    fn persistence_test() {
-        let persistence = InMemoryPersistence::new();
-        let client_a = ClientBuilder::new()
-            .persistence(persistence)
-            .account("foo".to_string())
-            .unwrap()
-            .build();
+    // #[test]
+    // fn persistence_test() {
+    //     let persistence = InMemoryPersistence::new();
+    //     let client_a = ClientBuilder::new()
+    //         .persistence(persistence)
+    //         .account("foo".to_string())
+    //         .unwrap()
+    //         .build();
 
-        let client_b = ClientBuilder::new()
-            .persistence(persistence)
-            .account("foo".to_string())
-            .unwrap()
-            .build();
+    //     let client_b = ClientBuilder::new()
+    //         .persistence(persistence)
+    //         .account("foo".to_string())
+    //         .unwrap()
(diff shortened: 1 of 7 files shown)


xmtp/src/builder.rs
```diff
@@ -33,14 +36,47 @@ where
         self
     }
 
+    pub fn account(mut self, wallet_address: String) -> Result<Self, String> {
```
neekolas, line 39: Maybe should rename this to `find_or_create_account`?

michaelx11: down for that name!

michaelx11 (approved):
LGTM, we can switch over to protos after #89

michaelx11:
Actually @neekolas immediately after #89, I'm going to rip out most of my code and move it into your `account.rs` file

xmtp/src/builder.rs
```diff
@@ -33,14 +36,47 @@ where
         self
     }
 
+    pub fn find_or_create_account(mut self, wallet_address: String) -> Result<Self, String> {
+        let key = get_account_storage_key(wallet_address);
+        let persistence = self.persistence.as_mut().ok_or_else(|| {
```
jazzz, line 41: super nit: the mut ref persistence is only needed for the write operation in the 2nd match arm. 

Since the object is owned you could only take the &mut if needed which would relax constraints from the borrow checker. Not super applicable in this code, but since mut refs are exclusive, this could cause borrow errors in the future/other cirumstances

xmtp/src/builder.rs
```diff
@@ -33,14 +36,47 @@ where
...
+                persistence.write(key, data.as_bytes())?;
+
+                self.account = Some(account)
+            }
+            Err(e) => return Err(format!("Failed to read from persistence: {}", e)),
+        }
+
+        Ok(self)
+    }
+
```
jazzz, line 66: This seems like it should be on the client not the builder.

Using the builder pattern it shouldn't matter than order the params are built. and the result return type will make chaining need an unwrap. 

Following the pattern, the builder would accept a Option<wallet_address> and then the client would fetch and store.

neekolas: So, the reason I didn't want to put it on the client is because I didn't want the client to have an optional account field. Feels like it makes life easier to have as many fields as possible be required at the time of creation, so we don't have to have a bunch of option checks in the client code. 

But maybe the juice isn't worth the squeeze here.

neekolas: One alternative would be to have the `build()` method of the builder do the work to load the data. Then you still get a clean client.

PR #98, opened 2023-04-27
Clean up builder

neekolas:
## Summary

Small PR addressing some of the comments in https://github.com/xmtp/libxmtp/pull/95.

- Replaces `find_or_create_account` with a simple function to set the wallet_address
- Actually finds or creates the account as part of the `build` function
- Only takes a mutable reference to persistence when needed

This does have the effect of making `.build()` return a `Result<Client<P>, String>` instead of just the client.

### DIFF
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -41,7 +41,7 @@ where
         self
     }
 
-    pub fn find_or_create_account(&mut self) -> Result<VmacAccount, String> {
+    fn find_or_create_account(&mut self) -> Result<VmacAccount, String> {
         let wallet_address = self
             .wallet_address
             .as_ref()
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -71,7 +71,7 @@ where
 
                 Ok(account)
             }
-            Err(e) => return Err(format!("Failed to read from persistence: {}", e)),
+            Err(e) => Err(format!("Failed to read from persistence: {}", e)),
         }
     }
 
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -48,7 +48,7 @@ where
             .ok_or_else(|| "Wallet address must be set before setting the account".to_string())?;
 
         let key = get_account_storage_key(wallet_address.to_string());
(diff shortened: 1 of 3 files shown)


xmtp/src/builder.rs
```diff
@@ -48,28 +58,35 @@ where
                 let data_string = std::str::from_utf8(&data).map_err(|e| format!("{}", e))?;
                 let account: VmacAccount =
                     serde_json::from_str(data_string).map_err(|e| format!("{}", e))?;
-                self.account = Some(account)
+                Ok(account)
```
michaelx11, line 61: In a world with Wallet<>Account associations, the deserialized VmacAccount should be checked against the set wallet address, otherwise it would request a signature (or have some intermediate state like "unlinked")?

Took me a second to grok the reasoning behind setting a wallet address, but it makes sense that we wouldn't set a wallet. We just need to know what wallet this VmacAccount is for or will be for.

michaelx11 (approved):

xmtp/src/builder.rs
```diff
@@ -48,28 +58,35 @@ where
...
             }
             Ok(None) => {
                 let account = VmacAccount::generate();
                 let data = serde_json::to_string(&account).map_err(|e| format!("{}", e))?;
 
-                persistence.write(key, data.as_bytes())?;
+                self.persistence
+                    .as_mut()
+                    .unwrap()
+                    .write(key, data.as_bytes())?;
```
jazzz, line 70: Does the client require that that the Account is in the persistence  object? What happens if the Client is constructed directly outside of the builder?

neekolas: I would think not. The reason we write to persistence is for future client instantiations, but the client shouldn't care since it already has the account.

xmtp/src/builder.rs
neekolas: We also just need to store it somewhere consistent so we can look it up later and re-use on future client instantiations. Wallet address is a good, stable, identifier.

xmtp/src/builder.rs
jazzz: Awesome.

jazzz (approved):
👍

PR #100, opened 2023-05-01
Persistence binding for web using web-sys based localStorage, browser-based unit tests

richardhuaaa:
In the short-term, let's commit to using web-sys for our web binding - this gives us localStorage/indexedDB, fetch, console log, etc. If we need to support node in the future, we can split the wasm crate into one with web-sys and one without.

This PR implements Persistence using web-sys localStorage (potentially substitutable with indexedDB in the future). I started with @michaelx11's [example](f8e9f1aa72d896241641240c62e1e5dbc308b782), tidied it, and made it work with a variety of tests.

Tested via:
1. Example web app - run `run_example.sh` in the `bindings_js/example` directory
2. Unit tests against JS running Chrome in headless mode - run `npm test` in the `bindings_js` directory
3. Unit tests against wasm running Chrome in headless mode - run `./test.sh` in the `bindings_wasm` directory

https://github.com/xmtp/xmtp-js/issues/348


jazzz (approved):
Like this change a lot. Particularly adding the js example -- thats a pattern I think could be used much more in the rest of the packages.  

There are however many panics in this code which need to be avoided at all costs. Using Results and avoiding `.expect` would remove  these are make the code more stable. 

I'd recommend that expects/unwraps only be used in tests unless there is a concrete reason a panic is preferred. wdyt?

PR #101, opened 2023-05-01
Use namespaced persistence

richardhuaaa:
When writing to persistence, have it clearly be marked as xmtp data, and namespaced per-account. This PR introduces a `NamespacedPersistence` class that wraps `Persistence` and automatically prefixes the keys used to read and write data. For example, serialized account data will automatically be stored using the key `xmtp/account_<address>/vmac_account`.

Also, use `&str` instead of `String` for cleaner code/to prevent unnecessary copies.

Tested via unit tests in `xmtp`, `xmtp_wasm`, `xmtp_js`.

https://github.com/xmtp/xmtp-js/issues/348


xmtp/src/builder.rs
```diff
@@ -41,18 +41,11 @@ where
...
-        let persistence = self.persistence.as_ref().ok_or_else(|| {
-            "Persistence engine must be set before setting the account".to_string()
-        })?;
-
-        let existing = persistence.read(key.clone());
+    fn find_or_create_account(
+        persistence: &mut NamespacedPersistence<P>,
+    ) -> Result<VmacAccount, String> {
+        let key = "vmac_account";
+        let existing = persistence.read(&key);
```
jazzz, line 48: Compiler warning here

xmtp/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 pub mod account;
 pub mod builder;
 pub mod client;
+pub mod in_memory_persistence;
```
jazzz, line 4: may want to add this file to the persistence module rather than creating a new one.

jazzz (approved):
These are some great changes. Simplified structure and removed mem copies. Love it see it

Qiuck fix for the compiler errors and this is 👍

Issue #104, opened 2023-05-04
Custom JS error classes/enum


richardhuaaa:
wasm_bindgen expects a JsValue to be returned from any function that can throw. 

Currently, we handle this by returning JsError from all of our wasm bindings. This has the effect of auto-converting from Rust std::err::Error to JsError automatically via JsError::from<Error>, and then into JsValue via JsError::into<JsValue>.

The implementation of JsError::from<Error> is the JS equivalent of calling new Error(...) with the string generated from calling to_string() on the Rust error, which doesn't give consumers the opportunity to differentiate between different types of errors in code.

A workaround would be to define custom JS Error subclasses and import them into Rust, or define a special JS XMTPError type with an enum field and mappings from all known errors to this enum.

More info at https://github.com/rustwasm/wasm-bindgen/issues/1742#issuecomment-984149894

PR #105, opened 2023-05-04
Better error handling

richardhuaaa:
This achieves a few things:
1. Strongly typed Rust errors, using the thiserror crate. Info [here](https://nick.groenen.me/posts/rust-error-handling/) and [here](https://dev.to/seanchen1991/a-beginner-s-guide-to-handling-errors-in-rust-40k2). I think this is the most commonly used way today, coincidentally vodozemac also uses the same pattern.
2. Allow implementors of `Persistence` bindings to define their own `Error` type depending on the platform, via an associated type on the `Persistence` trait. When an error is encountered in `Persistence` within libxmtp, it will be wrapped with a Rust error and included on the `native_error` field, so that the platform-specific binding code can do whatever they want with it after it bubbles up.
3. In the LocalStorage implementation of Persistence, serialize from bytes to string via base64 instead of utf8.

I think this unblocks us on error handling within the core Rust `xmtp` crate, but the error handling in JS-land is far from perfect. The Rust error is strongly typed, but before it passes over the wasm_bindgen interface, it's effectively wrapped into a JS Error in the following way: `new Error(<rust_error.to_string()>)`. More details and possible solution in https://github.com/xmtp/libxmtp/issues/104, but I think it's fine to punt this for now.

Added tests in `xmtp` and `bindings_js`.

### DIFF
--- a/bindings_js/example/index.html
+++ b/bindings_js/example/index.html
@@ -5,23 +5,39 @@
     </head>
     <body>
         <p id="preamble">Running test...</p>
-        <p id="result"></p>
+        <p id="result1"></p>
+        <p id="result2"></p>
 
         <script type="module">
             import { Client } from './dist/es/index.js';
+            window.localStorage.clear();
 
-            async function run() {
+            await (async function testReadWrite() {
                 Client.resetAll();
                 let client = await Client.create();
                 let input = [65, 65, 65];
                 client.writeToPersistence("foo", input);
                 if (client.readFromPersistence("foo").toString() === input.toString()) {
-                    document.getElementById("result").innerText = "Test passed!";
+                    document.getElementById("result1").innerText = "Read/write test passed!";
                 } else {
-                    document.getElementById("result").innerText = "Test failed";
+                    document.getElementById("result1").innerText = "Read/write test failed";
                 }
-            }
-            run();
+            })();
+
+            await (async function testError() {
+                Client.resetAll();
+                let setItem = Storage.prototype.setItem;
+                Storage.prototype.setItem = () => { throw new Error("test error"); };
+                try {
+                    let client = await Client.create();
+                    client.writeToPersistence("abc", [65, 65, 65]);
+                    document.getElementById("result2").innerText = "Error test failed";
+                } catch (e) {
+                    document.getElementById("result2").innerText = "Error test passed! This is how an error looks: " + e.message;
+                }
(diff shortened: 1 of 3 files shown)


bindings_wasm/src/local_storage_persistence.rs
```diff
@@ -1,5 +1,17 @@
...
 
+#[derive(Error, Debug)]
+pub enum LocalStoragePersistenceError {
+    #[error("Failed to read/write from local storage")]
+    ReadWriteError { native_error: JsValue },
+
+    #[error("Failed to deserialize from local storage")]
+    DeserializationError(#[from] base64::DecodeError),
+}
+
```
jazzz, line 14: Like the idea of using this enum structure. Very clear about how a module can fail.

From a handling POV is there a mechanism to inform developers which Errors are possible for a given method/function? 

Looking at `fn write(&mut self, key: &str, value: &[u8]) -> Result<(), Self::Error>`

the handler code would look like:
```rust
match writer.write("KEY", "bytes".as_bytes()) {
  Ok(v) => success(v),
  Err(LocalStoragePersistenceError::ReadWriteError) => retry(),
  Err(LocalStoragePersistenceError::DeserializationError) => notPossible!(),
}
```

It's not immediately clear which functions return which errors without looking at the source. Worse if developers ignore part of the pattern matching:

```rust
match writer.write("KEY", "bytes".as_bytes()) {
  Ok(v) => success(v),
  Err(LocalStoragePersistenceError::ReadWriteError) => retry(),
  Err(_) => Ignore(),
}
```

Then potentially uncaught errors will only be discoverable at runtime. Any guidance on how these errors should be handled from the calling perspective?

xmtp/src/persistence/mod.rs
```diff
@@ -1,8 +1,9 @@
 pub mod in_memory_persistence;
 
 pub trait Persistence {
-    fn write(&mut self, key: &str, value: &[u8]) -> Result<(), String>;
-    fn read(&self, key: &str) -> Result<Option<Vec<u8>>, String>;
+    type Error;
```
jazzz, line 4: I can get behind using Associated Types for Errors. Seems like in most cases the Errors could be standardized at the trait level which would simplify handling, thought I imagine we expect there to be many error types specific to the Trait Impl that need handling?

jazzz (approved):
👍

PR #109, opened 2023-05-10
Account Scaffolding

jazzz:
This PR adds plumbing for a Higher Order account object. 

- Creates `Account` struct
- Adds `AccountCreator` to help create accounts
- Adds Mechanism for Implementors to sign keys
- Adds Association struct to add flexible signatures
- updates WasmBindings

Work saved for followup PRs

- Signature generation and validation
- Update Client and ClientBuilder to use `Account` not `VmacAccount`

### DIFF
--- a/xmtp/src/account.rs
+++ b/xmtp/src/account.rs
@@ -7,17 +7,17 @@ use crate::Signable;
 
 type Address = String;
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
 pub enum AssociationFormat {
     Eip191,
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
 pub enum AssociationType {
     Static,
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
 pub struct Association {
     pub addr: Address,
     pub key_bytes: String,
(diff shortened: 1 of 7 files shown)


xmtp/src/lib.rs
```diff
@@ -7,6 +7,10 @@ pub mod vmac_protos;
 pub use builder::ClientBuilder;
 pub use client::Client;
 
+pub trait Signable {
+    fn bytes_to_sign(&self) -> Vec<u8>;
+}
+
```
jazzz, line 13: Creates a common interface for signable objects

xmtp/src/account.rs
```diff
@@ -34,8 +83,87 @@ impl<'de> Deserialize<'de> for VmacAccount {
...
+
         Ok(Self::new(account))
     }
 }
+
+#[derive(Serialize, Deserialize)]
+pub struct Account {
+    keys: VmacAccount,
+    assoc: Association,
+}
```
jazzz, line 97: VmacAccount remains as an wrapper around the OlmAccount providing abstraction + Serialization

Account integrates all the signing and application specific context with the keys. In most cases Account replaces VmacAccount

xmtp/src/account.rs
```diff
@@ -1,19 +1,68 @@
...
+    pub fn test() -> Self {
+        Self {
+            addr: "ADDR".to_string(),
+            key_bytes: "KEY_BYTES".to_string(),
+            text: "TEXT".to_string(),
+            format: AssociationFormat::Eip191,
+            proof: vec![0, 2, 3, 5, 67],
+            proof_type: AssociationType::Static,
+        }
+    }
```
jazzz, line 40: Not fully implemented - Currently uses a Mock for now

xmtp/src/account.rs
neekolas: I like this

neekolas (approved):
This feels like good progress

xmtp/src/account.rs
```diff
@@ -1,19 +1,68 @@
...
+type Address = String;
+
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+pub enum AssociationFormat {
+    Eip191,
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+pub enum AssociationType {
+    Static,
```
neekolas, line 17: What are some other `AssociationType`s you have in mind. I don't understand the distinction between `AccountType` and `AccountFormat`

jazzz: In V2 the text used in link wallets and Identity keys was static. To support 1) a wider range of Associations and 2) versioning of the signature text its helpful to capture how the signature text ought to be validated. Expected values include:

- XmtpStaticV1
- SIWE 
- CAIP-122

**AssociationFormat** - The set of cryptographic standards to generate the Signature( EIP191, ED25519_BLAKE3, BIP340, etc)
**AssociationType** - The application level standards used to generate/validate the text which was signed.

*(The names could definitely use some work. Feedback welcome)*

jazzz:
@richardhuaaa 

> I wonder if we could eventually restructure in such a way that the app does not need to know what an Account is,

This is the end goal. 

>hide the Account struct within the xmtp crate and not have that Vec<Account> within bindings_wasm at all.

This seems harder. The Vec<Account> will be removed  however moving the js_sy::Function into the core rust library would bleed abstraction layers. The core library should not be aware of wasm particulars (That is the intent of bindings_wasm). We could look into a abstraction layer that wraps the platform specific constructions like js_sys::Function , or ensure that platform specific operations occur in the bindings.

In my mind bindings_wasm is only intended to be consumed by other SDKS, and bindings_js provides the JS specific sugar around the bindings_wasm.

PR #113, opened 2023-05-16
Copy bindings_swift to new crate

richardhuaaa:
This is a literal copy of `bindings_swift` to a new crate called `bindings_ffi` with no additional changes. Splitting this as a separate change so that subsequent PR's are easier to review.

Rough plan:
1. Next PR exposes the v3 libxmtp client to swift on this new `bindings_ffi` crate. The new crate saves us from worrying about API compatibility with the existing V2 SDK's. I considered renaming `bindings_swift` to `bindings_swift_v2` or `bindings_swift_legacy` but couldn't decide.
2. At some point soon switch the new crate from using [`swift-bridge`](https://github.com/chinedufn/swift-bridge) to Mozilla [uniffi](https://github.com/mozilla/uniffi-rs) as per @michaelx11's recommendation, which will give us Android support as well as a more mature feature set.
3. When v3 is released and used by all the SDK's, delete the `bindings_swift` crate.


neekolas:
Uniffi is tempting, given it has a common language to codegen bindings for both iOS and Android. And may just have less cracks and weird holes than `swift-bridge` (to be seen). Do want to call out that it's going to be a fair bit of work in the iOS side to replace all the generated swift types with new Uniffi types. /cc @nakajima

neekolas:
I'm not really sure I understand the concern about breaking API compatibility. Are we thinking of changing the bindings for the existing networking and `secp256k1` types (before the move to uniffi)? I would think that as long as our changes can compile correctly, we shouldn't really have a risk of breaking anything by adding methods.

Swift already needs to pin to a specific version of `libxmtp`, so it's not like they are going to automatically upgrade to a new version that is broken.

michaelx11:
> I'm not really sure I understand the concern about breaking API compatibility. Are we thinking of changing the bindings for the existing networking and secp256k1 types (before the move to uniffi)? I would think that as long as our changes can compile correctly, we shouldn't really have a risk of breaking anything by adding methods.

I do think it's possible to just have V3 code live alongside the small React Native changes. It'll just require some care to make sure we leave the existing stuff mostly untouched, both directly (raw code) and indirectly (any dependencies, xcframework inclusions, etc)

> Swift already needs to pin to a specific version of libxmtp, so it's not like they are going to automatically upgrade to a new version that is broken.

Yup, there won't be any automatic breakage. The case I can envision is we need a bugfix in React Native V2 SDK, and we end up paying a large lump sum of work of repairing the flow (bindings_swift -> XMTPRustSwift.xcframework -> xmtp-ios -> xmtp-react-native). Actually that might be an argument for just keeping the same flow (same `bindings_swift` crate and the same XMTPRustSwift.xcframework )and making sure we continue to test both old and new (V3) code regularly.

neekolas (approved):

neekolas (approved):

PR #114, opened 2023-05-17
Add networking to XMTP client and expose it in FFI crate

richardhuaaa:
Note that half of the changes are in `bindings_ffi/src/lib.rs`, which gets automatically hidden as a large diff.

This is the starting point for exposing our libxmtp client to the platform layer (iOS/Android). I added the networking trait from @neekolas's PR with some slight modifications and hooked up `xmtp_networking` to it.

There's still a lot of gaps to fill here but figured I should land this first so there's some initial structure:
1. Have not tested whether this actually works with `swift-bridge` at the platform layer. I will switch to Mozilla uniffi first and test the integration at that point.
2. Need to add proper error handling.
3. Have not exposed account creation to the platform layer yet.
4. Have not hooked up the API auth token.

I will work on (1) next - if anyone wants to pull any of the others feel free!

### DIFF
--- a/xmtp/src/networking.rs
+++ b/xmtp/src/networking.rs
@@ -48,16 +48,18 @@ impl Default for MockXmtpApiClient {
 impl XmtpApiClient for MockXmtpApiClient {
     async fn publish(
         &mut self,
-        token: String,
+        _token: String,
         envelopes: Vec<Envelope>,
     ) -> Result<PublishResponse, String> {
-        let mut existing: Vec<Envelope> = match self.messages.get(&token) {
-            Some(envelopes) => envelopes.clone(),
-            None => vec![],
-        };
-        existing.append(envelopes.clone().as_mut());
-        self.messages.insert(token, envelopes);
-
+        for envelope in envelopes {
+            let topic = envelope.content_topic.clone();
+            let mut existing: Vec<Envelope> = match self.messages.get(&topic) {
+                Some(existing_envelopes) => existing_envelopes.clone(),
+                None => vec![],
+            };
+            existing.push(envelope);
+            self.messages.insert(topic, existing);
+        }
         Ok(PublishResponse {})
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4626,6 +4626,7 @@ dependencies = [
  "serde_json",
  "thiserror",
  "tokio",
+ "uuid 1.3.1",
  "vodozemac",
  "xmtp_cryptography",
  "xmtp_networking",
--- a/xmtp/Cargo.toml
+++ b/xmtp/Cargo.toml
@@ -20,3 +20,4 @@ hex = "0.4.3"
 ethers = "2.0.4"
 ethers-core = "2.0.4"
 tokio = {version="1.28.1", features=["rt", "macros"]}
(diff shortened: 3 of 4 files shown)


xmtp/src/builder.rs
```diff
@@ -56,8 +66,8 @@ where
         self
     }
 
-    pub fn wallet_address(mut self, wallet_address: String) -> Self {
-        self.wallet_address = Some(wallet_address);
+    pub fn wallet_address(mut self, wallet_address: &str) -> Self {
```
neekolas, line 69: Curious why this change was necessary, given it's being converted back to a string

xmtp/src/client.rs
```diff
@@ -11,16 +12,18 @@ pub enum Network {
     Prod,
 }
 
-pub struct Client<P>
+pub struct Client<A, P>
 where
+    A: XmtpApiClient,
     P: Persistence,
 {
+    pub api_client: A,
```
neekolas, line 20: Would be good to have a test that exercises this

neekolas (approved):
Great work! This unblocks a whole lot of workstreams

xmtp/src/builder.rs
```diff
@@ -91,7 +101,13 @@ where
         }
     }
 
-    pub fn build(&mut self) -> Result<Client<P>, ClientBuilderError<P::Error>> {
+    pub fn build(&mut self) -> Result<Client<A, P>, ClientBuilderError<P::Error>> {
+        let api_client =
+            self.api_client
+                .take()
```
jazzz, line 107: Taking the value of the Option leaves `self` in a changed state -- and multiple calls will produce different configurations of Clients. `build` should consume self if its going to consume it's members, to avoid confusion

bindings_ffi/src/lib.rs
```diff
@@ -337,25 +158,28 @@ mod tests {
...
         let topic = Uuid::new_v4();
-        let mut client = super::create_client(ADDRESS.to_string(), false)
+        let mut client = super::create_client(account, ADDRESS, false).await.unwrap();
+        let mut sub = client
+            .api_client
+            .subscribe(vec![topic.to_string()])
             .await
             .unwrap();
-        let mut sub = client.subscribe(vec![topic.to_string()]).await.unwrap();
         std::thread::sleep(std::time::Duration::from_millis(100));
```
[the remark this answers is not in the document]
neekolas: Good catch. I've run into this one before too

xmtp/src/networking.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[async_trait]
+impl XmtpApiClient for MockXmtpApiClient {
+    async fn publish(
+        &mut self,
+        token: String,
+        envelopes: Vec<Envelope>,
+    ) -> Result<PublishResponse, String> {
+        let mut existing: Vec<Envelope> = match self.messages.get(&token) {
```
neekolas, line 54: This is my half-finished garbage code. It shouldn't be looking up the messages by token, but going through each envelope and finding them by topic. You can see a correctly working version [here](https://github.com/xmtp/libxmtp/pull/118/files#diff-1c0e183cafb21ff8ca78b4d5a769750eb7df51979ca1f3316bd9a32786c9e645R54-R62).

PR #118, opened 2023-05-19
Contact publishing and retrieval

neekolas:
## Summary

- Allows for publishing contact bundles (as they are currently defined in `xmtp/proto`
- Allows for retrieving contact bundles from the network
- Replaces the `VmacAccount` on `Client` with an `Account`
- Fixes some bugs in the `MockXmtpClient`

## Next up on my list
- Wrap VmacContactBundle proto in a struct that has some additional methods for things like comparing equality, getting wallet address, validation, etc
- Session management

## Notes

This PR is littered with TODOs. We have a bunch of interrelated workstreams here and the sequencing is tricky. The goal of this PR is to get things into `main` with the correct types, as imperfect as the implementations may be, and then fix as we get other pieces together.

For example, we use a randomly generated wallet to sign the key bundle right now, instead of an actual wallet signer.


xmtp/Cargo.toml
```diff
@@ -15,9 +15,10 @@ xmtp_networking = { path = "../xmtp_networking" }
 async-trait = "0.1.68"
 xmtp_cryptography = { path = "../xmtp_cryptography"}
 hex = "0.4.3"
-
-[dev-dependencies]
 ethers = "2.0.4"
```
neekolas, line 18: This will go back to a dev dependency once we have some way of passing in an external signer.

xmtp/src/account.rs
```diff
@@ -28,7 +32,9 @@ impl VmacAccount {
     }
 
     pub fn generate() -> Self {
-        Self::new(OlmAccount::new())
+        let mut acc = OlmAccount::new();
+        acc.generate_fallback_key();
```
neekolas, line 36: Realized I have to do this manually to ensure that a fallback key is present after generating.

xmtp/src/client.rs
```diff
@@ -20,7 +24,8 @@ where
     pub api_client: A,
     pub network: Network,
     pub persistence: NamespacedPersistence<P>,
-    pub account: VmacAccount,
+    // TODO: Make account private. Just need to figure out how to access for tests
+    pub account: Account,
```
jazzz, line 28: You can use fine grain visibility scoping https://doc.rust-lang.org/reference/visibility-and-privacy.html

You could use `pub(self)` if the test are contained in the same module.

jazzz (approved):
Love it. Step by Step

xmtp/src/client.rs
neekolas: Did not know `pub(self)` was a thing in Rust. That sounds very useful.

PR #119, opened 2023-05-22
Add Example

jazzz:
During development its important to thinking about how a library will be consumed/used in the wild. This example dogfoods the coreSDK with the intention of highlighting friction, and understand what is needed to get an actual client up and running.

Currently many features are not available yet, however this example can grow with the external API. 

Note there are already many optimizations to make, however this PRs intention is to capture what the interface looks like today.


neekolas (approved):
Like the idea of having a bare-bones CLI example. Will be very useful for testing

PR #120, opened 2023-05-24
Builder Interface

jazzz:
Refactors the Builder interface to include passing in a signer

## Highlights
- InboxOwner Trait
  - Provides a common interface for Signers, which can be passed into the account builder
  - Currently implemented for ethers::LocalWallet, will implement for other signers when we tackle platformSDKs
- Adds AccountStrategy enum to ClientBuilder, which allows control of the account fetching behavior
  - Passing in a impl InboxOwner ->   CreateIfNotFound,
  - Passing in a Address -> LoadCachedOnly
  - Passing in an Account -> Use  ExternalAccount
  - This was left over from Integrating with the FFI bindings, however is not strictly needed after the refactor. Open to dropping.
- Simplifies the api for creating a Client
- Ported to use Account, rather than VmacAccount

### DIFF
--- a/bindings_ffi/src/lib.rs
+++ b/bindings_ffi/src/lib.rs
@@ -30,13 +30,9 @@ async fn create_client(
     // TODO proper error handling
 ) -> Result<xmtp::Client<FfiApiClient, InMemoryPersistence, UnencryptedMessageStore>, String> {
     let api_client = FfiApiClient::new(host, is_secure).await?;
-    let persistence = InMemoryPersistence::new();
-    let store = UnencryptedMessageStore::default();
 
     let xmtp_client = xmtp::ClientBuilder::new(owner.into())
         .api_client(api_client)
-        .persistence(persistence)
-        .store(store)
         .build()
         .map_err(|e| format!("{:?}", e))?;
 
--- a/bindings_ffi/src/lib.rs
+++ b/bindings_ffi/src/lib.rs
@@ -1,8 +1,9 @@
 use async_trait::async_trait;
 use xmtp::{
-    account::Account, networking::XmtpApiClient,
-    persistence::in_memory_persistence::InMemoryPersistence, storage::UnencryptedMessageStore,
+    networking::XmtpApiClient, persistence::in_memory_persistence::InMemoryPersistence,
+    storage::UnencryptedMessageStore,
 };
+use xmtp_cryptography::utils::LocalWallet;
 use xmtp_networking::grpc_api_helper::{self, Subscription};
 use xmtp_proto::xmtp::message_api::v1::{Envelope, PagingInfo, PublishResponse, QueryResponse};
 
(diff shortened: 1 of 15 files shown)


bindings_ffi/src/lib.rs
```diff
@@ -11,33 +12,27 @@ pub type FfiXmtpClient = xmtp::Client<FfiApiClient, InMemoryPersistence, Unencry
...
 
-    let xmtp_client = xmtp::ClientBuilder::new()
-        .wallet_address(&account.addr())
-        .account(account)
+    let xmtp_client = xmtp::ClientBuilder::new(owner.into())
         .api_client(api_client)
-        .persistence(persistence)
-        .store(store)
         .build()
         .map_err(|e| format!("{:?}", e))?;
```
jazzz, line 38: Heres an example of what the ClientBuilder API looks like

jazzz, line 38: Heres an example of what the ClientBuilder API looks like

xmtp/src/builder.rs
```diff
@@ -21,39 +20,74 @@ pub enum ClientBuilderError<PE, SE> {
...
+
+    #[error("Associating an address to account failed")]
+    AssociationFailed(#[from] AssociationError),
+    // #[error("Error Initalizing Store")]
+    // StoreInitialization(#[from] SE),
+    #[error("Error Initalizing Account")]
+    AccountInitialization(#[from] AccountError),
+}
+
+pub enum AccountStrategy<O: InboxOwner> {
```
neekolas, line 34: This is neat

xmtp/src/builder.rs
```diff
@@ -77,88 +111,108 @@ where
...
+    pub fn build(mut self) -> Result<Client<A, P, S>, ClientBuilderError<P::Error>> {
+        let api_client = self.api_client.take().unwrap_or_default();
+        let wallet_address = self.get_address();
+        let persistence = self.persistence.take().unwrap_or_default();
         let mut persistence =
-            NamespacedPersistence::new(&get_account_namespace(wallet_address), persistence);
-        let account = Self::find_or_create_account(&mut persistence)?;
+            NamespacedPersistence::new(&get_account_namespace(&wallet_address), persistence);
+
+        // Fetch the Account based upon the account strategy.
```
neekolas, line 204: So clean

xmtp/src/client.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
 };
 use prost::Message;
 use xmtp_proto::xmtp::{message_api::v1::Envelope, v3::message_contents::VmacContactBundle};
-
+//TODO: Remove Comment
```
neekolas, line 12: Think you can remove this TODO

neekolas (approved):
This is looking slick

xmtp/src/client.rs
jazzz: If you think we don't need it 👍

xmtp/src/builder.rs
```diff
@@ -21,39 +20,74 @@ pub enum ClientBuilderError<PE, SE> {
...
+    // #[error("Error Initalizing Store")]
+    // StoreInitialization(#[from] SE),
+    #[error("Error Initalizing Account")]
+    AccountInitialization(#[from] AccountError),
+}
+
+pub enum AccountStrategy<O: InboxOwner> {
+    CreateIfNotFound(O),
+    CachedOnly(Address),
+    ExternalAccount(Account),
```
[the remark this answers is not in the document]
jazzz: changed it so its only available in test builds to make it more clear

PR #121, opened 2023-05-24
Client Async Registration + Error Types

jazzz:
Adds registration flow for Clients. Mostly type changes, concrete errors for get the client ready for async. 

## Highlights
- Adds method for clients to publish their contact bundle if its not already found on the network. 

## Out of Scope
- Forced Registrations
  - I've made an Abstraction layer that wraps the client and forces it to be registered before being used. However SwiftBridge Errors are slowing progress. Given Swiftbridge is on its way out, I'll revisit once the new abstraction layer is in. 
  - The end goal is that developers cannot forget to init.

### DIFF
--- a/xmtp/examples/cli-client.rs
+++ b/xmtp/examples/cli-client.rs
@@ -12,7 +12,8 @@ use xmtp_cryptography::utils::LocalWallet;
 
 /// A complete example of a minimal xmtp client which can send and recieve messages.
 /// run this example from the cli:  `RUST_LOG=DEBUG cargo run --example cli-client`
-fn main() {
+#[tokio::main]
+async fn main() {
     env_logger::init();
     info!("Starting CLI Client....");
 
@@ -27,14 +28,21 @@ fn main() {
         .store(msg_store)
         .build();
 
-    let _client = match client_result {
+    let uninit_client = match client_result {
         Err(e) => {
             error!("ClientBuilder Error: {:?}", e);
             return;
         }
         Ok(c) => c,
     };
-    warn!("Client Is not properly initialized at this point -- Signed account not present");
+
+    let client = match uninit_client.init().await {
+        Err(e) => {
+            error!("Initialization Failed: {}", e.to_string());
+            panic!("Could not init");
+        }
+        Ok(c) => c,
+    };
 
     // Application logic
     // ...
--- a/xmtp/examples/cli-client.rs
+++ b/xmtp/examples/cli-client.rs
@@ -28,7 +28,7 @@ async fn main() {
         .store(msg_store)
         .build();
 
-    let uninit_client = match client_result {
+    let mut client = match client_result {
         Err(e) => {
             error!("ClientBuilder Error: {:?}", e);
             return;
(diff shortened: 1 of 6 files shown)


xmtp/src/client.rs
```diff
@@ -34,71 +42,111 @@ where
...
     pub fn wallet_address(&self) -> Address {
         self.account.addr()
     }
 
-    pub async fn get_contacts(
+    async fn noop(&self) {}
+    pub async fn init(&mut self) -> Result<(), ClientError> {
+        // Register Contact Bundles
+        let registered_bundles = self.noop().await; // Causes #[swift_bridge::bridge] to fail when noop is a member fn
+
```
jazzz, line 79: Calling an async member function in `init` causes errors in the swift bridge. It appears that the self parameter included in the async call adds a reference of Client to the future. 

This results in client being required to implement Sync. 

The current test configuration uses an ephemeral Diesel implementation which does not implement Sync (because it is not threadsafe).

jazzz: https://github.com/xmtp/libxmtp/actions/runs/5072644086/jobs/9110613123?pr=121

xmtp/examples/cli-client.rs
```diff
@@ -21,20 +22,24 @@ fn main() {
...
             return;
         }
         Ok(c) => c,
     };
-    warn!("Client Is not properly initialized at this point -- Signed account not present");
+
+    if let Err(e) = client.init().await {
+        error!("Initialization Failed: {}", e.to_string());
+        panic!("Could not init");
+    };
```
jazzz, line 42: Example: Interim Registration

xmtp/src/client.rs
```diff
@@ -34,71 +42,110 @@ where
...
     }
 
-    pub async fn get_contacts(
+    pub async fn init(&mut self) -> Result<(), ClientError> {
+        // Register Contact Bundles
+        let registered_bundles = self.get_contacts(&self.wallet_address()).await?;
+
+        if registered_bundles.is_empty() {
+            self.publish_user_contact().await?;
+        }
```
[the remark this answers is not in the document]
neekolas: Good catch. Looks like it would

xmtp/src/client.rs
```diff
@@ -34,71 +42,110 @@ where
...
+
+        self.is_initialized = true;
+        Ok(())
+    }
+
+    async fn get_contacts(
         &self,
         wallet_address: &str,
-    ) -> Result<Vec<VmacContactBundle>, String> {
+    ) -> Result<Vec<VmacContactBundle>, ClientError> {
```
neekolas, line 90: Thanks

xmtp/src/client.rs
```diff
@@ -34,71 +42,110 @@ where
...
+            .await
+            .map_err(ClientError::QueryError)?;
 
         let mut contacts = vec![];
         for envelope in response.envelopes {
             // TODO: Wrap the proto in some special struct
             // TODO: Handle errors better
-            let contact_bundle = VmacContactBundle::decode(envelope.message.as_slice())
-                .map_err(|e| format!("{}", e))?;
+            let contact_bundle =
```
neekolas, line 102: I foresee some merge conflicts with what I'm working on. I have a new `Contact` struct in my branch. I'll work out the conflicts on my end, but I'll probably steal some of your error handling code to reduce the burden.

neekolas: Sorry, beat you to the punch merging in the `Contact` struct, which'll mean you'll have to deal with a few merge conflicts.

xmtp/src/client.rs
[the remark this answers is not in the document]
jazzz: Added check for matching ID's ( which are deterministic )

PR #124, opened 2023-05-26
Encrypted Store Impl

jazzz:
Enables Message storage encryption for local message history. 

## Highlights
- Enables SqlCipher builds for on disk encryption, removing the pareto dominated unencryptedMessageStore

## Out of Scope
- Project Hierarchy:
  - Its clear that these implementations ought to exist in a dedicated components package rather than being baked into the client. Will move in subsequent PR's

## CheckList
- [x] Tests Verifying encryption
- [x]  ~Feature configuration for build~ (Defering for now. Conditional complication leads to a scenario where the datastore is not encrypted thought it appears it is. )


neekolas (approved):
LGTM

PR #125, opened 2023-05-29
Uniffi-backed crate for Android on XMTPv2

richardhuaaa:
We need to expose the Rust-backed Diffie-Hellman operation to unblock performance improvements in Android, context [here](https://github.com/xmtp/xmtp-react-native/issues/33#issuecomment-1564569524). To keep binary size and build-times low, as well as to avoid significant time overhauling `xmtp_swift`, I've made a separate crate called `xmtp_dh` that only contains the single `diffie_hellman_k256` API that is needed for now.

There are three pieces here:
- The `xmtp_dh` crate. The `README.md` describes how to integrate it into an Android app.
- An example Android app (`xmtp_dh/examples`). Used to run an end-to-end test.
- The `uniffi_bindgen_generator` crate. A boilerplate binary crate used to [generate the platform-specific bindings](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html#multi-crate-workspaces) in the other crates. Will be used for `bindings_ffi` soon as well.

This review is not as big as it looks - **you can ignore everything in the Android example app under `xmtp_dh/examples` except for `MainActivity.kt`**.

Credit to @michaelx11 - I reused a lot of the structure from his vmac prototype here.


nplasterer (approved):
Thanks for the great readme and example! LGTM.

PR #128, opened 2023-05-31
Make proto reference explicit

jazzz:
## Problem

Updates to https://github.com/xmtp/proto/tree/xmtpv3 cause local development environments to break, due the `xmtp_proto` package being refetched everytime the package is built. 

## Solution
This PR makes the reference explicit so that it's tracked in git. In the future any changes to the Proto repo will not be fetched until the appropriate code has been merged into main, and a commit updates the reference in build.rs.

### DIFF
--- a/xmtp_proto/build.rs
+++ b/xmtp_proto/build.rs
@@ -29,7 +29,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let status = Command::new("buf")
         .arg("generate")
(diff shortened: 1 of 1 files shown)


neekolas (commented):
I think @richardhuaaa is working on an alternate solution that replaces the build file with a shell script that will be run on demand.

neekolas:
FWIW, almost through fixing up the package to use the new protos

PR #129, opened 2023-05-31
Update to use new protos

neekolas:
## Summary

- Updates all proto usages to be compatible with latest protos


jazzz (approved):
Looks good. 

Noticed that the Contact bundle doesn't contain the signing key which might cause issues down the line. AEAD can be relied on for awhile however we'll eventually want the signing identity as part of the bundle to ensure that the signature chain can be validated for user level actions.  #FutureWork

neekolas:
Good catch. Will add in the next round of proto updates.

PR #130, opened 2023-06-01
Generate protos via shell script

richardhuaaa:
The original `build.rs` makes some assumptions about what is available on the host operating system, as well as fetching from github. This breaks cross-compilation which happens via docker containers, means the build can't run when offline, and causes breaking changes to `xmtp/proto` to break the build in `libxmtp`. This PR commits the generated files instead, and replaces the `build.rs` with a shell script that we can run whenever we explicitly want to pull in changes.

### DIFF
--- a/dev/gen_protos.sh
+++ b/dev/gen_protos.sh
@@ -1,3 +1,4 @@
+pushd xmtp_proto > /dev/null
 if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
     if ! cargo install protoc-gen-prost-crate; then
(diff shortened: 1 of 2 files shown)


xmtp_proto/gen_protos.sh
```diff
@@ -0,0 +1,10 @@
+if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
```
neekolas, line 1: Would be nicer if we had a `dev` folder in the repo, to match the convention in other xmtp projects

neekolas (approved):

PR #131, opened 2023-06-01
feat: use proto encoding for rpc requests across bridge

dmccartney:
This changes `bindings_swift` to use protobuf to encode/decode req/res messages as bytes across the rust/swift bridge.
This deletes lots of piecemeal glue code and parallel types on both sides of the bridge: e.g.
```rs
// rust implementation
async fn query(&mut self, req: Vec<u8>) -> Result<Vec<u8>, String> {
    let request: QueryRequest = Message::decode(&req[..]).map_err(|e| format!("{}", e))?;
    let result: QueryResponse = self.client.query(request).await.map_err(|e| format!("{}", e))?;
    Ok(result.encode_to_vec())
}
```
```swift
// swift usage
func query(request: QueryRequest) async throws -> QueryResponse {
  let req = RustVec<UInt8>(try request.serializedData());
  let res: RustVec<UInt8> = try await rustClient.query(req);
  return try QueryResponse(serializedData: Data(res));
}
```

Along the way, I also normalized the `xmtp_networking` API helper to receive and return the corresponding `*Request` and `*Response` messages: e.g.
```rs
pub async fn query(
    &self,
    request: QueryRequest,
) -> Result<QueryResponse, tonic::Status>;
```
(so there were some signature changes to the `bindings_ffi` as part of that ^)


nplasterer (approved):
LGTM!

neekolas:
Do you have any benchmarks on what this does to performance?

We previously had serialized the requests/responses as JSON to go over the bridge, and saw a _very_ significant performance improvement moving to shared types. Protobufs are faster than JSON to encode/decode, but I don't know how much faster in this case.

nplasterer:
Thats a great call Micheal had some benchmarking branches in place https://github.com/xmtp/xmtp-react-native/tree/michaelx-benchmark_v1.0.1_conversation_list_with_XMTPRust_v0.2.2-beta0 if you'd like to take some of that code and run it through and see what it does @dmccartney. I'm assuming you have this all wired together locally. 
The current numbers without these changes are

`iOS: Loaded 2001 conversations in 10157ms (10.157s)`
`Android: Loaded 2001 conversations in 19488ms (19.4s)`

dmccartney:
I repeated that benchmark (on the wifi here, with the same user connected to the "dev" network) and the results look about the same:

Before: (manually serializing across the bridge)
>  Loaded 2001 conversations in 8453ms
>  Loaded 2001 conversations in 8494ms
>  Loaded 2001 conversations in 8478ms

After: (protobuf serializing across the bridge)
>  Loaded 2001 conversations in 8582ms
>  Loaded 2001 conversations in 8710ms
>  Loaded 2001 conversations in 8636ms

(They're actually so similar that I had to triple check I didn't swap it out properly.)

neekolas (approved):

PR #133, opened 2023-06-04
Cross platform builds in CI

snormore:
This PR adds a GH workflow that builds across platforms, specifically the android platforms referenced in the cross_build.sh scripts, but can be extended to build for others. It seems like one of the next steps is to hook up version publishing in general for these packages/platforms, but this PR just focuses on getting the builds running in CI for validation/testing and further iteration. The builds are still relatively slow in CI, taking about 15 minutes to complete, but not nearly as slow as when running locally on a mac via qemu.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,53 @@
+name: Build
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [
+            aarch64-unknown-linux-gnu,
+            x86_64-linux-android,
+            i686-linux-android,
+            armv7-linux-androideabi,
+            aarch64-linux-android
+          ]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: "Cache"
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            ./target/${{ matrix.target }}
+          key: ${{ matrix.target }}-${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
+
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: ${{ matrix.target }}
+
+      # Install latest cross to mitigate unwind linking issue on android builds.
+      # See https://github.com/cross-rs/cross/issues/1222
+      - name: Install rust cross
+        run: |
+          cargo install cross --git https://github.com/cross-rs/cross
+  
+      - name: Build target
+        uses: actions-rs/cargo@v1
+        with:
+          use-cross: true
+          command: build
+          args: --release --target ${{ matrix.target }}
--- a/xmtp/Cargo.toml
+++ b/xmtp/Cargo.toml
@@ -43,4 +43,4 @@ url = "2.3.1"
 
 [dependencies.libsqlite3-sys]
 version = "0.26.0"
-features = ["bundled-sqlcipher"]
+features = ["bundled-sqlcipher-vendored-openssl"]
--- a/xmtp/src/bin/update-schema.rs
+++ b/xmtp/src/bin/update-schema.rs
@@ -13,7 +13,7 @@ use xmtp::storage::{EncryptedMessageStore, StorageOption};
 const DIESEL_TOML: &str = "./diesel.toml";
 
 /// This binary is used to to generate the schema files from a sqlite database instance
-/// and update the appropriate file. The destination is read from the `diesel.toml`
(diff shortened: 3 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: ${{ matrix.target }}
+
+      # Install latest cross to mitigate unwind linking issue on android builds.
+      # See https://github.com/cross-rs/cross/issues/1222
```
snormore, line 43: Example of when this fails: https://github.com/xmtp/libxmtp/actions/runs/5165258906/jobs/9304664428

xmtp/Cargo.toml
```diff
@@ -43,4 +43,4 @@ url = "2.3.1"
 
 [dependencies.libsqlite3-sys]
 version = "0.26.0"
-features = ["bundled-sqlcipher"]
+features = ["bundled-sqlcipher-vendored-openssl"]
```
snormore, line 46: Fixes https://github.com/xmtp/libxmtp/actions/runs/5165745304/jobs/9305449137

.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
...
+            x86_64-linux-android,
+            i686-linux-android,
+            armv7-linux-androideabi,
+            aarch64-linux-android
+          ]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: "Cache"
```
snormore, line 25: This doesn't seem to be super effective in reducing run time of the workflow, so could probably be removed.

.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
+name: Build
+on:
+  pull_request:
```
snormore, line 3: Do we want this to run on every PR? The workflow takes about 15-20 minutes to complete.

https://github.com/bitcoindevkit/bdk-ffi/blob/master/.github/workflows/publish-android.yaml is a decent example of the full build + publish workflow that we'd want to turn this into eventually, which wouldn't run on every PR.

xmtp/src/bin/update-schema.rs
```diff
@@ -37,7 +37,7 @@ fn update_schemas_encrypted_message_store() -> Result<(), std::io::Error> {
 
     {
         // Initalize DB to read the latest table definitions
-        let _ = EncryptedMessageStore::new(StorageOption::Peristent(tmp_db.clone()), EncryptedMessageStore::generate_enc_key() ).unwrap();
+        let _ = EncryptedMessageStore::new(StorageOption::Persistent(tmp_db.clone()), EncryptedMessageStore::generate_enc_key() ).unwrap();
```
snormore, line 40: Unrelated to cross platform builds. Just fixes a typo I came across while digging through things. Can be split into separate PR if desired.

.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [
+            aarch64-unknown-linux-gnu,
+            x86_64-linux-android,
```
snormore, line 16: Might want to comment this one out until https://github.com/xmtp/libxmtp/issues/134 is addressed, if we want to merge with 🟢 CI anyway

snormore:
> Are we able to save and/or commit the build artifacts?

We can publish releases for it from a GH workflow using something like https://github.com/marketplace/actions/publish-a-release - and similarly publish releases for the kotlin/swift packages that use them, in GH releases and/or the maven/gradle package manager or whatever makes sense for those.

.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
...
+      # See https://github.com/cross-rs/cross/issues/1222
+      - name: Install rust cross
+        run: |
+          cargo install cross --git https://github.com/cross-rs/cross
+  
+      - name: Build target
+        uses: actions-rs/cargo@v1
+        with:
+          use-cross: true
+          command: build
```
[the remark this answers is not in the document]
snormore: I think it's building from root atm

xmtp/Cargo.toml
jazzz: This features forces libsqlite to statically build a copy of openssl into the project, however will cause version mismatch issues when other crates are using a non vendored version.

.github/workflows/build.yml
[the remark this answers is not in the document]
neekolas: I think we only care about cross-compiling for the FFI bindings.

jazzz (approved):
👍

neekolas:
I'd suggest we make this a non-blocking workflow (can be done in the repo settings) so that we don't slow down every PR waiting for this build step to complete.

PR #135, opened 2023-06-05
Account Persistence

jazzz:
This PR makes changes to the persistence and storage to store accounts in the encrypted datastore.

### DIFF
--- a/xmtp/migrations/2023-06_05-134021_accounts/up.sql
+++ b/xmtp/migrations/2023-06_05-134021_accounts/up.sql
@@ -1,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS accounts (
+    id INTEGER PRIMARY KEY NOT NULL,
+    serialized_key TEXT NOT NULL
+);
--- a/xmtp/migrations/2023-06_05-134021_accounts/up.sql
+++ b/xmtp/migrations/2023-06_05-134021_accounts/up.sql
@@ -1,5 +1,5 @@
 CREATE TABLE IF NOT EXISTS accounts (
     id INTEGER PRIMARY KEY NOT NULL,
     created_at INTEGER NOT NULL,
-    serialized_key TEXT NOT NULL
+    serialized_key BLOB NOT NULL
 );
--- a/xmtp/migrations/2023-06_05-134021_accounts/up.sql
+++ b/xmtp/migrations/2023-06_05-134021_accounts/up.sql
@@ -1,4 +1,5 @@
 CREATE TABLE IF NOT EXISTS accounts (
     id INTEGER PRIMARY KEY NOT NULL,
+    created_at INTEGER NOT NULL,
     serialized_key TEXT NOT NULL
 );
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -1,11 +1,13 @@
+use crate::storage::EncryptedMessageStore;
 use crate::{
     account::{Account, AccountError},
     association::{Association, AssociationError, AssociationText},
     client::{Client, Network},
     networking::XmtpApiClient,
     types::Address,
-    Errorer, InboxOwner,
+    Errorer, InboxOwner, KeyStore,
 };
+use crate::{Fetch, StorageError, Store};
 use thiserror::Error;
 
 #[derive(Error, Debug)]
@@ -25,6 +27,9 @@ pub enum ClientBuilderError {
     // StoreInitialization(#[from] SE),
     #[error("Error Initalizing Account")]
     AccountInitialization(#[from] AccountError),
+
+    #[error("Storage Error")]
+    StorageError(#[from] StorageError),
 }
 
 pub enum AccountStrategy<O: InboxOwner> {
@@ -55,7 +60,7 @@ where
 pub struct ClientBuilder<A, S, O>
 where
     A: XmtpApiClient + Default,
-    S: Default + Errorer,
+    S: KeyStore,
     O: InboxOwner,
 {
     api_client: Option<A>,
@@ -68,8 +73,9 @@ where
 impl<A, S, O> ClientBuilder<A, S, O>
 where
     A: XmtpApiClient + Default,
-    S: Default + Errorer,
+    S: KeyStore,
     O: InboxOwner,
+    S: Fetch<Account>,
 {
     const ACCOUNT_KEY: &str = "xmtp_account";
 
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -111,7 +111,7 @@ where
 
     /// Fetch account from peristence or generate and sign a new one
     fn find_or_create_account(owner: &O, store: &mut S) -> Result<Account, ClientBuilderError> {
-        let account = Self::retrieve_peristed_account(store)?;
+        let account = Self::retrieve_persisted_account(store)?;
 
         match account {
             Some(a) => {
@@ -129,7 +129,7 @@ where
     }
 
     /// Fetch Account from persistence
-    fn retrieve_peristed_account(store: &mut S) -> Result<Option<Account>, ClientBuilderError> {
+    fn retrieve_persisted_account(store: &mut S) -> Result<Option<Account>, ClientBuilderError> {
         let mut accounts = store.fetch()?;
         Ok(accounts.pop())
     }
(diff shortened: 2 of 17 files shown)


xmtp/src/lib.rs
```diff
@@ -28,12 +30,25 @@ pub trait Errorer {
...
 pub trait Store<I> {
-    fn store(&self, into: &mut I) -> Result<(), String>;
+    fn store(&self, into: &mut I) -> Result<(), StorageError>;
+}
+
+pub trait KeyStore: Default + Errorer {
+    fn get_account(&mut self) -> Result<Option<Account>, StorageError>;
+
+    fn set_account(&mut self, account: &Account) -> Result<(), StorageError>;
+}
```
jazzz, line 40: Adds trait, as the previous approach led to type issues

xmtp/src/lib.rs
```diff
@@ -28,12 +30,25 @@ pub trait Errorer {
...
+
+#[derive(Debug, Error)]
+pub enum StorageError {
+    #[error("could not Fetch: {0}")]
+    Fetch(String),
+    #[error("could not Store: {0}")]
+    Store(String),
```
jazzz, line 47: Rather than using an associated error type, the trait impls all return a `StorageError` for easier type management

neekolas: Great minds think alike. I did almost the exact same thing in my branch.

xmtp/src/builder.rs
```diff
@@ -261,25 +199,51 @@ mod tests {
...
-                .persistence(persistence)
+                .store(store_a)
                 .build()
                 .unwrap();
+        let keybytes_a = client_a.account.get_keys().curve25519.to_bytes();
+        drop(client_a);
 
-        let client_b: Client<MockXmtpApiClient, InMemoryPersistence, EncryptedMessageStore> =
+        // Reload the existing store and wallet
+        let store_b = EncryptedMessageStore::new_unencrypted(StorageOption::Peristent(
```
neekolas, line 220: Just realized we have a typo in the enum. Should be `Persistent`

xmtp/src/client.rs
```diff
@@ -30,35 +32,34 @@ pub enum ClientError {
...
 
-pub struct Client<A, P, S>
+pub struct Client<A, S>
 where
     A: XmtpApiClient,
-    P: Persistence,
 {
     pub api_client: A,
     pub network: Network,
-    pub persistence: NamespacedPersistence<P>,
```
neekolas, line 40: Love unifying on a single persistence engine for everything.

xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -132,9 +149,24 @@ impl EncryptedMessageStore {
     }
 }
 
+impl KeyStore for EncryptedMessageStore {
+    fn get_account(&mut self) -> Result<Option<Account>, StorageError> {
+        let mut account_list: Vec<Account> = self.fetch()?;
```
neekolas, line 154: Would be helpful to throw an error if this list is ever longer than a single item.

xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -145,17 +177,51 @@ impl Store<EncryptedMessageStore> for NewDecryptedMessage {
...
+impl Store<EncryptedMessageStore> for Account {
+    fn store(&self, into: &mut EncryptedMessageStore) -> Result<(), StorageError> {
+        let mut conn_guard = into
+            .conn
+            .lock()
+            .map_err(|e| StorageError::Store(e.to_string()))?;
+        diesel::insert_into(accounts::table)
+            .values(NewStoredAccount::new(json!(self).to_string()))
+            .execute(conn_guard.deref_mut())
+            .expect("Error saving account");
```
neekolas, line 202: Do we really want to panic here?

xmtp/src/storage/encrypted_store/schema.rs
```diff
@@ -9,3 +17,5 @@ diesel::table! {
         content -> Binary,
     }
 }
+
+diesel::allow_tables_to_appear_in_same_query!(accounts, messages,);
```
neekolas, line 21: Are we currently taking advantage of this?

neekolas (approved):
I will take on the brain damage of dealing with the merge conflicts between our PRs after this goes in.

I think this looks great. We can weigh out the pros and cons of my DB changes on my branch tomorrow.

xmtp/migrations/2023-06_05-134021_accounts/up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE IF NOT EXISTS accounts (
+    id INTEGER PRIMARY KEY NOT NULL,
+    created_at INTEGER NOT NULL,
+    serialized_key TEXT NOT NULL
```
neekolas, line 4: I've been storing serialized sessions as blobs. I can change to use TEXT for consistency's sake. It shouldn't really matter one way or the other.

xmtp/src/lib.rs
neekolas: FWIW, I am using the store and fetch traits in my session PR. I like them, and can see the utility for the most boring methods that have the same semantics for every model. I'll be adding some more specific methods too

xmtp/migrations/2023-06_05-134021_accounts/up.sql
jazzz: I don't have a strong preference, though blob makes more sense. In diesel the difference is whether you get a string or a Vec<u8>. I'll change

xmtp/src/lib.rs
```diff
@@ -28,12 +30,25 @@ pub trait Errorer {
 }
 
 pub trait Store<I> {
-    fn store(&self, into: &mut I) -> Result<(), String>;
+    fn store(&self, into: &mut I) -> Result<(), StorageError>;
+}
+
+pub trait KeyStore: Default + Errorer {
```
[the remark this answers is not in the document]
jazzz: My intention is to create as small of Traits as possible. 

The KeyStore trait is trait for anything that wants to store sensitive Keys.
MsgStore trait would govern controlling messages. 

It's the same reason why Store and Fetch are separate traits. Smaller scopes = More flexibility

xmtp/src/lib.rs
jazzz: @richardhuaaa The type issues stem from Client not knowing that Store<EncryptedStore> has been implemented for Account. It cannot reasonably know this unless EncryptedStore is in scope which defeats the purpose of making it abstract. 

Either use a trait to define expected behavior, or remove the Generic and make the client aware of EncryptedStore

You are right that I'm still juggling two potential paths while the direction is decided

xmtp/src/storage/encrypted_store/mod.rs
jazzz: I was planning for a future where KeyRotation is implemented and there may be more than one historical account. 

Enforcing at the Schema level would mean throwing away that data.  -- I'm not particularly opposed to that path -- just haven't given it much thought.

jazzz: @neekolas is an error preferred over "Warning + Only return the oldest"?

My fear about the an error is that developers will need a way to resolve the issue, or there install will be bricked

neekolas: That's a fair concern. I'm alright with "warning + return the newest"

xmtp/migrations/2023-06_05-134021_accounts/up.sql
jazzz: @neekolas what binary format are you using for serde? CBOR?

neekolas: I've just been naively using `serde_json::to_vec`, but I could be convinced we should standardize on CBOR. Given this database is only meant to be read by a single application, trading performance for compatibility seems reasonable.

jazzz: Awesome thats where I was leaning, but wanted to match. I'll set the tone and we can debate later

jazzz:
Github Outage is blocking this PR from getting merged. Will be merged once latest commits are showing.

https://www.githubstatus.com/incidents/1g1gkh0qpyvs

xmtp/migrations/2023-06_05-134021_accounts/up.sql
jazzz: FYI: I avoided CBOR as serde_cbor is EOL. https://crates.io/crates/serde_cbor

Should probably think a little before jump down that rabbit hole

PR #137, opened 2023-06-08
Uniffi interface for xmtpv3 and Android example

richardhuaaa:
This is a pretty barebones, scrappy initial version of running xmtpv3 end-to-end through an example Android app. The only methods exposed in this PR are creating the client and getting the account address from the client, which does a little bit under the hood, including networking calls.

I've tried to make the README.md in `bindings_ffi` as clear as possible, but please leave comments for anything I need to explain that I've missed.

The build is pretty slow. For now, dev workflow might be to just run `gen_kotlin.sh` to generate the bindings without doing any cross-compilation. This is relatively fast and can reveal any mistakes in the bindings interface. I'll add this to CI eventually.

Coming up next:
- Mutable function calls
- Pass wallet down from app
- Error handling strategy
- Run gen_kotlin.sh from CI (synchronously)
- Cross-compile from CI (asynchronously) - and make it faster ideally
- Run android integration test from CI (asynchronously)

Screen Shot 2023-06-09 at 3 43 47 PM


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "rust-analyzer.linkedProjects": ["bindings_ffi/Cargo.toml"]
```
neekolas, line 2: 🤯 

Didn't know this was an option. So, we can have separate builds but still have VSCode run Rust Analyzer on all the projects

xmtp_dh/Cargo.toml
```diff
@@ -18,10 +18,14 @@ strip = true        # Strip symbols from binary*
...
 crate-type = ["cdylib"]
 
+[[bin]]
+name = "xmtp-dh-gen"
+path = "src/bin.rs"
+
 [dependencies]
-uniffi = { version = "0.23.0" }
-uniffi_macros = "0.23.0"
+uniffi = { git = "https://github.com/mozilla/uniffi-rs", rev = "755049ebaa967fdfbb192d5b9c111353a7fee99f", features = ["tokio", "cli"] }
```
jazzz, line 26: Would it be possible to track this version information in the workspace a la :https://github.com/rust-lang/cargo/issues/3931?

Worried managing the git refs might get annoying

bindings_ffi/README.md
```diff
@@ -1,41 +1,42 @@
...
+
+## Async and concurrency
+
+We use Tokio as our multi-threaded [async runtime](https://rust-lang.github.io/async-book/08_ecosystem/00_chapter.html). Uniffi can use this runtime on async methods and objects using the annotation `#[uniffi::export(async_runtime = ‘tokio’)]`. Uniffi plumbs up an executor (scheduler) in the foreign language to the Tokio runtime in Rust. More details [here](https://github.com/mozilla/uniffi-rs/blob/734050dbf1493ca92963f29bd3df49bb92bf7fb2/uniffi_core/src/ffi/rustfuture.rs#L11-L18).
+
+Because the foreign language may be multi-threaded, any objects passed to the foreign language must be `Send` and `Sync`, and [no references to `&mut self` are permitted](https://mozilla.github.io/uniffi-rs/udl/interfaces.html#concurrent-access). TODO: interior mutability pattern we should use, sync vs async mutex/rwlock
```
neekolas, line 42: This restriction is going to be tricky to work with, but seems like it's necessary with uniffi. I assume this goes all the way down the chain (if I expose a struct that has another struct as a property, that child struct can't have any `&mut self` references either).

neekolas: Or can we maybe work around that by wrapping the child in an Arc or something?

bindings_ffi/examples/MainActivity.kt
```diff
@@ -0,0 +1,27 @@
...
+// Run setup_android_example.sh to set it up
+class MainActivity : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        runBlocking {
+            super.onCreate(savedInstanceState)
+            setContentView(R.layout.activity_main)
+
+            val textView: TextView = findViewById<TextView>(selftest_output)
+            try {
+                val client = uniffi.xmtpv3.createClient("http://10.0.2.2:5556", false);
```
neekolas, line 19: Is this an emulator specific thing? Is there ever a case where it would need to be changed to something like 127.0.0.1?

neekolas (approved):
This seems like huge progress. Great work!

PR #138, opened 2023-06-13
Proposal: Use interior mutability within Client

richardhuaaa:
## Background

We need to support some degree of concurrency; even if we avoid concurrency in the library itself, we cannot avoid multi-threading in any foreign-language callers.

We currently access the `Client` struct from multiple places: the `Conversations` struct, the `Conversation` struct, and across the uniffi interface in `bindings_ffi` (as well as any other foreign language bindings in the future).

Rust mandates that a `mut` reference is *exclusive*, which means no other location can hold a reference for as long as a mut reference is held in a single location. Our way around mutable operations on Client so far has been to wrap all references to Client in a `Mutex` everywhere we use it.

## Proposal

An alternative would be to only ever use non-`mut` references to Client everywhere. Unlike `mut` references, this is non-exclusive, which removes the need for a `Mutex` on all of the locations where `Client` is consumed. This can be achieved by upholding a new invariant:

**No public-facing method on Client can use `&mut self`**

In order to do this, we simply wrap any mutable state within Client itself with a `Mutex` or `rwlock`. In effect, we are simply pushing the mutex lower in the stack. I would suggest that in all cases of concurrent mutable access, we push it as far down as we can - for example, this PR modifies the `XmtpApiClient` trait to only use `&self` instead of `&mut self`, forcing all implementers to encapsulate any mutability internally.

## Benefits

1. **Simpler uniffi interface**: With the old approach, every method across the uniffi boundary would need to be individually wrapped with a mutex acquisition operation, turning all method calls into async calls even if they weren't originally async, and introducing boilerplate code for every method.
2. **Granularity of access**: This allows safe parallelism in calls to `Client`, as calls that do not need access to mutable data can execute in parallel to calls that do. Of course, this is not a major concern for us right now.
3. **Push complexity lower in the stack**: Rather than force all callers of `Client` to deal with mutexes, we can handle it from within `Client` only in the places that need it.
4. **Easier to send `Client` across `await`**: With non-mutable references, `Client` does not need to be `sync` to cross an `await` (assuming a multi-threaded runtime; this does not necessarily matter for Uniffi consumers).

### DIFF
--- a/xmtp/src/conversations.rs
+++ b/xmtp/src/conversations.rs
@@ -24,7 +24,7 @@ where
         wallet_address: String,
     ) -> Result<SecretConversation<A>, ConversationError> {
         let contacts = self.client.get_contacts(wallet_address.as_str()).await?;
-        let conversation = SecretConversation::new(self.client.clone(), wallet_address, contacts);
+        let conversation = SecretConversation::new(self.client, wallet_address, contacts);
 
         Ok(conversation)
     }


neekolas:
I like this a lot. That mutex on the `Client` was not aging well in its week of existence. This is a clear pattern we can use everywhere.

One thing worth noting: I don't think `Client` is the only thing we will need to expose to the FFI layer. `Conversation` is going to encapsulate a lot of the functionality applications are ultimately going to care about. We should treat that the same way we treat the `Client`

xmtp/src/networking.rs
```diff
@@ -47,18 +47,19 @@ impl Default for MockXmtpApiClient {
 #[async_trait]
 impl XmtpApiClient for MockXmtpApiClient {
     async fn publish(
-        &mut self,
+        &self,
         _token: String,
         envelopes: Vec<Envelope>,
     ) -> Result<PublishResponse, String> {
+        let mut messages = self.messages.lock().unwrap();
```
jazzz, line 54: nice

jazzz (commented):
I am torn here. 

1) removing mutability is a fantastic change. Particularly the change set implies that MockXmtpApi was the source of the issue, and this change set localized those concerns to that structure. Love it, and the outcome. 

2) I also simultaneously disagree with the premise for this change. 

> "We need to support some degree of concurrency; even if we avoid concurrency in the library itself, we cannot avoid multi-threading in any foreign-language callers."

Is this inherently true? Looking at the structure of similar libraries, the core central library is simple and the complexity is owned by the implementors. Given the intention is to consume this library in different platforms I would assume that each platform should determine what it needs.  

> We currently access the Client struct from multiple places: the Conversations struct, the Conversation struct, and across the uniffi interface in bindings_ffi (as well as any other foreign language bindings in the future).

I know this is an issue today, but I'm not convinced that it needs to be that way. Passing the Client around is a rust antipattern as it violates the single owner principal and requires more complicated solutions- Arc, Mutexs etc. Using lifetimed references is one option that would reduce conflict with the borrow checker. 

Further more if the intention is to support wasm in the future then passing around large complicated objects across the wasm boundary is effectively a non-starter. 

----

I'd like to better understand why exactly the client needs to be moved between contexts and where exactly the friction is coming from so that the source can be addressed rather than symptoms.

jazzz (approved):
I really like where this ended up! Great Work!

PR #141, opened 2023-06-16
Add integration test for xmtpv3 on Android

richardhuaaa:
This can tell us if there are any regressions in the end-to-end flow, and can eventually be added to CI.

Right now the erroring test case is hanging indefinitely rather than throwing an error, will investigate that next.

### DIFF
--- a/bindings_ffi/examples/ExampleInstrumentedTest.kt
+++ b/bindings_ffi/examples/ExampleInstrumentedTest.kt
@@ -18,7 +18,7 @@ class ExampleInstrumentedTest {
     @Test
     fun testHappyPath() {
         runBlocking {
-            val client = uniffi.xmtpv3.createClient("http://10.0.2.2:5556", false);
+            val client = uniffi.xmtpv3.createClient(EMULATOR_LOCALHOST_ADDRESS, false);
             assertNotNull("Should be able to construct client", client.walletAddress())
         }
     }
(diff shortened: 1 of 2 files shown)


bindings_ffi/examples/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,38 @@
...
+ * Instrumented test, which will execute on an Android device.
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
+    @Test
+    fun testHappyPath() {
+        runBlocking {
+            val client = uniffi.xmtpv3.createClient("http://10.0.2.2:5556", false);
```
neekolas, line 21: Can we move this address into a shared constant somewhere?

examples/android/xmtpv3_example/.gitignore
```diff
@@ -36,4 +36,5 @@ google-services.json
 jniLibs/
 xmtp_dh.kt
 xmtpv3.kt
-MainActivity.kt
\ No newline at end of file
+MainActivity.kt
+ExampleInstrumentedTest.kt
```
neekolas, line 40: Curious why we need to .gitignore this file. Is that because it gets copied over as part of the build?

neekolas (approved):

PR #142, opened 2023-06-16
Renamed contact.id to installation_id

jazzz:
When trying to work thought the contact path I found the Contact::Id concept to be hard to follow, and figured I would split this out into it's own PR for discussion. 

This PR is a small proposal to rename it to Contact::Installation_id to add clarity and match the other places it is used. Open to other suggestions however in the newer objects its referred to as a "installation_id" and in other places referred to as the Contact::Id which makes tracing the source code difficult.

### DIFF
--- a/xmtp/src/client.rs
+++ b/xmtp/src/client.rs
@@ -86,7 +86,7 @@ where
 
         if !registered_bundles
             .iter()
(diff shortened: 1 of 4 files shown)


neekolas (approved):
Makes sense to me. This will be less confusing

PR #145, opened 2023-06-19
Create Components package

jazzz:
The TopicApiClient implementation is currently only available in `bindings_ffi` however is useful in native rust Clients (Specifically in `Cli-Client`).

This PR creates a package called `xmtp_components` which makes useful implementations available in other packages.


neekolas (approved):
Makes sense to me

At some point we should think about which crates we'll want to consolidate, given how many there are. But we'll probably do a better job of that if we punt it till later where we have a full inventory of what's needed.

jazzz:
@richardhuaaa  left some notes on #140 and pre-emptively gave it a 👍 to unblock you. 

I'm not opposed to this plan; main driver is to be able to access a network implementation in as many places as possible so the solution you describe fits the bill perfectly.

I think there is an open question around if there should be 1 Package, where each domain is a module (e.g. Components/networking) or if each domain ought to get there own package (e.g. xmtp_networking, xmtp_storage, xmtp_wallet). Though I'm not picky here -- I agree with @neekolas's comments above, this is very changeable in the future.

PR #147, opened 2023-06-20
Add kotlin tests for uniffi binding

richardhuaaa:
This uses a hidden uniffi feature to run tests against our binding code [in native Kotlin](https://sourcegraph.com/github.com/mozilla/uniffi-rs@73caaf7c78c89866603bd7c4fca2bf783698bf71/-/blob/uniffi_bindgen/src/bindings/kotlin/test.rs?L29-74). This doesn't run end-to-end in a full Android environment, however it will let us know if our Uniffi interface is syntactically correct, which should be 90% of what we need (and run in a fraction of the time of a cross-platform build, ~10s).

This needs some setup on the host machine (described in the README), we probably need a docker image - excluding from CI for now.

### DIFF
--- a/bindings_ffi/README.md
+++ b/bindings_ffi/README.md
@@ -35,7 +35,7 @@ You'll need to do the following one-time setup to run Kotlin tests:
 - Install the JRE from `www.java.com`
 - Run `make install-jar` and add both jars to your CLASSPATH environment variable, for example add `export CLASSPATH=$HOME/jna/jna.jar:$HOME/jna/kotlinx-coroutines-core-jvm.jar` to your `.zshrc`
 
-If you want to skip the setup, you can also run `cargo test -- --skip kts` to only run Rust unit tests. We are currently skipping this in CI.
+If you want to skip the setup, you can also run `cargo test -- --skip kts` to only run Rust unit tests. CI will run all tests regardless.
 
 # Uniffi
 


bindings_ffi/README.md
```diff
@@ -25,6 +25,18 @@ The generated bindings (`xmtp_dh.kt`) and the cross-compiled binaries (`jniLibs`
...
+Ensure a local API host is running - run `dev/up` from the repo root.
+
+You'll need to do the following one-time setup to run Kotlin tests:
+
+- Run `brew install kotlin` to get `kotlinc`
+- Install the JRE from `www.java.com`
+- Run `make install-jar` and add both jars to your CLASSPATH environment variable, for example add `export CLASSPATH=$HOME/jna/jna.jar:$HOME/jna/kotlinx-coroutines-core-jvm.jar` to your `.zshrc`
+
+If you want to skip the setup, you can also run `cargo test -- --skip kts` to only run Rust unit tests. We are currently skipping this in CI.
+
```
jazzz, line 39: This seems reasonable to me given `bindings_ffi` is exlcuded from the workspace.

bindings_ffi/tests/test_android.kts
```diff
@@ -0,0 +1,27 @@
+import kotlinx.coroutines.*
+import kotlin.system.*
+import uniffi.xmtpv3.*
+
+runBlocking {
+    try {
+        val client = uniffi.xmtpv3.createClient("http://localhost:5556", false)
```
jazzz, line 7: There will eventually need to be a envvar override for the url in the future. But thats not a now problem.

jazzz (approved):
I can't say I full get it, but I like tests. 🚢

PR #150, opened 2023-06-26
Pass InboxOwner down from Android

richardhuaaa:
This passes the equivalent of a signer (which implements the Rust `InboxOwner`) down from our Android example app, integration test, and Kotlin test.

Next step is to improve the error handling - right now we are just passing untyped strings.

### DIFF
--- a/bindings_ffi/tests/test_android.kts
+++ b/bindings_ffi/tests/test_android.kts
@@ -35,7 +35,6 @@ runBlocking {
      }
 }
 
-// This test does not pass yet - pending https://github.com/mozilla/uniffi-rs/issues/1611
 runBlocking {
     try {
         val client = uniffi.xmtpv3.createClient(inboxOwner, "http://malformed:5556", false);


bindings_ffi/examples/MainActivity.kt
```diff
@@ -1,29 +1,50 @@
...
 
+class Web3jInboxOwner(private val credentials: Credentials) : FfiInboxOwner {
+    override fun getAddress(): String {
+        return credentials.address
+    }
+
+    override fun sign(text: String): ByteArray {
+        val messageBytes: ByteArray = text.toByteArray(StandardCharsets.UTF_8)
+        val signature = Sign.signPrefixedMessage(messageBytes, credentials.ecKeyPair)
+        return signature.r + signature.s + signature.v
```
jazzz, line 25: Nothing wrong here. But in the future we'll need to watch out which encoding the signing library is using for the V value. In this case bouncycastle is using eip155. but not always the case

jazzz (approved):
Looks good form what I can see

PR #151, opened 2023-06-26
Proposal: State machine for sending messages

richardhuaaa:
This is a very ham-fisted (and incomplete) proposal for a state machine for message sending. Feel free to tear apart the details, mainly throwing something initial down to get to consensus on the overall approach. Suggested principles here:

1. We should persist data to the DB when possible rather than holding it in-memory. Some data, like Olm session data, would be single-use - not held in memory longer than the duration of a function call. Because:
    1. We need singletons - it would be bad to have separate copies of the same session in-memory on different conversations. We can use the DB as the synchronization layer here.
    1. Devices can shut off at any time - we don’t want to lose data.
1. We need a clearly defined state machine for the data we store in the DB. Because:
    1. We need to be able to continue from where we were if the device shuts off mid-operation. Message sending can involve multiple network round-trips and there is a large permutation of possible session/conversation/user states. Resetting encryption state on decryption failure will be similarly complicated.
    1. This makes it easier to understand the overall system, separate data into different structs, and split up the work required.

This file won't actually be committed (I'm modifying a generated diesel file here) - but if we end up agreeing to something like this, I'll move the description to a README and make the required code changes to generate the schema.

### DIFF
--- a/xmtp/README.md
+++ b/xmtp/README.md
@@ -15,29 +15,26 @@
 
 ### States
 
+```
 Conversation:
-
-- UNINITIALIZED: No invites have been sent
-- INVITED: Invites have been sent
+    - UNINITIALIZED: No invites have been sent
+    - INVITED: Invites have been sent
 
 User:
-
-- LAST_REFRESHED: The local timestamp at which an updated list of installations and pre-keys was requested for that user (and successfully received)
+    - LAST_REFRESHED: The local timestamp at which an updated list of installations and pre-keys was requested for that user (and successfully received)
 
 Installation:
-
-- UNINITIALIZED: There is no session state with that installation (no prekey messages were sent yet)
-- SESSION_CREATED: There is existing session state with that installation
+    - UNINITIALIZED: There is no session state with that installation (no prekey messages were sent yet)
+    - SESSION_CREATED: There is existing session state with that installation
 
 Message:
-
-- UNINITIALIZED: The message has not been encrypted yet
-- LOCALLY_COMMITTED: The outbound payloads have been constructed
+    - UNINITIALIZED: The message has not been encrypted yet
+    - LOCALLY_COMMITTED: The outbound payloads have been constructed
 
 Outbound Payload:
-
-- PENDING: The payload has not been confirmed as sent yet
-- SERVER_ACKNOWLEDGED: The payload has been acknowledged by the server
+    - PENDING: The payload has not been confirmed as sent yet
+    - SERVER_ACKNOWLEDGED: The payload has been acknowledged by the server
+```
 
 ### Creating a conversation
 


xmtp/src/storage/encrypted_store/schema.rs
```diff
@@ -1,5 +1,75 @@
...
+                     Send invite as prekey message
+                     Set installation state to PREKEY_MESSAGE_SENT
+                 Else if installation.state == PREKEY_MESSAGE_SENT
+                     Send invite as ratchet message
+         Set conversation.state = INVITES_SENT
+    If conversation.state == INVITES_SENT:
+        For each user in the conversation (including self):
+            For each installation of the user:
+                installation.state must be PREKEY_MESSAGE_SENT:
+                    Send message as ratchet message
```
neekolas, line 38: If the conversation was initiated by you, wouldn't you need to continue sending prekey messages until you've received a reply?

neekolas (commented):
Appreciate the clarity around the proposal. Completely agree message sending is a complicated enough flow we should treat it as a state machine.

> We should persist data to the DB when possible rather than holding it in-memory. Some data, like Olm session data, would be single-use - not held in memory longer than the duration of a function call. Because:

I like this proposal. Having the session objects be short-lived and database driven is cleaner. And we don't have to worry about shared ownership, since sessions can be discarded as soon as we are done with them.

I'll give the specifics of the proposal a more thorough look tomorrow.

xmtp/src/storage/encrypted_store/schema.rs
```diff
@@ -1,5 +1,75 @@
 // @generated automatically by Diesel CLI.
 
+/*
+SIMPLIFYING ASSUMPTIONS:
+- The installations in a conversation do not change after the conversation is initialized. Can sketch out plan for adding/removing installations from existing conversations later.
+- We do not include message data on invites. Ideally we start including messages directly on invites later.
+- Every state update is a DB write. This allows us to resume on cold start.
+- On cold start, we can scan the DB for UNINITIALIZED messages and resume sending them.
+- Every message and invite has a randomly generated id. This is used to deduplicate on the receiving side. If a message is received with an id that already exists in the DB, it is ignored (after advancing the ratchet state).
```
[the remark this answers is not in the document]
neekolas: If we are hashing the encrypted content, they should always be unique because of nonces. Even if we don't include an inner envelope that includes a timestamp like we do for V1/V2.

xmtp/src/storage/encrypted_store/schema.rs
neekolas: Maybe it's possible in principle, but I can't make Vodozemac behave that way. If you create a new outbound session and encrypt two messages in a row, I can confirm both are `PreKey` messages.

You can see it in the Vodozemac code here:

```rust
    pub fn encrypt(&mut self, plaintext: impl AsRef<[u8]>) -> OlmMessage {
        let message = match self.config.version {
            Version::V1 => self.sending_ratchet.encrypt_truncated_mac(plaintext.as_ref()),
            Version::V2 => self.sending_ratchet.encrypt(plaintext.as_ref()),
        };

        if self.has_received_message() {
            OlmMessage::Normal(message)
        } else {
            let message = PreKeyMessage::new(self.session_keys, message);

            OlmMessage::PreKey(message)
        }
    }
```

xmtp/src/storage/encrypted_store/schema.rs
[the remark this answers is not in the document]
jazzz: > we should store individual copies of the encrypted payloads for each receiver in the DB until they are confirmed as sent,

The other option is to only persist the Session when the Network Accepts the message. That way state is rolled back appropriately on an error and can be retried later.

jazzz:
Love the push towards a state machine here. I think it will help Identify error cases in the design phase and manage complexity through implementation.

xmtp/README.md
```diff
@@ -0,0 +1,104 @@
...
+    return success
+
+processMessages():
+    For each message in UNINITIALIZED state, processing in order of timestamp:
+        Fetch the conversation's users from the DB (including self)
+        For each user:
+            If user.last_refreshed is uninitialized or more than THRESHOLD ago:
+                refreshUserInstallations()    // Could be kicked off asynchronously or synchronously
+                return  // refreshUserInstallations() will call back into processMessages() when ready
+        Fetch the installations of both users from the DB
```
jazzz, line 42:
```suggestion
        Fetch the installations of all users from the DB
```

xmtp/README.md
```diff
@@ -0,0 +1,104 @@
...
+        For each installation:
+            // Build the plaintext payload
+            If conversation.state == UNINITIALIZED:
+                Construct the payload as an invite with the message attached
+            Else if conversation.state == INVITED:
+                Construct the payload as a message
+            // Encrypt the payload
+            If installation.state == UNINITIALIZED:
+                Create an outbound session (hold it in memory)
```
jazzz, line 51: What should happen in the case there is no session for channel but there is a conversation?

xmtp/README.md
```diff
@@ -0,0 +1,104 @@
...
+            Use the existing session to encrypt a payload containing an invite with the message attached (hold it in memory)
+        In a single transaction:
+            Push all encrypted payloads to outbound_payloads table with outbound_payload.state = UNINITIALIZED
```
jazzz, line 54:
```suggestion
            Push all encrypted payloads to outbound_payloads table with outbound_payload.state = PENDING
```

The state is technically initialized it just hasn't been sent yet

xmtp/README.md
```diff
@@ -0,0 +1,104 @@
...
+            Commit all updated session states to the DB with installation.state = SESSION_CREATED
+            Set message.state = LOCALLY_COMMITTED
+            Set conversation.state = INVITED
+            Delete all data from memory
```
jazzz, line 58: Does this need to be a zeroize operation? or would a drop be sufficent?

xmtp/README.md
```diff
@@ -0,0 +1,104 @@
...
+refreshUserInstallations(user):
+    Fetch installations/contact bundles for the user from the network
+    Fetch installations/contact bundles for the user from the DB
+    In a single transaction:
+        For each installation from the DB:
+            If it doesn't exist in the network contact bundles or is expired or revoked, delete it from the DB
+        For each installation from the network:
+            If it doesn't exist in the DB, insert it with installation.state = UNINITIALIZED
+        Set user.last_refreshed to NOW
+    processMessages();  // Could be kicked off asynchronously or synchronously
```
jazzz, line 71: In this function the outer loop is over `UNINITIALIZED` messages, which in this case there are not guaranteed to be any -- Which would mean the initialization of installations would not be processed until a message is sent.

Is this intentional?

xmtp/README.md
```diff
@@ -0,0 +1,104 @@
...
+    return
+
+processPayloads():
+    For each outbound payload in in UNINITIALIZED state, processing in order of sequential ID (possibly batched):
```
jazzz, line 75: > processing in order of sequential ID

Is this critical(MUST)? or a suggestion (SHOULD)?

jazzz (approved):
I like where this is headed. Think there are more corner cases to discover, but I like that this could be the living document of the processing logic

PR #153, opened 2023-06-28
Make swift release artifact from tagged commits

mkobetic:
Automate the process described in https://github.com/xmtp/libxmtp/blob/main/bindings_swift/README.md#workflow. Introduces a "Release" github workflow that in response to a tagged commit:
* runs the xcframework generation
* updates the xmtp-rust-swift repo
* tags the update in xmtp-rust-swift with the same tag

Slight modification of the process is that instead of mixing the source file munging with copying over to the `xmtp-rust-swift` repo the tweaked shell script only does the munging on a copy of the source files and the whole thing is packaged into a zip file that can be unzipped over the `xmpt-rust-swift` repo to update it.

The workflow also patches the version in XMTPRust.podspec to match the new tag. If the format of the spec file changes the sed rule may need to be updated.

To allow the workflow to push to another repo I followed [this guide](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow#authenticating-with-a-github-app) and installed [this app on the org](https://github.com/organizations/xmtp/settings/apps/libxmtp-release). The relevant secrets are stored only [in this repo](https://github.com/xmtp/libxmtp/settings/secrets/actions). If additional repos are added to this workflow they MUST be added to [this installation](https://github.com/organizations/xmtp/settings/installations/39118494) of the app.

I tested this by modifying the workflow to fetch/push to [`test-auto-release` branch in the `xmtp-rust-swift` repo](https://github.com/xmtp/xmtp-rust-swift/tree/test-auto-release). I was a bit puzzled by the [diffs shown for that branch](https://github.com/xmtp/xmtp-rust-swift/compare/test-auto-release), they seem as if the binary file for ios-arm64 was dropped, but I checked it's there, so I'm not sure what the binary diffs mean there.

Note that a [successful run](https://github.com/xmtp/libxmtp/actions/runs/5424322027/jobs/9863517364) of this workflow takes about 20 minutes, largely spent compiling the framework for the different architectures. I tried to make use of the 3 cores which should be available on macos runners by trying to enable concurrency in make `make -j 3 swift`, but that seemed to make things worse for some reason. So I gave up.

Anyway, we presumably want to push these to the `main` branch but I didn't want to push to `main` from this dev branch, so I propose to make a final test with the `main` branch after we merge this. For that we should pick a version tag that we want to put on the current version of libxmtp (the latest tag from `xmtp-rust-swift` is [`0.3.0-beta0`](https://github.com/xmtp/xmtp-rust-swift/releases/tag/0.3.0-beta0). Also note that we need to keep pushes to `main` open on the `xmtp-rust-swift` repo, otherwise we will break this workflow.

I intend to update the readme in `xmtp-rust-swift` to indicate that the repo is updated automatically from this repo after we merge this.

# Notes

I was wondering if we could reuse the artifacts from the Build workflows, instead of rebuilding everything here again, but I'm not sure if those are the same thing that we need for this.


.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+    - '*'
+
+jobs:
+
+  swift:
+    runs-on: macos-12
+    permissions:
+      contents: write
+    env:
+      RELEASE_BRANCH: test-auto-release # do we want to push to main directly?
```
mkobetic, line 15: Don't forget to fix this.

bindings_swift/update_spec_version.sed
```diff
@@ -0,0 +1 @@
+/^ *s.version *=/s/= *'.*'/= 'VERSION'/
```
mkobetic, line 1: I put this into a file because the quoting issues trying to inline this into the action were a pain to deal with and the debug cycle is painful and slow.

bindings_swift/Makefile
```diff
@@ -50,11 +50,11 @@ framework: move_gen $(LIB)
 		-library ./target/aarch64-apple-ios/release/libxmtp_rust_swift.a \
 		-headers ./include/ \
 		-output XMTPRustSwift.xcframework
-	rm -f bundle.zip
```
mkobetic, line 53: moved this to the swift target

neekolas:
> Also note that we need to keep pushes to main open on the xmtp-rust-swift repo, otherwise we will break this workflow.

I wonder if we can setup Github permissions to only allow the bot user to push to main, but not any other users

neekolas (approved):

PR #156, opened 2023-07-06
Disable hanging Android test

richardhuaaa:
Tests that initialize a second client on Android sometimes hang and never complete. Disabling this for now so that it doesn't block anyone, but I will come back to investigate this later.


jazzz (approved):
⏩

PR #157, opened 2023-07-11
Persist conversations and users

richardhuaaa:
Simple PR to create the conversation and user rows in the database when `Conversations::new_secret_conversation()` is called.

### DIFF
--- a/examples/cli/cli-client.rs
+++ b/examples/cli/cli-client.rs
@@ -175,7 +175,7 @@ async fn send(client: Client, addr: &String, msg: &String) -> Result<(), CliErro
 
         let om = session.encrypt(msg.as_bytes());
         info!("{:?} ", om);
-        session.save(&client._store).unwrap();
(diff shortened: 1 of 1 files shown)


xmtp/src/conversation.rs
```diff
@@ -40,17 +43,37 @@ impl<'c, A> SecretConversation<'c, A>
...
+            .insert_or_ignore_conversation(StoredConversation {
+                peer_address: obj.peer_address(),
+                convo_id: obj.convo_id(),
+                created_at: now(),
+                convo_state: ConversationState::Uninitialized as i32,
+            })?;
+        Ok(obj)
+    }
+
+    pub fn convo_id(&self) -> String {
```
neekolas, line 73: We might want to move this into a utility function. I can see convo ids being created from places other than the conversation (to find conversations from an invite, for example).

neekolas (approved):

PR #159, opened 2023-07-11
Invite reading flow

neekolas:
## Summary

Proposed state machine for reading invites

### DIFF
--- a/xmtp/README.md
+++ b/xmtp/README.md
@@ -129,45 +129,49 @@ init():
 
 ### Receiving invites
 
+```plaintext
 downloadInvites():
-    get the `refresh_jobs` record with an id of `invites`, and obtain a lock on the row:
-        store `now()` in memory to mark the job execution time
-        fetch all messages from invite topic with timestamp > refresh_job.last_run - PADDING_TIME # PADDING TIME accounts for eventual consistency of network. Maybe 30s.
-        for each message in topic:
-            save (or ignore if already exists) raw message to inbound_invite table with status of PENDING
-            update `refresh_jobs` record last_run = current_timestamp
+    Get the `refresh_jobs` record with an id of `invites`, and obtain a lock on the row:
+        Store `now()` in memory to mark the job execution time
+        Fetch all messages from invite topic with timestamp > refresh_job.last_run - PADDING_TIME # PADDING TIME accounts for eventual consistency of network. Maybe 30s.
+        For each message in topic:
+            Save (or ignore if already exists) raw message to inbound_invite table with status of PENDING
+        Update `refresh_jobs` record `last_run = current_timestamp`
 
 processInvites():
-    for each inbound_invite in PENDING state:
-        if an existing session exists with the `inviter.installation_id`:
-            decrypt the inner invite using the existing session
-            if decryption fails:
-                set inbound_invite state to DECRYPTION_FAILURE
+    For each inbound_invite in PENDING state:
+        If an existing session exists with the `inviter.installation_id`:
+            Decrypt the inner invite using the existing session
+            If decryption fails:
+                Set inbound_invite state to DECRYPTION_FAILURE
                 continue
+            Update session in the database
         else:
-            create a new inbound session with the inviter
-            decrypt the inner invite using the new session
-            if decryption fails:
-                set inbound_invite state to DECRYPTION_FAILURE
+            Create a new inbound session with the inviter
+            Decrypt the inner invite using the new session
+            If decryption fails:
+                Set inbound_invite state to DECRYPTION_FAILURE
                 continue
-            persist the session to the database
-        
-        if invite validation fails:
-            set inbound_invite state to INVALID
+            Persist the session to the database
+
+        If invite validation fails:
+            Set inbound_invite state to INVALID
             continue
-        
-        fetch the existing conversation with convo_id derived from inner invite
-        if conversation with convo_id doesn't already exist in DB:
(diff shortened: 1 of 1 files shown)


xmtp/README.md
```diff
@@ -13,12 +13,24 @@
 - Repeated sends of the same payload should be idempotent. When receiving a message or invite, the receiving side will store the hash of the encrypted payload alongside the decrypted result. If a message is received with an id that already exists in the DB, it is ignored.
 - We have ignored race conditions for now (as network requests may take different amounts of time). The receiver side should be tolerant of out-of-order payloads. If ordering is a must, it is possible to use multi-producer, single-consumer queues, or singleton threads for processMessages() and processPayloads() that run on an interval.
 
+### Helper database models
+
+`refresh_jobs`:
+
+```sql
+CREATE TABLE IF NOT EXISTS refresh_jobs (
```
neekolas, line 21: I'll remove this from the README once the table actually exists. Just figured it was helpful for clarity now.

xmtp/README.md
```diff
@@ -13,12 +13,24 @@
...
+)
+```
+
 ### States
 
 ```
 Conversation:
     - UNINITIALIZED: No invites have been sent
     - INVITED: Invites have been sent
+    - INVITE_RECEIVED: You have been invited to this conversation by another installation
```
neekolas, line 33: Technically we could just re-use `INVITED` and everything would be fine. Just don't want to throw out the context as to whether you were the inviter or invitee in a convo. Feels like it might be useful later.

xmtp/README.md
```diff
@@ -30,10 +42,17 @@ Installation:
...
 
 Outbound Payload:
     - PENDING: The payload has not been confirmed as sent yet
     - SERVER_ACKNOWLEDGED: The payload has been acknowledged by the server
+
+Inbound invite:
+    - PENDING: The payload has been downloaded from the network but has not been processed yet
+    - PROCESSED: The invite has been successfully processed and the conversation has been created
+    - DECRYPTION_FAILURE: The inner invite failed to decrypt
+    - INVALID: The invite failed validation
```
jazzz, line 55: For clarity this state the message decrypted successfully, however the inner packet could not be decoded?

xmtp/README.md
```diff
@@ -108,7 +127,50 @@ init():
...
+                Set inbound_invite state to DECRYPTION_FAILURE
+                continue
+            Update session in the database
+        else:
+            Create a new inbound session with the inviter
+            Decrypt the inner invite using the new session
+            If decryption fails:
+                Set inbound_invite state to DECRYPTION_FAILURE
+                continue
+            Persist the session to the database
```
jazzz, line 155: In the case there the decryption fails for a new session/PreKeyMessage, shouldn't the session be persisted regardless.

The session is valid outside of the invite, and subsequent messages from the remote_installation will continue to use the same session, albeit with increasing chain index values. By throwing away the session information does this not mean that all subsequent messages will need to be discarded even if they are valid?

neekolas: I don't think there is anything to save or update when decryption fails.

For creating a new inbound session, you need to decrypt the first inbound message as part of session creation. If that fails, there is no session to persist.

For decrypting a message using an existing session, if decryption fails the ratchet state would be unchanged so there is no need to update anything. The next message would just be treated as an out-of-order message from the perspective of the ratchet state.

xmtp/README.md
```diff
@@ -108,7 +127,53 @@ init():
     return
 ```
 
-### Receiving an invite
+### Receiving invites
+
+```plaintext
+downloadInvites():
+    Get the `refresh_jobs` record with an id of `invites`, and obtain a lock on the row:
```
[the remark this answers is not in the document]
neekolas: That's right. It'll block if we use a SQL `SELECT FOR UPDATE ...`.

Not strictly necessary, but it'll help avoid situations where we download the same invites twice (although the state machine should be idempotent in the case that we did).

xmtp/README.md
```diff
@@ -108,7 +127,53 @@ init():
...
+        Store `now()` in memory to mark the job execution start
+        Fetch all messages from invite topic with timestamp > refresh_job.last_run - PADDING_TIME # PADDING TIME accounts for eventual consistency of network. Maybe 30s.
```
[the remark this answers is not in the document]
neekolas: I thought about that. We do similar things in `xmtp-js`. The issue we run into is that with the eventual consistency of the network, we need to pad it by some number of seconds. That means we would be constantly reprocessing the most recent invite on subsequent calls (since the last timestamp - PADDING will always return at least one result).

xmtp/README.md
```diff
@@ -108,7 +127,53 @@ init():
...
+            continue
+        If an existing session exists with the `inviter.installation_id`:
+            Decrypt the inner invite using the existing session
+            If decryption fails:
+                Set inbound_invite state to DECRYPTION_FAILURE
+                continue
+            Update session in the database
+        else:
+            Create a new inbound session with the inviter
+            Decrypt the inner invite using the new session
```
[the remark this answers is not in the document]
neekolas: Hmm. Good point.

If we chose to overwrite it in this race condition, we could run into a case where the two installations actually swap sessions (since both would have non-matching invites and would overwrite their session with the other's). That's just as bad as doing nothing.

But having to support multiple sessions would mean we would need some sort of `session_id` on all messages to disambiguate.

Think I'll look into this some more. Olm must have run into something similar and had to resolve it somehow.

PR #163, opened 2023-07-18
release action for android

mkobetic:
This PR extends the `Release` action (triggered by tagging a commit) to create a release in this repo containing a zip archive with the shared libraries and the kotlin glue code that can be unzipped over the example app or the android sdk to update it to the released version of libxmtp.

The archive structure follows the pattern established by the `setup_android_example.sh` script, i.e. a properly structured `jniLibs` directory and the kotlin glue code in `java/xmtpv3.kt`. We may want to replace the zip archive with a different packaging approach in the future, but given that bare libxmtp is not intended for public consumption a more opaque approach may not be a bad choice.

The process is implemented in two stages. First it expands the existing `Build` action to not just build all the shared libraries, but to also capture them as short-lived artifacts, adds another step `kotlin` that builds the kotlin glue (relying on the `gen_kotlin.sh` script), and then `package` combines all these artifacts in a zip archive which is uploaded as a longer living artifact. The build action is executed for every push and main merge, so most of the time the archives end up unused and dropped after the set amount of time (GH default 90 days).

Second stage is associated with the `Release` action (triggered by tagging a commit, usually on the main branch). The `ffi` step will find the artifacts produced by the build action for the associated commit, downloads the zip archive and creates a libxmtp release with the archive as its artifact. This makes the archive accessible from anywhere for as long as the release exists. Obviously the release needs to be triggered while the corresponding zip archive artifact exists (i.e. within 90 days). We may want to shorten this period, I'm not sure what the storage costs will be for 90 days worth of commits.

Sample result of the release action can be seen in https://github.com/xmtp/libxmtp/releases/tag/test-release-android. The release uses the tag used to trigger it, in this case the tag was `test-release-android`. 

Additional notes inline.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,3 +51,73 @@ jobs:
           use-cross: true
           command: build
           args: --release --target ${{ matrix.target }} --manifest-path bindings_ffi/Cargo.toml --target-dir bindings_ffi/target
+
+      - name: Upload binary
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ matrix.target }}.so
+          path: bindings_ffi/target/${{ matrix.target }}/release/libbindings_ffi.so
+          retention-days: 1
+
+  kotlin:
+    name: kotlin
+    runs-on: macos-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            ./target/release
+          key: ${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: x86_64-apple-darwin
+      - name : Install kotlin
+        run: brew install ktlint 
+      - name: Generate Bindings
+        working-directory: bindings_ffi
+        run: ./gen_kotlin.sh
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: xmtpv3.kt
+          path: bindings_ffi/src/uniffi/xmtpv3/xmtpv3.kt
+          retention-days: 1
+
+  package:
+    name: package
+    needs: [ build, kotlin ]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          path: .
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -51,3 +51,73 @@ jobs:
           use-cross: true
           command: build
           args: --release --target ${{ matrix.target }} --manifest-path bindings_ffi/Cargo.toml --target-dir bindings_ffi/target
+
+      - name: Upload binary
```
mkobetic, line 55: We need these artifacts only briefly until the `package` job bellow is able to collect them all and put them into the zip archive.

.github/workflows/build.yml
```diff
@@ -51,3 +51,73 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ matrix.target }}.so
```
mkobetic, line 58: Note that I'm using the target name as the library name so that I can disambiguate the library artifacts in the `package` step. They are renamed again when they are assembled into the zip archive.

.github/workflows/build.yml
```diff
@@ -51,3 +51,73 @@ jobs:
...
+          retention-days: 1
+
+  package:
+    name: package
+    needs: [ build, kotlin ]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
```
mkobetic, line 102: We're not specifying artifact name, so all artifacts are downloaded.

.github/workflows/release.yml
```diff
@@ -59,3 +59,21 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # for the release action
+    steps:
+      - uses: actions/checkout@v3
+      - name: download package
+        env:
+          HEAD_SHA: ${{ github.sha }}
+          GITHUB_REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ github.token }}
```
mkobetic, line 73: Token is needed to be able to use `gh` in the script.

nplasterer (approved):
This looks good from my perspective. So the final steps here is downloading that zip, removing the existing files that were previously in the android sdk, and then unzipping it in the android sdk? Seems easy and straight forward.

mkobetic:
> This looks good from my perspective. So the final steps here is downloading that zip, removing the existing files that were previously in the android sdk, and then unzipping it in the android sdk? Seems easy and straight forward.

I assume that most of the time it's safe to just unzip over the target without removing anything. The structure should be pretty stable, the set of supported architectures shouldn't change much and there's only one kotlin file generated. We'll see.

PR #165, opened 2023-07-20
Schema change: Add outbound_payloads table, message state column

richardhuaaa:
Small PR to update the schema in preparation for message sending work.


xmtp/migrations/2023-07-19-223017_add_message_state/down.sql
```diff
@@ -0,0 +1,2 @@
+-- Columns cannot be dropped in sqlite without dropping the whole table first
+-- Instead of removing the column, simply do nothing and ignore the column in code
```
jazzz, line 2: 🧠

xmtp/src/storage/encrypted_store/models.rs
```diff
@@ -36,6 +42,7 @@ pub struct DecryptedMessage {
     pub convo_id: String,
     pub addr_from: String,
     pub content: Vec<u8>,
+    pub state: i32,
```
jazzz, line 45: For Clarity: Am I correct in that DecryptedMessage is going to be used for local never encrypted messages? Outbound messages are also going be DecryptedMessages? Nothin against this, just attempting to stay in the loop

jazzz, line 45: Out of scope for now, but Decrypted Message will want to have a field 'install_from' to record which installation the message originated from.

jazzz (approved):

PR #166, opened 2023-07-21
Persist plaintext message when sending

richardhuaaa:
Most of this PR is renaming, unit tests, and CLI updates. The only concrete change is we add a method called `sendMessage()` to `Conversation` that persists the message in the DB with UNPROCESSED state.


jazzz (approved):
🐚

PR #167, opened 2023-07-24
Download invites

neekolas:
## Summary

- Creates table for `refresh_jobs`
- Creates helper function for getting the latest refresh job in a transaction and doing something with it
- Adds code for downloading invites

## Next up
- Storing the invites in the DB
- Figure out a better solution for transactionality. See [this thread](https://xmtp-labs.slack.com/archives/C05GBHEAGPP/p1690192829199419)

## Notes
- I realized that row level locking in SQLite isn't a thing...instead the entire database is locked for the transaction. Given how little concurrency is needed in our app, that may be fine, but is something we should keep an eye on.

### DIFF
--- a/xmtp/src/storage/encrypted_store/mod.rs
+++ b/xmtp/src/storage/encrypted_store/mod.rs
@@ -564,7 +564,7 @@ mod tests {
         .unwrap();
 
         store
-            .lock_refresh_job(RefreshJobKind::Invite, |conn, job| {
+            .lock_refresh_job(RefreshJobKind::Invite, |_, job| {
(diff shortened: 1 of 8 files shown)


xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -246,6 +246,36 @@ impl EncryptedMessageStore {
 
         Ok(msg_list)
     }
+
+    pub fn lock_refresh_job<F>(&self, kind: RefreshJobKind, cb: F) -> Result<(), StorageError>
```
neekolas, line 250: This feels right on the line of being too application specific to live in the store. Might move to somewhere else.

PR #168, opened 2023-07-25
Inbound invite table

neekolas:
## Summary

- Creates table for `inbound_invites`
- Saves invites to the database from envelopes, rather than converting to `Invitation` structs
- Creates a more generic method for getting the latest messages from a topic


jazzz (approved):
Very clean. 👍

PR #169, opened 2023-07-25
Refresh Sessions

jazzz:
This PR adds the ability to detect new installations, so they can be included in message fanout.

Not covered in this PR:
- generating sessions for new installations
- integration of refresh_installations

### DIFF
--- a/xmtp/migrations/2023-07-10-223309_add_installations/up.sql
+++ b/xmtp/migrations/2023-07-10-223309_add_installations/up.sql
@@ -3,7 +3,6 @@ CREATE TABLE installations (
     user_address TEXT NOT NULL,
     first_seen_ns BIGINT NOT NULL,
     contact BLOB NOT NULL,
-    contact_hash TEXT NOT NULL,
     expires_at_ns BIGINT,
     FOREIGN KEY(user_address) REFERENCES users(user_address)
 );
--- a/xmtp/src/client.rs
+++ b/xmtp/src/client.rs
@@ -1,6 +1,8 @@
 use core::fmt;
 use std::fmt::Formatter;
 
+use diesel::Connection;
+use log::info;
 use thiserror::Error;
 use vodozemac::olm::OlmMessage;
 
@@ -8,12 +10,13 @@ use crate::{
     account::Account,
     contact::{Contact, ContactError},
     session::SessionManager,
-    storage::{EncryptedMessageStore, StorageError, StoredInstallation},
+    storage::{EncryptedMessageStore, StorageError, StoredInstallation, StoredSession},
     types::networking::{PublishRequest, QueryRequest, XmtpApiClient},
     types::Address,
-    utils::{build_envelope, build_user_contact_topic},
+    utils::{build_envelope, build_user_contact_topic, key_fingerprint},
     Store,
 };
+use std::collections::HashMap;
 use xmtp_proto::xmtp::message_api::v1::Envelope;
 
 #[derive(Clone, Copy, Default, Debug)]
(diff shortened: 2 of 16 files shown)


xmtp/src/client.rs
```diff
@@ -32,6 +35,8 @@ pub enum ClientError {
     PublishError(String),
     #[error("storage error: {0}")]
     Storage(#[from] StorageError),
+    #[error("dieselError: {0}")]
+    Ddd(#[from] diesel::result::Error),
```
jazzz, line 39: With the client now performing transactions directly, having diesel errors be converted directly to ClientErrors is easier. Open to another Error handing strategy

xmtp/src/client.rs
```diff
@@ -142,19 +147,56 @@ where
...
+                    self.store
+                        .insert_or_ignore_install(install, transaction_manager)?;
+                    self.store.insert_or_ignore_session(
+                        StoredSession::try_from(&session)?,
+                        transaction_manager,
+                    )?;
+                }
+
+                Ok(())
+            },
```
jazzz, line 183: This is the core of the refresh. The network calls are all done in advance to minimize work being done in the transaction lambda.

xmtp/src/storage/errors.rs
```diff
@@ -14,8 +15,10 @@ pub enum StorageError {
...
     #[error(transparent)]
     ImplementationError(#[from] anyhow::Error),
+    #[error("ContactError")]
+    ContactError(#[from] ContactError),
     #[error("serialization error")]
     SerializationError,
-    #[error("unknown storage error")]
-    Unknown,
+    #[error("unknown storage error: {0}")]
+    Unknown(String),
```
jazzz, line 23: UnknownErrors were not helpful. Added a string value to make them useful

xmtp/src/utils.rs
```diff
@@ -30,3 +32,7 @@ pub fn build_envelope(content_topic: String, message: Vec<u8>) -> Envelope {
 pub fn base64_encode(bytes: &[u8]) -> String {
     general_purpose::STANDARD_NO_PAD.encode(bytes)
 }
+
+pub fn key_fingerprint(key: &Curve25519PublicKey) -> String {
+    base64_encode(keccak256(key.to_string().as_str()).as_slice())
+}
```
jazzz, line 38: This makes the base64 encoded hash of a publickKey a named operation which can be used form multiple files consistently

xmtp_networking/src/lib.rs
```diff
@@ -3,12 +3,13 @@ pub mod grpc_api_helper;
 pub const LOCALHOST_ADDRESS: &str = "http://localhost:5556";
 pub const DEV_ADDRESS: &str = "https://dev.xmtp.network:5556";
 
+pub use grpc_api_helper::Client;
+
```
jazzz, line 7: Public reexport, which makes it easier to use in applications

xmtp/migrations/2023-07-10-223309_add_installations/up.sql
```diff
@@ -3,7 +3,6 @@ CREATE TABLE installations (
     user_address TEXT NOT NULL,
     first_seen_ns BIGINT NOT NULL,
     contact BLOB NOT NULL,
-    contact_hash TEXT NOT NULL,
```
jazzz, line 6: Contact hash doesn't really help at this moment. Removing. This is a breaking change as it was removed from the migration directly

xmtp/src/client.rs
jazzz: I don't see why it wouldn't be possible, however that would not be my current preference. I hear your desire to have a central data management interface. If going that route I'd prefer that a separate object be built for that purpose, rather than creating monolithic structs.

A Two Struct approach would be:
StorageObject: (EncryptedStore -> SqlcipherStore)
QueryManager:

Separating the two would allow ownership of the underlying storageObject to be more flexible, and better facilitate testing. 

Very open to discussion on this

jazzz:
SenGoing to merge this, open questions around storage can be pulled out into another change set

PR #171, opened 2023-07-31
Encode using text codec when sending message

richardhuaaa:
We need to be able to construct the [`EncodedContent` protobuf](https://github.com/xmtp/proto/blob/xmtpv3/proto/message_contents/content.proto#L30) when sending a message, which will allow us to add support for non-text formats in the future.

This PR implements a text codec, similar to the [pattern used in xmtp-js](https://github.com/xmtp/xmtp-js/blob/main/src/codecs/Text.ts#L17).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,3 +51,73 @@ jobs:
           use-cross: true
           command: build
           args: --release --target ${{ matrix.target }} --manifest-path bindings_ffi/Cargo.toml --target-dir bindings_ffi/target
+
+      - name: Upload binary
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ matrix.target }}.so
+          path: bindings_ffi/target/${{ matrix.target }}/release/libbindings_ffi.so
+          retention-days: 1
+
+  kotlin:
+    name: kotlin
+    runs-on: macos-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            ./target/release
+          key: ${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: x86_64-apple-darwin
+      - name : Install kotlin
+        run: brew install ktlint 
+      - name: Generate Bindings
+        working-directory: bindings_ffi
+        run: ./gen_kotlin.sh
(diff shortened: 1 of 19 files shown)


xmtp/src/codecs/text.rs
```diff
@@ -0,0 +1,66 @@
...
+            authority_id: TextCodec::AUTHORITY_ID.to_string(),
+            type_id: TextCodec::TYPE_ID.to_string(),
+            version_major: 1,
+            version_minor: 0,
+        }
+    }
+
+    fn encode(text: String) -> Result<EncodedContent, CodecError> {
+        Ok(EncodedContent {
+            r#type: Some(TextCodec::content_type()),
```
jazzz, line 25: Am I correct that a raw string literal is used here because type is a keyword?

jazzz: Found my own answer. This is not a raw string literal, as the grammar does not match. This is a "keyword Identifier"

xmtp/src/codecs/text.rs
```diff
@@ -0,0 +1,66 @@
...
+    const AUTHORITY_ID: &str = "xmtp.org";
+    const TYPE_ID: &str = "text";
+    const ENCODING_KEY: &str = "encoding";
+    const ENCODING_UTF8: &str = "UTF-8";
+}
+
+impl ContentCodec<String> for TextCodec {
+    fn content_type() -> ContentTypeId {
+        ContentTypeId {
+            authority_id: TextCodec::AUTHORITY_ID.to_string(),
```
jazzz, line 16: If all content types need to be defined at compile time then it doesn't seem like Strings are needed in this case. Types could be changed to use `&'static str`

jazzz (approved):

neekolas:
@richardhuaaa @jazzz 

Are we sure we need to handle content encoding in Rust? Given we already have codecs written in Swift, Kotlin, and JS, I wonder if we really just want to take in `EncodedContent` as bytes and leave it up to the platform SDKs to handle content encoding/decoding. It doesn't seem strictly necessary to have that in scope for this project.

For the fancier formats (like remote attachments), even if we implemented fully in Rust there would have to be a lot of wrapper code in the native languages anyways to enable creation.

We could still keep this implementation in for the CLI and testing.

richardhuaaa:
@neekolas That's a good point! We should reduce the scope in Rust today and only increase over time as we gain confidence. Maybe conversation sending at the Rust level should just take in bytes for now. I'm on board, and will refactor the codecs out later.

neekolas:
We can keep the `EncodedContent` and `TextCodec` stuff somewhere in the repo, since it's useful for the CLI and tests.  As long as it's not part of the `send_message` flow, callers can choose not to use it if they are calling from the context of a full-featured SDK like iOS or Android.

PR #172, opened 2023-08-01
Create outbound payloads for sent messages

richardhuaaa:
Once messages have been committed to the messages table in `Unprocessed` state, `Conversations.process_messages()` will encrypt them into outbound payloads and write the result to the outbound_payloads table. Example usage is in the CLI.

Not included here as the PR is already chunky:
- Handling of conversation invites
- Pushing the outbound payloads onto the network
- Merging `send_message`, `process_messages`, `refresh_user_installations` and `send_outbound_payloads` into a single SDK method so end users don't have to call each one separately
- Refactoring the SDK so that multiple clients can share the same API client for unit tests
- Sealed sender and header signing

### DIFF
--- a/xmtp/src/conversations.rs
+++ b/xmtp/src/conversations.rs
@@ -154,7 +154,7 @@ where
                         continue;
                     }
                     let mut session = SessionManager::try_from(stored_session)?;
-                    let outbound_payload = self.create_outbound_payload(&mut session, &message)?;
+                    let outbound_payload = self.create_outbound_payload(&mut session, message)?;
                     let updated_session = StoredSession::try_from(&session)?;
                     outbound_payloads.push(outbound_payload);
                     updated_sessions.push(updated_session);
--- a/examples/cli/cli-client.rs
+++ b/examples/cli/cli-client.rs
@@ -213,7 +213,7 @@ async fn register(db: Option<PathBuf>, use_local: bool) -> Result<(), CliError> 
     Ok(())
 }
 
-async fn send(client: Client, addr: &String, msg: &String) -> Result<(), CliError> {
+async fn send(client: Client, addr: &str, msg: &String) -> Result<(), CliError> {
     let conversations = Conversations::new(&client);
     let conversation = conversations
         .new_secret_conversation(addr.to_string())


xmtp/src/utils.rs
```diff
@@ -21,6 +21,10 @@ pub fn build_user_invite_topic(public_key: String) -> String {
     format!("xmtp/1/invite-{}", public_key)
 }
 
+pub fn build_installation_message_topic(installation_id: &str) -> String {
+    format!("xmtp/1/message-{}", installation_id)
```
neekolas, line 25: I just realized none of our V3 topics have the `/proto` suffix, which all the existing (v1/v2) topics do. We should decide if we care

xmtp/src/conversations.rs
```diff
@@ -59,18 +71,134 @@ where
...
+            outbound_payloads,
+            updated_sessions,
+        )?;
+
+        Ok(())
+    }
+
+    pub async fn process_outbound_messages(&self) -> Result<(), ConversationError> {
+        let mut messages = self.client.store.get_unprocessed_messages()?;
+        messages.sort_by(|a, b| a.created_at.cmp(&b.created_at));
```
neekolas, line 164: Why not have the DB query just return the messages in the right order?

xmtp/src/conversations.rs
```diff
@@ -59,18 +71,134 @@ where
...
+        // TODO encrypted sealed metadata using sealed sender
+        let sealed_metadata = metadata.encode_to_vec();
+        let message_header = PadlockMessageHeader {
+            sent_ns: message.created_at as u64,
+            sealed_metadata,
+        };
+
+        let payload = PadlockMessagePayload {
+            message_version: PadlockMessagePayloadVersion::One as i32,
+            // TODO sign header - requires exposing a vmac method to sign bytes
```
neekolas, line 98: Or I guess we just base64 encode before signing?

xmtp/src/conversations.rs
```diff
@@ -59,18 +71,134 @@ where
...
+            payload: envelope.encode_to_vec(),
+            outbound_payload_state: OutboundPayloadState::Pending as i32,
+        })
+    }
+
+    pub fn process_outbound_message(
+        &self,
+        message: &StoredMessage,
+    ) -> Result<(), ConversationError> {
+        let my_sessions = self
```
neekolas, line 126: Having this live outside the transaction does open up the possibility that we could be operating with stale data here (if there were concurrent runs of this process). Ideally this entire function would be a single transaction.

xmtp/src/utils.rs
neekolas: Mostly because Waku had the suffix, but also I guess to allow for non-protobuf messages in the future.

xmtp/src/conversations.rs
neekolas: I ran into a similar issue with the invite loading last week. 

I have been solving by making methods take in a `conn` as an argument. Was even thinking about making the `Save` and `Store` traits work that way.

neekolas (approved):

PR #173, opened 2023-08-01
Codegen: padlock message protos

richardhuaaa:
Separating this out so that my PR's are smaller - this one is mostly code-genned changes from https://github.com/xmtp/proto/pull/92.

I temporarily pointed the codegen script at the branch that backs https://github.com/xmtp/proto/pull/93 as there is one change that hasn't landed yet. I will wait for that PR to land first, then point the script back at `xmtpv3` and re-generate before landing this one.


jazzz (approved):
🤖

PR #174, opened 2023-08-03
Publish outbound payloads onto the network

richardhuaaa:
This PR takes the payloads written into the `outbound_payloads` table and publishes them onto the network. We try to make sure that:

1. Payloads are published at most once - a row locking mechanism ensures at most one thread is trying to publish a given payload at any given time. 
   - The network (and the receiving client) should be able to deduplicate payloads regardless, if the lock expires
2. Payloads are published at least once - if the app goes offline or is force killed before confirmation is received from the server, it can be retried later by calling the `publish_outbound_payloads` method again.

We still want to merge `send_message`, `process_messages`,` refresh_user_installations` and `send_outbound_payloads` into a single SDK method so end users don't have to call each one separately - will handle this separately, but I'm thinking that send_message can spawn a separate thread to handle the rest of the process, and immediately return success to the layer above.

### DIFF
--- a/xmtp/src/storage/encrypted_store/mod.rs
+++ b/xmtp/src/storage/encrypted_store/mod.rs
@@ -367,13 +367,13 @@ impl EncryptedMessageStore {
         lock_duration_ns: i64,
     ) -> Result<Vec<StoredOutboundPayload>, StorageError> {
         let conn = &mut self.conn()?;
-        use self::schema::outbound_payloads::dsl::*;
+        use self::schema::outbound_payloads::dsl as schema;
         let now = now();
         // Must happen atomically
-        let payloads = diesel::update(outbound_payloads)
-            .filter(outbound_payload_state.eq(payload_state as i32))
-            .filter(locked_until_ns.lt(now))
-            .set(locked_until_ns.eq(now + lock_duration_ns))
+        let payloads = diesel::update(schema::outbound_payloads)
+            .filter(schema::outbound_payload_state.eq(payload_state as i32))
+            .filter(schema::locked_until_ns.lt(now))
+            .set(schema::locked_until_ns.eq(now + lock_duration_ns))
             .get_results::<StoredOutboundPayload>(conn)?;
         Ok(payloads)
     }


xmtp/src/conversations.rs
```diff
@@ -189,12 +195,44 @@ where
...
+        if unsent_payloads.is_empty() {
+            return Ok(());
+        }
+
+        let payload_ids = unsent_payloads
+            .iter()
+            .map(|payload| payload.created_at_ns)
+            .collect();
+        let envelopes = unsent_payloads
+            .into_iter()
```
jazzz, line 216: SAND: Moving the Envelopes in this context doesn't seem necessary. A reference would be sufficient and easier to understand. To keep a more consistent memory model using `iter` should be favored over `into_iter` unless there is a clear requirement

xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -360,6 +360,40 @@ impl EncryptedMessageStore {
...
         Ok(())
     }
+
+    pub fn fetch_and_lock_outbound_payloads(
+        &self,
+        payload_state: OutboundPayloadState,
+        lock_duration_ns: i64,
+    ) -> Result<Vec<StoredOutboundPayload>, StorageError> {
+        let conn = &mut self.conn()?;
+        use self::schema::outbound_payloads::dsl::*;
```
jazzz, line 370: DUST: From a readability level I would suggest using a scoped import rather than a wildcard, I've been finding it easy to run into name conflicts between fields in the model and params being passed into the function

https://github.com/xmtp/libxmtp/blob/9a4b8ccaddcea7cb76d3edf0177af55ad079fe1c/xmtp/src/storage/encrypted_store/mod.rs#L183

jazzz (approved):
Looks good. Included a few nits incase they are helpful to you.

xmtp/src/storage/encrypted_store/mod.rs
neekolas: Oh ya. I've definitely run into that as well

xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -360,6 +360,40 @@ impl EncryptedMessageStore {
...
+        let now = now();
+        // Must happen atomically
+        let payloads = diesel::update(outbound_payloads)
+            .filter(outbound_payload_state.eq(payload_state as i32))
+            .filter(locked_until_ns.lt(now))
+            .set(locked_until_ns.eq(now + lock_duration_ns))
```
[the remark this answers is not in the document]
neekolas: One thing we'll have to be mindful of with this approach is that we need to somehow retrigger this process if rows have been excluded because of the `locked_until_ns`. I wouldn't expect this to be running in a loop or on a cron schedule, but we'll somehow need to make sure this gets re-run in some way other than the user sending another message.

neekolas (approved):

neekolas (approved):

PR #175, opened 2023-08-03
Session Initiation on client initialization

jazzz:
The PR enables Installation refresh + session initialization on client initialization. 

This allows the client to operate with the latest set of sibling installations, by pulling the latest list everytime the client is created. A mechanism is needed to  ensure this list stays up to date over the course of the clients lifecycle however that is out of scope for now. 

On Send clients will also need to refresh the foreign installations to ensure messages are fanned out to the correct installations.

### DIFF
--- a/xmtp/src/client.rs
+++ b/xmtp/src/client.rs
@@ -2,7 +2,7 @@ use core::fmt;
 use std::fmt::Formatter;
 
 use diesel::Connection;
-use log::{debug, info};
+use log::info;
 use thiserror::Error;
 use vodozemac::olm::OlmMessage;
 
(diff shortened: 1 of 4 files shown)


xmtp/src/client.rs
```diff
@@ -162,7 +165,7 @@ where
 
         let new_installs = contacts
             .iter()
-            .filter(|contact| self_install_id == contact.installation_id())
+            .filter(|contact| self_install_id != contact.installation_id())
```
jazzz, line 168: This bug was annoying to catch. Real tests should be added as soon as the infra exists to configure test db states. Currently testing manually via the CLI

PR #176, opened 2023-08-04
Pass in DBConnection Explicitly

neekolas:
## Summary

I've been running into issues using code that calls our `Save` or `Store` traits inside a transaction. This is because the traits attempt to get a new connection from the pool when the pool is already being used for the transaction manager.

We need to make the connection used explicit instead of implicit to make our existing code work with transactions.

This does that by making the `Save`, `Fetch`, and `Store` traits expect a `DbConnection` instead of an `EncryptedMessageStore`. We'll also want to make the methods on `EncryptedMessageStore` take in a `DbConnection` instance explicitly, so that they can be used safely from inside a transaction. That can be handled in a subsequent PR.

## Alternatives considered

- We could ditch the `Save` and `Store` traits altogether, and just add methods on `EncryptedMessageStore` that take in a DbConnection. That ultimately works the same way, but I like the tidiness of using traits for these operations that are so similar. I do kinda want to get rid of the `Fetch` trait and move those operations to the `EncryptedMessageStore`, since I think in the fullness of time each entity's fetch is going to need slightly different parameters for querying. 
- Keep the existing traits and add additional `SaveWithConn`, `StoreWithConn`, `FetchWithConn` that take in an explicit connection. I don't hate that, but it feels unnecessary given that most of our code is ultimately going to be called from somewhere inside a transaction.

## Potential Challenges

- In order to make this approach work with other connections beyond `SQLite` we will need to do some trickery with features to redefine what the `DBConnection` type references depending on what database you are targeting.
- Because SQLite can only have a single concurrent connection, care needs to be taken to ensure that only one `conn`  exists at a time. `DbConnection`s should be created as early as possible and re-used. That means other functions that operate on the database will also need to take in a `DbConnection` as an argument whereas before they could simply rely on having access to an `EncryptedMessageStore`.

### DIFF
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -126,7 +126,7 @@ where
             None => {
                 info!("Creating new XMTP identity");
                 let new_account = Self::sign_new_account(owner)?;
-                new_account.store(store)?;
+                new_account.store(&mut store.conn().unwrap())?;
                 Ok(new_account)
             }
         }
@@ -136,7 +136,8 @@ where
     fn retrieve_persisted_account(
         store: &mut EncryptedMessageStore,
     ) -> Result<Option<Account>, ClientBuilderError> {
-        let mut accounts = store.fetch()?;
+        let conn = &mut store.conn()?;
+        let mut accounts = conn.fetch()?;
         Ok(accounts.pop())
     }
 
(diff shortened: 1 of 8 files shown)


neekolas:
> For non-sqlite DB's, my rough thought is we could move the methods from EncryptedMessageStore onto a MessageStore trait, and define the connection type as an [associated type](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html) on that trait. Then just implement any other databases against that trait, with the connection type defined by the implementation.

I like that too

>  But if I understand correctly, this PR doesn't lock us into having only a single connection necessarily, so it's fine regardless.

Maybe our current issue is just an implementation problem with our connection pool having a max size of 1. We can try turning that up.

PR #177, opened 2023-08-07
Process invites

neekolas:
## Summary

- Processes inbound invites from the database and creates the users/conversations
- Fixes a bug in saving sessions
- Adds new status to `InboundInvite` (`InviteReceived`)
- Adds a DB test for fetching inbound invites by status


jazzz (approved):
Few lints/clippies that should be addressed, but like the structure of this change

PR #178, opened 2023-08-08
Share networking client between XMTP clients for tests

richardhuaaa:
The `MockXmtpApiClient` stores network requests in memory. This means that if two separate api clients are constructed, they won't share any memory - which means that in unit tests, we are unable to validate any communication between two users that passes through the network. This PR enables us to share a `MockXmtpApiClient` between multiple XMTP clients, so that this kind of testing is unblocked.

### DIFF
--- a/xmtp/src/conversations.rs
+++ b/xmtp/src/conversations.rs
@@ -409,7 +409,7 @@ mod tests {
             InboundInvite, InboundInviteStatus, MessageState, OutboundPayloadState,
             StoredConversation, StoredMessage, StoredUser,
         },
-        test_utils::test_utils::{gen_test_client, gen_test_conversation},
+        test_utils::test_utils::{gen_test_client, gen_test_conversation, gen_two_test_clients},
         utils::{build_envelope, build_user_invite_topic},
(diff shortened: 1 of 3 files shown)


xmtp/src/conversations.rs
```diff
@@ -480,10 +479,9 @@ mod tests {
         let unprocessed_messages = alice_client.store.get_unprocessed_messages().unwrap();
         assert_eq!(unprocessed_messages.len(), 1);
 
-        // TODO replace with Client.refresh_user_installations. Requires us to refactor the SDK so that
-        // two XMTP clients can share the same API client
         alice_client
-            .create_outbound_session(&bob_client.account.contact())
+            .refresh_user_installations(&bob_client.wallet_address())
+            .await
             .unwrap();
```
richardhuaaa, line 485: `refresh_user_installations` fetches Bob's contacts from the network before creating a session with them. Before this change, it would have fetched 0 contacts because the API client that Bob published to on initialization was different from the API client that Alice is fetching from here

neekolas (approved):
Great change. Going to make testing so much easier

PR #179, opened 2023-08-08
Change Primary key of OutboundPayloads Table

jazzz:
Expecting some pushback on this one, but thought I'd put forward a solution to start the conversation. 

## Problem
There is a Primary key error being emitted when sending to an address with multiple installs.
The outbounds_payloads table uses `created_at_ns` as a primary key. This value is pulled from the timestamp which the message was created. Message fanout leads to multiple outbound payloads being created with the same key.

image

## Solution
This PR adds a BREAKING CHANGE to change the primary key of the `outbound_payload` table to use a msg_id which is calculated via  Sha256("\<timestamp\>:\<topic\>");

image

## Notes
- I first thought about changing the definition of `created_at_ns` to be when the outbound payload was created rather than the message, however that would lead to potential duplicate messages as double sends would not be caught or detected.

### DIFF
--- a/xmtp/migrations/2023-07-19-232818_create_outbound_payloads/up.sql
+++ b/xmtp/migrations/2023-07-19-232818_create_outbound_payloads/up.sql
@@ -1,5 +1,5 @@
 CREATE TABLE outbound_payloads (
-    msg_id TEXT NOT NULL,
+    payload_id TEXT NOT NULL,
     created_at_ns BIGINT KEY NOT NULL,
     content_topic TEXT NOT NULL,
     payload BLOB NOT NULL,
--- a/xmtp/migrations/2023-07-19-232818_create_outbound_payloads/up.sql
+++ b/xmtp/migrations/2023-07-19-232818_create_outbound_payloads/up.sql
@@ -1,5 +1,6 @@
 CREATE TABLE outbound_payloads (
-    created_at_ns BIGINT PRIMARY KEY NOT NULL,
+    msg_id TEXT NOT NULL,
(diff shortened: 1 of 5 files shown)


neekolas (approved):
No objections from me

richardhuaaa (approved):
Great catch! Could we call it `payload_id` or just `id` instead of `message_id`, seeing as the message identifier (which multiple payloads exist for) would potentially be something different? Also, are we intending this identifier to be globally identifiable across installations? Just curious why we didn't use an auto-incrementing int here, but it's not a big deal.

jazzz:
>Just curious why we didn't use an auto-incrementing int here, but it's not a big deal.

My understanding was that locally deterministic `id`'s are used for deduping messages in the database. If thats true then auto incrementing would not be useful, as each message would get a new 'id' rather than conflicting on insert. 

Could be wrong here

richardhuaaa:
Oh I see, we want to use the ID to prevent an identical payload being inserted twice. Fair enough!

xmtp/migrations/2023-07-19-232818_create_outbound_payloads/up.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE outbound_payloads (
-    created_at_ns BIGINT PRIMARY KEY NOT NULL,
+    payload_id TEXT NOT NULL,
```
neekolas, line 2: Shouldn't this field be labeled as `PRIMARY KEY`?

PR #180, opened 2023-08-09
List Conversations

neekolas:
## Summary
- Adds a method to list conversations (and optionally refresh first)
- Fixes `newConversation` to store installations in the database instead of loading them ephemerally
- Fixes a bug where installations were being created before the corresponding user
- Removes `members` from `SecretConversation` struct, instead pulling directly from the DB


xmtp/src/client.rs
```diff
@@ -191,9 +192,18 @@ where
 
         self.store.conn().unwrap().transaction(
             |transaction_manager| -> Result<(), ClientError> {
+                self.store.insert_or_ignore_user_with_conn(
```
neekolas, line 195: Feels a little wasteful to insert this and then update it a few lines later. Should probably consolidate into a single function that tries to insert and updates the `refresh_timestamp` on conflict

xmtp/src/conversation.rs
```diff
@@ -128,13 +137,23 @@ where
...
     pub async fn initialize(&self) -> Result<(), ConversationError> {
         let inner_invite_bytes = Invitation::build_inner_invite_bytes(self.peer_address.clone())?;
-        for contact in self.members.iter() {
+        let conn = &mut self.client.store.conn()?;
+        for contact in self.members(conn)?.iter() {
             let id = contact.installation_id();
 
             // TODO: Persist session to database
-            let session = self.client.create_outbound_session(contact)?;
+            let session = self.client.get_session(contact, conn)?;
```
neekolas, line 156: This was causing issues with duplicating sessions

xmtp/src/conversation.rs
```diff
@@ -128,19 +137,33 @@ where
...
+    fn members(&self, conn: &mut DbConnection) -> Result<Vec<Contact>, ConversationError> {
+        let my_installations = self.client.my_other_devices(conn)?;
+        let peer_installations = self
+            .client
+            .get_contacts_from_db(conn, self.peer_address().as_str())?;
+
+        Ok(vec![my_installations, peer_installations].concat())
+    }
+
     pub async fn initialize(&self) -> Result<(), ConversationError> {
```
richardhuaaa, line 149: Should we check the conversation state before doing everything else under initialize? And are we calling this every time `SecretConversation` is constructed, or at some other point?

Just wondering where this will fit into https://github.com/xmtp/libxmtp/blob/main/xmtp/README.md?plain=1#L85-L88 (modifications welcome if this is no longer accurate)

xmtp/src/client.rs
```diff
@@ -137,32 +138,33 @@ where
 
     pub fn get_session(
         &self,
-        contact: &Contact,
         conn: &mut DbConnection,
+        contact: &Contact,
```
jazzz, line 142: fair enough, this feels more natural

xmtp/src/client.rs
```diff
@@ -137,32 +138,33 @@ where
...
+            .filter(|c| c.installation_id() != my_installation_id)
             .collect())
     }
 
     /// Fetch Installations from the Network and create unintialized sessions for newly discovered contacts
     // TODO: Reduce Visibility
     pub async fn refresh_user_installations(&self, user_address: &str) -> Result<(), ClientError> {
         // Store the timestamp of when the refresh process begins
-        let refresh_timestmap = now();
+        let refresh_timestamp = now();
```
jazzz, line 167: 😞

xmtp/src/conversation.rs
neekolas: That's a good call-out. I had to make some changes to `initialize` for compatibility, but I haven't really taken a proper look at what needs to change to make it compliant with the state machine spec.

One thing is definitely to check the state and make sure it hasn't already been initialized.

xmtp/src/conversation.rs
```diff
@@ -152,8 +175,16 @@ where
...
                 .map_err(|_| ConversationError::Unknown)?;
+
+            session.save(conn)?;
         }
 
+        self.client.store.set_conversation_state(
+            conn,
+            self.convo_id().as_str(),
+            ConversationState::Invited,
+        )?;
```
jazzz, line 186: Calling out for future discussions: Conversation State is tracked at the peer_address layer, however the messages are fanned out at the installation_id level. This means that any installation that registers after this function was called would not have access to the invite.

xmtp/src/conversations.rs
```diff
@@ -44,12 +44,42 @@ where
         Self { client }
     }
 
-    pub async fn new_secret_conversation(
+    pub fn new_secret_conversation(
```
jazzz, line 47: Is there a distinction between secret_conversations and conversations?

jazzz (approved):

xmtp/src/conversations.rs
neekolas: It's the only conversation type we have right now. Legacy of padlock

xmtp/src/conversation.rs
neekolas: Ya. Definitely an issue we'll have to solve

jazzz: On the positive side: A newly registered installation will be able to send out its own invite thought which would allow them to join in -- Convo_id's are deterministic so it should just work

PR #181, opened 2023-08-09
Add ClientError details

jazzz:
While debugging I needed some extra details from clientError::unknown.

This PR:
- Removed Unknown in favor of Generic(String) so that some details must always be provided
- adds ergonomics for string types


neekolas (approved):
Great quality of life improvement

PR #182, opened 2023-08-10
Refresh user installations if stale

richardhuaaa:
Small PR with two changes:
1. When processing messages, refresh user installations if they are stale
2. Publish the outbound messages immediately after processing them.

To summarize, these are the steps when sending a message:
1. Write the message to the database
2. Encrypt the message into payloads (possibly refreshing installations along the way)
3. Publish the messages to the server

Steps (2) and (3) were originally separated out because they involve network calls that may potentially fail (e.g. device is offline, the network request takes a long time, device is killed, etc). So there is a DB write between each step, and each step can be picked up independently. The change in this PR means that (3) automatically starts after (2).

In a future PR I'd like to also kick off (2) from (1) in a background thread, so that clients don't have to wait for a server request to get confirmation that their send is happening. However, there is some refactoring that is needed because `Conversation` doesn't hold a reference to `Conversations`. In a later PR I'm thinking I can move the methods on `Conversations` onto `Conversation`.

### DIFF
--- a/xmtp/src/conversations.rs
+++ b/xmtp/src/conversations.rs
@@ -27,7 +27,10 @@ use crate::{
     Client,
 };
 
-const PADDING_TIME_NS: i64 = 30 * 1000 * 1000 * 1000;
+const SECOND_IN_NS: i64 = 1000 * 1000 * 1000;
+const MINUTE_IN_NS: i64 = 60 * SECOND_IN_NS;
+const PADDING_TIME_NS: i64 = 30 * SECOND_IN_NS;
+const INSTALLATION_REFRESH_INTERVAL_NS: i64 = 10 * MINUTE_IN_NS;
 
 pub struct Conversations<'c, A>
 where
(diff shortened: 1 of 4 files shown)


xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -224,10 +224,10 @@ impl EncryptedMessageStore {
         conn: &mut PooledConnection<ConnectionManager<SqliteConnection>>,
         user_address: &str,
         timestamp: i64,
-    ) -> Result<usize, StorageError> {
-        diesel::update(users::table.filter(users::user_address.eq(user_address)))
+    ) -> Result<StoredUser, StorageError> {
+        diesel::update(users::table.find(user_address))
             .set(users::last_refreshed.eq(timestamp))
-            .execute(conn)
+            .get_result::<StoredUser>(conn)
```
richardhuaaa, line 230: This change just ensures that an error is thrown if no row exists and no update was made (`get_result` requires there to be one result).

xmtp/src/conversations.rs
```diff
@@ -385,6 +393,7 @@ where
             }
         }
 
+        self.publish_outbound_payloads().await?;
```
jazzz, line 396: Unwrapping an error from publish_outbound_payloads is going to propagate all the way to the client app when something goes wrong - Is this desired? 

Is and error on publish critical or can it be logged and ignored?

jazzz (approved):
Thats a lot of change in a small PR. Nicely done

xmtp/src/conversations.rs
richardhuaaa: So it's not the case right now, but ideally end clients won't be calling these methods directly. I see two places where `process_outbound_messages()` will be called from the SDK:

1. On initialization after a cold start
2. On message send (`conversation.send_message()`) - likely spun off as a separate thread so that it is non-blocking to the app

In both of these places, I think an error in `publish_outbound_messages` should be propagated to `process_outbound_messages`, and then handled at the call-site inside the SDK.

PR #183, opened 2023-08-10
XLI tidying

richardhuaaa:
Some quick tidying of the CLI in preparation for end-to-end testing:

- Combine `reg` and `testReg` into `register`, with an optional `seed` param to allow for multi-installation wallet testing
- Remove WalletConnect option from `register` as this is no longer functional after WalletConnect V1 was deprecated (at least until there is a viable v2 rust client)
- Update example commands and comments


jazzz (approved):
Looks good. I agree the XLI isn't the best name open to others that are short and unique.

PR #184, opened 2023-08-11
Proposal: Move v2 crates to separate branch

richardhuaaa:
As this repo becomes more publicly visible, the distinction between v2 and v3 will create unnecessary confusion (see README.md in this PR).

Our strategy until this point has been to make separate crates for v2 and v3 work, to avoid making breaking changes to existing dependencies. In addition to being confusing this will quickly become unmanageable as we introduce more versions in the future. Separating major (breaking) version updates by branches, and then merging into main as needed seems like a more viable strategy

As the focus of this repo is on v3, I've created a `v2` branch to house our v2 crates, and this PR would remove them from main. This lets us continue to iterate on v3 without breaking v2, and vice versa.


nakajima (commented):
Lacking most context here but seems good to me.

jazzz (approved):
I very much agree with this. Removing old cruft will make things easier to understand.

there will need to be a set of PR which lock the consuming repos to a branch, though it is unclear to me where these packages are consumed.

Approving assuming the consumers are set to point to the dedicated branch

neekolas:
Maybe @mkobetic can figure out the branch locking strategy. It's tricky because there is an intermediate repo for Swift that would also need to be aware of the branching strategy.

richardhuaaa:
Great catch. So we have a release process that pushes changes from the `bindings_swift` crate in this repo to the `xmtp-rust-swift` repo, rather than the other way around. Can check with @mkobetic what needs to be done to trigger it from the new branch. For the `xmtp_dh` crate, it is manual (a shell script which copies the compiled binaries to a location of your choice).

PR #185, opened 2023-08-11
Added Message Receiving 

jazzz:
This PR Adds the functionality for Receiving messages.

The code is in need of a refactor (Messages and invites could share the same pathway) but this work is blocking further integration. Follow up Item

## Notes
- Currently replying to a message results in NoSession being found. So there is an issue somewhere in the state management. Will handle in follow up
- Tested working under fanout


xmtp/src/conversations.rs
```diff
@@ -82,6 +95,146 @@ where
...
+                .ok_or(ConversationError::NoSessions(
+                    "Message was not PreKey message, but no existing session".into(),
+                ))?;
+
+            let olm_message: OlmMessage = shuck_ok!(
+                serde_json::from_slice(&message_envelope.ciphertext),
+                InboundInviteStatus::DecryptionFailure
+            );
+
+            session.decrypt(olm_message, conn)?
```
jazzz, line 218: This portion is untested due to an error in sending

xmtp/src/conversations.rs
```diff
@@ -457,6 +619,10 @@ mod tests {
         ClientBuilder, Fetch,
     };
 
+    fn init() {
+        let _ = env_logger::builder().is_test(true).try_init();
+    }
```
jazzz, line 624: This allows conditional logging inside of tests. If called from a test: All Errors are logged, and log entries are included if the test fails

xmtp/src/conversations.rs
```diff
@@ -647,6 +813,70 @@ mod tests {
...
+
+        a_convos.receive().unwrap();
+
+        let _alice_messages = alice_client
+            .store
+            .get_stored_messages(&mut alice_client.store.conn().unwrap())
+            .unwrap();
+
+        // TODO: This is currently failing with a NoSession error for unknown reasons
+        // assert_eq!(alice_messages.len(), 2);
```
jazzz, line 877: This is a follow up piece of work. The sending side is not finding a session when replying to a message.

xmtp/src/conversations.rs
```diff
@@ -304,12 +465,11 @@ where
...
         };
         let olm_message = session.encrypt(&payload.encode_to_vec());
-        let ciphertext = match olm_message {
-            OlmMessage::Normal(message) => message.to_bytes(),
-            OlmMessage::PreKey(prekey_message) => prekey_message.to_bytes(),
-        };
 
-        let envelope = PadlockMessageEnvelope {
+        let ciphertext =
+            serde_json::to_vec(&olm_message).map_err(|_| ConversationError::Unknown)?;
```
jazzz, line 470: Uses the built in serializer for the OlmMessage rather manually handling each type.

xmtp/src/conversations.rs
```diff
@@ -82,6 +95,146 @@ where
...
+            unsealed_header.sender_user_address,
+            message_obj.content_bytes,
+            MessageState::Received as i32,
+        );
+
+        self.client
+            .store
+            .insert_or_ignore_message(conn, stored_message)?;
+
+        Ok(InboundInviteStatus::Processed)
```
neekolas, line 235: Messages should probably have their own status enum, even if the values are the same as `InboundInviteStatus` today

xmtp/src/conversations.rs
neekolas: The nice thing about `serde_json::to_vec` is that it doesn't require an additional field to distinguish pre-key and non-pre-key messages. The bad thing is that it will serialize it in a much bulkier format (JSON string) rather than `to_bytes` which actually uses the protobuf representation.

I was thinking I should actually update the invite handling to use `to_bytes` for more compact messages.

jazzz: I waffled on this one. for the same reasons. I'll leave the encodings as they are for now then. And we can look at changing the Invite Protos later then

xmtp/src/conversations.rs
jazzz: My intention is to merge the code paths for receiving Invites and messages, in which case the statuses would be shared across a InboundPayloadStatus enum. 

For today I'll add a copy, and we can revisit later

xmtp/src/client.rs
```diff
@@ -360,6 +365,24 @@ where
 
         Ok(response.envelopes)
     }
+
+    /// Search network for a specific InstallationContact
+    /// This function should be removed as soon as possible given it is a potential DOS vector.
+    /// Contacts for a message should always be known to the client
+    pub async fn download_contact_for_installation(
```
jazzz, line 367: If the first message received by the client is a PreKey payload which is not an invite, then the client doesn't have the necessary contact bundle to create a valid session. This results in a linear search through the contact bundles

xmtp/src/client.rs
```diff
@@ -293,24 +293,21 @@ where
         &self,
         conn: &mut DbConnection,
         contact: &Contact,
-        // Message MUST be a pre-key message
-        message: &Vec<u8>,
+        prekey_message: PreKeyMessage,
```
jazzz, line 296: This is the easiest solution for now, since messages and invites are encoded differently. The caller is now responsible for passing in a decoded Prekey Message.

Required updates to [`process_inbound_invite`](https://github.com/xmtp/libxmtp/pull/185/files#diff-43fed88baca0e696206ee0e68177689c3149e6d0bbdd34fa8d8b4b9b53555271R332)

xmtp/src/conversations.rs
```diff
@@ -176,17 +329,25 @@ where
...
             None => {
-                (_, plaintext) = match self.client.create_inbound_session(
-                    conn,
-                    &invitation.inviter,
-                    &invitation.ciphertext,
-                ) {
-                    Ok((session, plaintext)) => (session, plaintext),
-                    Err(err) => {
-                        log::error!("Error creating session: {:?}", err);
+                let prek_key = match olm_message {
```
neekolas, line 332: OK. So, we are expecting that there will either be an existing session or you will be receiving a prekey message. 

We might have to revisit this if we run into cases where the first reply isn't delivered and one side is sending regular messages but the other side doesn't have an active session. Should be a pretty fringe edge case.

xmtp/src/conversations.rs
neekolas: Ooh. That's very helpful. I've been replacing our log statements with `println`s when debugging tests, and then switching them back.

neekolas (approved):

xmtp/src/client.rs
richardhuaaa: Would it help to have the sender contact attached to the pre-key message?

richardhuaaa (approved):

xmtp/src/conversations.rs
richardhuaaa: > one side is sending regular messages but the other side doesn't have an active session

Just want to clarify this. An initiator will create a session and keep sending prekey messages until a reply is received. The only case where the initiator will send a non-prekey message is if the peer has replied, in which case the peer should have a session?

examples/cli/cli-client.rs
```diff
@@ -212,6 +224,16 @@ async fn send(client: Client, addr: &str, msg: &String) -> Result<(), CliError>
     Ok(())
 }
 
+async fn recv(client: &Client) -> Result<(), CliError> {
+    let conversations = Conversations::new(client);
+    // client.fetch_new_messages().await?;
+    // conversations.save_invites()?;
```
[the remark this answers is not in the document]
jazzz: Eventually revc will need to handle the invite flow as well. Left these as a reminder, but removed for now.

xmtp/src/client.rs
jazzz: Yes. I am mulling over some larger changes in the protos. Namely the addition of a "Payload" type which unifies Invites and Messages -- There seems to be a lot of overlap between these.

The payload type would contain the same message header from PadlockMessages, and integrate the contact bundle as needed. That way the processing for all packets could be the same, albeit with different contents.

richardhuaaa: Excellent, I was wondering if we should do that too :+1:

xmtp/src/conversations.rs
jazzz: Correct the processing flow has been updated to:

Is there a Session:
- Yes: Decrypt using existing session
- No: is this a PrekeyMessage:
     - YES: create inbound session
     - No: Error

PR #187, opened 2023-08-14
End-to-end test scenarios

richardhuaaa:
Suggested test scenarios for us to run through - please comment with anything missing!

### DIFF
--- a/TEST_SCENARIOS.md
+++ b/TEST_SCENARIOS.md
@@ -1,0 +1,73 @@
+# End-to-end test scenarios
+
+These can be performed manually and/or automated. More scenarios will be added as they are implemented (XMTP v3 is a work in progress).
+
+## How to read the tests
+
+### Identifiers
+
+- Accounts (in this case wallets) are identified by letters (A, B, C)
+- Installations are identified by numbers (1, 2, 3)
+
+For example, Client A1 is the first installation of account A, A2 is the second installation of account A, and B1 is the first installation of account B.
+
+### Registering
+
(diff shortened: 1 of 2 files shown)


xmtp/src/client.rs
```diff
@@ -22,7 +22,7 @@ use crate::{
 use std::collections::HashMap;
 use xmtp_proto::xmtp::message_api::v1::Envelope;
 
-const INSTALLATION_REFRESH_INTERVAL_NS: i64 = 10 * 60 * 1_000_000_000; // 10 minutes
+const INSTALLATION_REFRESH_INTERVAL_NS: i64 = 0;
```
richardhuaaa, line 25: This is a hack in order to make test case (1) pass in a timely way. I have a proposal for a clean/better way of handling installation syncing that I will put up in a bit.

TEST_SCENARIOS.md
```diff
@@ -0,0 +1,73 @@
...
+Register client A2
+    Send a message from A1 to B
+    Verify message was received in all clients
+Register client B2
+    Send a message from A1 to B
+    Verify message was received in all clients
+Register client A3
+    Send a message from A3 to B
+    Verify message was received in all clients
+```
```
jazzz, line 49: I like this formulation

Test:
- the nominal case
- That clients detect new self-installations
- That clients fanout to new remote-installations
- New installations can be heard. 


Think this is fantastic scope

jazzz (approved):

PR #188, opened 2023-08-14
Ensure an error is thrown if no row was found to update

richardhuaaa:
Our update operations don't throw an error if no row was found to update. Small PR to make sure we catch this, in case it helps @jazzz diagnose why replies currently give a NoSession error.


neekolas (approved):

jazzz (approved):
😍

PR #189, opened 2023-08-14
List Messages API

neekolas:
## Summary
- Adds a new `sent_at_ns` field to properly sort `StoredMessage` results. `created_at` time is only accurate for sent messages, but will show the time the message was received (instead of sent) for inbound messages.
- Adds a more advanced query method for stored messages
- Adds support for listing messages to `SecretConversation`


xmtp/src/conversation.rs
```diff
@@ -140,11 +157,30 @@ where
             self.client.account.addr(),
             content_bytes,
             MessageState::Unprocessed as i32,
+            now(),
         )
         .store(&mut self.client.store.conn().unwrap())?;
         Ok(())
     }
 
+    pub async fn list_messages(
```
neekolas, line 166: I made this async so that we can refresh the message store, once there is a way to do that from inside a conversation.

xmtp/src/conversation.rs
```diff
@@ -140,11 +157,29 @@ where
...
+        &self,
+        opts: ListMessagesOptions,
```
jazzz, line 168: Can the listMessageOptions struct be passed by ref in this case? would avoid having to create multiple objects for successive queries.

jazzz (approved):

xmtp/src/conversation.rs
neekolas: That's a good idea

richardhuaaa (approved):
Very clean

PR #190, opened 2023-08-14
Fix duplicate sessions

jazzz:
## Problem

When replying to message the client was creating a new unnecessary session. The root cause was that refresh_user_installations creates a session, which was allowed because the primary key on the sessions table was set to `session_id`.

### Path 

- Initializing a conversation causes a refresh of the users contact bundles
- New contact creates a new session via `insert_or_ignore_session`.
- since the primary key was set to `session_id` there was never a conflict and sqlite allowed the operation

## Solution

This PR changes the primary key to `peer_installation_id` such that only one session can be created for a given installation. This results in no extra sessions being added.

### DIFF
--- a/xmtp/migrations/2023-06-03-011250_create_sessions/up.sql
+++ b/xmtp/migrations/2023-06-03-011250_create_sessions/up.sql
@@ -1,6 +1,6 @@
 CREATE TABLE IF NOT EXISTS sessions (
-    session_id TEXT PRIMARY KEY NOT NULL,
+    peer_installation_id TEXT PRIMARY KEY NOT NULL,
+    session_id TEXT NOT NULL,
     created_at BIGINT NOT NULL,
-    peer_installation_id TEXT NOT NULL,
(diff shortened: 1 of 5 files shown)


richardhuaaa (approved):
> Initializing a conversation causes a refresh of the users contact bundles
> New contact creates a new session via insert_or_ignore_session.

If this is triggered when the client is replying to a message, shouldn't the client have saved the contact when receiving a message from it?

Good fix regardless as we want to maintain the invariant of 1 session for 1 peer installation

jazzz:
@richardhuaaa - Great point. Withotu saving the contact the system still functions however would be best to synchronize state earlier. I'll add.

neekolas (approved):

jazzz:
@richard Contact saving from within `process_inbound_messages` got messy. Given the intention is refactor the Protos to add a Contact to a prekey message, I think this is best to revist after that change.

PR #192, opened 2023-08-15
Default CLI to dev network

richardhuaaa:
This PR:
- Defaults the CLI to use the XMTP dev network
- Adds a `--local` option for pointing to localhost instead
- Use `/xmtp/3` endpoints instead of `/xmtp/1` endpoints for consistency with client versioning
- Adds detail to networking errors

This means that external consumers don't have to run a local node in order to run the CLI, and it means the CLI can be tested across multiple machines.

Will land https://github.com/xmtp/xmtp-node-go/pull/286 and verify that the dev network accepts CLI requests before landing this.

### DIFF
--- a/examples/cli/cli-client.rs
+++ b/examples/cli/cli-client.rs
@@ -187,17 +187,14 @@ async fn create_client(
     account: AccountStrategy<Wallet>,
 ) -> Result<Client, CliError> {
     let msg_store = get_encrypted_store(db).unwrap();
-    let mut builder = ClientBuilder::new(account)
-        .store(msg_store);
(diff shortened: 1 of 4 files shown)


neekolas (approved):

jazzz (commented):
Are we sure this is what we want?

publishing to Dev means message payloads are going to be persisted, and any changes are going to be breaking. 

Are we committing to rollout plans/network wipes for breaking changes?

neekolas:
> publishing to Dev means message payloads are going to be persisted, and any changes are going to be breaking.

That can be mitigated by _strongly_ encouraging developers to use disposable wallets for testing here. I still think we would want to be able to make breaking changes without wiping anything.

PR #194, opened 2023-08-15
Docs: Update root and CLI READMEs for v3-alpha

jhaaaa:


examples/cli/README.md
```diff
@@ -1,22 +1,89 @@
...
-`./xli.sh --db user2.db3 info`
+1. In the root directory of this repo, run:
 
-### Send message
+   ```bash
+   dev/up
+   ```
 
-`./xli.sh --db user1.db3 send <user2_address> "hello"`
+2. Create a sender wallet account (user1). Create an [installation key bundle](#installation-key-bundles) and store it in the database. Grant the installation key bundle permission to message on behalf of the sender address. This will allow the CLI to message on behalf of the sender address.
```
neekolas, line 20: Sorry, I thought this part of the document was going to be in the `examples/cli` folder. We should explicitly tell people to move to that folder between step 1 and 2, since the commands from 2 onwards will only work in that folder.

examples/cli/README.md
```diff
@@ -1,22 +1,95 @@
...
+   ./xli.sh --db user1.db3 send <user2_address> "hello"
+   ```
+
+6. List conversations.
+
+   ```bash
+   ./xli.sh --db user1.db3 list-conversations
+   ```
+
+## Double ratchet messaging
```
mkobetic, line 56: This part (to the end) could probably be in the top level README. Doesn't seem specific to the CLI example.

examples/cli/README.md
jhaaaa: Such a great callout @neekolas - thank you! The README is in `examples/cli`, but the `dev/up` command needs to be run at the root. So absolutely important to make going to `examples/cli` an explicit step.

richardhuaaa: We no longer need to run `dev/up`, because the CLI points to the dev network by default, which allows messaging across multiple machines.

If people explicitly want to run against localhost, they can run `dev/up` and then specify the `--local` flag on the CLI.

examples/cli/README.md
```diff
@@ -1,22 +1,89 @@
...
+
+   ```bash
+   ./xli.sh --db user1.db3 send <user2_address> "hello"
+   ```
+
+6. List conversations.
+
+   ```bash
+   ./xli.sh --db user1.db3 list-conversations
+   ```
```
richardhuaaa, line 54: @neekolas Should we add the upcoming list-messages command here as well?

examples/cli/README.md
jhaaaa: Thank you @richardhuaaa! Updating now..... <3

examples/cli/README.md
```diff
@@ -1,22 +1,46 @@
...
+
+If you want to run the CLI against localhost, go to the root directory and run `dev/up`. Then run the CLI commands using the `--local` flag.
```
jhaaaa, line 46: @richardhuaaa Did I capture this info correctly? =) Just making sure...

examples/cli/README.md
```diff
@@ -1,22 +1,46 @@
...
+   ./xli.sh --db user2.db3 register
+   ```
+
+4. Get the recipient's wallet address.
+
+   ```bash
+   ./xli.sh --db user2.db3 info
+   ```
+
+5. Send a message into the conversation. The message is sent using [one session between each installation](https://github.com/xmtp/libxmtp/blob/main/README.md#installation-key-bundles) associated with the sender and recipient. The message is encrypted using a session key dedicated to this single message per the [double ratchet algorithm](https://github.com/xmtp/libxmtp/blob/main/README.md#double-ratchet-messaging).
```
richardhuaaa, line 34:
```suggestion
5. Send a message into the conversation. The message is sent using [one session between each installation](https://github.com/xmtp/libxmtp/blob/main/README.md#installation-key-bundles) associated with the sender and recipient. The message is encrypted using a per-message encryption key derived using the [double ratchet algorithm](https://github.com/xmtp/libxmtp/blob/main/README.md#double-ratchet-messaging).
```

Just want to be careful using the term `session key` - in Signal's [spec](https://signal.org/docs/specifications/doubleratchet/), `session key` refers to the initial root key of the double ratchet algorithm, and `message key` refers to the encryption key used per-message.

examples/cli/README.md
richardhuaaa, line 46:
```suggestion
If you want to run the CLI against localhost, go to the root directory and run `dev/up` to start a local server. Then run the CLI commands using the `--local` flag.
```

README.md
```diff
@@ -1,36 +1,89 @@
...
+- [`bindings_js`](https://github.com/xmtp/libxmtp/tree/main/bindings_js): JS bindings (in progress)
+- [`bindings_wasm`](https://github.com/xmtp/libxmtp/tree/main/bindings_wasm): Wasm bindings (in progress)
+
+## Key features of XMTP v3-alpha
+
+### Double ratchet messaging
+
+**What**: XMTP v3-alpha introduces the use of the [double ratchet algorithm](https://signal.org/docs/specifications/doubleratchet/) to keep past and future messages secure.
+
+**Why**: With double ratchet messaging, each message is encrypted by its own session key. If the session key a sender used to send a message becomes compromised:
```
richardhuaaa, line 51:
```suggestion
**Why**: With double ratchet messaging, each message is encrypted by its own key. If a message key becomes compromised:
```
As with the other comment, being a little bit specific about the session/message key terminology, sorry I didn't bring this up earlier.

Some degree of forward secrecy/PCS is actually maintained if any of the message/session/installation keys are compromised, but would require more nuance to explain below, so happy with this level of granularity as more detail can be found in the double ratchet spec.

examples/cli/README.md
neekolas: We can do it as part of that PR

README.md
```diff
@@ -1,36 +1,89 @@
...
+
+With XMTP v3-alpha, there is no XMTP user key bundle. Instead, the user signs with their blockchain account to grant permission to an installation key bundle to send messages from their account address. The user signs to grant permission to an installation key bundle for each installation of an app built with XMTP. The installation key bundle is stored on each device and is not required to be stored on the XMTP network.
+
+**Why**: Storing keys on the XMTP network makes them potentially available to anyone. Moving key storage to devices significantly improves security, as keys never leave a user’s device. Additionally, installation key bundles provide separate and revocable keys for each app installation on a device, simplifying key management. When a user revokes (deletes) a key from a device, you can be sure that it is gone.
```
richardhuaaa, line 77:
```suggestion
**Why**: Storing keys on the XMTP network makes them potentially available to anyone. Moving key storage to devices significantly improves security, as keys never leave a user’s device. When a user deletes a key from a device, you can be sure that it is gone. Additionally, installation key bundles provide separate and revocable keys for each app installation on a device, simplifying key management.
```
Small nuance here - deletion and revocation are different:
- Deletion means destroying the private key for an installation, which is something you might do if you control the installation and want to clean up after yourself.
- Revocation (not implemented yet) means publishing something onto the network that declares a certain installation key is no longer valid, which is something you might do for a compromised installation on a device that you do not control.

neekolas (approved):

richardhuaaa (approved):
This is crystal clear, thanks so much for doing this!

PR #200, opened 2023-08-16
Change session handling

jazzz:
## Problem
The receive path is not currently handing messages correctly, which is resulting in errors when receiving messages. 

The core issue is that new Prekey messages are not creating new sessions when needed, which is leading to decryption failures. The end result is that two installations can never get synchronized if two sessions were created simultaneously.

## Solution
I'm proposing this new receive path which will handle the case where two peers are sending from different sessions. The change here is that PreKey messages should be checked against existing sessions: If no match create a new session.

### DIFF
--- a/xmtp/README.md
+++ b/xmtp/README.md
@@ -192,28 +192,31 @@ process_inbound_messages():
         If the payload is malformed and the proto cannot be decoded:
             Set inbound_invite state to INVALID
             continue
-        If an existing session exists with the `sender.installation_id`:
-            Decrypt the ciphertext using the existing session
+
+        If session from get_session:
+            Decrypt ciphertext using session
             If decryption fails:
                 Set inbound_message state to DECRYPTION_FAILURE
                 continue
-            Update session in the database
-        else:
-            Create a new inbound session with the sender
-            Decrypt the ciphertext using the new session
-            If decryption fails:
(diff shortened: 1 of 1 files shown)


jazzz:
@neekolas 

>If either side has already received a message, the old flow works, right?
Yes. Mostly. 

> this is handling the case where Alice and Bob both initiate a conversation with one another before either side has received a message from the other?

Yes and self fanout in the case of multiple installations.

---

If an installation is learning about the other client via a message then everything will work. In the A->B if B has never heard of A then it will create the correct session and it messages can flow for a time being. 

Since the implementation incorrectly assumes that only one session will between two installations, we quickly get into trouble.

Currently this trouble state can be reach in the following ways:
- On initialization clients check for any new "self" installations and optimistically create a session. Assuming installations A1,A2 if A1 discovers A2 from a send initialed refresh ( e.g. A1 was sending to B) then sessions between A1 and A2 will be corrupted
- Simultaneous session creation.

In normal operation a session would see that it is receiving a PreKeyMessage for a session it does not have and would create a new one, allowing the clients to synchronize

neekolas (approved):

richardhuaaa (approved):
We'll need some way of cleaning old sessions, so that their keys don't last forever, but sounds good to me

PR #201, opened 2023-08-16
Allow Multiple Sessions

jazzz:
This PR implements the approach in #200 

The major change is that the sessions table now contains multiple sessions for a given peer_installation_id. As a heuristic the most recently updated sessions is the the one that should be used for outdoing messages.

Note: This breaks existing installations. All previous installations should be deleted to ensure consistency

### DIFF
--- a/xmtp/migrations/2023-06-03-011250_create_sessions/up.sql
+++ b/xmtp/migrations/2023-06-03-011250_create_sessions/up.sql
@@ -2,6 +2,6 @@ CREATE TABLE IF NOT EXISTS sessions (
     session_id TEXT PRIMARY KEY NOT NULL,
     created_at BIGINT NOT NULL,
     updated_at BIGINT NOT NULL,
-    peer_installation_id TEXT  NOT NULL,
+    peer_installation_id TEXT NOT NULL,
     vmac_session_data BLOB NOT NULL
 )
--- a/xmtp/migrations/2023-06-03-011250_create_sessions/up.sql
+++ b/xmtp/migrations/2023-06-03-011250_create_sessions/up.sql
@@ -1,6 +1,7 @@
 CREATE TABLE IF NOT EXISTS sessions (
     session_id TEXT PRIMARY KEY NOT NULL,
     created_at BIGINT NOT NULL,
+    updated_at BIGINT NOT NULL,
     peer_installation_id TEXT  NOT NULL,
     vmac_session_data BLOB NOT NULL
 )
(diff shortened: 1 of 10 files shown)


xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -161,35 +163,56 @@ impl EncryptedMessageStore {
...
+        let session_list = sql_query("Select sessions.* from
+                    (
+                        select distinct
+                            first_value(session_id) over (partition by peer_installation_id order by updated_at desc) as session_id
+                        from sessions
+                        where user_address = ?
+
+                    )as ids
+                left join sessions on ids.session_id = sessions.session_id
+        ").bind::<Text,_>(user_address).load::<StoredSession>(conn).map_err(|e| StorageError::Unknown(e.to_string()))?;
```
jazzz, line 214: Would love to use diesel here, but couldn't find another way. Minimally the inputs are checked for length, and characters for added safety.

An approach that tags sessions as 'active' could be looked at but has more opportunity for bugs in managing the state

jazzz:
With this change messages can now flow between self installations:

B:
image
A2
image

xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -161,35 +163,56 @@ impl EncryptedMessageStore {
...
+            ));
+        }
+        let session_list = sql_query("Select sessions.* from
+                    (
+                        select distinct
+                            first_value(session_id) over (partition by peer_installation_id order by updated_at desc) as session_id
+                        from sessions
+                        where user_address = ?
+
+                    )as ids
```
neekolas, line 212: Might want to run this through a SQL formatter for readability.

xmtp/migrations/2023-06-03-011250_create_sessions/up.sql
```diff
@@ -1,6 +1,7 @@
 CREATE TABLE IF NOT EXISTS sessions (
-    peer_installation_id TEXT PRIMARY KEY NOT NULL,
-    session_id TEXT NOT NULL,
+    session_id TEXT PRIMARY KEY NOT NULL,
     created_at BIGINT NOT NULL,
+    updated_at BIGINT NOT NULL,
+    peer_installation_id TEXT  NOT NULL,
```
neekolas, line 5: nit: looks like an extra space in here

neekolas (approved):

xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -161,35 +163,68 @@ impl EncryptedMessageStore {
...
+
+        sessions
+            .filter(peer_installation_id.eq(installation_id))
+            .order(updated_at.desc())
+            .get_results(conn)
+            .map_err(|e| StorageError::Unknown(e.to_string()))
     }
 
-    pub fn get_sessions(
+    pub fn get_latest_sessions(
```
richardhuaaa, line 195: Not needed for this PR, but would it be worth adding a unit test on this one, seeing as the SQL is complicated?

richardhuaaa (approved):
:+1:

PR #202, opened 2023-08-16
Refresh only on send

jazzz:
## Problem 
Currently libxmtp refreshes installations On Client:initialization, conversation:Initialization and message sending. 

This change isn't strictly a problem but does add complexity around when sessions are initiated.

## Solution 
Refresh installations only on send. Currently Send is the only time when new installations are needed. 
- Receive uses existing sessions or creates a new one from a prekey message. 
- Installation management is not implemented yet. 

To cut down on complexity and help with debugging for the time being, this PR updates the library to only refresh on send.

### DIFF
--- a/xmtp/src/conversation.rs
+++ b/xmtp/src/conversation.rs
@@ -190,9 +190,6 @@ where
     }
 
     pub async fn initialize(&self) -> Result<(), ConversationError> {
-        self.client
-            .refresh_user_installations(self.peer_address().as_str())
-            .await?;
         let inner_invite_bytes = Invitation::build_inner_invite_bytes(self.peer_address.clone())?;
         let conn = &mut self.client.store.conn()?;
         for contact in self.members(conn)?.iter() {
--- a/examples/cli/cli-client.rs
+++ b/examples/cli/cli-client.rs
@@ -64,8 +64,6 @@ enum Commands {
         msg: String,
     },
     Recv {},
-
-    Refresh {},
     ListContacts {},
     Clear {},
 }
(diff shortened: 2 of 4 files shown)


xmtp/src/conversation.rs
```diff
@@ -190,9 +190,6 @@ where
     }
 
     pub async fn initialize(&self) -> Result<(), ConversationError> {
-        self.client
```
neekolas, line 193: Don't we want to keep this one so we can send invites to the most up-to-date set of participant installations?

richardhuaaa: I wonder if it makes sense to separate out refreshing installations (i.e. pre-fetching contacts from the network whenever is optimal), and creating sessions with those installations (which ideally happens only on send).

richardhuaaa: But currently we're fetching on every send regardless (refresh interval of 0), and a low interval might actually be ideal, so maybe it matters less

richardhuaaa: Agree - the invite is effectively a message, ideally we can use similar code-paths for invites compared to messages

richardhuaaa (approved):
I wonder if it makes sense to separate out refreshing installations (i.e. pre-fetching contacts from the network whenever is optimal), and creating sessions with those installations (which ideally happens only on send).

But currently we're fetching on every send regardless (refresh interval of 0), and a low interval might actually be ideal, so maybe it matters less.

Pre-approving regardless of what you decide, assuming @neekolas's comment is addressed

xmtp/src/conversation.rs
jazzz: Thats a fair point. I got a little over zealous in this case. will update

jazzz:
@richardhuaaa I agree that refresh needs some work. In the long term having an offline cache would be helpful for sending messages when clients don't have network connections. So separating those would he helpful. 

Given the number of high priorities I'm inclined to revisit this later, though acknowledging thats where it needs to be in the future

PR #204, opened 2023-08-18
Add conversation bindings

neekolas:
## Summary

- Adds binding objects for `Conversation` and `Conversations`
- Adds method for listing conversations
- Adds method for sending messages (as raw bytes expected to be `EncodedContent`)
- Adds method for listing messages
- Renames `send_message` to `send_text` and creates a new lower-level method for sending raw bytes `conversation.send()`

## Notes
- The tests only pass when `tokio::test` is run with the flavor of `multi_thread`. More testing is required to see if this is compatible with the actual runtime used by Android and iOS in Uniffi. If that creates deadlocks (which happens with the default tokio test runner) we will have to remove all use of `block_on` in our code.

## Next up
- We need to figure out the API ergonomics for when to call `conversations.receive`. Right now, this never gets called. Do we want to do this as part of `conversation.list_messages()`? Keep it as a separate function that must be triggered manually?


xmtp/src/conversation.rs
```diff
@@ -103,7 +103,7 @@ impl<'c, A> SecretConversation<'c, A>
 where
     A: XmtpApiClient,
 {
-    pub(crate) fn new(client: &'c Client<A>, peer_address: Address) -> Self {
+    pub fn new(client: &'c Client<A>, peer_address: Address) -> Self {
```
neekolas, line 106: Need to expose this to the bindings so that we can create Conversations on demand.

bindings_ffi/src/lib.rs
```diff
@@ -76,25 +78,167 @@ pub async fn create_client(
...
+            inner_client: self.inner_client.clone(),
+            peer_address: convo.peer_address(),
+        });
+
+        Ok(out)
+    }
+
+    pub async fn list(&self) -> Result<Vec<Arc<FfiConversation>>, GenericError> {
+        let inner = self.inner_client.as_ref();
+        let conversations = xmtp::conversations::Conversations::new(inner);
```
neekolas, line 131: The lifetimes on our `Conversation` and `Conversations` method are incompatible with Uniffi exported objects, so instead we never store the structs on anything that gets exposed and just build them on demand. That sidesteps the need for compatible lifetimes with objects that move over the FFI boundary.

xmtp/src/storage/encrypted_store/mod.rs
```diff
@@ -348,6 +348,31 @@ impl EncryptedMessageStore {
         Ok(())
     }
 
+    pub fn get_refresh_job(
```
richardhuaaa, line 351: Is this meant to be part of this PR? Don't see it called anywhere

richardhuaaa (approved):
Not an issue right now - the approach of making a new Conversations object each time may break in the future if Conversations is ever stateful (e.g. keeps a list of conversations).

But we can work that out if we ever get there, probably making some kind of wrapper object that keeps the `Client` and `Conversations` together so that one can't be cleaned up without the other.

xmtp/src/storage/encrypted_store/mod.rs
neekolas: I added it when I thought I was going to have to remove the transaction callback. Can remove until we need it.

neekolas:
Btw, I've tested this on an actual Android device and don't see any signs of deadlock. Promising sign, but will keep monitoring as we add more bindings.

PR #206, opened 2023-08-22
Simplify external API interface

richardhuaaa:
There are two changes here:

1. Make all methods of `Conversations` static, and accept a `Client` argument, so that `Conversations` doesn't need to be instantiated anywhere. `Conversations` doesn't really hold state anyway and effectively functions as a wrapper for `Client` right now, so it doesn't make sense for consumers to have to instantiate it and hold onto it.
2. Call `process_outbound_messages()` on init and on message send, so that consumers don't have to manually call it.

This simplifies the client API quite a bit, as can be seen in `bindings_ffi` and in the `CLI`.

Ideally, we could have spawned a separate thread to call `process_outbound_messages` so that the existing thread is not blocked, but that requires additional refactoring. This requires us to enforce `XmtpApiClient` is `Send + Sync` (which is easy), but also comes up against issues with lifetimes for `Client`, for which a reference is needed in the other thread that may outlive the current thread. This possibly requires us to start using `Arcs` again.

### DIFF
--- a/xmtp/src/client.rs
+++ b/xmtp/src/client.rs
@@ -1,27 +1,37 @@
 use core::fmt;
-use std::fmt::Formatter;
+use std::{fmt::Formatter, time::Duration};
 
 use diesel::Connection;
 use log::{debug, info};
+use prost::Message;
 use thiserror::Error;
-use vodozemac::olm::PreKeyMessage;
+use vodozemac::olm::{self, PreKeyMessage};
 
 use crate::{
     account::Account,
     contact::{Contact, ContactError},
-    conversations::Conversations,
+    conversation::peer_addr_from_convo_id,
     session::SessionManager,
     storage::{
-        now, DbConnection, EncryptedMessageStore, StorageError, StoredInstallation, StoredSession,
-        StoredUser,
+        now, DbConnection, EncryptedMessageStore, MessageState, OutboundPayloadState, StorageError,
+        StoredInstallation, StoredMessage, StoredOutboundPayload, StoredSession, StoredUser,
     },
     types::networking::{PublishRequest, QueryRequest, XmtpApiClient},
     types::Address,
-    utils::{build_envelope, build_user_contact_topic, key_fingerprint},
+    utils::{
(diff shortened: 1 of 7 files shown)


xmtp/src/client.rs
```diff
@@ -119,6 +120,13 @@ where
         }
 
         self.is_initialized = true;
+
+        // Send any unsent messages
+        let conversations = Conversations::new(&self);
```
neekolas, line 125: Self is already a reference here

xmtp/src/conversation.rs
```diff
@@ -158,15 +159,20 @@ where
         )
         .store(&mut self.client.store.conn().unwrap())?;
 
+        let conversations = Conversations::new(&self.client);
+        if let Err(err) = conversations.process_outbound_messages().await {
```
neekolas, line 163: I could get on board with moving `process_outbound_messages` to `Client`. It's really not a "conversations" specific function (like `list` or `newConversation` are), so much as a general processing workload for messages.

neekolas (approved):

richardhuaaa:
There is a failing unit test - as part of the refactor I made `Conversations::list()` call the `Conversation` constructor that calls `insert_or_ignore_user` where it didn't before, and for some reason the `User` table is not recognized. Will fix this before landing, but the CLI and all other tests are passing.

neekolas (approved):

neekolas:
I've run into table not recognized issues before. It has to do with having multiple concurrent open connections to the ephemeral SQLite. Apparently that causes it to create a new in-memory SQLite instance.

richardhuaaa:
@neekolas Thank you for the pointer! The behavior in the ephemeral case is unexpected, but understandable. I've put in a change to limit the pool size to 1 for the ephemeral instance, so that we won't run into this case again. In the event that a connection is requested from the pool while another one is ongoing, it will wait for a time limit before expiring with an error.

An added plus of this is it's much easier for tests to detect deadlocks. In the latest commit I fixed a deadlock that I introduced (which ordinarily self-resolves if the pool size is 2 or greater).

PR #208, opened 2023-08-29
Update uniffi version

jazzz:
- Updates the uniffi crate the the latest version; Incorporating https://github.com/mozilla/uniffi-rs/pull/1692.
- updates the package name from `bindings_ffi` to `xmtpv3` to provide consistent module paths. 

```
---- uniffi_foreign_language_testcase_test_android_kts stdout ----
Creating testing out_dir: /Users/jazz/dev/libxmtp/bindings_ffi/target/tmp/bindings_ffi-3492e72b41fe83b6
thread 'uniffi_foreign_language_testcase_test_android_kts' panicked at 'assertion failed: `(left == right)`
  left: `CallbackInterface { module_path: "bindings_ffi", name: "FfiLogger" }`,
 right: `CallbackInterface { module_path: "xmtpv3", name: "FfiLogger" }`', /Users/jazz/.cargo/git/checkouts/uniffi-rs-6f89edd2a1ffa4bd/cae8edc/uniffi_bindgen/src/interface/universe.rs:42:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -186,10 +186,6 @@ jobs:
           path: .
       - name: Build archive
         run: |
-          # not sure why we're building this one
-          # mv aarch64-unknown-linux-gnu/libxmtpv3.so jniLibs/arm64-v8a/
-          # mkdir -p jniLibs/x86_64
-          # mv x86_64-linux-android/libxmtpv3.so jniLibs/x86_64/
           mkdir -p jniLibs/x86
           mv i686-linux-android/libxmtpv3.so jniLibs/x86/
           mkdir -p jniLibs/armeabi-v7a
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: ${{ matrix.target }}
-          path: bindings_ffi/target/${{ matrix.target }}/release/libbindings_ffi.so
+          path: bindings_ffi/target/${{ matrix.target }}/release/libxmtpv3.so
           retention-days: 1
 
   build-macos:
@@ -107,7 +107,7 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: ${{ matrix.target }}
-          path: bindings_ffi/target/${{ matrix.target }}/release/libbindings_ffi.dylib
+          path: bindings_ffi/target/${{ matrix.target }}/release/libxmtpv3.dylib
           retention-days: 1        
 
   kotlin:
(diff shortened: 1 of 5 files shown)


jazzz:
Getting a failure on: https://github.com/xmtp/libxmtp/blob/573bdc669084bf068dd4eb34158d28958dca8599/.github/workflows/build.yml#L189

Comments seem to suggest its not relevant. Is this an important build step or should it be removed?

richardhuaaa:
That is a comment applying to the commented out lines underneath, so not directly related to the error. There are commands being run under that, specifically the moving of compiled binaries named `libbindings_ffi` etc. After renaming the module from `bindings_ffi` to `xmtpv3`, all of the build needs to be updated to reference the new name. I can help with that if it is tricky

richardhuaaa:
Also, I didn't originally do this because I was wondering whether it is correct to name the `bindings_ffi` crate `xmtpv3`? Wondering if that presents a naming conflict or other confusion in the future, although happy to sign off on this.

jazzz:
>Also, I didn't originally do this because I was wondering whether it is correct to name the bindings_ffi crate xmtpv3? Wondering if that presents a naming conflict or other confusion in the future, although happy to sign off on this.

I had the same question, specifically around the package name (not the root crate).  I figured I'd get this PR compiling, and then we could consider a name change, which would break some more references.

Currently the import in a future Android Platform SDK would looks like `import uniffi.xmtpv3.*`. Would `uniffi.xmtp_ffi.*` be preferred? Something else perhaps? Happy to include that change in this set if there is some agreement on the name

jazzz:
@richardhuaaa Thanks for the direction on the build fix. I was misunderstanding the library generation. All references have been updated

bindings_ffi/Cargo.toml
```diff
@@ -11,14 +11,19 @@ crate-type = ["lib", "cdylib"]
 [dependencies]
 log = { version = "0.4", features = ["std"] }
 thiserror = "1.0.40"
-uniffi = { git = "https://github.com/mozilla/uniffi-rs", rev = "6938c262293802f00a373f0d1aa16a2079071b24", features = ["tokio", "cli"] }
-uniffi_macros = { git = "https://github.com/mozilla/uniffi-rs", rev = "6938c262293802f00a373f0d1aa16a2079071b24" }
+uniffi = { git = "https://github.com/mozilla/uniffi-rs", rev = "cae8edc45ba5b56bfcbf35b60c1ab6a97d1bf9da", features = [
+    "tokio",
+    "cli",
+] }
+uniffi_macros = { git = "https://github.com/mozilla/uniffi-rs", rev = "cae8edc45ba5b56bfcbf35b60c1ab6a97d1bf9da" }
```
mkobetic, line 18: Do we want to stick with hardcoded commits or should we switch to using tags? They would at least communicate more clearly which version we are using. I realize that tags can move, but we presumably assume responsible maintainer regardless and the fact that it can move would be a feature, not a bug.

mkobetic:
> That is a comment applying to the commented out lines underneath, so not directly related to the error. 

Indeed, I put that in because I didn't see a point of building that target and commented it out. If we agree we can probably take it out completely.

.github/workflows/build.yml
```diff
@@ -187,15 +187,15 @@ jobs:
       - name: Build archive
         run: |
           # not sure why we're building this one
-          # mv aarch64-unknown-linux-gnu/libbindings_ffi.so jniLibs/arm64-v8a/
+          # mv aarch64-unknown-linux-gnu/libxmtpv3.so jniLibs/arm64-v8a/
           # mkdir -p jniLibs/x86_64
-          # mv x86_64-linux-android/libbindings_ffi.so jniLibs/x86_64/
+          # mv x86_64-linux-android/libxmtpv3.so jniLibs/x86_64/
```
mkobetic, line 192: This one didn't seem to build when I tried, do we want this target eventually?

mkobetic (approved):

bindings_ffi/Cargo.toml
jazzz: In general I'd be in favor of this -- though it does look like the last [release](https://github.com/mozilla/uniffi-rs/releases/tag/v0.24.3) doesn't include the given change. 

 > the fact that it can move would be a feature, not a bug.
 
This is not always true in rust, particularly when it comes to package version resolution. Changing of a dependency version can cause conflicts in other packages if they have strict dependency requirements.

.github/workflows/build.yml
jazzz: These appear to adding confusion more than anything. I've removed them for now.

PR #217, opened 2023-08-31
build: wasm build with xmtp_proto and xmtp_cryptography

dmccartney:
This sets up the test CI for `bindings_wasm`.
It also makes some progress towards getting the other crates wasm-friendly:
- [x] `xmtp_cryptography` works as-is
- [x] `xmtp_proto` works now w/ a new `grpc` feature that triggers the not-wasm-friendly grpc client generation

For future PRs:
- [ ] `xmtp_networking` will probably split into
  - the existing `xmtp_api_grpc` (not wasm-friendly) and 
  - a new `xmtp_api_grpc_gateway` (wasm-friendly)
- [ ] `xmtp` will need some surgery to allow more configuration of the storage system


xmtp_proto/buf.gen.yaml
```diff
@@ -16,6 +16,8 @@ plugins:
       - no_server=true
       - compile_well_known_types
       - extern_path=.google.protobuf=::pbjson_types
+      # Exclude it from non-tonic builds (so we can use the rest in Wasm)
+      - client_mod_attribute=xmtp.message_api.v1=#[cfg(feature = "tonic")]
```
dmccartney, line 20: This was the sneaky buf change to include a `cfg` that excludes the tonic client from non-tonic builds.

xmtp_proto/src/gen/xmtp.message_api.v1.tonic.rs
```diff
@@ -1,5 +1,6 @@
 // @generated
 /// Generated client implementations.
+#[cfg(feature = "tonic")]
 pub mod message_api_client {
```
dmccartney, line 4: this is the effect of that ^ buf config change on the generated output (i.e. exclude it unless `tonic` feature is enabled)

dmccartney: (the `xmtp_proto/Cargo.toml` already had the `grpc` feature configured to imply the `tonic` feature)

xmtp_proto/buf.gen.yaml
neekolas: Clever

neekolas (approved):
This is looking like great progress. Love it

PR #219, opened 2023-09-02
Support persisted store in FFI

nakajima:
This PR exposes the ability for FFI clients to specify where the local DB should live. See https://github.com/xmtp/xmtp-swift/pull/2 for an implementation in Swift.

I know very little about Rust so feel free to tear this to shreds.


bindings_ffi/src/lib.rs
```diff
@@ -65,8 +70,21 @@ pub async fn create_client(
...
+
+        None => {
+            info!("Using ephemeral store");
+            EncryptedMessageStore::new(StorageOption::Ephemeral, static_enc_key())
+        }
+    };
+
     let mut xmtp_client: RustXmtpClient = xmtp::ClientBuilder::new(inbox_owner.into())
         .api_client(api_client)
+        .store(store.unwrap())
```
nakajima, line 87: Unwrap is bad, right?

bindings_ffi/src/lib.rs
```diff
@@ -302,6 +323,44 @@ mod tests {
...
+        .await
+        .unwrap();
+
+        let other_installation_id = client_b.inner_client.installation_id();
+        drop(client_b);
+
+        assert!(
+            installation_id == other_installation_id,
+            "did not use same installation ID"
+        )
```
nakajima, line 361: I wasn't sure if there was a better thing to assert here, but this test does fail if you don't pass a location for the local DB so I think it's at least verifying something.

bindings_ffi/src/lib.rs
neekolas: Unwrap will panic if an error is found. Would be nicer to have a `.map_err` and convert the error to a `GenericError`, which will then get returned to the native iOS/Android caller.

bindings_ffi/src/lib.rs
```diff
@@ -65,8 +70,21 @@ pub async fn create_client(
         .await
         .map_err(|e| stringify_error_chain(&e))?;
 
+    let store = match db {
+        Some(path) => {
+            info!("Using persistent storage: {} ", path);
+            EncryptedMessageStore::new_unencrypted(StorageOption::Persistent(path))
```
neekolas, line 76: Shouldn't we also take in an encryption key here from the caller, and provide an encrypted DB by default?

bindings_ffi/src/lib.rs
richardhuaaa: If you'd like, you can send a message from Client A, and then list messages on Client B, and verify that the sent message is in the list (the DB is used to store it)

bindings_ffi/src/lib.rs
nakajima: Cool, fixed in https://github.com/xmtp/libxmtp/pull/219/commits/9ee92904cfd5f838913c1d0efa5c6a5faed539a5.

bindings_ffi/src/lib.rs
nakajima: Yep you're totally right. Added the ability to pass a key in https://github.com/xmtp/libxmtp/pull/219/commits/54bcfeaaa7bac73db12d9aa33ac8feea33e9e579. We can ensure clients _always_ pass a key in the SDKs.

richardhuaaa (approved):

PR #220, opened 2023-09-08
Remove invites and convo state

richardhuaaa:
We did some fantastic work on invites, but with inbox topics the concept doesn't quite make as much sense in v3 as it did in v2. We were thinking it might be a placeholder for a future consent mechanism, but it brought up some complex design decisions that we're not ready to make yet.

Some changes to call out here:
1. We no longer automatically send invites when we create a conversation, which means the other side will not be aware that a conversation exists until you send the first message. This is analogous to how most messaging apps work.
2. When refreshing our list of conversations, we need to receive all incoming messages, rather than receive all incoming invites.


xmtp/src/conversations.rs
```diff
@@ -191,6 +180,8 @@ impl<A: XmtpApiClient> Conversations<A> {
 
         //TODO: Validate message
 
+        // TODO move this logic into a Conversation::save_message() method
+        SecretConversation::ensure_conversation_exists(client, conn, &message_obj.convo_id)?;
```
richardhuaaa, line 184: We're not guaranteed to have received an invite first (which would have created the conversation row in the DB), so we need to add the row when receiving messages.

neekolas (approved):
Love to see PRs that look like this 
CleanShot 2023-09-08 at 15 12 06@2x

PR #221, opened 2023-09-08
Fix test failures caused by two tests fighting over the same file.

nakajima:
These tests were passing in isolation but failed when run concurrently. This PR fixes that and adds a comment to hopefully prevent it from happening in the future.

### DIFF
--- a/bindings_ffi/src/lib.rs
+++ b/bindings_ffi/src/lib.rs
@@ -275,7 +275,10 @@ mod tests {
     };
     use tempfile::TempPath;
     use xmtp::InboxOwner;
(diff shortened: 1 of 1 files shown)


neekolas (approved):
Approving, although maybe we should just make this generate a random filename for the DB so that we don't need the comment/warning

nakajima:
> although maybe we should just make this generate a random filename for the DB 

You're right. Fixed in https://github.com/xmtp/libxmtp/pull/221/commits/d1c861191b57601f53e370ab79f818e2b623842d.

neekolas (approved):

Issue #227, opened 2023-09-12
Ramp-up task: Detect DB errors


richardhuaaa:
### Is your feature request related to a problem?

This ramp-up task will get you familiar with the database, with Rust error handling, and with writing unit tests.

We are using Diesel ORM with a sqlite database. Most of the DB operations are in `mod.rs` for encrypted store. We are using `Diesel::insert_or_ignore` in a few places, for example: https://github.com/xmtp/libxmtp/blob/4494d4fe6a536292663ac1f82a6a9cdeaa797a1f/xmtp/src/storage/encrypted_store/mod.rs#L455.

Originally, the reason for this was to ignore duplicate insertion errors, so that the operation is idempotent if it is called twice and the row already exists. However, a problem with `Diesel::insert_or_ignore` is it will ignore all errors, including foreign key constraint violations, for example - which results in the row silently not being inserted at all. This makes debugging difficult.

### Describe the solution to the problem

My rough idea is to just use the normal `Diesel::insert`, and then add code specifically to detect when a duplicate primary key error happens and just return `Ok()`, while surfacing the rest of the errors normally. This could be put in a helper method that is used everywhere `insert_or_ignore` was used. I also thought about using upsert, but we don't want to update any columns (e.g. state) if a row already exists. It's up to you what the best way of doing this is.

 Ideally we add a unit test for this as well.

PR #228, opened 2023-09-13
Update insert_or_ignore

daria-github:
This PR updates usages of `insert_or_ignore` to just insert and then run `Ok()` in the event of a duplicate primary key error. Updated var names as well as updated the order for consistency in the imports of similar methods. 

I wanted to export this to a helper method and add a unit test, but couldn't get Rust to compile with any variation I tried.  

This should close out https://github.com/xmtp/libxmtp/issues/227.

### DIFF
--- a/xmtp/src/builder.rs
+++ b/xmtp/src/builder.rs
@@ -170,7 +170,7 @@ where
             #[cfg(test)]
             AccountStrategy::ExternalAccount(a) => a,
         };
-        store.insert_or_ignore_user(StoredUser {
+        store.insert_user(StoredUser {
             user_address: account.addr(),
             created_at: now(),
             last_refreshed: 0,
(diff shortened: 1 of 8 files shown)


xmtp/src/builder.rs
```diff
@@ -170,7 +170,7 @@ where
             #[cfg(test)]
             AccountStrategy::ExternalAccount(a) => a,
         };
-        store.insert_or_ignore_user(StoredUser {
+        store.insert_user(StoredUser {
```
richardhuaaa, line 173: Would it make sense to name this something like `ensure_user_exists` or `create_user_if_needed`, to clarify that it might already exist?

richardhuaaa (approved):
This looks great! It might be a good learning experience to add a unit test, but totally up to you.

Re: helper method discussion, it's not strictly necessary, but if you want to do it it could be structured as follows. Agree it wouldn't be worth it to wrap the entire function in a helper method.
```
{
    let result = match diesel::insert_into(users::table)
        .values(user)
        .execute(conn);
    ignore_unique_violation(result)
}
```

PR #254, opened 2023-09-26
Refactor store

neekolas:
## Summary

Refactors the `EncryptedMessageStore` so that all methods that access the database explicitly take a `conn` as the first argument. We were doing this in most places, but had some older code that didn't follow the newer pattern. This makes all methods on the store safe to be run from inside a transaction.

This change removes any implicit creation of new connections, which can lead to deadlocks or strange behaviour when these methods are run from inside transactions.


richardhuaaa (approved):
Nice, this is much cleaner!

PR #255, opened 2023-09-27
Set timeout for Tonic requests

richardhuaaa:
Not sure if this will help the issue with hanging requests, possibly we can do some kind of exponential backoff thing. Need to test against the repro


neekolas (approved):
Can't hurt to try

richardhuaaa:
Going to land this so that the artifacts can be built and deployed for us to test

PR #256, opened 2023-09-27
Update insert_or_ignore

daria-github:
This PR updates usages of insert_or_ignore to consume a helper that runs Ok() in the event of a duplicate primary key error. Updated var names and added some tests for the helper.

This should close out https://github.com/xmtp/libxmtp/issues/227.


richardhuaaa (approved):
Very clean!

PR #258, opened 2023-09-29
MLS Experiment

neekolas:
## Summary

Hacky prototype of getting MLS to work with XMTP. I used the [`cli` example](https://github.com/openmls/openmls/tree/main/cli) from OpenMLS as reference, and a lot of code is borrowed from there. This is not meant to be used anywhere and is purely for educational purposes.

All of my code lives in `client.rs` and `group.rs`. The rest is from the CLI example.

## Features

- Uses XMTP as a delivery service to store welcome messages, proposals, and application messages.
- Serializes the client's keystore to disk as JSON files
- Supports group creation
- Supports adding new members to a group
- Supports sending messages to a group

## Important missing things

- Removing members from a group
- ~Any concept of wallet addresses. Clients are identified by a random UUID and are not authenticated beyond that~ Wallet address identities are now supported
- Any concept of having multiple devices per user
- Proper SQLite database, or any way of persisting messages to disk
- Support for one-time KeyPackages. I just create a single KeyPackage per user and re-use it. Pretty sure that means this will break if we have multiple groups per user.
- Any real error handling. The code is littered with `expect`s and is designed to crash if anything goes wrong
- Any support for generic `ApiClient` or storage types
- Any pagination of API requests
- Without a database to hold things like group info, there are some actions that will only work once. For example, if you create a group and run `client.load_groups()` twice, it will panic since the SDK will have already deleted the associated key package after the first run.

## Learnings

- KeyPackages in MLS are the equivalent of PreKeys in Double Ratchet. We will need a system to somehow cross off KeyPackages
- Welcome Messages in MLS are the equivalent of V2 invites
- Because our delivery service is just a topic, users will have to read their own commit messages. OpenMLS doesn't like that (since the commits have already been processed) and some of the errors you'll see logged when running the tests are caused by this. 
- There is a real shortage of examples using OpenMLS. If we go down this path, we are going to have to become true experts in how the package works and likely contributors to it.

## Known issues
- I get some scary warnings when receiving messages from the network. Seems to be related to mismatched epochs. The tests still pass, but it's something that we should properly understand (which I definitely don't). Some are caused by the client reading it's own commits, but that may not be all the issues. Running the `test_send_multiple_clients` prints all this to the console...but still seems to work. New members get added, existing members can see the new members, and messages are sent and received. But I'm skeptical.
```
[2023-09-29T13:55:45Z ERROR openmls::group::public_group::validation] Wrong Epoch: message.epoch() 0 != 1 self.group_context().epoch()
error processing message: ValidationError(WrongEpoch)
adding third member
[2023-09-29T13:55:46Z ERROR openmls::group::public_group::validation] Wrong Epoch: message.epoch() 0 != 1 self.group_context().epoch()
error processing message: ValidationError(WrongEpoch)
[2023-09-29T13:55:46Z ERROR openmls::tree::secret_tree] This is the wrong ratchet type.
[2023-09-29T13:55:46Z ERROR openmls::framing::private_message_in]   Ciphertext generation out of bounds 0
        RatchetTypeError
error processing message: ValidationError(UnableToDecrypt(GenerationOutOfBound))
[2023-09-29T13:55:46Z ERROR openmls::group::public_group::validation] Wrong Epoch: message.epoch() 0 != 2 self.group_context().epoch()
error processing message: ValidationError(WrongEpoch)
error processing message: ValidationError(UnableToDecrypt(AeadError))
[2023-09-29T13:55:46Z ERROR openmls::group::public_group::validation] Wrong Epoch: message.epoch() 1 != 2 self.group_context().epoch()
error processing message: ValidationError(WrongEpoch)
test group::tests::test_send_multiple_clients ... ok
```

- I had to make this its own crate separate from the workspace due to dependency conflicts with VMAC
- I'm not sure about the way we are using signatures as credentials in this implementation. It's not clear to me if OpenMLS will protect against a case where a member of a group sends messages using the credentials of another member. This will need testing.


jac18281828:
This is really cool!  Thanks for doing this!  🔥🚀😅

neekolas:
Some more thoughts on MLS

### Wallet ownership

The MLS specification defines two types of identity that can be included in a credential: `Basic` and `X.509`. Basic identities are simply an opaque byte array. X.509 includes a [full certificate](https://en.wikipedia.org/wiki/X.509) used to associate the signing key with another public key by signature, and can include additional information.

X.509 feels like the most standards-driven and rich approach, but X.509 support is currently [unimplemented in OpenMLS](https://github.com/openmls/openmls/issues/134).

We need the wallet association to be included in the credential so that it gets included in each leaf of the tree and can be verified by all members of the group.

The short-term solution is to use a wallet signature of the signature public key bytes as the identity. This can be verified by all members of the group. X.509 support feels like a solid long-term solution. If any member attempts to add a key package with an invalid or missing wallet signature, we need to make sure that the commit is consistently ignored so that it doesn't corrupt the group state or give the invalid member access to the group. 

We _could_ consider proposing a DID document as a third identity type, but that would need to make its way into the IETF spec and would be an even longer term project.

Particularly curious on your thoughts @jac18281828.

### WASM support

There has been an open PR for [WASM support](https://github.com/openmls/openmls/pull/1316) since March, so it seems possible, but I would imagine it would take some investment to make work.

### Eventual consistency

The [MLS architecture document](https://messaginglayersecurity.rocks/mls-architecture/draft-ietf-mls-architecture.html#name-eventually-consistent) offers two possible options:

> Upon receiving a Commit from the Delivery Service, clients can either:

> Pause sending new messages for a short amount of time to account for a reasonable degree of network latency and see if any other Commits are received for the same epoch. If multiple Commits are received, the clients can use a deterministic tie-breaking policy to decide which to accept, and then resume sending messages as normal.

> Accept the Commit immediately but keep a copy of the previous group state for a short period of time. If another Commit for a past epoch is received, clients use a deterministic tie-breaking policy to decide if they should continue using the Commit they originally accepted or revert and use the later one. Note that any copies of previous or forked group states must be deleted within a reasonable amount of time to ensure the protocol provides forward-secrecy.

The second option feels the most robust. The first is clearly the easiest. If we wanted to continue working on MLS, I wonder if we could implement the first option immediately to unblock work and pay an existing OpenMLS contributor to add capabilities for the second to the library. FWIW, Matrix is [working on the second option](https://gitlab.matrix.org/uhoreg/matrix-dmls/-/blob/develop/matrix-dmls/src/dmls_group.rs) in their DMLS branch as well.

neekolas:
I've implemented wallet identities via Basic Credentials in https://github.com/xmtp/libxmtp/pull/258/commits/8e7c4fde74fc0b1824534c40adc92458a6616d7b. Recipients can now validate that the message was coming from a given wallet address.

Think that's the end of this experiment for me, unless there is something specific we want to try out on top of this base.

jac18281828:
> ### Wallet ownership
> The MLS specification defines two types of identity that can be included in a credential: `Basic` and `X.509`. Basic identities are simply an opaque byte array. X.509 includes a [full certificate](https://en.wikipedia.org/wiki/X.509) used to associate the signing key with another public key by signature, and can include additional information.

X.509 seems intriguing because it is close to the model we are trying to build, to “bind an identity to a public key through a signature”.  However X.509 typically becomes trusted because it is hierarchical and a central trusted authority signs the highest level certificate.   I can think of some interesting hybrid models where we post a signing certificate to the users on-chain identity, but ultimately it will boil down to either self sign or trusted and may end up too ceremonial for our needs.

I like the idea of the Basic identity because we can still bind the identity to the public key by signing it with the users wallet, with their Ethereum public key (EPK).   We need to sort out the details of the contact bundle but I would see it containing the users DID and installation public key at a minimum.    

The on chain storage model is still a WIP and this is something that needs some design effort.   @richardhuaaa  There are three possible solutions.    The current solution in did:eth is not to store the key bundle at all.   These bundles may be looked up through the Ethereum logs.   This is a storage optimization and the process is a bit tedious for my tastes.   Considering the size of the contact bundle we would have two additional strategies that we could propose and contribute to the did:eth registry.  First we could directly store several small attributes such as an elliptic curve public key and a DID.    Alternatively, if we believe our use of the registry will eventually grow we could store a Merkle proof of the contact bundle.    I am starting to like this last option both due to its storage efficiency and its opacity to Eve.

So the protocol proceeds as follows:

1. Basic identity is signed by EPK (wallet signature required @richardhuaaa)
2. Identity bundle is forwarded on chain, probably via our subsidy gateway, where the signature is validated at the EVM level and authenticity assured 
3. Identity bundle is sent to counter party off chain via MLS
4. Counterparty receives the Basic identity and confirms proof of the identity bundle through the on chain registry

PR #260, opened 2023-10-04
V2 ecies library

neekolas:
## Summary

- Adds base functionality for ECIES encryption and decryption
- Adds library for signing the inner payload in ECIES encryption, and verifying signature at the time of decryption
- Wraps encrypted results in the `EciesMessage` proto

This is meant to be used for end-to-end encryption and decryption, including packaging everything in the appropriate protos. The results of `encrypt_message` are meant to be published directly to the network, and can be decrypted from the `envelope.message` directly.

Messages that go through this process are verified to have been signed by the public key passed in to `decrypt_message`.

## Usage

```rust
let some_bytes_i_want_to_send = vec![1,2,3,4];

let encrypted = encrypt_message(my_public_key, my_private_key, some_bytes_i_want_to_send);
send_to_network(encrypted)

let result_bytes = read_from_network()
let decrypted = decrypt_message(my_public_key, my_private_key, result_bytes);
// decrypted === some_bytes_i_want_to_send
```

## Next steps for me
- Add more tests
- Look at better error types
- Uniffi bindings for iOS/Android
- WASM bindings for JS
- Integrate into the JS keystore API


dev/docker/docker-compose.yml
```diff
@@ -5,13 +5,14 @@ services:
...
-      - --message-db-connection-string=postgres://postgres:xmtp@db:5432/postgres?sslmode=disable
-      - --message-db-reader-connection-string=postgres://postgres:xmtp@db:5432/postgres?sslmode=disable
+      - --store.enable
+      - --store.db-connection-string=postgres://postgres:xmtp@db:5432/postgres?sslmode=disable
+      - --store.reader-db-connection-string=postgres://postgres:xmtp@db:5432/postgres?sslmode=disable
       - --wait-for-db=30s
       # Disable authn until we have reliable support for generating auth tokens
       # - --api.authn.enable
     ports:
+      - 5555:5555
```
richardhuaaa, line 15: Are these changes related?

xmtp_ecies/src/signed_payload.rs
```diff
@@ -0,0 +1,170 @@
...
+};
+
+use crate::{
+    ecies::{decrypt_bytes, encrypt_bytes},
+    hash::sha3_256,
+};
+
+pub fn encrypt_message(
+    public_key: &[u8],
+    private_key: &[u8],
```
richardhuaaa, line 17: Would it be worth specifying that these must be secp256k1 keys?

richardhuaaa (approved):
LGTM

richardhuaaa:
Are you planning to make separate libraries from the existing Rust ones used in V2?

dev/docker/docker-compose.yml
neekolas: No, but they're necessary for the tests to pass. The v2 branch didn't get the fixes for the docker-compose arg updates, I guess.

xmtp_ecies/src/signed_payload.rs
neekolas: Great call out. Will add

neekolas:
> Are you planning to make separate libraries from the existing Rust ones used in V2?

Looking at that next. It's straightforward to add to the Swift library that is being used in the iOS SDK, but we need Android and JS support as well. We don't currently have anything building for Android or WASM from the V2 branch.

richardhuaaa:
@neekolas Not certain, but I think Android is using `xmtp_dh`

neekolas:
@richardhuaaa Ohhh. Interesting. I hadn't even looked in that folder. In that case, I'll add it to `bindings_swift` and `xmtp_dh` and then try and figure out something for WASM

richardhuaaa:
It's all a bit of a mess, if it works then great but if you end up wanting to make structural changes feel free! Currently the swift side needs grpc but the Android side does not, whereas Android needs DH but not grpc. The swift side uses swift-bridge and the Android side uses uniffi

PR #261, opened 2023-10-04
Set timeout and keep alive options

richardhuaaa:
https://github.com/xmtp/libxmtp/pull/255 effectively attempted to set a timeout on the server-side, which is the wrong approach to use for https://github.com/xmtp/xmtp-react-native/issues/129.

Due to the extremely slow cycle time for testing fixes I've set every available client-side timeout option here, as well as enabling keep-alive pings that might be able to detect a disconnected socket sooner. Thanks @dmccartney and @neekolas for the suggestions.


neekolas (approved):

dmccartney (commented):
LGTM 🤞

PR #203, opened 2023-08-16
v2 build and release actions

mkobetic:
This is some additional cleanup after splitting off the v2 stuff to the `v2` branch.

## Build Action:
* build v2 artifacts instead of v3 artifacts
* package up the `xmtp_dh` bits for android, produce `xmtp_dh.zip` artifact that can be unzipped over the target repo (xmtp/xmtp-android)
* package up the `bindings_swift` bits for ios, produces `xmtp-rust-swift.zip`, that can be unzipped over the target repo (xmtp/xmtp-rust-swift)
* build everything that is timeconsuming in the Build stage in parallel (this primarily moved some steps from the swift-v2 release action up to the build stage)

## Release Action
* create a release with the `xmtp_dh.zip` artifact
* update `xmtp/xmtp-rust-swift` repo with the build results and tag the new commit.

The build artifacts can be viewed in the Actions tab. The test release artifacts can be seen in
* https://github.com/xmtp/libxmtp/releases/tag/test-v2
* https://github.com/xmtp/xmtp-rust-swift/releases/tag/test-v2

## Questions

* Should we drop the artifact directories from the repo since they are now built automatically? 
* Should we delete bindings_ffi/ and bindings_wasm/ from this branch?
* Anything else that can be removed to reduce confusion?

PR #263, opened 2023-10-05
Restore old build process for v2

richardhuaaa:
The v2-specific automation updates from https://github.com/xmtp/libxmtp/pull/203 didn't seem to work correctly when @nplasterer and I tested it, it seemed like the v3 and v2 releases (which publish to different repos) get mixed together, and we're not sure if this is because of user error or a genuine issue.

I was going to go back and figure out what is wrong later, but as @neekolas needs this now, I'm reverting the build scripts to what they were in 7f78aded3da3e671742c167fb1055d57c0179d8e, which has been working for me. Later, we can revert this PR and debug the automation.


nplasterer (approved):

neekolas (approved):

neekolas:
Thanks for tackling this @richardhuaaa

neekolas:
@richardhuaaa think you need to roll back the Github workflows to match the old Makefile

PR #266, opened 2023-10-06
Initial SealedSender Implementation

neekolas:
## Summary

Working implementation of Signal's SealedSender algorithm.

The algorithm is [described in this blog post](https://signal.org/blog/sealed-sender/) as follows.
```
message_ciphertext             = signal_protocol_encrypt(message)
e_pub, e_priv                  = X25519.generateEphemeral()
e_chain, e_cipherKey, e_macKey = HKDF(salt="UnidentifiedDelivery" || recipientIdentityPublic || e_pub, ikm=ECDH(recipientIdentityPublic, e_priv), info="")
e_ciphertext                   = AES_CTR(key=e_cipherKey, input=senderIdentityPublic)
e_mac                          = Hmac256(key=e_macKey, input=e_ciphertext)

s_cipherKey, s_macKey = HKDF(salt=e_chain || e_ciphertext || e_mac, ikm=ECDH(recipientIdentityPublic, senderIdentityPrivate), info="")
s_ciphertext          = AES_CTR(key=s_cipherKey, input=sender_certificate || message_ciphertext)
s_mac                 = Hmac256(key=s_macKey, input=s_ciphertext)

message_to_send = s_ciphertext || s_mac
```

To implement SealedSender I tried to marry the algorithm as described with [the implementation in `libsignal`](https://github.com/signalapp/libsignal/blob/d1f9dff273e6da059af699c6afe860fb93406032/rust/protocol/src/sealed_sender.rs#L671) to get to a workable version. 

## Notes

- In the algorithm description above, we generate an `e_mac` via Hmac256 and use it in the salt of the static cipherkey. The implementation in `libsignal` doesn't do that at all, as far as I can tell. So I left it out. Instead, the salt for the HKDF is based on just the chain key and the ciphertext.
- This is more complicated than we likely need for our purposes. There is a part of me that would rather just replace this with ECIES using x25519 keys, which would have fewer point multiplications and HKDFs. But for now it feels the safest to use a very well-vetted algorithm, and maybe that decision becomes a fun getting-started project for a future applied cryptographer hire.
- `x25519_dalek` has a `EphemeralSecret` type that feels more appropriate for use in these ephemeral keypairs. I just need to figure out how to create one from bytes instead of randomly being generated, since in some cases it actually comes from a long-term identity key.
- Everything is based on raw `x25519-dalek` keys instead of Vmac types. That makes the implementation workable for both Vmac and MLS. To make it work with Vmac, we are going to have to add some converters to our fork, since the underlying `x25519-dalek` keys are not accessible outside of the Vmac crate.

## ToDo
- Write tests for some of the unhappy paths.
- Improve error handling
- Deal with serialization of the messages


xmtp/Cargo.toml
```diff
@@ -36,8 +38,12 @@ ethers-core = "2.0.4"
 prost = { version = "0.11", features = ["prost-derive"] }
 futures = "0.3.28"
 base64 = "0.21.1"
-tokio = "1.28.1"
+tokio = { version = "1.32.0", features = ["full"] }
```
neekolas, line 41: One of the other deps seems to have upgraded us to a version of Tokio that needs a feature to include the `tokio::test` macro, so I added that.

xmtp/Cargo.toml
```diff
@@ -36,8 +38,12 @@ ethers-core = "2.0.4"
...
 anyhow = "1.0.71"
+x25519-dalek = "2.0.0"
```
neekolas, line 43: Added a bunch of the underlying cryptography crates that we are already using in Vmac.

xmtp/src/cryptography.rs
```diff
@@ -0,0 +1,80 @@
...
+    let zero_nonce = [0u8; 16];
+    let mut cipher = ctr::Ctr32BE::<Aes256>::new(key[..].into(), zero_nonce[..].into());
+
+    let mut ctext = ptext.to_vec();
+    cipher.apply_keystream(&mut ctext);
+    Ok(ctext)
+}
+
+fn aes_256_ctr_decrypt(ctext: &[u8], key: &[u8]) -> Result<Vec<u8>, DecryptionError> {
+    aes_256_ctr_encrypt(ctext, key).map_err(|e| match e {
```
neekolas, line 38: This is what `libsignal` does. I can't quite get my head around why this works...but it does.

neekolas:
I'm going to close this PR and replace with a simpler [Noise Pattern N](https://noiseexplorer.com/patterns/N/) implementation generated from Noise Explorer.

It's ultimately going to be more work to take this and re-write it in a way where there is 0 chance it'll inherit the [GPL terms](https://github.com/signalapp/libsignal/blob/main/LICENSE) of libsignal. And we'll end up with a system that is more complicated, for benefits that are unclear in our specific context.

PR #267, opened 2023-10-06
Lower the timeout values for the rust connection

nplasterer:
This PR https://github.com/xmtp/libxmtp/pull/261 greatly improved the disconnection issue https://github.com/xmtp/xmtp-react-native/issues/129 in react native. But hovering around 10 seconds to reconnect was still not ideal. Lets drop this timeout down even lower for the time being to improve this for current integrators while we look to find a better improvement here. The only draw back is that this may cause a lot more unnecessary network pings or maybe it won't be that noticeable.

Since I'm not super familiar with Rust I looked to other repos with hyper implementations to see what timeouts they used for these fields
pool_idle_timeout: https://github.com/risingwavelabs/risingwave/blob/a46292eedc9a6d1ad10b5fb3e7d89495edfccad2/src/connector/src/common.rs#L423-L431
Otherwise just halved them. The common theme was 5 seconds for all the timeouts so maybe I dropped some of these down too low. Open to thoughts.

### DIFF
--- a/xmtp_networking/src/grpc_api_helper.rs
+++ b/xmtp_networking/src/grpc_api_helper.rs
@@ -46,7 +46,7 @@ fn get_tls_connector() -> HttpsConnector<tower::timeout::Timeout<HttpConnector>>
                 .enable_http2()
                 .wrap_connector(s)
         })
-        .timeout(Duration::from_secs(2))
+        .timeout(Duration::from_secs(5))
         .service(http)
 }
 
@@ -76,7 +76,7 @@ impl Client {
             let tls_conn = hyper::Client::builder()
                 .pool_idle_timeout(Duration::from_secs(5))
                 .http2_keep_alive_interval(Duration::from_secs(3))
-                .http2_keep_alive_timeout(Duration::from_secs(2))
+                .http2_keep_alive_timeout(Duration::from_secs(5))
                 .build(connector);
 
             let uri =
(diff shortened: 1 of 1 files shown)


xmtp_networking/src/grpc_api_helper.rs
```diff
@@ -74,9 +74,9 @@ impl Client {
             let connector = get_tls_connector();
 
             let tls_conn = hyper::Client::builder()
-                .pool_idle_timeout(Duration::from_secs(30))
-                .http2_keep_alive_interval(Duration::from_secs(10))
-                .http2_keep_alive_timeout(Duration::from_secs(5))
+                .pool_idle_timeout(Duration::from_secs(5))
+                .http2_keep_alive_interval(Duration::from_secs(3))
+                .http2_keep_alive_timeout(Duration::from_secs(2))
```
neekolas, line 79: I wonder if we want to keep the timeout higher here, just so that it doesn't incorrectly abort connections if you're on a high latency device (bad mobile connection). I don't think this number has any bearing on the speed we detect disconnects (pretty sure the keepalive will fail immediately if you are offline).

xmtp_networking/src/grpc_api_helper.rs
```diff
@@ -74,9 +74,9 @@ impl Client {
             let connector = get_tls_connector();
 
             let tls_conn = hyper::Client::builder()
-                .pool_idle_timeout(Duration::from_secs(30))
-                .http2_keep_alive_interval(Duration::from_secs(10))
-                .http2_keep_alive_timeout(Duration::from_secs(5))
+                .pool_idle_timeout(Duration::from_secs(5))
```
neekolas, line 77: Are we sure this value is related to the bug in RN? If we kill idle connections faster, that could result in a lot more connections to our server and worse performance. But I don't really know.

xmtp_networking/src/grpc_api_helper.rs
nplasterer: Ya I wasn't really sure which one of these numbers actually did the trick from the previous PR so totally cool to keep some of these higher.

richardhuaaa: Yeah so the interval is how often a grpc ping is sent, the timeout is the delay after which the connection is killed. More info here: https://grpc.io/docs/guides/keepalive/#common-situations-where-keepalives-can-be-useful. If the delay while testing was always 5s or less, then it seems likely that it fails immediately on the next ping, like neekolas says.

richardhuaaa (approved):
Pre-approving assuming the feedback is addressed! Also edited my comment - when I said 5s I meant 10s

PR #222, opened 2023-09-11
feat: implement grpc_gateway api client

dmccartney:
This begins the implementation of the wasm-friendly grpc-gateway API client.
It continues where #217 left off.

- [x] split `xmtp_networking` into `xmtp_api_grpc` and `xmtp_api_grpc_gateway`
- [x] setup CI for `xmtp_api_grpc_gateway`
- [x] implement grpc-gateway `Query`
- [x] implement grpc-gateway `Publish`
- [x] implement grpc-gateway `BatchQuery`

Future PR
- [ ] implement grpc-gateway `Subscribe`
- [ ] `xmtp` surgery to make storage layer swappable
- [ ] basic sample app (to feel concurrency/perf beyond automated tests)

PR #268, opened 2023-10-11
fix: add .subscribe / async support to wasm client

dmccartney:
This picks up where #222 left off implementing `Subscribe` in the wasm/grpc-gateway client.

### DIFF
--- a/xmtp_api_grpc/src/grpc_api_helper.rs
+++ b/xmtp_api_grpc/src/grpc_api_helper.rs
@@ -245,12 +245,13 @@ impl Subscription {
     }
 }
 
-impl XmtpApiSubscription for Subscription {
+#[async_trait]
+impl<'a> XmtpApiSubscription for Subscription {
     fn is_closed(&self) -> bool {
         self.closed.load(Ordering::SeqCst)
     }
 
-    fn get_messages(&self) -> Vec<Envelope> {
+    async fn get_messages(&mut self) -> Vec<Envelope> {
         let mut pending = self.pending.lock().unwrap();
         let items = pending.drain(..).collect::<Vec<Envelope>>();
         items
--- a/xmtp_api_grpc/src/grpc_api_helper.rs
+++ b/xmtp_api_grpc/src/grpc_api_helper.rs
@@ -1,15 +1,14 @@
+use futures::stream::BoxStream;
+use futures_util::StreamExt;
 use http_body::combinators::UnsyncBoxBody;
 use hyper::{client::HttpConnector, Uri};
 use hyper_rustls::HttpsConnector;
 use std::str::FromStr;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Mutex}; // TODO switch to async mutexes
-use tokio::sync::oneshot;
 use tokio_rustls::rustls::{ClientConfig, OwnedTrustAnchor, RootCertStore};
 use tonic::async_trait;
 use tonic::Status;
-use tonic::{metadata::MetadataValue, transport::Channel, Request, Streaming};
-use xmtp_proto::api_client::{Error, ErrorKind, XmtpApiClient, XmtpApiSubscription};
+use tonic::{metadata::MetadataValue, transport::Channel, Request};
+use xmtp_proto::api_client::{Error, ErrorKind, XmtpApiClient};
 use xmtp_proto::xmtp::message_api::v1::{
     message_api_client::MessageApiClient, BatchQueryRequest, BatchQueryResponse, Envelope,
     PublishRequest, PublishResponse, QueryRequest, QueryResponse, SubscribeRequest,
@@ -18,7 +17,7 @@ use xmtp_proto::xmtp::message_api::v1::{
 fn tls_config() -> ClientConfig {
     let mut roots = RootCertStore::empty();
     // Need to convert into OwnedTrustAnchor
-    roots.add_server_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.0.iter().map(|ta| {
+    roots.add_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.0.iter().map(|ta| {
         OwnedTrustAnchor::from_subject_spki_name_constraints(
             ta.subject,
             ta.spki,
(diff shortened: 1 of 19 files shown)


xmtp_api_grpc_gateway/src/lib.rs
```diff
@@ -36,67 +47,153 @@ impl XmtpApiClient for XmtpGrpcGatewayClient {
...
     fn is_closed(&self) -> bool {
-        true
+        return self.stream.is_none();
     }
 
-    fn get_messages(&self) -> Vec<Envelope> {
-        vec![]
+    // HACK: this consumes from the stream whatever is already ready.
+    // TODO: implement a JS-friendly promise/future interface instead
+    async fn get_messages(&mut self) -> Vec<Envelope> {
```
dmccartney, line 159: After much-mashing to get this working, I think we should scrap this and change the abstraction for `XmtpApiSubscription` across the library.  I suspect it will be cleaner either as a proper `Stream` (i.e. ✂️  the `get_messages()` stuff and use `.collect()` and other stream tools) and/or by designing outside-in to use JS promises (qua rust `Futures`) and callbacks.

xmtp_api_grpc_gateway/tests/wasm_test.rs
```diff
@@ -129,3 +135,64 @@ pub async fn test_batch_query_publish_batch_query() {
...
+    assert_eq!(false, sub.is_closed());
+
+    // HACK: this tugs at the stream w/ poll_next() to initiate the HTTP req.
+    //       we need to do this _before_ we publish, otherwise the
+    //       publish will happen before the subscription is ready.
+    // TODO: implement a JS-friendly promise/future interface instead
+    let mut cx = Context::from_waker(noop_waker_ref());
+    let stream = sub.stream.as_mut().unwrap();
+    let init = stream.as_mut().poll_next(&mut cx);
+    assert_eq!(true, init.is_pending());
```
dmccartney, line 166: :(

xmtp_api_grpc_gateway/src/lib.rs
neekolas: I have no objections to this

xmtp_proto/src/api_client.rs
```diff
@@ -72,11 +73,9 @@ impl StdError for Error {
...
 }
 
-pub trait XmtpApiSubscription {
-    fn is_closed(&self) -> bool;
-    fn get_messages(&self) -> Vec<Envelope>;
-    fn close_stream(&mut self);
-}
+pub trait XmtpApiSubscription {}
+
+impl<T: Stream<Item = Envelope>> XmtpApiSubscription for T {}
```
dmccartney, line 78: Beyond the WASM implementation, this is conceptually the big change to `subscribe()` across both `grpc` and `grpc-gateway` clients. Now the result is just an ordinary `Stream` of `Envelope`s.

dmccartney, line 78: Beyond the WASM implementation, this is conceptually the big change to `subscribe()` across both `grpc` and `grpc-gateway` clients. Now the result is just an ordinary `Stream` of `Envelope`s.

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -57,12 +56,12 @@ pub enum InnerApiClient {
     ),
 }
 
-pub struct Client {
+pub struct XmtpGrpcClient {
```
dmccartney, line 59: I renamed to be a little easier to understand alongside the other `XmtpApiClient`:
`xmtp_api_grpc` -> `XmtpGrpcClient`
`xmtp_api_grpc_gateway` -> `XmtpGrpcGatewayClient`

xmtp_api_grpc_gateway/src/lib.rs
```diff
@@ -36,67 +42,89 @@ impl XmtpApiClient for XmtpGrpcGatewayClient {
...
-        &self,
-        _request: SubscribeRequest,
-    ) -> Result<XmtpGrpcGatewaySubscription, Error> {
-        // TODO
-        Err(Error::new(ErrorKind::SubscribeError))
+    async fn subscribe(&self, request: SubscribeRequest) -> Result<Self::Subscription, Error> {
+        // grpc-gateway streams newline-delimited JSON bodies
+        let response = self
+            .http
+            .post(&format!("{}/message/v1/subscribe", self.url))
```
neekolas, line 63: This doesn't look like it needs to be a reference, does it?

xmtp_proto/src/api_client.rs
richardhuaaa: Do we need the `XmtpApiSubscription` trait at all, seeing as it's empty? Can we just mandate that `subscribe()` returns a `Stream` instead of an `XmtpApiSubscription`? I think that lets us call the `Stream` methods on consumers even if the concrete type for `XmtpApiClient` is not known.

xmtp_api_grpc_gateway/tests/wasm_test.rs
```diff
@@ -129,3 +135,53 @@ pub async fn test_batch_query_publish_batch_query() {
...
+                content_topic: topic.to_string(),
+                message: vec![1, 2, 3, 4],
+                timestamp_ns: 1234,
+            }],
+        },
+    )
+    .await
+    .expect("published");
+
+    if let Ok(env) = rx.await {
```
richardhuaaa, line 181: Nooby question, but what is the point of the `channel` here? If we are `awaiting` anyway, what difference does it make if we do `rx.await` rather than `stream.next().await` directly?

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -162,9 +161,12 @@ impl XmtpApiClient for Client {
...
                 .map_err(|e| Error::new(ErrorKind::SubscribeError).with(e))?
                 .into_inner(),
-        };
-
-        Ok(Subscription::start(stream).await)
+        }
+        // Discard any messages that we can't parse.
+        // TODO: consider surfacing these in a log somewhere
+        .filter_map(|r| async move { r.ok() })
+        .boxed();
```
richardhuaaa, line 168: Another nooby question - why do we need `boxed()` here? Can we use `tonic::Streaming<Envelope>` instead of `BoxStream<'static, Envelope>`?

richardhuaaa:
This looks much better, thanks for cleaning it up! Just left some nooby questions about Streams

PR #271, opened 2023-10-19
Add Rust-Toolchain, refactor generics, typos/cleanup

insipx:
Fixed a TODO that required `Default` on `XmtpApiClient`. 

I also generally did some refactoring of type parameters, although this is subjective, so I did not do it everywhere (mostly in the builder and ClientBuilder). I wanted to get thoughts on it. IMO, it's much easier to understand generics in code + documentation like `Widget<Frame, Foo, Crypto>` rather than the terse version `Widget<Q, N, O, X, P>`, especially as usage of generics increases. This does have the downside of occasionally breaking the line if Generic names are very long, or if there are lots and lots of generics, EX:

```
Widget<
      FrameForTheGpuWindow,
      CryptographicPrimitivesForY,
      GenericB,
      GenericX,
>
```

Although I've found this to be uncommon, and the benefit of readable generics to be really nice

Generally just some things I noticed, can bring over to #272 once merged if desired

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,5 @@
+[toolchain]
+channel = "stable"
+components = [ "rustc", "cargo", "clippy", "rustfmt", "rust-analyzer" ]
+targets = [ "wasm32-unknown-unknown" ]
+profile = "default"
--- a/xmtp/src/account.rs
+++ b/xmtp/src/account.rs
@@ -36,13 +36,13 @@ pub enum AccountError {
     Unknown,
 }
 
+/// Holds an account and adds some serialization methods on top
 pub struct VmacAccount {
     account: OlmAccount,
 }
 
-// Struct that holds an account and adds some serialization methods on top
 impl VmacAccount {
-    // Create a new instance
+    /// Create a new instance
     pub fn new(account: OlmAccount) -> Self {
         Self { account }
     }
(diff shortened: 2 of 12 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "stable"
```
insipx, line 2: We can also pin to a specific version, not sure what the preference is

neekolas: `stable` works for me. We have enough crazy build-time stuff happening in the bindings, last thing we need is to be doing that on a nightly build

neekolas:
Thank you for tackling this @insipx. These look like great quality of life improvements. Things like readable generics will definitely be helpful as we onboard more devs to this repo.

It's some really unfortunate timing on this one. A few days ago, it wasn't clear at all that we were going to be sunsetting this crate and copying the code into a new one. But here we are. Unfortunately, you're going to duplicate these changes on the `mls` branch. Apologies for not announcing the move to a new crate earlier to save you some duplicate work.

richardhuaaa (approved):
Thank you for doing this cleanup!! I anticipate us also moving more of your changes across over time as the new crate becomes fleshed out, so I think it will be useful either way.

insipx:
doing changes to MLS in new PR to keep things small

PR #272, opened 2023-10-19
Add skeleton of xmtp_mls crate

richardhuaaa:
This creates an `xmtp_mls` crate, and copies over some of the common things we'll need from the `xmtp` crate:
- Networking
- Database
- Client
- ClientBuilder
- Associations

There is no MLS-specific logic in this PR. I've made the `Account` object an empty struct for now, `Client` does no business logic other than holding the API client and DB, and there are no models in the database. The only changes are stripping out vmac-specific logic and wrangling to make the whole thing compile.

Happy to port over #268 once it lands, if it's a hassle @dmccartney


neekolas (approved):
Looking very tidy

xmtp_mls/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+async-trait = "0.1.68"
+diesel = { version = "2.0.4", features = ["sqlite", "r2d2", "returning_clauses_for_sqlite_3_35"] }
+diesel_migrations = { version = "2.0.0", features = ["sqlite"] }
+ethers = "2.0.4"
+ethers-core = "2.0.4"
+futures = "0.3.28"
+hex = "0.4.3"
+libsqlite3-sys = { version = "0.26.0", optional = true}
+log = "0.4.17"
+openmls = { path = "../../openmls/openmls", features = ["test-utils"] }
```
neekolas: We should probably make a fork and just point at that everywhere

xmtp_mls/src/account.rs
```diff
@@ -0,0 +1,18 @@
...
+#[derive(Debug, Error)]
+pub enum AccountError {
+    #[error("generating new account")]
+    BadGeneration(#[from] SignatureError),
+    #[error("bad association")]
+    BadAssocation(#[from] AssociationError),
+    #[error("mutex poisoned error")]
+    MutexPoisoned,
+    #[error("unknown error")]
+    Unknown,
```
insipx, line 15: Where does this error occur? I can't find anywhere it is referenced

xmtp_mls/src/association.rs
```diff
@@ -0,0 +1,227 @@
...
+    TextMismatch,
+    #[error(
+        "Address mismatch in Association: Provided:{provided_addr:?} != signed:{signing_addr:?}"
+    )]
+    AddressMismatch {
+        provided_addr: Address,
+        signing_addr: Address,
+    },
+    #[error("unknown association error")]
+    Unknown,
```
insipx, line 24: Same here

I think we should avoid adding errors that won't trigger

insipx (approved):
nits w.r.t errors. I have some other possible changes in #271, so I won't hold this one up

PR #278, opened 2023-10-23
Create validation service

neekolas:
## Summary

- Creates a new GRPC server using Tonic
- Handlers for validating Key Packages and Group Messages, and extracting identities and wallet addresses from credentials using the code in `xmtp_mls`
- Dockerfile for building the server
- Generates new Protobufs for both the MLS API and the RPC interface between our nodes <> validation service
- Adds an API Wrapper for the new MLS nodes


xmtp_proto/buf.gen.yaml
```diff
@@ -13,11 +13,16 @@ plugins:
   - plugin: buf.build/community/neoeinstein-tonic
     out: src/gen
     opt:
-      - no_server=true
+      - no_server=false
```
neekolas, line 16: We now need to generate servers everywhere. We may want to add a second feature to the server attributes so that we can toggle them off as needed.

xmtp_mls/src/client.rs
```diff
@@ -40,21 +42,24 @@ impl From<&str> for ClientError {
...
+{
     pub fn new(
         api_client: ApiClient,
         network: Network,
         identity: Identity,
         store: EncryptedMessageStore,
     ) -> Self {
         Self {
-            api_client,
+            api_client: ApiClientWrapper::new(api_client),
```
neekolas, line 62: I wrap the `ApiClient` into a prettier interface that abstracts away the Protobuf nonsense. Open to better names for  it than `ApiClientWrapper`.

xmtp_mls/Cargo.toml
```diff
@@ -32,3 +32,4 @@ anyhow = "1.0.71"
 
 [dev-dependencies]
 tempfile = "3.5.0"
+xmtp_api_grpc = { path = "../xmtp_api_grpc" }
```
neekolas, line 35: I'm proposing for this MLS work we ditch the `MockApiClient` and just use a local server for all testing.

dev/gen_protos.sh
```diff
@@ -5,7 +5,7 @@ if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
         exit 1
     fi
 fi
-if ! buf generate https://github.com/xmtp/proto.git#branch=xmtpv3,subdir=proto; then
+if ! buf generate https://github.com/xmtp/proto.git#branch=nmolnar/mls-verification-service,subdir=proto; then
```
neekolas, line 8: Will switch this back to `main` once we merge https://github.com/xmtp/proto/pull/107

mls_validation_service/src/handlers.rs
```diff
@@ -0,0 +1,256 @@
...
+    ) -> Result<Response<ValidateGroupMessagesResponse>, Status> {
+        let out: Vec<ValidateGroupMessageValidationResponse> = request
+            .into_inner()
+            .group_messages
+            .into_iter()
+            .map(|message| {
+                match validate_group_message(message.group_message_bytes_tls_serialized) {
+                    Ok(res) => ValidateGroupMessageValidationResponse {
+                        group_id: res.group_id,
+                        epoch: res.epoch,
```
richardhuaaa, line 70: This would be used for proposals/commits too right? If so, is this the epoch going forward, or the epoch the message was expected to be sent on - don't we need both?

And is the golang server going to validate that the current epoch matches the expected epoch outside of this one?

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -57,8 +63,18 @@ pub enum InnerApiClient {
     ),
 }
 
+pub enum InnerMlsClient {
```
richardhuaaa, line 66: Confirming I understood here - the MlsClient is NOT the validation service, it is a client for the golang gRPC server?

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -73,10 +89,12 @@ impl Client {
             let uri =
                 Uri::from_str(&host).map_err(|e| Error::new(ErrorKind::SetupError).with(e))?;
 
-            let tls_client = MessageApiClient::with_origin(tls_conn, uri);
+            let tls_client = MessageApiClient::with_origin(tls_conn.clone(), uri.clone());
+            let mls_client = ProtoMlsApiClient::with_origin(tls_conn, uri);
 
             Ok(Self {
                 client: InnerApiClient::Tls(tls_client),
+                mls_client: InnerMlsClient::Tls(mls_client),
```
richardhuaaa, line 97: Is there any consumer that would need both of these clients? Wondering if we should just return one or the other based on a flag? Or do we rely on endpoints from both of these clients?

xmtp_mls/Cargo.toml
richardhuaaa: It means we always have to run a server in order to run tests, which is fine by me. The reason is because the server logic is a little more complex and it would be duplicative to mock out?

richardhuaaa (approved):
Summarizing what I understood:

1. The validation service is a rust server that is consumed by the golang gRPC server, NOT the libxmtp client. It is included in libxmtp for now because it shares the Rust validation logic used by libxmtp. In future when libxmtp is a published crate the validation server could be moved to a separate repo.

2. Separately this PR adds a client for the golang gRPC server entry point that libxmtp uses.

FYI @snormore

mls_validation_service/src/handlers.rs
neekolas: This will get used for any group message (application, proposal, commit).

We don't do any validation on the epoch. I probably should just remove the field from the response. It predates our convo with @Bren2010 around relying on total ordering rather than epoch validation.

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: That's right.

`xmtp_mls` talks to `xmtp-node-go` which talks to `mls_validation_service`

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: `xmtp_mls` is going to need both. My intention is to rely on all of our existing `Query`, `BatchQuery` and `Stream` methods for reading messages from topics, since we don't need to do anything special on the read side.

xmtp_mls/Cargo.toml
neekolas: That's basically it. The mock API client was already incomplete (didn't really support pagination). Would be even harder to properly mock out the new APIs. This way, we get real validation of bad messages etc.

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: They are implemented as separate GRPC services on the node with separate protobuf services, so they need separate underlying clients (although the clients can share a single connection).

PR #279, opened 2023-10-23
Rename account->identity

richardhuaaa:
Separating out this quick PR to make later PR's easier to review.

The word 'account' is overloaded - we previously decided that this would refer to the *blockchain account* or other user-level identity, rather than an individual installation under that account. Renamed to 'identity' here, which is what the openMLS CLI calls this, and is consistent with the concept of a *cryptographic identity* commonly referred to in docs.


neekolas (approved):
Great change

PR #280, opened 2023-10-23
Create MLS identity

richardhuaaa:
This creates an MLS credential, signing key pair, and last resort key package.

We treat the public key of the signature key pair as the installation's identity, and the address of the blockchain account as the user's identity. The MLS credential is a serialized `Eip191Association` proto, which is a wallet signature for the signature public key.

For now we store the keys in an in-memory key store.

Still to be done:
- Unit tests for credential generation - can add this once credential validation is done
- (I will work on this next) Use persistent sqlite store instead of in-memory store. We can use a `key_store` table with `key` and `value` columns for the OpenMLS storage provider. We should also define our own custom tables for storing things like the identity that we represent.
- Upload credential and key packages to network
- Better protos (currently reusing Eip191Association as the credential)

### DIFF
--- a/xmtp_mls/src/configuration.rs
+++ b/xmtp_mls/src/configuration.rs
@@ -1,4 +1,5 @@
 use openmls_traits::types::Ciphersuite;
 
 // TODO confirm ciphersuite choice
-pub const CIPHERSUITE: Ciphersuite = Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
+pub const CIPHERSUITE: Ciphersuite =
+    Ciphersuite::MLS_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519;
--- a/xmtp_mls/src/configuration.rs
+++ b/xmtp_mls/src/configuration.rs
@@ -1,0 +1,4 @@
+use openmls_traits::types::Ciphersuite;
+
+// TODO confirm ciphersuite choice
+pub const CIPHERSUITE: Ciphersuite = Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
--- a/xmtp_mls/src/identity.rs
+++ b/xmtp_mls/src/identity.rs
@@ -4,11 +4,17 @@ use openmls::{
     versions::ProtocolVersion,
 };
 use openmls_basic_credential::SignatureKeyPair;
-use openmls_traits::{types::Ciphersuite, OpenMlsProvider};
+use openmls_traits::{
+    types::{Ciphersuite, CryptoError},
+    OpenMlsProvider,
+};
 use thiserror::Error;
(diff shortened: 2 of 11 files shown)


xmtp_mls/src/configuration.rs
```diff
@@ -0,0 +1,4 @@
+use openmls_traits::types::Ciphersuite;
+
+// TODO confirm ciphersuite choice
+pub const CIPHERSUITE: Ciphersuite = Ciphersuite::MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519;
```
neekolas, line 4: I wonder if we actually want to just start with `MLS_128_DHKEMX25519_CHACHA20POLY1305_SHA256_Ed25519` given both are already implemented and we can choose whichever one we want.

richardhuaaa: Happy to do this for now, we should still validate this choice though

xmtp_mls/src/identity.rs
```diff
@@ -1,14 +1,108 @@
...
+        // Generate association
+        let assoc_text = AssociationText::Static {
+            blockchain_address: owner.get_address(),
+            installation_public_key: signature_keys.to_public_vec(),
+        };
+        let signature = owner.sign(&assoc_text.text())?;
+        let association = Eip191Association::new(signature_keys.public(), assoc_text, signature)?;
+
+        // Create AccountLinkedKey proto
+        // TODO: Rename/revise protos. Eip191Associations are relatively big, we should decide if we want smaller credentials or not
```
neekolas, line 92: What fields would we omit from Eip191Association? Don't we need pretty much everything to reconstruct the verification text on the other side.

neekolas (approved):
I can take "unit tests for credential generation". I need to get the generation and validation in sync with https://github.com/xmtp/libxmtp/pull/278 anyways

xmtp_mls/src/identity.rs
```diff
@@ -1,14 +1,108 @@
...
+        let installation_key_proto: ProtoWrapper<VmacUnsignedPublicKey> =
+            signature_keys.to_public_vec().into();
+        let account_linked_key = VmacAccountLinkedKey {
```
neekolas, line 95: At least right now, the Credential only ever is accessed inside a KeyPackage. The KeyPackage already contains the pub key of the signing keypair. So, it might be superfluous to include the `installation_key` in the credential a second time. We would still need to look up the signing key from the KeyPackage or leaf node to properly validate a credential, otherwise you could be signing some random public key unrelated to your account or borrowing someone else's credential.

If we want to use the credentials outside of the context of leaf nodes/key packages, we would need to include the signing public key. But I haven't run into those use-cases yet.

richardhuaaa: Great call-out! I'll also double-check tomorrow, but if that's the case we can throw out the `VmacAccountLinkedKey` proto altogether

PR #283, opened 2023-10-24
Identity persistence + OpenMLS key store

richardhuaaa:
- Implements persistence for identities
- Implements key store needed by openMLS - this is simply a table called `key_store` with `key` and `value` columns


xmtp/Cargo.toml
```diff
@@ -36,7 +36,7 @@ ethers-core = "2.0.4"
 prost = { version = "0.11", features = ["prost-derive"] }
 futures = "0.3.28"
 base64 = "0.21.1"
-tokio = "1.28.1"
```
richardhuaaa, line 39: These changes are needed to make the tests pass when running `cargo test` from inside the xmtp crate directly

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -0,0 +1,118 @@
...
+        entry.store(&mut self.store.conn()?)?;
+        Ok(())
+    }
+
+    /// Read and return a value stored for ID `k` that implements the
+    /// [`FromKeyStoreValue`] trait for deserialization.
+    ///
+    /// Returns [`None`] if no value is stored for `k` or reading fails.
+    fn read<V: MlsEntity>(&self, k: &[u8]) -> Option<V> {
+        let conn_result = self.store.conn();
```
neekolas, line 51: This is going to get a little tricky when we need to interact with OpenMLS inside a transaction.

We don't control the interface for `OpenMlsKeyStore`, so we can't add an argument for a connection. I guess we could generate new `OpenMlsKeyStore` instances that hold a `conn` as a property if we needed a version that was transaction-friendly.

richardhuaaa: Good point - do you have an example of where we might need to do that? That might help with thinking through this. I've been hoping that we can treat OpenMLS data as separate and opaque to us - they handle their stuff, we handle ours, but aware that this abstraction can break

richardhuaaa: I guess we can wait and see, if it's not immediately obvious

xmtp_mls/src/builder.rs
```diff
@@ -107,21 +101,36 @@ where
...
+    }
+
+    fn initialize_identity(
+        self,
+        store: &EncryptedMessageStore,
+        provider: &XmtpOpenMlsProvider,
+    ) -> Result<Identity, ClientBuilderError> {
+        let conn = &mut store.conn()?;
+        let identity_option: Option<Identity> = conn.fetch(())?.map(|i: StoredIdentity| i.into());
+        match self.identity_strategy {
```
insipx, line 118: could possibly move this into an `impl` on `IdentityStrategy`, if it's used more often

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -163,8 +167,155 @@ fn warn_length<T>(list: &Vec<T>, str_id: &str, max_length: usize) {
...
+
+            StoredIdentity::new("dummy_address".to_string(), rand_vec(), rand_vec())
+                .store(&mut store.conn().unwrap())
+                .unwrap();
+        } // Drop it
+
+        enc_key[3] = 145; // Alter the enc_key
+        let res = EncryptedMessageStore::new(StorageOption::Persistent(db_path.clone()), enc_key);
+        // Ensure it fails
+        match res.err() {
```
insipx, line 294: could use

```suggestion
        assert!(matches!(res.err(), Some(StorageError::DbInitError(_)), "Expected DbInitError");
```

xmtp_mls/src/storage/errors.rs
```diff
@@ -14,6 +14,8 @@ pub enum StorageError {
     Store(String),
     #[error("serialization error")]
     SerializationError,
+    #[error("deserialization error")]
+    DeserializationError,
```
insipx, line 18: This is more something that doesn't _really_ matter, but imo I prefer errors like 
`Error::Deserialization`

Rather then including the redundant 'Error' on the end, since the enum will  be used in-code, and saves some line-space.

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -163,8 +165,155 @@ fn warn_length<T>(list: &Vec<T>, str_id: &str, max_length: usize) {
...
+            let account_address = "address";
+            StoredIdentity::new(account_address.to_string(), rand_vec(), rand_vec())
+                .store(conn)
+                .unwrap();
+
+            let fetched_identity: StoredIdentity = conn.fetch(()).unwrap().unwrap();
+            assert_eq!(fetched_identity.account_address, account_address);
+        }
+
+        fs::remove_file(db_path).unwrap();
```
neekolas, line 268: Not a blocker, but we should probably have some sort of `defer` like struct that lets us run cleanup tasks like this even if the code above it errors, which should run any time the struct gets dropped from memory. We could do it with something like this:

```rust
struct FileGuard {
    path: String,
}

impl FileGuard {
    fn new(path: String) -> Self {
        FileGuard { path }
    }
}

impl Drop for FileGuard {
    fn drop(&mut self) {
        if let Err(e) = fs::remove_file(&self.path) {
            eprintln!("Failed to remove {}: {}", self.path, e);
        }
    }
}
```

xmtp_mls/src/storage/sql_key_store.rs
neekolas: The one I'm thinking about is processing an incoming message from the group topic. Ideally we would wrap each message processing task into a transaction, and roll the whole thing back if something went wrong. But some of the messages are going to need writes to the Keystore as group state gets updated.

neekolas (approved):

xmtp_mls/src/storage/encrypted_store/mod.rs
neekolas: Or we could use a tempfile. I just know that I have a bunch of random `.db3` files sitting around in my local copy of this repo

insipx: +1 for tempfile

PR #284, opened 2023-10-25
Add README

neekolas:
I took a first pass at the database schema and state machine for MLS.

We can rely on the total ordering of the Query endpoint for our network to get to a consistent state on every read of the group topic. All messages are read in ascending order, and any message sent for a previous epoch must be discarded. This flow introduces a new concept "group_intents" to handle cases where you might send a message in one epoch, and then upon subsequently reading the network need to re-send in another epoch.

Intents are not bound to any epoch and can therefore be safely retried in cases where conflicts occur.

Some actions can either be done optimistically (send the message then refresh the group state) or pessimistically (refresh group state, send message, refresh again). I have opted to handle regular message sending optimistically and anything that modifies the ratchet tree (add/remove members and key updates) pessimistically. Conflicts are possible either way, but are less likely with the pessimistic approach.


neekolas:
One thing I'm wondering is whether this is more conservative than necessary.

For sending application messages, which doesn't advance the epoch, we should very rarely need to re-send them. If we set `max_past_epochs` to something like 3 or 5, a lot of things would have had to happen in the group since your last sync to render a message unreadable. Having to read the group state on every message send feels very heavy-handed, so maybe we can relax that and make some publishes more "fire and forget".

Syncing the group state _before_ and _after_ adding/removing new members is also a lot. We could just send optimistically and re-send later. There would be a higher rate of re-sends with that approach, but it may be a worthwhile trade-off for speed and efficiency in the base case.

xmtp_mls/README.md
```diff
@@ -0,0 +1,136 @@
...
+Simplified high level flow for adding members to a group:
+
+1. Consume Key Packages for all new members
+1. Create a `group_intent` for adding the members
+1. Sync the state of the group with the network
+1. Convert the intent into concrete commit and welcome messages for the current epoch
+1. Publish commit message
+1. Sync the state of the group with the network
+1. If no conflicts: Publish welcome messages to new members.
+   If conflicts: Go back to step 4 and try again
```
richardhuaaa, line 116: So at a high level all of these actions involve publishing a payload, and then only applying the change once you read the payload back from the network and confirm it has been placed in-order without any conflicts with other changes. We're essentially placing actions onto a queue.

Once you read the payload back from the queue, you need some kind of mechanism to correlate the payload back to your intent, and apply any remaining actions associated with the intent (e.g. send out welcome messages). 

Ideally:
1. Any data needed to correlate the payload back to the intent is in the database, rather than held in-memory (in the event that the app is killed before the payload is read).
2. The whole process should not block the user from taking further actions, as it involves multiple network round-trips. For example, you could add a member and then send a message (before receiving a response from the network), and then apply both changes when you read the corresponding payloads.

I'm wondering if we can support this kind of asynchronicity and robustness in the face of unreliable network conditions and randomly killed app lifetimes by splitting these into smaller atomic steps roughly mapping to the `intent_state` you enumerated above. Each of these steps can be picked up at any time (e.g. after an app crash, after coming back online) and don't block on any other step:

**1. User action loop**
User actions can be thought of as an ordered stream of events - for each user action we create a corresponding intent with `TO_SEND` state and store it in the database. The intents MUST be ordered in the database.

**2. Intent publishing loop**
Read through each `TO_SEND` intent in order and map it into the corresponding MLS messages (including downloading key packages if necessary). Publish them onto the network, and then once published update the state to `PUBLISHED` and move onto the next intent.

You may also need to persist some state, such as welcome messages, onto the intents table so that it can be retrieved later - perhaps there are separate `publish_data` and `commit_data` columns.

**3. Payload receiving loop**
Read payloads from the network in-order. Most of them will be messages or payloads from other installations. For payloads from your own installation, map them back to your intent, apply any remaining actions, and then update the intent state to `COMMITTED`. This loop can be a pull-based loop or a push-based loop, it doesn't matter.

These three loops can be running in three separate threads, or they can be methods that are manually invoked at regular intervals or on-demand from the same thread, it doesn't matter.

richardhuaaa (commented):
I love this proposal - the `intent` concept is powerful and gives a foundation for all of our future work, essentially solving the issue of retries for conflicting commits. This is also laid out incredibly clearly and seems very well thought out.

Building on this concept, my main suggestion would be to increase the asynchronicity and robustness of each of these operations using event loops.

Some other thoughts that came up for me, not necessarily addressable here:

1. In the event that an update fails due to conflicts, I'm not sure what happens with the cryptographic state, as we have already updated ours when constructing the original payload. I'm not sure if there is forking logic or rewinding logic already built into OpenMLS.
2. All retries in this proposal happen automatically without user involvement. It may be worth thinking through any cases where the user should manually retry instead - for example if they sent a sensitive message for a set of group members, but another member was added before the message was committed.
3. The event loop I suggested preserves the ordering of all user actions, but I'm not sure if this ordering is preserved if an update fails due to conflicts.
4. A malicious group member can add a member to the group, but not send them any welcome messages. This prevents anyone else from adding the member, while keeping the group undiscoverable from the member. It's not the biggest issue in practice, but I don't see anything in the MLS spec for how to address this, curious if @Bren2010 has thoughts

neekolas:
@richardhuaaa great questions

> In the event that an update fails due to conflicts, I'm not sure what happens with the cryptographic state, as we have already updated ours when constructing the original payload. I'm not sure if there is forking logic or rewinding logic already built into OpenMLS.

Unlike the demo I had done previously, we don't update our cryptographic state until we read the message from the network in this model. Nothing changes in the local ratchet tree/group state at the time of sending the commit.

> All retries in this proposal happen automatically without user involvement. It may be worth thinking through any cases where the user should manually retry instead - for example if they sent a sensitive message for a set of group members, but another member was added before the message was committed.

It's a good point. Feels like something we can table until post-launch. Retries of application messages should be very rare if we take advantage of `max_past_epochs` and only retry if a message is from an epoch so old that the keys have already been discarded (maybe a new group where members are being added in rapid succession by different members). 

> The event loop I suggested preserves the ordering of all user actions, but I'm not sure if this ordering is preserved if an update fails due to conflicts.

I would expect that if one intent fails due to conflicts all later intents will also fail. That may not be true if we set `max_past_epochs > 0`, since application messages could actually succeed even if they were from a past epoch. It may be a narrow enough edge case we don't have to worry too much about it.

xmtp_mls/README.md
neekolas: I like where you're going with this. Making things as parallelizable and re-entrant as possible is a worthwhile goal.

You want to try and fork the Excalidraw and see what these changes would look like in the state machine?

xmtp_mls/README.md
```diff
@@ -0,0 +1,136 @@
...
+## State Machine
+
+The [following diagram](https://app.excalidraw.com/s/4nwb0c8ork7/6pPH1kQDoj3) illustrates some common flows in the state machine
+
+![MLS State Machione](../img/mls-state-machine.png "MLS State Machine")
+
+### Known missing items from the state machine
+
+- Key updates
+- Processing incoming welcome messages
```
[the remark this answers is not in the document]
neekolas: That's a good point. I've totally left out the notion of permissioning. But yes, I think having the group creator (wallet address) be the only one allowed to add/remove members is a good place to start.

xmtp_mls/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Add members to a group
+
+Simplified high level flow for adding members to a group:
+
+1. Consume Key Packages for all new members
+1. Create a `group_intent` for adding the members
+1. Sync the state of the group with the network
```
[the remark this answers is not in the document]
neekolas: 1. We don't strictly need to sync the state. It just makes conflicts less likely. Given that adds/removes are relatively rare it might make sense. Also might be totally overkill.
2. I mean process all the unread messages on the group topic
3. I've left subscriptions out of the equation here, but I have been thinking about them a bit. Subscriptions don't guarantee ordering or completeness of results, and that's unlikely to change. For us to support subscriptions in MLS I think the flow needs to be:
- Open a subscription to the group topic
- For any application message for an epoch you have keys for, process it immediately
- When you receive a commit, trigger a call to the network to read from the group topic in the committed order. Discard any application messages you have already saved, and process all commits in order.

That lets us have real-time messaging without negating the benefits of total ordering we get from the query API.

xmtp_mls/README.md
```diff
@@ -0,0 +1,143 @@
...
+- TO_SEND // Either has never been sent to the network or needs to be re-sent
+- PUBLISHED // Sent to the network but has not been read back or committed
+- COMMITTED // Committed messages could be deleted
+
+### INTENT_KIND
+
+- SEND_MESSAGE // An intent to send a message to the group
+- ADD_MEMBERS // An intent to add members to the group
+- REMOVE_MEMBERS // An intent to remove members from the group
+- KEY_UPDATE // An intent to update your own group key
```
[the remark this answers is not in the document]
Bren2010: I would propose these intent kinds:
- SEND_MESSAGE: Send a message to members of the group
- COMMIT_NEW: Send a commit covering new actions (update / remove / add) to the group
- COMMIT_PENDING: Send a commit covering pending proposals

A "COMMIT_PENDING" intent is created when the application is unable to send an application message because there are pending proposals. Then you can specify that intents are always processed in this order:
- COMMIT_PENDING intents always get executed first
- Then SEND_MESSAGE and COMMIT_NEW intents are executed in the order they were inserted in the table (note: requires adding timestamp column to group_intents)

It is probably easiest for now to handle each intent as blocking; you can't handle the second one until the first one is complete and deleted.

xmtp_mls/README.md
```diff
@@ -0,0 +1,143 @@
...
+}
+
+-- This table is required to retry messages that do not send successfully due to epoch conflicts
+CREATE TABLE group_intents {
+    id SERIAL PRIMARY KEY NOT NULL,
+    kind INT NOT NULL, -- INTENT_KIND
+    group_id BLOB NOT NULL,
+    "data" BLOB NOT NULL, -- Some sort of serializable blob that can be used to re-try the message if the first attempt failed due to conflict
+    "state" INT NOT NULL, -- INTENT_STATE,
+    message_hash BLOB -- The hash of the encrypted, concrete, form of the message if it was published.
```
Bren2010, line 47: You may want to store the epoch in which the message is intended to land in this table. If the message lands in a subsequent epoch, it needs to be retried.

xmtp_mls/README.md
```diff
@@ -0,0 +1,143 @@
...
+}
+
+CREATE TABLE outbound_welcome_messages {
+    id BLOB PRIMARY KEY NOT NULL, -- Derived via SHA256(CONCAT(group_id, welcome_message, ratchet_tree, installation_id))
+    state INT NOT NULL, -- OUTBOUND_WELCOME_STATE
+    installation_id BLOB PRIMARY KEY NOT NULL,
```
Bren2010, line 53: Curious what the installation_id here is referring to? I don't think it should be an individual recipient, as Welcome messages are shared by all recipients

xmtp_mls/README.md
```diff
@@ -0,0 +1,143 @@
...
+    commit_hash BLOB NOT NULL, -- The hash of the commit which created this
+    group_id BLOB NOT NULL,
+    welcome_message BLOB NOT NULL,
+    ratchet_tree BLOB NOT NULL,
```
Bren2010, line 57: The ratchet tree should be included in the Welcome

xmtp_mls/README.md
```diff
@@ -0,0 +1,143 @@
...
+    "data" BLOB NOT NULL, -- Some sort of serializable blob that can be used to re-try the message if the first attempt failed due to conflict
+    "state" INT NOT NULL, -- INTENT_STATE,
+    message_hash BLOB -- The hash of the encrypted, concrete, form of the message if it was published.
+}
+
+CREATE TABLE outbound_welcome_messages {
+    id BLOB PRIMARY KEY NOT NULL, -- Derived via SHA256(CONCAT(group_id, welcome_message, ratchet_tree, installation_id))
+    state INT NOT NULL, -- OUTBOUND_WELCOME_STATE
+    installation_id BLOB PRIMARY KEY NOT NULL,
+    commit_hash BLOB NOT NULL, -- The hash of the commit which created this
```
Bren2010, line 54: Does this correspond to message_hash in group_intents? Same column name would make that clearer

xmtp_mls/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+-- Successfully processed messages meant to be returned to the user
+CREATE TABLE group_messages {
+    id BLOB PRIMARY KEY NOT NULL -- Derived via SHA256(CONCAT(decrypted_message_bytes, conversation_id, timestamp))
+    group_id BLOB NOT NULL,
+    decrypted_message_bytes BLOB NOT NULL, -- Message contents after decryption
+    sent_at_ns BIGINT NOT NULL, -- Based on the timestamp of the message
+    sender_installation_id TEXT NOT NULL,
+    sender_wallet_address TEXT NOT NULL, -- Could remove this if we added a table mapping installation_ids to wallet addresses
+}
```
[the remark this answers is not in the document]
neekolas: That's a very good point. This doesn't really consider the membership change history, which we do want to store somewhere.

Let me think about how we want to store that at rest.

xmtp_mls/README.md
neekolas: I was kinda hoping `openmls` would handle the logic of deciding "this is from the wrong epoch" rather than us having to do the bookkeeping separately. That way it can transparently handle things like setting `max_past_epochs` for the group to a value > 0.

xmtp_mls/README.md
neekolas: An `installation_id` is the term we are using as the unique identifier for particular app installation. It is just the public key of the signing key for their MLS identity.

While welcome messages are shared between all recipients, we are going to need to send multiple copies of the message to each installation's topic on our network so that they can go to the correct recipient. Every installation has a topic where their welcome messages are stored.

It's a little wasteful to keep so many copies of each pending message, but it lets us modify the state if some sends succeed and some fail.

xmtp_mls/README.md
neekolas: Good call out. I'll rename.

xmtp_mls/README.md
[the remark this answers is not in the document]
neekolas: If we have a generic "commit" intent, my concern is how hard it would be to untangle the commit and reconstruct it in a later epoch. Do we still have access to the Key Packages of new members etc in a format that is usable for generating a new commit? I'd have to dig in a little deeper to OpenMLS to figure out just how accessible that information is.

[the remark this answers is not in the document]
neekolas: > Then SEND_MESSAGE and COMMIT_NEW intents are executed in the order they were inserted in the table (note: requires adding timestamp column to group_intents)

I was imagining using the SERIAL primary key to handle sequencing, but I should have called that out more clearly in the doc.

> It is probably easiest for now to handle each intent as blocking; you can't handle the second one until the first one is complete and deleted.

That feels like a reasonable place to start

xmtp_mls/README.md
neekolas: Sorry. I forgot to update this README. I realized the same thing and removed elsewhere

xmtp_mls/README.md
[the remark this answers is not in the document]
richardhuaaa: > COMMIT_PENDING: Send a commit covering pending proposals

@Bren2010 I didn't understand what this means. What happens when this intent is processed?

> A "COMMIT_PENDING" intent is created when the application is unable to send an application message because there are pending proposals

Can't we simply store an `intent` with the plaintext message, and then later on send the message after all previous intents have been executed?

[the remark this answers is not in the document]
richardhuaaa: Oh I think I understood your intention here. The current idea is to not use proposals at all, only commits. Are there drawbacks to this?

xmtp_mls/README.md
Bren2010: OpenMLS is not going to be able to decrypt messages that you yourself sent. So when you are seeing what order messages got put in on the topic, you will not be able to "black box" run your own messages through OpenMLS to see if they ended up decryptable

xmtp_mls/README.md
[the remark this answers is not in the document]
richardhuaaa: Summary of chat: We can choose to send proposals and commits in the same payload, but in order to be standards-compliant we can't assume this is the case when receiving. Therefore, we may have pending proposals at any time. This will in turn block the sending of application messages

xmtp_mls/README.md
neekolas: I'd love to learn more about that. My understand was that I would be able to validate my own messages, and in the case of commits be able to apply them. Is that not the case?

xmtp_mls/README.md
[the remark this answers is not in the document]
richardhuaaa: > A "COMMIT_PENDING" intent is created when the application is unable to send an application message because there are pending proposals

What problem arises if we simply commit pending proposals as part of the `send message` intent (and possibly also every other intent)?

xmtp_mls/README.md
```diff
@@ -0,0 +1,172 @@
...
+- SEND_MESSAGE // An intent to send a message to the group
+- ADD_MEMBERS // An intent to add members to the group
+- REMOVE_MEMBERS // An intent to remove members from the group
+- KEY_UPDATE // An intent to update your own group key
+
+### OUTBOUND_WELCOME_STATE
+
+- PENDING // Needs to wait for commit to be applied before sending
+- READY_TO_SEND
+- SENT // Messages may be deleted at this point. We may decide to remove this state altogether.
```
Bren2010, line 101: I believe this can be removed

xmtp_mls/README.md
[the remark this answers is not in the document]
Bren2010: > If we have a generic "commit" intent, my concern is how hard it would be to untangle the commit and reconstruct it in a later epoch.

The reason I proposed a generic "commit" intent was to allow actions of several different types to be batched together. I see below that you're scoping that out for now, so this should be fine!

xmtp_mls/README.md
```diff
@@ -0,0 +1,172 @@
+# XMTP MLS
```
jhaaaa, line 1: Thank you for this README, @neekolas! What do you think of updating the title and adding a bit of intro text for context? Some suggested starter text here:

```suggestion
# XMTP group chat using MLS

This document describes plans for how XMTP can provide group chat using [Messaging Layer Security](https://messaginglayersecurity.rocks/) (MLS).
```

xmtp_mls/README.md
```diff
@@ -0,0 +1,192 @@
...
+Simplified high level flow for sending a group message:
+
+1. Create a `group_intent` for sending the message
+1. Convert the intent into a concrete message for the current epoch
+1. Publish message to the network
+1. Sync the state of the group with the network (can be debounced or otherwise only done periodically)
+1. If no conflicts: Mark the message as committed.
+   If conflicts: Go back to step 2 and try again (reset the intent's state to `TO_SEND` and clear the `publish_data` and `post_commit_data` fields)
+
+### Syncing group state
```
richardhuaaa, line 152: @neekolas and I tentatively discussed something like this for handling concurrency.

Would love feedback @neekolas, @Bren2010, @insipx! I wrote up a lot of the rationale so that it's easier to follow, but not sure if the rationale belongs in this doc.

neekolas: This makes sense to me. I like it, and it feels straightforward enough to implement. Love removing the locking mechanism from the `topic_refresh_state` table.

xmtp_mls/README.md
neekolas: The library today only handles group chat, but in future it's gong to include 1:1 as well. 

I will add the intro text, which helps set the reader up for the onslaught of technical detail that follows.

neekolas:
@richardhuaaa you mind approving this so we can merge it? We can always amend it with more PRs as plans change or details get fleshed out.

PR #288, opened 2023-10-29
read_topic API

neekolas:
## Summary

The only query patterns I can see in our use-cases are to read a topic to completion, after a given timestamp. So, I made a method in the `ApiWrapper` that does that and handles pagination for the caller.


xmtp_mls/src/api_client_wrapper.rs
```diff
@@ -37,9 +40,56 @@ where
...
+
+            for envelope in &result.envelopes {
+                out.push(envelope.clone());
+            }
+
+            if result.envelopes.len() < page_size as usize || result.paging_info.is_none() {
+                break;
+            }
+
+            cursor = match result.paging_info.unwrap().cursor {
```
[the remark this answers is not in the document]
neekolas: There is a chance that there is a null paging info for the last page. But that gets checked above and we break if it's `None`. I like that little tweak. Will make the change.

richardhuaaa:
Nooby question - is there a max page size our API supports? If our client is always reading to completion, what is the purpose of paging under the hood?

neekolas:
@richardhuaaa 100 item max page size

xmtp_mls/src/api_client_wrapper.rs
```diff
@@ -34,6 +40,51 @@ where
...
+                        cursor,
+                        limit: page_size,
+                        direction: SortDirection::Ascending as i32,
+                    }),
+                })
+                .await?;
+
+            for envelope in &result.envelopes {
+                out.push(envelope.clone());
+            }
```
richardhuaaa, line 68: Small nit: I think you can do `out.append(result.envelopes)` if you do `let mut result` above and store the count beforehand, that might save a number of clones

xmtp_mls/src/api_client_wrapper.rs
```diff
@@ -34,6 +40,51 @@ where
...
+
+            if result.envelopes.len() < page_size as usize || result.paging_info.is_none() {
+                break;
+            }
+
+            cursor = result
+                .paging_info
+                .expect("Empty paging info")
+                .cursor
+                .map(|wrapper| wrapper);
```
richardhuaaa, line 78: Don't forget this lint!

richardhuaaa (approved):
👍

xmtp_mls/src/api_client_wrapper.rs
neekolas: That's pretty slick. Done

PR #289, opened 2023-10-30
Add database models

neekolas:
## Summary

- Adds a first pass at the database schema
- Adds initial models matching the schema
- Breaks up models into one file per table

## Notes

I decided to break up the DB models into multiple files. The `xmtp` crate's `EncryptedMessageStore` was getting pretty long and unwieldy, especially for tests. I figure we can break up the `impl` blocks for the store methods and all the Trait implementations into the respective per-table files and things should be more manageable. Lmk what you all think.


xmtp/migrations/2023-06-30-063813_create_users_and_conversations/up.sql
```diff
@@ -1,8 +1,9 @@
 CREATE TABLE users (
-    user_address TEXT PRIMARY KEY NOT NULL,
-    created_at BIGINT NOT NULL,
-    last_refreshed BIGINT NOT NULL
+  user_address TEXT PRIMARY KEY NOT NULL,
```
insipx, line 2: are we re-using some of the tables from XMTP?

insipx (approved):
LGTM

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
```diff
@@ -0,0 +1,72 @@
...
+
+-- Allow for efficient sorting of groups
+CREATE INDEX groups_created_at_idx ON groups(created_at_ns);
+
+CREATE INDEX groups_membership_state ON groups(membership_state);
+
+-- Successfully processed messages meant to be returned to the user
+CREATE TABLE group_messages (
+    "id" BLOB PRIMARY KEY NOT NULL,
+    -- Derived via SHA256(CONCAT(decrypted_message_bytes, conversation_id, timestamp))
```
richardhuaaa, line 18: I think this comment was intended for the field above

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
```diff
@@ -0,0 +1,72 @@
...
+    "group_id" BLOB NOT NULL,
+    -- Message contents after decryption
+    "decrypted_message_bytes" BLOB NOT NULL,
+    -- Based on the timestamp of the message
+    "sent_at_ns" BIGINT NOT NULL,
+    -- Enum GROUP_MESSAGE_KIND
+    "kind" INT NOT NULL,
+    -- Could remove this if we added a table mapping installation_ids to wallet addresses
```
richardhuaaa, line 26: I would be in favor of this - we probably will need something eventually to track validity/revocation/expiries

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
```diff
@@ -0,0 +1,72 @@
...
+    -- Serial ID auto-generated by the DB
+    "id" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    -- Enum INTENT_KIND
+    "kind" INT NOT NULL,
+    "group_id" BLOB NOT NULL,
+    -- Some sort of serializable blob that can be used to re-try the message if the first attempt failed due to conflict
+    "data" BLOB NOT NULL,
+    -- INTENT_STATE,
+    "state" INT NOT NULL,
+    -- The hash of the encrypted, concrete, form of the message if it was published.
```
richardhuaaa, line 51: Why is this needed?

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
```diff
@@ -0,0 +1,72 @@
...
+    "message_hash" BLOB,
+    FOREIGN KEY (group_id) REFERENCES groups(id)
+);
+
+CREATE INDEX group_intents_group_id_id ON group_intents(group_id, id);
+
+CREATE TABLE outbound_welcome_messages (
```
richardhuaaa, line 58: 1. Do we need this as a separate table, as opposed to `data` on `group_intents`?
2. Can this be created on-demand as opposed to pre-created in the beginning?

xmtp_mls/src/storage/encrypted_store/key_store.rs
```diff
@@ -0,0 +1,38 @@
+use super::DbConnection;
```
richardhuaaa, line 1: For consistency I think this file should be called `key_store_entry.rs`

richardhuaaa (approved):

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
neekolas: You're right. Will move.

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
neekolas: I'm still on the fence about this. It's a lot simpler to just rely on the MLS group state and the network to tell you when something has changed or is expired. I think a lot of the data we'd be tracking there would be duplicative. But I haven't dug in too deeply. I'd suggest we leave it in the "maybe" column until we're closer to implementation or actually have a concrete need for it.

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
neekolas: When we read a message back off the network we need to link back to the intent that generated it to update the intent state. In the case of your own messages, they aren't even decryptable.

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
neekolas: I'm mostly basing this off of the APIs provided by OpenMLS today. There may be ways of reconfiguring OpenMLS to work the way you propose.

I'm suggesting the flow works as such:
1. Create an intent to add members, with just a list of `installation_id`s as the intent data
2. When processing the intent (after having processed all previous intents) we call `group.add_members(...)` with the key packages of all the new members. This will give us back the commit and the welcome message that has to be sent to all members. It also saves a pending commit in the MLS store. The welcome message doesn't get generated anywhere else in OpenMLS.
3. We read the group state from the network and see if the commit will be accepted or rejected. If accepted we merge the pending commit stored in the MLS state.
4. If accepted, we send the welcome messages created in step 2

To do it on demand, we'd need to somehow generate the welcome messages in step 3 or 4. But based on my reading of the OpenMLS docs, it has to happen at Step 2. That's why I'm suggesting a separate table. Alternatively, we could have a mutable field on the intent (maybe called `meta` or something) that we could write the welcome message data to in step 2. That would remove the need for a separate table, but would make more of the intent mutable.

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
```diff
@@ -0,0 +1,72 @@
+CREATE TABLE groups (
+    -- Random ID generated by group creator
+    "id" BLOB PRIMARY KEY NOT NULL,
+    -- Based on the timestamp of the welcome message
+    "created_at_ns" BIGINT NOT NULL,
+    -- Enum of GROUP_MEMBERSHIP_STATE
+    "membership_state" INT NOT NULL
```
insipx, line 7: could we use TEXT as the enum instead of an integer? 

Something like 

` membership_state      TEXT CHECK( membership_state IN ('allowed','rejected','pending') )`

neekolas: I'm easy there. A few bytes of extra state isn't going to kill us. It would require a database migration every time we add an allowed enum value to update that check, as opposed to the int where it's just a code change. But it's more readable in the DB which is worth something.

insipx: that's true, I didn't consider the migration aspect. Maybe we just leave it as int for now. I'm not strongly for either way I was just writing up the rust part and had the idea

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
richardhuaaa: Ahh gotcha. Do we intend for this to apply only to application messages, or just generically all payloads? If so, would `payload_hash` be more accurate?

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
richardhuaaa: Gotcha, so we can't create them on-demand. Yes, I think it might be better to use a `data` or `meta` or `post_commit_data` field on `group_intents` - it creates less complexity in the schema, saves us models and having to marshal them in and out per-field, and saves us having to write separate queries to fetch the `outbound_welcome_messages` later.

neekolas: I'm on board for that

xmtp_mls/src/storage/encrypted_store/key_store.rs
neekolas: Good call

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
neekolas: All payloads. I'm fine with `payload_hash`

xmtp/migrations/2023-06-30-063813_create_users_and_conversations/up.sql
neekolas: This schema below is all we have a need for at the moment. Because OpenMLS has its own database structure with the KeyValue store, some of the tables from the old `xmtp` schema simply aren't necessary.

richardhuaaa (approved):

Issue #291, opened 2023-10-31
Feature request: `group_messages` database table


neekolas:
### Is your feature request related to a problem?

We need the following features for the Group messages database table

- [x] Create schema and models
- [x] Implement `Fetch` and `Store` traits for model
- [ ] Add methods to `EncryptedMessageStore` for querying for messages
- [ ] Add tests for inserting and retrieving group messages and updating membership state
### Describe the solution to the problem

The query APIs should look something like

```rust
pub fn get_group_messages(&self, conn: &mut DbConnection, group_id: &[u8], sent_after: Option<i64>, sent_before: Option<i64>, kind: Option<i64>) -> Result<Vec<StoredGroupMessage>, StorageError> {}

pub fn get_group_message(&self, conn: &mut DbConnection, id: &[u8]) -> Result<Option<StoredGroupMessage>, StorageError> {}
```

Issue #286, opened 2023-10-27
Feature request: Groups Database Table


neekolas:
### Is your feature request related to a problem?

We need to access the groups table in `xmtp_mls`

### Describe the solution to the problem

- [x] Add migration for creating the groups table according to the [schema](https://github.com/xmtp/libxmtp/pull/284/files#diff-5050c06d09f185c8204da5aa643f691fcd8e90011542ed42cf7a7a24da6ddda9R18)
- [x] Implement `Store` and `Fetch` traits for the table (I don't think we need Delete here, but maybe I'm wrong)
- [x] Create `GroupMembershipState` enum
- [x] Add method to `EncryptedMessageStore` to update group membership state
- [ ] Add tests for inserting and retrieving group messages and updating membership state

neekolas:
I checked off the first item in https://github.com/xmtp/libxmtp/pull/289

PR #300, opened 2023-10-31
Group Database Table Implementations

insipx:
finishing out #286 

This also impls Fetch & Store for most tables

Tests for inserting & retreiving messages will be apart of #291


insipx:
This PR is ready to be opened, but i'm going to wait on the other PRS to merge (built this on database models branch) and then run cargo fmt on the codebase (adding a ./dev/fmt)

xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -1,11 +1,119 @@
...
+pub enum GroupMembershipState {
+    /// User is allowed to interact with this Group
+    Allowed = 1,
+    /// User has been Rejected from this Group
+    Rejected = 2,
+    /// User is Pending acceptance to the Group
+    Pending = 3,
+}
+
+impl ToSql<Integer, Sqlite> for GroupMembershipState
```
neekolas, line 47: Ooh. This is great. No more manually converting enums into `i32`s

xmtp_mls/src/storage/encrypted_store/outbound_welcome_message.rs
```diff
@@ -0,0 +1,18 @@
+use super::schema::outbound_welcome_messages;
```
neekolas, line 1: If you update from `main` this file should now be gone, as well as the associated table

insipx: oo i probably screwed something up in the merge, good catch

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -10,19 +11,22 @@ use super::{
 };
 
 #[derive(Debug)]
+/// CRUD Operations for an [`EncryptedMessageStore`]
 pub struct SqlKeyStore<'a> {
-    store: &'a EncryptedMessageStore,
+    store: Cow<'a, EncryptedMessageStore>,
```
neekolas, line 16: I'd love to learn more about what the implications are for wrapping this in a `Cow` are, and what the advantages are. How do I make sure I don't misuse the smart pointer?

xmtp_mls/src/api_client_wrapper.rs
```diff
@@ -590,10 +587,7 @@ mod tests {
         mock_api
             .expect_query()
             .withf(|req| match req.paging_info.clone() {
-                Some(paging_info) => match paging_info.cursor {
-                    Some(_) => true,
-                    None => false,
-                },
+                Some(paging_info) => paging_info.cursor.is_some(),
```
neekolas, line 590: Nice 🏌️

neekolas (approved):

xmtp_mls/src/api_client_wrapper.rs
insipx: `cargo clippy --fix`

xmtp_mls/src/storage/sql_key_store.rs
neekolas: So long as Cow is always cloning the existing instance, it should be fine since the SQLite connection pool is safe to be cloned. 

My main worry is around anything that causes multiple SQLite pools to exist at the same time pointing to the same DB file. That will cause conflicts and failures on db writes.

PR #302, opened 2023-11-01
Remove complicated build flow from PRs

neekolas:
## Summary

We are building a ton of different binaries on every push to a PR. These are really a production release process and are _very_ slow. I propose we remove the build steps from PRs and just do it on pushes to main (relying on the tests to validate whether something works in a PR). This should speed up our CI time dramatically, since we have so many build steps that some get queued on each commit.

The trade-off is that if we make a change that breaks the iOS/Android build we will only find out when something is merged to main. To my knowledge, this has never happened in a way that wouldn't also show up in the tests. But it's possible.

I also removed the `bindings_js` tests, since that package isn't used at all. Should probably follow up to just remove the whole folder.


.github/workflows/build.yml
```diff
@@ -1,11 +1,9 @@
 name: Build
 on:
-  pull_request:
   push:
     branches:
       - main
 jobs:
-
```
neekolas, line 8: All changes below this are just formatting. It appears this file never went through a formatter.

insipx (approved):

Issue #292, opened 2023-10-31
Feature request: `topic_refresh_state` database table


neekolas:
### Is your feature request related to a problem?

We need some database methods for interacting with the `topic_refresh_state`.

- [x] `Store` and `Fetch` trait implementations
- [ ] `lock_topic`
- [ ] `unlock_topic` method on `EncryptedMessageStore`

### Describe the solution to the problem

The way this needs to work is that we want to lock the row for a given topic while we are loading messages from it, and unlock the row when the job is complete.

We could use a pessimistic locking strategy, like wrapping updates in a transaction or otherwise locking the database. But we don't want to block read access for that long given we are waiting for asynchronous network requests to complete.

So, it may make more sense to just add a `locked_until` field on the table which we can use to do concurrency control. The way that would work would look something like this:

When we begin a refresh of the topic we would use the `lock_topic` method to check the current row and ensure that it either
A) Doesn't exist
B) Has a `locked_until` value set in the past

If the check fails (the topic is currently locked) we would retry in a loop until it becomes unlocked. If either of the conditions above are true, we would continue to set the `locked_until` value to something like 5 or 10 seconds into the future.

When the operation has completed we would update the row and set the `locked_until` value to NULL using the `unlock_topic` method and set the `last_message_timestamp_ns` field to the timestamp of the newest entry in the topic (if any). These two operations should be done in the same database update. 

### Additional details

I'm open to other implementations that give us the same properties.

1. We don't completely lock up the database to other callers while the topic refresh job is in progress
2. We can ensure a single process is refreshing a given topic at any time
3. The `last_message_timestamp_ns` field is updated at the same time or before the topic is unlocked for the next caller

I'm open to solutions that just use application-level mutexes to get to the same result, which would be fine given that we expect only a single `libxmtp` instance will be operating on a database at any time.

insipx:
I also think it would be possible to use a `RWLock` here and enforce updates in a single thread

richardhuaaa:
@insipx I wrote up a possible method for handling concurrency here: https://github.com/xmtp/libxmtp/pull/284. The advantage is it's relatively simple to implement and non-blocking for callers. Would love your thoughts!

PR #303, opened 2023-11-01
Suggestions for MLS state machine

richardhuaaa:
Will try to keep these suggestion PR's small, for this one:

1. Serialize welcome messages to `post_commit_data` field on intent rather than separate table
1. Remove some group sync steps (we can handle conflicts anyway)
1. EDIT: Drop pending proposals for now
1. Describe how payloads can be synced in a non-duplicative way

Other things worth considering:
1. Making some of the steps generic rather than specific to intent type (a lot of it is duplicative/identical)
1. Whether to make commit processing strictly sequential (commit must be fully done before next one begins) or concurrent. If we make it strictly sequential, we need to think about how to handle concurrent syncs (as the application also needs to sync incoming messages, either via pull or subscription)


richardhuaaa:
AFAIK, MLS lets you define the logic for which proposals are accepted or rejected. Another possibility could be to mandate that each client will only commit its own proposals. OpenMLS has a `[remove_pending_proposal](https://openmls.tech/openmls/doc/openmls/group/struct.MlsGroup.html#method.remove_pending_proposal)` method that we could use instead of committing pending proposals

xmtp_mls/README.md
```diff
@@ -52,7 +52,9 @@ CREATE TABLE group_intents (
     "kind" INT NOT NULL,
     "group_id" BLOB NOT NULL,
     -- Some sort of serializable blob that can be used to re-try the message if the first attempt failed due to conflict
-    "data" BLOB NOT NULL,
+    "publish_data" BLOB NOT NULL,
+    -- Data needed after applying a commit, such as welcome messages
+    "post_commit_data" BLOB NOT NULL,
```
neekolas, line 57: Thanks. This is also in the DB models PR now.

xmtp_mls/README.md
```diff
@@ -122,41 +116,58 @@ The [following diagram](https://app.excalidraw.com/s/4nwb0c8ork7/6pPH1kQDoj3) il
...
-1. Sync the state of the group with the network
-1. Convert the intent into concrete commit and welcome messages for the current epoch
+1. Consume Key Packages for all new members
+1. Convert the intent (and any pending proposals) into concrete commit and welcome messages for the current epoch
+   1. Write the welcome messages to the `post_commit_data` field for later
 1. Publish commit message
 1. Sync the state of the group with the network
 1. If no conflicts: Publish welcome messages to new members.
-   If conflicts: Go back to step 4 and try again
+   If conflicts: Go back to step 2 and try again (reset the intent's state to `TO_SEND` and clear the `data` and `post_commit_data` fields)
```
neekolas, line 134: This works for me

xmtp_mls/README.md
```diff
@@ -122,41 +116,58 @@ The [following diagram](https://app.excalidraw.com/s/4nwb0c8ork7/6pPH1kQDoj3) il
...
+### Syncing group state
+
+The server maintains an inbound topic for each group, and a single inbound topic for the client's identity. For each topic, the client maintains a row that stores `last_synced_payload_id` and `lock_sync_until_ns` fields.
+
+1. In a single transaction, validate that `lock_until_ns` is not set to a value greater than `now()`, set it to a timeframe in the future, and fetch the `last_message_timestamp_ns`
+1. Fetch all payloads greater than the timestamp from the server
+1. Sequentially process each payload. For each payload, update the `last_message_timestamp_ns` and any corresponding database writes for that payload in a single transaction
+1. When the sync is complete, release the lock by setting `lock_until_ns` to 0
+
+This flow will be similar regardless of if the sync happens via a poll-based or subscription-based mechanism. For a subscription-based mechanism, the lock will be obtained at the start of the subscription, and extended on a heartbeat time interval, until the subscription is closed.
```
neekolas, line 170: In the absence of really robust subscriptions (which I think is going to be quite hard to implement for V1) my recommendation would be to not adjust any locks on the subscription and allow state syncs to happen in parallel.

xmtp_mls/README.md
```diff
@@ -122,41 +116,58 @@ The [following diagram](https://app.excalidraw.com/s/4nwb0c8ork7/6pPH1kQDoj3) il
...
 1. Sync the state of the group with the network (can be debounced or otherwise only done periodically)
-1. If no conflicts: Mark the message as committed. If conflicts: Go back to step 2.
+1. If no conflicts: Mark the message as committed. If there are welcome messages in the `post_commit_data` field, publish them.
+   If conflicts: Go back to step 2 and try again (reset the intent's state to `TO_SEND` and clear the `data` and `post_commit_data` fields)
+
+### Syncing group state
+
+The server maintains an inbound topic for each group, and a single inbound topic for the client's identity. For each topic, the client maintains a row that stores `last_synced_payload_id` and `lock_sync_until_ns` fields.
+
+1. In a single transaction, validate that `lock_until_ns` is not set to a value greater than `now()`, set it to a timeframe in the future, and fetch the `last_message_timestamp_ns`
```
neekolas, line 165: I also have this logic broken out as a separate ticket [here](https://github.com/xmtp/libxmtp/issues/292)

xmtp_mls/README.md
```diff
@@ -122,41 +116,58 @@ The [following diagram](https://app.excalidraw.com/s/4nwb0c8ork7/6pPH1kQDoj3) il
...
-   If conflicts: Go back to step 3 and try again
+1. If no conflicts: Done. If there are welcome messages in the `post_commit_data` field, publish them.
+   If conflicts: Go back to step 2 and try again (reset the intent's state to `TO_SEND` and clear the `data` and `post_commit_data` fields)
 
 ### Send a message
 
 Simplified high level flow for sending a group message:
 
 1. Create a `group_intent` for sending the message
+1. Commit any pending proposals
```
neekolas, line 153: One thing that gets a little weird about this flow is that committing any pending proposals will result in a new message to the network containing the commit in addition to the one containing the message. All other intents map to a single unprocessed message on the network. That breaks the model a bit (what `payload_hash` do we store on the intent?). Maybe the thing to do here is actually to spawn a new intent committing the change? Open to other ideas.

neekolas:
> OpenMLS has a [remove_pending_proposal](https://openmls.tech/openmls/doc/openmls/group/struct.MlsGroup.html#method.remove_pending_proposal) method that we could use instead of committing pending proposals

That would allow us to check the box of "supporting proposals" in the short term, without actually having to support them. We'd have to think through what it would look like to change this later. Maybe having a single member of the group on a newer version that didn't remove the proposals, but instead created a commit, would be fine. Would just require a bit of investigation.

neekolas:
> Whether to make commit processing strictly sequential (commit must be fully done before next one begins) or concurrent. If we make it strictly sequential, we need to think about how to handle concurrent syncs (as the application also needs to sync incoming messages, either via pull or subscription)

The simplest is to always make processing sequential. Gives us the same guarantees no matter where the message came from. Push notifications and Subscriptions that contain a commit message would just trigger a sync using the Query API that would ensure all previous commits were applied first. Application messages received via Push or Subscription are probably safe to attempt processing immediately (although they may fail if you had somehow missed a prior commit). Failed processing could be remedied with a sync-by-query.

It feels like anything else breaks the total ordering assumption baked into MLS.

We are going to need to design a solution to the total ordering problem for decentralization, but we've already decided to table that for a later iteration of the protocol.

richardhuaaa:
> The simplest is to always make processing sequential. Gives us the same guarantees no matter where the message came from. Push notifications and Subscriptions that contain a commit message would just trigger a sync using the Query API that would ensure all previous commits were applied first.

I think I understand - we can't process the same payload twice, so the suggestion here is to have an application sync trigger commit processing first. But what happens if you have an application sync in progress at the time that you make a commit, or halfway through the commit process?

xmtp_mls/README.md
richardhuaaa: If subscriptions are not robust, I feel like it causes a lot of downstream problems.
1) If a commit message is lost, subsequent payloads are unreadable. You can trigger a pull at this point, but it's added complexity.
2) If we don't lock syncs while subscriptions are in progress, then we have concurrent processing of the same payloads, meaning one side or the other will fail. It becomes unclear if failures are due to bad payloads/missing epoch messages, or if they are due to parallel processing.
3) It's not the best dev experience for consumers of the SDK.

I have some suggestions:
1) We could make subscriptions more robust. We could pass the `last_synced_payload_timestamp` to the server when initiating subscriptions. Alternatively, we could have the server record the previous payload timestamp on each payload, so that the client can detect when it is missing payloads and re-sync.
2) We could omit subscriptions from the initial MLS milestones and fix them up in later milestones.

neekolas:
> But what happens if you have an application sync in progress at the time that you make a commit, or halfway through the commit process?

We _could_ have some sort of locking mechanism that prevents this. These updates should be relatively fast and simultaneous syncs should be infrequent. But I don't think we need that. If a message fails to decrypt because it's already been processed and the keys have been discarded, we just move on to the next message. As long as both processes are operating sequentially and talking to the same DB, everything should just work.

The system is going to need to be very resilient to decryption failures since anyone can publish junk to the group topic, plus the fact that your own messages are unreadable.

neekolas (approved):

xmtp_mls/README.md
neekolas: I do like the third option. I hadn't considered adding the `previous_payload_timestamp` to all our streamed envelopes.

Given that all options look like option 1 in some cases, I wonder if we start with that to unblock everything and give us streaming methods quite easily. Then we can layer on option 2 or 3 depending on how we feel about the scope of changes.

richardhuaaa: That's a really good framing, thanks for writing it up. Okay, this makes sense to me!

richardhuaaa:
This was a really helpful discussion. I'll try to succinctly describe the various edge cases we discussed, at least so that I understand how to implement the thing, but with that in mind I think I'm happy with this!

PR #306, opened 2023-11-02
Document XMTP identities

richardhuaaa:
We are reviewing the XMTP identity structure and need a document to share. This is mostly similar to what we wrote up for the double-ratchet implementation, with adjustments for the MLS work.

I'm still working on the mapping from conversations to accounts and will put up the options for this shortly.

### DIFF
--- a/xmtp_mls/IDENTITY.md
+++ b/xmtp_mls/IDENTITY.md
@@ -1,0 +1,82 @@
+# XMTP Identity Structure
+
+In XMTP v3, a messaging account is represented by an Ethereum wallet address. An account consists of multiple app installations that may send and receive messages on behalf of it. Each installation is a separate cryptographic identity with its own set of keys.
+
+```
+Amal's account (Ethereum wallet address)
+│
+├── Converse app (mobile phone)
+│   └── Installation key bundle 1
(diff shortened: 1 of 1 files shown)


xmtp_mls/IDENTITY.md
```diff
@@ -0,0 +1,82 @@
...
+       }
+   }
+   ```
+
+4. A last resort KeyPackage (signed by the installation key per the MLS spec) is generated and stored on the device.
+5. The app publishes the public signing key, credential, and last resort key package to the account's contact topic on the server. This enables other apps to understand that the installation is on the network, associated with the account, and how to contact it.
+
+**Installation management**
+
+At any time, the user may enumerate active installations by querying the contact topic for the account. The user may identify each installation by the creation time as well as the installation key from the signing text.
```
neekolas, line 59: We no longer have a contact topic, and in MLS all installations are stored in the MLS db but not on the network

neekolas (approved):

neekolas:
This looks great. Thanks for putting it together. Feels like the right level of granularity for newcomers to get an understanding of how identities work in this system.

PR #307, opened 2023-11-03
Break up fetch and store macros

neekolas:
## Summary

1. I broke up the `impl_fetch` and `impl_store` macros to better support the `group_intent` table, and others where fetching and storing might be done against separate models. 
2. I moved the macro definitions into the `encrypted_store` module. Even if the macros actually get exposed at the root, it feels odd to have code that is specific to one datastore sitting in `lib.rs`. Especially since when we get to WASM support, we will likely need to exclude this code from compilation via feature.

## Notes

I'm sure I could have made things work with a single macro that supported more complex arguments to support something like the `group_intents` table. But macros are complicated enough on their own - might as well keep each one small and readable.

### DIFF
--- a/xmtp_mls/src/lib.rs
+++ b/xmtp_mls/src/lib.rs
@@ -36,55 +36,6 @@ pub trait Delete<Model> {
     fn delete(&mut self, key: Self::Key) -> Result<usize, StorageError>;
 }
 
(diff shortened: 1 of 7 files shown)


insipx (approved):
nice, this makes sense!

PR #309, opened 2023-11-03
`group_intents` table methods

neekolas:
## Summary

- Implements Store and Fetch for `group_intents` tables
- Adds enum support for tables
- Adds methods for intent state transitions with validation
- Adds methods for finding intents by group, state, or kind


xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
```diff
@@ -49,10 +49,10 @@ CREATE TABLE group_intents (
     -- INTENT_STATE,
     "state" INT NOT NULL,
     -- The hash of the encrypted, concrete, form of the message if it was published.
-    "payload_hash" BLOB,
+    "payload_hash" BLOB UNIQUE,
```
neekolas, line 52: This feels safe to be unique. We should never have more than one intent pointing to a single payload

neekolas:
It looks like something is broken with the wasm-pack installer, which is breaking our CI. We are following the instructions on their website pretty directly. Hoping they fix it soon. If not, we can figure out a new way to install wasm-pack.

xmtp_mls/src/storage/encrypted_store/group_intent.rs
```diff
@@ -1,25 +1,551 @@
...
+
+        if let Some(allowed_kinds) = allowed_kinds {
+            query = query.filter(dsl::kind.eq_any(allowed_kinds));
+        }
+
+        Ok(query.load::<StoredGroupIntent>(conn)?)
+    }
+
+    // Set the intent with the given ID to `Published` and set the payload hash. Optionally add `post_commit_data`
+    pub fn set_group_intent_published(
```
richardhuaaa, line 102: Why not just mutate the struct and call `save()` on it, rather than having these three methods below? Not convinced the validation is worth the increased code surface/lower flexibility for change, but it's up to you

richardhuaaa (approved):

xmtp_mls/src/storage/encrypted_store/group_intent.rs
neekolas: We are going to have to enforce the state transitions somewhere. If we enforced elsewhere and saved, we would have to re-fetch the row from the db first since the local state might be out of date (or always use in a complicated tx that includes validation). This does everything in one SQL statement, which feels tidy. Can be used in a tx with the read or separately.

insipx:
Was going to say, like the separation of the `Impl EncryptedMessageStore` into it's own file w/ the model

PR #310, opened 2023-11-03
Group Messages Table/Impl/Tests

insipx:


xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -231,6 +215,18 @@ macro_rules! impl_store {
     };
 }
 
+impl<T> Store<DbConnection> for Vec<T>
```
neekolas, line 218: This is slick

xmtp_mls/src/lib.rs
```diff
@@ -47,4 +47,54 @@ mod tests {
...
+    /// wrapper over assert! macros for Ok's
+    ///
+    /// Make sure something is Ok(_) without caring about return value.
+    /// assert_ok!(fun());
+    ///
+    /// Against an expected value, e.g Ok(true)
+    /// assert_ok!(fun(), true);
+    ///
+    /// or the message variant,
+    /// assert_ok!(fun(), Ok(_), "the storage is not ok");
```
neekolas, line 76: Neat

xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -43,6 +42,23 @@ impl StoredGroup {
...
+        &self,
+        conn: &mut DbConnection,
+        id: GroupId,
+        state: GroupMembershipState,
+    ) -> Result<StoredGroup, StorageError> {
+        use super::schema::groups::dsl;
+
+        diesel::update(dsl::groups.find(id.as_ref()))
+            .set(dsl::membership_state.eq(state))
+            .get_result::<StoredGroup>(conn)
```
neekolas, line 57: Do we actually need to return the value here? I've been working on the state machine, and it's all very much "pull from the DB when you need to know something" so there's no chance of out-of-date state in memory when doing concurrent work.

xmtp_mls/src/storage/encrypted_store/group_message.rs
```diff
@@ -18,12 +25,282 @@ pub struct StoredGroupMessage {
...
+        }
+
+        if let Some(kind) = kind {
+            query = query.filter(dsl::kind.eq(kind));
+        }
+        Ok(query.load::<StoredGroupMessage>(conn)?)
+    }
+
+    /// Get a particular group message
+    pub fn get_group_message<MessageId: AsRef<[u8]>>(
```
neekolas, line 102: How is this different from the `Find` trait?

xmtp_mls/src/storage/encrypted_store/group.rs
insipx: Yeah don't think so, I just tried to mirror what was in `xmtp` as far as these updates look like (lacking some in-code context on state machine overall atm, but hope to get up to speed on that this week with a ticket) -- I'm all for less complexity wherever possible

xmtp_mls/src/storage/encrypted_store/group_message.rs
insipx: First time I'm seeing `Find`, but it relates specifically to `diesel` ORM and the `find` method, tbh not deeply familiar with diesel ORM, but it seems at first glance the `FindDsl` trait would be most useful if you want to be generic over the key type. We could maybe experiment with refactoring the `Fetch` trait to make use of it somehow, and possibly remove the `Key` associated type. 

AsRef is mostly ergonomic/api -- it lets us pass a `&[u8]` or `Vec<u8>` without .to_vec() or constraining the type to something in that regard. It helps reduce cloning, and allows us to write

```rust
let id = vec![0x0];
// these are all valid
get_group_message(id)
get_group_message(id.as_slice());
get_group_message(&id);
```

It is a conversion trait, but meant to be v. cheap at-runtime

xmtp_mls/src/lib.rs
insipx: my goal here (as well as with `with_store` to make writing tests as ergonomic as possible. I did take some inspo from `tokio-test` and `frame-support` macros of the same name, but imo the more readable and easier it is to write tests the better

xmtp_mls/src/storage/encrypted_store/group_message.rs
neekolas: Sorry sorry sorry, I meant the `Fetch` trait. Although this is pretty slick and I'm now sold on the value of keeping it as-is.

neekolas (approved):

PR #313, opened 2023-11-03
312: introduce remote container development environment

### LINKED ISSUE #312: Feature request: add remote development container

Key benefits of using development containers in Visual Studio Code include:

1. **Consistency**: Development containers ensure that everyone on the team is using the same environment, reducing "works on my machine" issues.

2. **Isolation**: Containers provide isolation from the host system, preventing conflicts between different software versions.

3. **Reproducibility**: Containers can be versioned, making it easy to replicate the exact development environment in different stages of the project.

4. **Portability**: Development containers can be shared and run on different machines, making it easier to onboard new team members or work across multiple devices.

5. **Dependency Management**: Containers encapsulate dependencies, eliminating the need to install and manage them directly on the host system.

To use this development container in Visual Studio Code, specify the `Dockerfile` as the target for the Remote Containers module

Dockerfile is provided

_No response_

### Additional details

_No response_

jac18281828:
closes #312 

* introduces a development container with cargo and rust
* check code and rust fmt
* fix format issues currently committed on main
* tasks.json, GitHub workflow, support files etc.


jac18281828:
I discussed this change with @insipx to ensure the workflow is correct.   All checks and tests are passing.   All code changes address existing formatting issues because cargo fmt is currently not checked by the ci.   Check is now required as of this PR.

.github/workflows/ci-image.yml
```diff
@@ -0,0 +1,33 @@
+name: Build Dev Image CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - "*"
```
neekolas, line 10: Do we need this on every branch push?

jac18281828: Why not?   This is significantly faster than the existing ci integration suite.  Also doesn't the fast feedback help if you have a commit with issues/warnings/fmt problems?   Is there any harm to that we should consider?

jac18281828: As an aside, considering that we have existing code format issues committed on main, I suspect that running these checks more frequently is a good thing for the overall health of the code.

neekolas: One reason is that we seem to be pushing all images to the tag of LATEST, so commits on feature branches will override main.

Agree the checking formatting on all branches is a good thing, but we could easily add that into our existing lint CI workflow.

jac18281828:
Added Node 20.9.0 toolchain for yarn and npm

insipx:
I like the docker image + but adding FMT to the current workflow is better. As far as speed for 1 vs the other I'm probably not the one to know

In general I am for reproducible environments, so like the docker container and think it will make it easier to onboard devs in the future + have an easy way to get an environment up for example if we ever have a `CONTRIBUTING` section as the project grows. It also serves as a good enforced dependency list even if not using docker

.github/workflows/ci-image.yml
jac18281828: @neekolas  and I discussed offline and clarified that we both prefer to have fewer GitHub workflows.   I will take a look at the lint workflow and attempt to merge that into the containerized build.   

Going forward we should be able to achieve having all fmt/lint/unit testing in the containerized build and all integration tests (mls) in its own workflow.   The goal would be eventually there are about 2 workflows.     We won't achieve that in this PR, but that is a good target for the future.

neekolas (approved):

Dockerfile
```diff
@@ -0,0 +1,105 @@
...
+COPY --chown=xmtp:xmtp . .
+
+ENV PATH=~xmtp/.cargo/bin:$PATH
+ENV USER=xmtp
+
+RUN ~xmtp/.cargo/bin/cargo check
+RUN ~xmtp/.cargo/bin/cargo +nightly --version
+RUN ~xmtp/.cargo/bin/cargo +nightly fmt --check
+# some tests are setup as integration tests 👀 xmtp_mls
+RUN for crate in xmtp xmtp_api_grpc xmtp_api_grpc_gateway xmtp_cryptography xmtp_proto xmtp_v2; do cd ${crate}; ~xmtp/.cargo/bin/cargo test; done
```
neekolas, line 94: Is it normal practice to run the tests inside the Dockerfile, instead of with the code mounted as a volume? How does this work when I make a change to the code? It seems like every code change will go back to line 85 causing this step to rebuild all dependencies from scratch.

Dockerfile
```diff
@@ -0,0 +1,105 @@
...
+
+
+## Rust from builder
+COPY --chown=xmtp:xmtp --from=builder /home/xmtp/.cargo /home/xmtp/.cargo
+COPY --chown=xmtp:xmtp --from=builder /home/xmtp/.rustup /home/xmtp/.rustup
+
+USER xmtp
+
+# fmt coming from nightly
+RUN ~xmtp/.cargo/bin/rustup toolchain install nightly
```
neekolas, line 81: One thing that's funky here is that the Docker caching is going to lock you to whichever nightly version you first built this with. Subsequent builds will not re-download a newer nightly version.

If the goal is reproducibility, we should really have this locked to a particular stable version and make updates explicit.

richardhuaaa:
> To use this development container in Visual Studio Code, specify the Dockerfile as the target for the Remote Containers module

Can we include this in the README for future devs? And we mean the Dev Containers extension right?

And is it correct to say that this is optional? The alternative is to configure your environment yourself?

Dockerfile
jac18281828: Visual Studio Code handles the magic and mounts it as a volume.  On GitHub it acts as a containerized build.  The changes in code are reflected on your filesystem.

Dockerfile
jac18281828: It looks like @insipx created a ticket to remove the nightly.   I agree I think stable seems like the route to take.

insipx: In the fmt nightly ticket, i'll add to lock to a particular stable version. I think there should be a way with rustup to point it to the rust-toolchain file and install all the necessary components through that

insipx (approved):

PR #315, opened 2023-11-07
Publish intents flows

neekolas:
## Summary

- Scaffolds initial code for `SEND_MESSAGE` and `ADD_MEMBERS` intent publishing
- Removes usage of ephemeral storage and uses a tempfile instead for everything. This way we get the same concurrency as real applications. The ephemeral store was always weird, since it needed a pool size of 1. Will remove entirely in a subsequent PR.
- Scaffolds `MlsGroup` struct
- Group initialization, including saving to the database
- Adds some structs for serializing group intents to the database
- Fixes an issue overwriting entries in the SqlKeyStore

## Notes
- I am using Protobuf to serialize the intent data. It's not the nicest for ergonomics, but it's a set of tools that we already have setup and it does allow us to make breaking schema changes and handle past versions without migrations. Also more performant/compact than JSON.
- This needs a more robust set of e2e tests, but that's not so straightforward without the ability to properly read and parse messages from the network (which doesn't yet exist).


mls_validation_service/src/handlers.rs
```diff
@@ -64,13 +64,11 @@ impl ValidationApi for ValidationService {
                 match validate_group_message(message.group_message_bytes_tls_serialized) {
                     Ok(res) => ValidateGroupMessageValidationResponse {
                         group_id: res.group_id,
-                        epoch: res.epoch,
```
neekolas, line 67: Epoch was removed from the protos previously, but was still floating around here

xmtp_mls/src/builder.rs
```diff
@@ -136,7 +136,7 @@ where
                 parameter: "api_client",
             })?;
         let network = self.network;
-        let store = self.store.take().unwrap_or_default();
+        let store = self.store.take().unwrap();
```
neekolas, line 139: I'm not sure having a default here is really helpful outside of tests. If an actual caller isn't specifying a store their application is almost certainly misconfigured and panicking makes sense.

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -41,11 +33,11 @@ impl OpenMlsKeyStore for SqlKeyStore<'_> {
     ///
     /// Returns an error if storing fails.
     fn store<V: MlsEntity>(&self, k: &[u8], v: &V) -> Result<(), Self::Error> {
-        let entry = StoredKeyStoreEntry {
-            key_bytes: k.to_vec(),
-            value_bytes: db_serialize(v)?,
-        };
-        entry.store(&mut self.store.conn()?)?;
+        self.store.insert_or_update_key_store_entry(
```
neekolas, line 36: This does an INSERT OR REPLACE instead of just a plain INSERT, so that we can overwrite existing values. I ran into this when trying to save the updated group state.

xmtp_mls/src/builder.rs
insipx: As far as panicking goes, I think we should let the caller decide whether to panic or not and just stick to errors. We have more control over the error context and can return more helpful results than with panics. Additionally it allows any external crate or service using the library to recover if they choose to

xmtp_mls/src/utils/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod hash;
+#[cfg(test)]
```
neekolas, line 2: I added some global test utils (mostly just moving existing ones from the store), but they are only exported in `cfg(test)`

xmtp_mls/src/builder.rs
neekolas: If someone were to accidentally use the default (having the data stored in memory or in a tempfile), the result is total data loss.

insipx: I don't disagree that we can't continue, but a couple lines above this we return an error instead of panicking on missing an api_client, which is a similar case to this. i think we should choose a way here and be consistent throughout

At the very least, we should use `.expect` here to inform the client library what went wrong, and document it on the `ClientBuilder` fn that it panics under some circumstances

insipx: If we don't have good defaults for these, it might be better to fit them into the constructor, and then we avoid this issue altogether b/c it will be impossible to enter an invalid state with the builder

neekolas: That's a good point. I've updated to be consistent with above.

xmtp_mls/src/groups/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+            conn,
+            self.group_id.clone(),
+            Some(vec![IntentState::ToPublish]),
+            None,
+        )?;
+
+        for intent in intents {
+            // TODO: Wrap in a transaction once we can synchronize with the MLS Keystore
+            let result = self.get_publish_intent_data(&provider, &mut openmls_group, &intent);
+            match result {
```
insipx, line 150: This is OK, but could remove some indentation 

```rust
if let Err(e) = result {
    log::error!(/* ... */);
    continue;
}

let (payload, post_commit_data) = result.expect("Checked for error; qed");
self.client.api_client.publish_to_group(vec![payload.as_slice()]).await?;
// ...
```

What happens if one of the intents errors & isn't published? I'm assuming we don't bubble the error b/c we need to ensure all intents are processed no matter what

xmtp_mls/src/groups/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+};
+use openmls_traits::OpenMlsProvider;
+use thiserror::Error;
+use tls_codec::Serialize;
+use xmtp_proto::api_client::{XmtpApiClient, XmtpMlsClient};
+
+use self::intents::{AddMembersIntentData, IntentError, PostCommitAction};
+
+#[derive(Debug, Error)]
+pub enum GroupError {
```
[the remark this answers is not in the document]
neekolas: Afraid so. OpenMLS has a lot of different error types. I think the only one we have any real discretion over is `Generic`.

xmtp_mls/src/groups/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub fn new(group_id: Vec<u8>, client: &'c Client<ApiClient>) -> Self {
+        Self { client, group_id }
+    }
+
+    pub fn load_mls_group(
+        &self,
+        provider: &XmtpOpenMlsProvider,
+    ) -> Result<OpenMlsGroup, GroupError> {
+        let mls_group =
+            match OpenMlsGroup::load(&GroupId::from_slice(&self.group_id), provider.key_store()) {
```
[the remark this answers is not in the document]
neekolas: Thanks. That is nicer.

xmtp_mls/src/groups/mod.rs
neekolas: Great call-out. It's a bit of a tough situation. We don't want to have one bad intent completely brick your client because you can never move past it. On the other hand, we want the caller to be aware of the issue. I'm logging for now, but we probably should think about adding more state to keep track of the number of failed attempts or something.

xmtp_mls/src/groups/intents.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub(crate) fn to_bytes(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
+        PostCommitActionProto {
+            kind: Some(PostCommitActionKind::SendWelcomes(SendWelcomesProto {
+                installation_ids: self.installation_ids.clone(),
+                welcome_message: self.welcome_message.clone(),
+            })),
+        }
+        .encode(&mut buf)
+        .unwrap();
```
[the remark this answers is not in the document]
neekolas: In this case, we are dealing with data 100% created from inside the application and the generated proto code. It really should never fail. I can replace with an expect to give more context, but I don't think this is an error the caller should have to worry about.

xmtp_mls/src/groups/intents.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub fn new(key_packages: Vec<VerifiedKeyPackage>) -> Self {
+        Self { key_packages }
+    }
+
+    pub(crate) fn to_bytes(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
+        let key_package_bytes: Vec<Vec<u8>> = self
+            .key_packages
+            .iter()
+            .map(|kp| kp.inner.tls_serialize_detached().unwrap())
```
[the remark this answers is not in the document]
neekolas: Good catch. I'll update. This is data that comes over the network. It should be pre-validated, but it is possible for it to be malformed in some way that doesn't serialize.

insipx (approved):

insipx:
make sure to `cargo fmt` until we add that to CI

neekolas:
@insipx running that script gives me a bunch of errors. Our `rust-toolchain.toml` has us on the `stable` channel, which feels like the right channel for us. If some of the formatting rules require a nightly toolchain, I'd suggest we axe them.

```
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `format_code_in_doc_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 100`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
```

PR #318, opened 2023-11-08
Add ability to remove members

neekolas:
## Summary

- Adds support for the `RemoveMembers` intent data
- Adds some methods for removing members from a group
- Adds a test case for removing members, including the complicated case where there are multiple messages that result in a commit queued up to publish. Only the first should succeed while the rest will fail until the network has been synced.

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -3,8 +3,8 @@ mod intents;
 use intents::SendMessageIntentData;
 use openmls::{
     prelude::{
-        CredentialWithKey, CryptoConfig, GroupId, MlsGroup as OpenMlsGroup, MlsGroupConfig,
-        WireFormatPolicy,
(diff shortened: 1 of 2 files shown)


xmtp_mls/src/groups/mod.rs
```diff
@@ -223,6 +245,36 @@ where
...
+
+                if num_leaf_nodes != intent_data.installation_ids.len() {
+                    return Err(GroupError::Generic(format!(
+                        "expected {} leaf nodes, found {}",
+                        intent_data.installation_ids.len(),
+                        num_leaf_nodes
+                    )));
+                }
+
+                // The second return value is a Welcome, which is only possible if there
```
richardhuaaa, line 266: It might make sense to have a utility method that just removes all pending proposals, using [this](https://openmls.tech/openmls/doc/openmls/group/struct.MlsGroup.html#method.remove_pending_proposal)

richardhuaaa (approved):

xmtp_mls/src/groups/mod.rs
neekolas: 100% on board

PR #323, opened 2023-11-09
Implement post-commit actions

neekolas:
## Summary

- Adds handler for running post-commit actions. Currently the only action is sending welcome messages
- This will be run as part of the sync process, after messages have been read off the network

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -132,6 +132,8 @@ where
         intent.store(&mut conn)?;
 
         self.publish_intents(&mut conn).await?;
+        // ... sync state with the network
+        self.post_commit(&mut conn).await?;
 
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


xmtp_mls/src/groups/mod.rs
```diff
@@ -296,6 +299,39 @@ where
...
+                            .map(|installation_id| WelcomeMessage {
+                                installation_id,
+                                ciphertext: action.welcome_message.clone(),
+                            })
+                            .collect();
+                        self.client.api_client.publish_welcomes(welcomes).await?;
+                    }
+                }
+            }
+            let deleter: &mut dyn Delete<StoredGroupIntent, Key = i32> = conn;
```
neekolas, line 328: I had to do some gymnastics to make the Delete trait work now that we have multiple structs implementing it. Now that it requires disambiguation it's...not beautiful. Maybe there's some better way to tell the compiler which model we're deleting, or maybe we should rethink the trait to be explicitly generic.

insipx (approved):

PR #324, opened 2023-11-09
Add find_groups method

neekolas:
## Summary

- Adds methods for reading groups from the database and mapping the results into `MlsGroup` instances
- Stores the `created_at_ns` time on the group instance itself, so that callers can use their latest returned group as a starting time for future queries.

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -94,6 +94,25 @@ where
         Ok(group)
     }
 
+    pub fn find_groups(
+        &self,
+        allowed_states: Option<Vec<GroupMembershipState>>,
+        created_at_ns_gt: Option<i64>,
+        limit: Option<i64>,
+    ) -> Result<Vec<MlsGroup<ApiClient>>, ClientError> {
+        Ok(self
+            .store
+            .find_groups(
+                &mut self.store.conn()?,
+                allowed_states,
+                created_at_ns_gt,
+                limit,
+            )?
+            .into_iter()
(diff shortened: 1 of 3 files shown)


xmtp_mls/src/client.rs
```diff
@@ -94,6 +94,25 @@ where
         Ok(group)
     }
 
+    pub fn find_groups(
```
richardhuaaa, line 97: Can we put this directly on the `Group` struct, with `&Client` as an argument? Same with `create_group()`. Would prefer not to have every DB-related method replicated on the `Client` class

xmtp_mls/src/groups/mod.rs
```diff
@@ -57,6 +57,7 @@ pub enum GroupError {
 
 pub struct MlsGroup<'c, ApiClient> {
     pub group_id: Vec<u8>,
+    pub created_at_ns: i64,
```
richardhuaaa, line 60: Nice thinking here

richardhuaaa (approved):

xmtp_mls/src/client.rs
neekolas: I think in the FFI world it's going to make life easier to have all the group administration (`list`, `create`) live on the Client that we return to the native SDK. That way they don't have to deal with a bunch of different foreign object types and passing around references.

We could still do that mapping in the bindings, but it's messier. How about we leave it for now and we can always move things around later if we decide that it isn't as ergonomic?

PR #326, opened 2023-11-10
Receive flow for application messages

richardhuaaa:
This reads in application messages, decrypts them, and stores them in the database.

There's still lots of missing pieces here:
1. Sync from last processed payload, rather than syncing all payloads - working on adding the `topic_state` table separately
2. Handle concurrency of syncing
3. Handle payloads coming [from yourself](https://github.com/xmtp/libxmtp/pull/326#discussion_r1391858686)
4. Better unit tests - we can't decrypt payloads coming from ourselves, so we need to support adding members first
5. This only reads payloads for a group and doesn't return anything - we still need a higher-level method to list all messages in a group by reading from the DB, we also need a higher-level method to read payloads for all groups.
6. [Distinguishing](https://github.com/xmtp/libxmtp/pull/326#discussion_r1391907242) between retriable and non-retriable errors

There's plenty of space left here also to add in the logic for other message types


xmtp_mls/src/utils/topic.rs
```diff
@@ -1,5 +1,11 @@
 pub fn get_group_topic(group_id: &Vec<u8>) -> String {
-    format!("/xmtp/3/g-{}/proto", hex::encode(group_id))
+    format!("/xmtp/3/g-{}/proto", serialize_group_id(group_id))
+}
+
+pub fn serialize_group_id(group_id: &[u8]) -> String {
+    // TODO: I wonder if we really want to be base64 encoding this or if we can treat it as a
```
neekolas, line 6: The nodes treat it as a string, so it's going to have to get serialized somewhere.

xmtp_mls/src/groups/mod.rs
```diff
@@ -132,6 +150,132 @@ where
...
+        &self,
+        openmls_group: &mut OpenMlsGroup,
+        decrypted_message: &ProcessedMessage,
+        envelope_timestamp_ns: u64,
+    ) -> Result<(String, Vec<u8>), MessageProcessingError> {
+        let mut sender_account_address = None;
+        let mut sender_installation_id = None;
+        if let Sender::Member(leaf_node_index) = decrypted_message.sender() {
+            if let Some(member) = openmls_group.member_at(*leaf_node_index) {
+                if member.credential == *decrypted_message.credential() {
```
neekolas, line 175: Credential should have `PartialEq` implemented on it, so you can just use `member.credential.eq(decrypte_message.credential())` without dereferencing.

xmtp_mls/src/identity.rs
```diff
@@ -1,5 +1,3 @@
-use std::default;
```
neekolas, line 1: Thanks. Don't know how I missed that lint warning in my PR

xmtp_mls/src/utils/topic.rs
richardhuaaa: I copied this comment from elsewhere - I think it's musing over whether base64 would be more efficient than hex

xmtp_mls/src/groups/mod.rs
```diff
@@ -397,6 +536,21 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_receive_self_message() {
+        let wallet = generate_local_wallet();
+        let client = ClientBuilder::new_test_client(wallet.into()).await;
+        let group = client.create_group().expect("create group");
+        group.send_message(b"hello").await.expect("send message");
+
+        let result = group.receive().await;
+        if let GroupError::ReceiveError(errors) = result.err().unwrap() {
+            assert_eq!(errors.len(), 1);
```
richardhuaaa, line 548: The OpenMLS manual states: `To guarantee the best possible Forward Secrecy, the key material used to encrypt messages is immediately discarded after encryption. This means that the message author cannot decrypt application messages. If access to the messages content is required after creating the message, a copy of the plaintext message should be kept by the application.`

So I think we'll need to do a few things in the commit flow:
1. Save the plaintext message when sending
2. Advance the last message timestamp in the topics table, to indicate the payload has already been consumed. I think we need to correlate this based on the hash of the encrypted bytes - I didn't see any way to detect the sender without decrypting the message, which we can't do here

Saving these for later PR's.

neekolas: The plaintext and the message hash are already saved in the intent to tee this flow up

xmtp_mls/src/groups/mod.rs
```diff
@@ -1,30 +1,37 @@
 mod intents;
 
+#[cfg(test)]
+use std::println as debug;
```
neekolas, line 4: There's also this crate, which allows for log statements to get printed in tests https://crates.io/crates/test-env-log

xmtp_mls/src/groups/mod.rs
neekolas: For application messages, we can just use the plaintext saved in the intent.

For commits, we just need to `merge_pending_commit`. Technically, there is some ambiguity around which pending commit is actually being merged since the function doesn't take any arguments. We might want some more metadata on the intent to ensure the pending commit in the mls store matches the intent. In practice it should be fine since we can only have one pending commit at a time.

As long as we properly clear out failed intents it'll always be committing the right one, but it does feel a bit brittle.

xmtp_mls/src/groups/mod.rs
```diff
@@ -132,6 +155,122 @@ where
...
+                let mls_message_in = MlsMessageIn::tls_deserialize_exact(&envelope.message)?;
+
+                match mls_message_in.extract() {
+                    MlsMessageInBody::PrivateMessage(message) => self.process_private_message(
+                        &mut openmls_group,
+                        &provider,
+                        message,
+                        envelope.timestamp_ns,
+                    ),
+                    _ => panic!("Unsupported message type"),
```
neekolas, line 246: Should just return an error here and abort

xmtp_mls/src/groups/mod.rs
```diff
@@ -132,6 +155,122 @@ where
...
+        match decrypted_message.into_content() {
+            ProcessedMessageContent::ApplicationMessage(application_message) => {
+                let message_bytes = application_message.into_bytes();
+                let message_id =
+                    get_message_id(&message_bytes, &self.group_id, envelope_timestamp_ns);
+                let message = StoredGroupMessage {
+                    id: message_id,
+                    group_id: self.group_id.clone(),
+                    decrypted_message_bytes: message_bytes,
+                    sent_at_ns: envelope_timestamp_ns as i64, // TODO validate that this timestamp is controlled by server
```
neekolas, line 211: Yes. These are database timestamps

neekolas (approved):

xmtp_mls/src/groups/mod.rs
richardhuaaa: I figure it might be better to loudly fail, seeing as this shouldn't ever happen? Might be good to know if we have any unexpected message types on the channel

richardhuaaa: Ah - if we leave this as-is then anyone can DOS a group by leaving an unsupported message type. Will change this

xmtp_mls/src/groups/mod.rs
richardhuaaa: So adding this (or more specifically the updated `test-log` crate) seems to cause half of the `encrypted_store` tests to fail - it looks like there's a conflict in how the logger is set up with one of our other dependencies. Might just skip the hassle for now

neekolas: Fine by me

xmtp_mls/src/groups/mod.rs
```diff
@@ -132,6 +158,124 @@ where
...
+        Ok(())
+    }
+
+    pub fn process_messages(&self, envelopes: Vec<Envelope>) -> Result<(), GroupError> {
+        let provider = self.client.mls_provider();
+        let mut openmls_group = self.load_mls_group(&provider)?;
+        let receive_errors: Vec<MessageProcessingError> = envelopes
+            .into_iter()
+            .map(|envelope| -> Result<(), MessageProcessingError> {
+                let mls_message_in = MlsMessageIn::tls_deserialize_exact(&envelope.message)?;
```
neekolas, line 240: The thing that gets messy here is error handling. There are some errors we want to just log and move on from (garbage message, conflicting epochs/race conditions).

But there are some errors that we need to retry. For example, any sort of IOError saving to the DB or getting a connection from the pool. If Client A processes a commit successfully and Client B fails on the same commit, their group state will fall out of sync and Client B will be unable to read future messages.

One way to solve this would be to have some sort of `IsRetryable` method available on `GroupError`s. Retryable errors would abort the `process_messages` flow entirely, since we don't want to apply any future commits or increment the `last_message_timestamp` until we've fully processed the preceding message. The caller could then retry later. We'd just have to be sure that retryable errors can eventually either succeed or turn into non-retryable errors (maybe by keeping track of the number of attempts). Otherwise it could be a never-ending retry loop.

Open to other solutions.

It also might be valid to completely ignore this edge case and just have really good mechanisms for re-adding yourself to the group once the client gets into this bad state, assuming the bad state is reliably detectable.

richardhuaaa: This is a great point - I added this to the list of things to follow up on above

neekolas: I have a similar issue to figure out for the message publishing flow as well.

xmtp_mls/src/groups/mod.rs
neekolas: Agreed. We are going to need optimistic sending. Especially if we don't intend to fully sync after every message publish, which I think makes the most sense.

PR #332, opened 2023-11-14
Read welcomes from the network

neekolas:
## Summary

I've blocked out the flow for reading welcomes and persisting them as groups. There are still a bunch of `TODO`s in the code, which'll get resolved after some dependent work gets wrapped up. They include
- Reading and writing from the `last_message_timestamp_ns` in the DB
- Proper separation of retryable errors
- Wrapping operations in a DB transaction

## Notes

I also fixed the `devcontainer.json` so that integration tests can run from inside the devcontainer and connect to the local node defined in the docker compose file.


.devcontainer/devcontainer.json
```diff
@@ -1,26 +1,31 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
 // README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-dockerfile
 {
-	"name": "Existing Dockerfile",
```
neekolas, line 4: Not sure whose formatter is correct here for .json files

insipx: we should add a some kindof config file for JSON tied to a preferred formatter to avoid this maybe

insipx (approved):

.devcontainer/devcontainer.json
neekolas: Ya. I'll make a separate PR to enforce the use of prettier on the repo. We already use it for all the JS projects.

PR #333, opened 2023-11-14
feat: introduce bindings_flutter crate

dmccartney:
This introduces a new `bindings_flutter` crate to be consumed by the `xmtp-flutter` SDK.
See https://github.com/xmtp/xmtp-flutter/pull/94


bindings_flutter/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 XMTP
```
nplasterer, line 3: Might as well hit that with a 2024 at this point
```suggestion
Copyright (c) 2024 XMTP
```

nplasterer (approved):
Hard to really get to the bottom of the important bits here. Most of the files are just Android and iOS example setup stuff though. But seems like you tested well and it's all contained to this new `bindings_flutter` folder.

PR #334, opened 2023-11-14
Handling of own messages

neekolas:
## Summary

- Adds support for matching intents to incoming messages in the receive flow
- Updates the intent state depending on whether the message matches the group epoch
- Adds a `sync` function to encapsulate publishing, receiving, and post-commit flows
- Refactors message processing to branch out to the new flow if an intent is found

## Notes
There are a few error cases that are hard to handle

1. Retryable vs non-retryable errors. When to abort
2. What happens if an intent can never be fully fulfilled? We probably need an `error` intent status for intents that are impossible (trying to add a member that doesn't exist, trying to add a member already present in the group)
3. Application messages from the wrong epoch. These may actually be fine depending on the `max_past_epochs` config

## TODO
- Add support for writing a transcript of members added/removed
- Better mapping between commits and payloads. Should include a reference to the staged commit hash or something to make sure that when `merge_pending_commit` gets called it is actually being called on the correct commit
- More tests around failure modes and recovery

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -7,6 +7,8 @@ use intents::SendMessageIntentData;
 #[cfg(not(test))]
 use log::debug;
 use openmls::{
+    framing::ProtocolMessage,
+    group::GroupEpoch,
     prelude::{
         CredentialWithKey, CryptoConfig, GroupId, LeafNodeIndex, MlsGroup as OpenMlsGroup,
         MlsGroupConfig, MlsMessageIn, MlsMessageInBody, PrivateMessageIn, ProcessedMessage,
@@ -76,6 +78,15 @@ pub enum MessageProcessingError {
     },
     #[error("openmls process message error: {0}")]
     OpenMlsProcessMessage(#[from] openmls::prelude::ProcessMessageError),
+    #[error("merge pending commit: {0}")]
+    MergePendingCommit(#[from] openmls::group::MergePendingCommitError<StorageError>),
+    #[error("wrong epoch. expected {group_epoch:?} and got {message_epoch:?}")]
+    WrongEpoch {
+        message_epoch: GroupEpoch,
+        group_epoch: GroupEpoch,
+    },
+    #[error("intent error: {0}")]
+    Intent(#[from] IntentError),
     #[error("storage error: {0}")]
     Storage(#[from] crate::storage::StorageError),
     #[error("tls deserialization: {0}")]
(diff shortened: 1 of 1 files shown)


xmtp_mls/src/groups/mod.rs
```diff
@@ -236,40 +329,72 @@ where
...
                 // intentionally left blank.
             }
             ProcessedMessageContent::ExternalJoinProposalMessage(_external_proposal_ptr) => {
                 // intentionally left blank.
             }
-            ProcessedMessageContent::StagedCommitMessage(_commit_ptr) => {
-                // intentionally left blank.
+            ProcessedMessageContent::StagedCommitMessage(staged_commit) => {
+                println!("[{}] received staged commit", self.client.account_address());
+                openmls_group.merge_staged_commit(provider, *staged_commit)?;
```
neekolas, line 342: I added this line to make my test case handling conflicting messages produce the right errors

xmtp_mls/src/groups/mod.rs
```diff
@@ -211,13 +224,93 @@ where
...
+            return Err(MessageProcessingError::WrongEpoch {
+                message_epoch,
+                group_epoch,
+            });
+        }
+
+        match intent.kind {
+            IntentKind::AddMembers | IntentKind::RemoveMembers | IntentKind::KeyUpdate => {
+                debug!("[{}] merging pending commit", self.client.account_address());
+                openmls_group.merge_pending_commit(provider)?;
```
richardhuaaa, line 274: Some off-the-cuff thoughts here, not necessarily for this PR immediately:

1. [merge_staged_commit](https://openmls.tech/openmls/doc/openmls/group/struct.MlsGroup.html#method.merge_staged_commit) clears any pending commits - we probably actually need to handle requeuing any pending intents/commits on the receive flow when merging other people's commits. 
2. Because of that, I'm not sure if the `group_epoch != message_epoch` above is necessary or not
3. No rush, but I think we need to handle at least some of the errors on this line in the same way you're handling the `group_epoch != message_epoch` case - in fact maybe that check is done inside this method

richardhuaaa (approved):

xmtp_mls/src/groups/mod.rs
```diff
@@ -297,61 +422,64 @@ where
...
+        let conn = &mut self.client.store.conn()?;
         let intent_data: Vec<u8> = SendMessageIntentData::new(message.to_vec()).into();
         let intent =
             NewGroupIntent::new(IntentKind::SendMessage, self.group_id.clone(), intent_data);
-        intent.store(&mut conn)?;
+        intent.store(conn)?;
 
-        self.publish_intents(&mut conn).await?;
+        // Skipping a full sync here and instead just firing and forgetting
+        self.publish_intents(conn).await?;
```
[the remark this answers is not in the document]
neekolas: That's right. Especially with `max_past_epochs` we should be able to assume that sending a message will "just work" in almost all cases and not need to be re-queued. There are some edge cases where your local state is way out of sync with the network where it would need a retry, but they should be rare enough I think the trade-off is worth it.

xmtp_mls/src/groups/mod.rs
neekolas: 1. You're totally right. We should also be explicitly checking that there is a pending commit to be merged here and erroring if there isn't. Will fix.
2. I'm down to move this logic into an error branch of `merge_pending_commit`. Does feel cleaner to just let openmls tell us when there is a problem instead of trying to predict it.

neekolas: OK. Changed the logic up to not check for epoch matches and instead handle errors or `merge_pending_commit` and also explicitly return an error if there is no pending commit.

I looked at the code and tested it out and was a little surprised to find that `merge_pending_commit` just returns `OK(())` when there is no pending commit. That seems like an odd choice, but it's how OpenMLS was designed.

```rust
    pub fn merge_pending_commit<KeyStore: OpenMlsKeyStore>(
        &mut self,
        provider: &impl OpenMlsProvider<KeyStoreProvider = KeyStore>,
    ) -> Result<(), MergePendingCommitError<KeyStore::Error>> {
        match &self.group_state {
            MlsGroupState::PendingCommit(_) => {
                let old_state = mem::replace(&mut self.group_state, MlsGroupState::Operational);
                if let MlsGroupState::PendingCommit(pending_commit_state) = old_state {
                    self.merge_staged_commit(provider, (*pending_commit_state).into())?;
                }
                Ok(())
            }
            MlsGroupState::Inactive => Err(MlsGroupStateError::UseAfterEviction)?,
            MlsGroupState::Operational => Ok(()),
        }
    }
```

PR #335, opened 2023-11-14
Topic syncing and concurrency

richardhuaaa:
1. Only pull down payloads we haven't already processed (via `pull_from_topic`)
2. Do all payload processing within an atomic transaction that also updates the `last_synced_payload_time` (via `process_for_topic`)

These methods are agnostic to which topic, so can be used for both welcomes and group message processing, as well as any other case in the future. Unsure if they belong on `Client` or somewhere else.

Some missing pieces:
1. More complicated concurrency test cases
1. Sophisticated error handling (when an error happens, do we update `last_synced_payload_time` or do we not?)
1. Will add a follow-up PR to improve the [safety](https://github.com/xmtp/libxmtp/pull/335#discussion_r1396469800) of borrowing the DB connection in a transaction

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -1,5 +1,6 @@
 use std::collections::HashSet;
 
+use log::debug;
 use openmls::{
     framing::{MlsMessageIn, MlsMessageInBody},
     messages::Welcome,
@@ -179,10 +180,19 @@ where
             .store
             .get_last_synced_timestamp_for_topic(&mut conn, topic)?;
 
-        Ok(self
+        let envelopes = self
             .api_client
             .read_topic(topic, last_synced_timestamp_ns as u64)
-            .await?)
+            .await?;
+
+        debug!(
+            "Pulled {} envelopes from topic {} starting at timestamp {}",
+            envelopes.len(),
+            topic,
+            last_synced_timestamp_ns
+        );
+
+        Ok(envelopes)
     }
 
     // Get a flat list of one key package per installation for all the wallet addresses provided.
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -216,7 +216,7 @@ where
 
         let envelopes = self
             .api_client
-            .read_topic(topic, last_synced_timestamp_ns as u64)
+            .read_topic(topic, last_synced_timestamp_ns as u64 + 1)
             .await?;
 
         debug!(
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -1,6 +1,5 @@
 use std::{collections::HashSet, mem::Discriminant};
 
-use diesel::Connection;
 use log::debug;
 use openmls::{
     framing::{MlsMessageIn, MlsMessageInBody},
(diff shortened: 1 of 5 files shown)


xmtp_mls/src/client.rs
```diff
@@ -165,9 +209,58 @@ where
...
+        envelope_timestamp_ns: u64,
+        process_envelope: ProcessingFn,
+    ) -> Result<ReturnValue, MessageProcessingError>
+    where
+        ProcessingFn: FnOnce(XmtpOpenMlsProvider) -> Result<ReturnValue, MessageProcessingError>,
+    {
+        // TODO: We can handle errors in the transaction() function to make error handling
+        // cleaner. Retryable errors can possibly be part of their own enum
+        XmtpOpenMlsProvider::transaction(&mut self.store.conn()?, |provider| {
+            let is_updated = {
```
richardhuaaa, line 244: Wrapping this in a block, because the borrow needs to be ended before it's borrowed again somewhere else (e.g. when OpenMLS writes to the key store), or else there will be a runtime error. Don't love this solution but it works

richardhuaaa: After discussion, will implement a `WrappedConnection` struct in the next PR that can solve this problem

xmtp_mls/src/groups/mod.rs
```diff
@@ -387,38 +377,36 @@ where
         }
     }
 
-    pub fn process_messages(&self, envelopes: Vec<Envelope>) -> Result<(), GroupError> {
+    pub(crate) fn process_messages(&self, envelopes: Vec<Envelope>) -> Result<(), GroupError> {
         let mut conn = self.client.store.conn()?;
         let provider = self.client.mls_provider(&mut conn);
         let mut openmls_group = self.load_mls_group(&provider)?;
```
neekolas, line 383: Could merge this with the two lines above now that we don't need access to the provider or conn below this point

xmtp_mls/src/client.rs
```diff
@@ -165,9 +209,58 @@ where
...
     }
 
+    pub(crate) async fn pull_from_topic(&self, topic: &str) -> Result<Vec<Envelope>, ClientError> {
+        let mut conn = self.store.conn()?;
+        let last_synced_timestamp_ns =
+            EncryptedMessageStore::get_last_synced_timestamp_for_topic(&mut conn, topic)?;
+
+        let envelopes = self
+            .api_client
+            .read_topic(topic, last_synced_timestamp_ns as u64)
```
neekolas, line 219: The network query API is "greater than or equals", so I think we actually want this to be `(last_synced_timestamp_ns + 1)` to avoid getting duplicate messages we will end up skipping

neekolas (approved):
I was using this branch in the demo this morning and everything was working.

xmtp_mls/src/groups/mod.rs
richardhuaaa: I'm not sure what you mean by merging - like put it all in the same line? I think that just makes it less readable :/

xmtp_mls/src/client.rs
richardhuaaa: Changing this, but wouldn't it make more sense to have the network query API just be "greater than"? That seems like a much cleaner system overall

neekolas: I definitely wish I had done it that way when I first wrote the query API last year. The rationale was compatibility with Waku, which worked that way. Since we are using the generic query interface, we'd have to make the change for everyone (maybe with a flag). But yes, that would be better.

xmtp_mls/src/groups/mod.rs
neekolas: It is less readable, but it does return the connection to the pool sooner. 

Not a big deal one way or the other.

Issue #336, opened 2023-11-16
list_users() - (listing participants at the user level)


neekolas:

neekolas:
I'm gonna take a stab at this, since I want it for the CLI

PR #339, opened 2023-11-17
Group members listing

neekolas:
## Summary

- Loads all members from the group tree and condenses into a single entry per wallet address
- Returns the wallet_address and a list of `installation_ids` for each member

## Resolves
https://github.com/xmtp/libxmtp/issues/336

### DIFF
--- a/xmtp_mls/src/groups/members.rs
+++ b/xmtp_mls/src/groups/members.rs
@@ -1,0 +1,83 @@
+use std::collections::HashMap;
+
+use xmtp_proto::api_client::{XmtpApiClient, XmtpMlsClient};
+
+use crate::identity::Identity;
+
+use super::{GroupError, MlsGroup};
+
+#[derive(Debug, Clone)]
+pub struct GroupMember {
+    pub wallet_address: String,
+    pub installation_ids: Vec<Vec<u8>>,
+}
+
+impl<'c, ApiClient> MlsGroup<'c, ApiClient>
+where
+    ApiClient: XmtpApiClient + XmtpMlsClient,
+{
+    // Load the member list for the group, merging together multiple installations into a single entry
+    pub fn members(&self) -> Result<Vec<GroupMember>, GroupError> {
+        let openmls_group =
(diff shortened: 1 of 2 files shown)


insipx (approved):
nice

xmtp_mls/src/groups/members.rs
```diff
@@ -0,0 +1,90 @@
...
+                    &member.signature_key,
+                )
+                .ok()
+                .map(|wallet_address| (wallet_address, member.signature_key.clone()))
+            })
+            .fold(
+                HashMap::new(),
+                |mut acc, (wallet_address, signature_key)| {
+                    acc.entry(wallet_address.clone())
+                        .and_modify(|e| e.installation_ids.push(signature_key.clone()))
```
insipx, line 38: didn't know about this method!

neekolas: I finally found a Rust thing @insipx doesn't know about! Gonna go call my mom

xmtp_mls/src/groups/members.rs
```diff
@@ -0,0 +1,90 @@
...
+            self.load_mls_group(&self.client.mls_provider(&mut self.client.store.conn()?))?;
+
+        let member_map: HashMap<String, GroupMember> = openmls_group
+            .members()
+            .filter_map(|member| {
+                Identity::get_validated_account_address(
+                    member.credential.identity(),
+                    &member.signature_key,
+                )
+                .ok()
```
richardhuaaa, line 31: Any reason we are filtering this? Doesn't this effectively make illegitimate group members invisible? Wondering if we should at least log something

neekolas: I can add some logging. I would expect we'll want to do the checks for legitimacy before actually accepting the commit to let them into the group, so illegitimate users should be impossible. But it would be good to have some visibility if we somehow screw that up.

richardhuaaa: Or like, just fail the group completely, seeing as it has been compromised. Your call

Issue #316, opened 2023-11-07
Figure out retry logic for errors in publishing


neekolas:
In some cases publish errors might be retryable, in which case we should bail out of the publishing flow and start from the beginning. In others, we probably want to move the intent to some sort of error state and never retry. The error should be surfaced to the originator of the intent somehow.

PR #341, opened 2023-11-20
Introduce retry logic strategy

insipx:
Introduces a `std::error::Error` supertrait, `RetryableError` which can be implemented on our error enums to specify a retryable error. All errors by default return false on invocation of this method.

introduces `retry!` and `retry_async!` which accept a retry strategy, `Retry::default()` and a closure to house the function to retry if it fails:

```rust
retry!(Retry::default(), || -> Result<(), GroupError> {
    my_fallable_function()?;
});
```

A macro was chosen to take advantage of [autoref specialization](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md), allowing us to make all errors false by default with the blanket `RetryableError` implementation on anything that implements the standard library Error trait. This strategy lets us pinpoint errors that should be retryable through implementing RetryableError, and can be extended into a proc-macro interface similar to `thiserror` in the future.

There are not any specific retry strategies like Exponential Backoff, but they can be supported in further PR's.

This should cover the first part of #316,
> In some cases publish errors might be retryable, in which case we should bail out of the publishing flow and start from the beginning. 

This also works with api_client calls,

```rust
            retry_async!(
                Retry::default(),
                (|| async {
                    self.api_client
                        .publish_to_group(PublishToGroupRequest { messages: to_send })
                        .await
                })
            )?;
```


xmtp_mls/src/groups/mod.rs
```diff
@@ -65,14 +68,26 @@ pub enum GroupError {
...
+    ReceiveErrors(Vec<MessageProcessingError>),
     #[error("generic: {0}")]
     Generic(String),
     #[error("diesel error {0}")]
     Diesel(#[from] diesel::result::Error),
 }
 
+impl crate::retry::RetryableError for GroupError {
+    fn is_retryable(&self) -> bool {
+        match self {
```
neekolas, line 83: This is a reasonable starting place. We'll have to do some diving into the OpenMLS error types to figure out which ones can be added to the list as a follow-on. For example, some of their internal error types wrap keystore errors, which are actually database errors and should be retried.

xmtp_mls/src/groups/mod.rs
```diff
@@ -378,37 +393,53 @@ where
...
             .map(|envelope| -> Result<(), MessageProcessingError> {
-                self.client.process_for_topic(
-                    &self.topic(),
-                    envelope.timestamp_ns,
-                    |provider| -> Result<(), MessageProcessingError> {
-                        self.process_message(&mut openmls_group, &provider, &envelope)?;
-                        openmls_group.save(provider.key_store())?;
-                        Ok(())
-                    },
+                retry!(
```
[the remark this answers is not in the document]
neekolas: We might be able to force a database error by exhausting the pool (opening N connections and not letting them go).

xmtp_mls/src/groups/mod.rs
```diff
@@ -518,10 +552,15 @@ where
             let (payload, post_commit_data) = result.expect("result already checked");
             let payload_slice = payload.as_slice();
 
-            self.client
-                .api_client
-                .publish_to_group(vec![payload_slice])
-                .await?;
+            retry_async!(
```
neekolas, line 555: WDYT about moving the retry logic into the `ApiClientWrapper` so that all callers don't have to think about this

insipx: That makes sense,

Actually, should we be retrying api client errors? There's an issue on the board talking about idempotency in the API

xmtp_mls/src/groups/mod.rs
insipx: Yeah, I wasn't sure which errors to mark here, i'll do some diggin and update

xmtp_mls/src/groups/mod.rs
neekolas: I'm fine with building it in for now and assume the idempotency is going to be there. If it starts making test cases unreliable we can always disable or set the number of retries to 0

neekolas: Also, it's only publish to group and publish welcomes that are not idempotent. Registering identities is, and all the query APIs are obviously safe.

neekolas (approved):
This looks slick. Like the approach, and how easy it will be for us to extend it for more cases that need retries.

PR #342, opened 2023-11-21
Lint warning cleanup

neekolas:
## Summary

- With the addition of the .devcontainer, we lost the annotations on PRs for lint warnings. Errors are caught by the Docker build, but warnings get lost in the logs of that Docker build and go nowhere. This PR fixes some of those warnings and re-adds the lint Github workflow to get the annotations back.
- I fixed a few warnings manually and the rest were fixed with `cargo clippy --fix`

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -25,9 +25,3 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           name: ffi
           args: --all-features --no-deps --manifest-path bindings_ffi/Cargo.toml
-      - uses: actions-rs/clippy-check@v1
-        name: Lint api_grpc_gateway
-        with:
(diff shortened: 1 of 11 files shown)


jac18281828 (commented):
Thanks for cleaning up those warnings.  That is super helpful!   Does it make sense to add the Link workflow to either build or test existing workflow.

It might make sense to have a PR on the lint workflow and one on the cleanup as I think the cleanup can just get merged immediately.

neekolas:
> Does it make sense to add the Link workflow to either build or test existing workflow.

The thing that I like about having lint as a separate workflow is that it just executes really really fast, since it doesn't have to build the entire app. We could add it to the beginning of the test workflow, but it still probably adds 5m to the time until you see a ✅

jac18281828 (approved):
Thanks for cleaning this up!

insipx (approved):

PR #343, opened 2023-11-21
Add ability to remove members

neekolas:
## Summary

- Adds feature for removing group members by account address
- Adds a test for this feature
- Adds member removal to the CLI

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -454,7 +454,7 @@ where
         self.sync_with_conn(conn).await
     }
 
-    pub async fn remove_members_by_installation_id(
+    pub(crate) async fn remove_members_by_installation_id(
         &self,
         installation_ids: Vec<Vec<u8>>,
     ) -> Result<(), GroupError> {
@@ -468,6 +468,20 @@ where
         intent.store(conn)?;
 
         self.sync_with_conn(conn).await
+    }
+
+    pub async fn remove_members(&self, wallet_addresses: Vec<String>) -> Result<(), GroupError> {
+        let installation_ids = self
+            .members()?
+            .into_iter()
+            .filter(|member| wallet_addresses.contains(&member.wallet_address))
+            .fold(vec![], |mut acc, member| {
+                acc.extend(member.installation_ids);
+                acc
+            });
+
+        self.remove_members_by_installation_id(installation_ids)
+            .await
     }
 
     pub async fn key_update(&self) -> Result<(), GroupError> {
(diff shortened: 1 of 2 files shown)


insipx (approved):

xmtp_mls/src/groups/mod.rs
```diff
@@ -470,6 +470,20 @@ where
         self.sync_with_conn(conn).await
     }
 
+    pub async fn remove_members(&self, wallet_addresses: Vec<String>) -> Result<(), GroupError> {
+        let installation_ids = self
+            .members()?
+            .into_iter()
+            .filter(|member| wallet_addresses.contains(&member.wallet_address))
```
richardhuaaa, line 477: @neekolas, we're constructing the list of installations here, and only later constructing the commit payload to remove them as part of the intent publishing flow.

We're relying on an epoch mismatch on the commit payload to detect if the list of installations in the conversations has changed. Is it possible for the epoch to have been updated between us constructing the list of installations, and the commit payload being constructed? Could we design our methods in a way such that this is not possible (e.g. store the wallet address on the intent rather than the installation ids)?

neekolas: I was mostly going for symmetry with the add members flow. In the case of adding members, I like that we only download the key packages once so that retries don't have to download a new set of KPs (which may introduce new surface area for bugs and weird behaviour).

But I guess in the case of removing members none of that matters. I'm open to just including the wallet address on the intent and deferring the work finding matching group members.

richardhuaaa: Yes, the risk here is that we don't exhaustively remove the installations for a member, and any remaining installation implies that the member is still a part of the group, causing the rest of their installations to be re-added invisibly. Especially if we insert a transcript of the removal, it may cause people to send messages to the removed member that they didn't expect to.

> we only download the key packages once so that retries don't have to download a new set of KPs (which may introduce new surface area for bugs and weird behaviour)

What kind of bugs/weird behavior would occur here? Was wondering whether it is also safest to operate off the latest data when adding participants. If we are using key packages with expiries rather than one-time key packages, it might also help us avoid using expired key packages.

neekolas: Mostly just indeterminacy. Requests to get key packages could fail, maybe new key packages were added and are bad. Probably not the end of the world. It's less bad now that we have long-lived key packages as opposed to one-time KPs that would slowly get eaten away at through retries.

I could be on board with switching both to just operate with wallet addresses.

PR #352, opened 2023-11-28
Add codecs support and one new codec

neekolas:
## Summary

- Copies over the `codecs` folder from the `xmtp` crate verbatim
- Generates the latest protos, which involved a few field renames
- Adds simple codec for `GroupMembershipChange`


xmtp_mls/src/groups/intents.rs
```diff
@@ -101,9 +100,9 @@ impl AddMembersIntentData {
         data: &[u8],
         provider: &XmtpOpenMlsProvider,
     ) -> Result<Self, IntentError> {
-        let msg = AddMembersPublishData::decode(data)?;
+        let msg = AddMembersData::decode(data)?;
```
neekolas, line 103: I had renamed this message in the proto repo but never updated in `libxmtp`, so I got forced into doing it now.

richardhuaaa (approved):

PR #353, opened 2023-11-30
Add detail and address feedback on identity doc

richardhuaaa:
Preview the doc here: https://github.com/xmtp/libxmtp/blob/rich/identity-doc-updates/xmtp_mls/IDENTITY.md

Adding more detail that was requested during review, as well as addressing feedback.

- Add context on ethereum wallets
- Add detail on signature formats and signature labelling
- Add installation_public_key to credentials
- Clean up time representation
- Add detail on credential validation and installation synchronization
- Add detail on revocation
- Add account-level membership description

### DIFF
--- a/xmtp_mls/IDENTITY.md
+++ b/xmtp_mls/IDENTITY.md
@@ -40,7 +40,7 @@ XMTP: <Label>\n\n
 
 | Label                   | Described in section                                    |
 | ----------------------- | ------------------------------------------------------- |
-| Grant messaging access  | [Installation registration](#installation-provisioning) |
+| Grant messaging access  | [Installation registration](#installation-registration) |
 | Revoke messaging access | [Installation revocation](#installation-revocation)     |
 
 ### Installation registration
--- a/xmtp_mls/IDENTITY.md
+++ b/xmtp_mls/IDENTITY.md
@@ -24,9 +24,11 @@ Using per-installation keys provides the following benefits:
 - The user may enumerate the installations that have messaging access to their account.
 - The user may revoke keys on a per-installation level.
 
-## Ethereum wallet
+## Identity lifecycle
 
-Today, an Ethereum wallet consists of a secp256k1 keypair, and is identified by a a public address, which is the hex-encoding of the last 20 bytes of the Keccak-256 hash of the public key, prepended by `0x`. The user is expected to have a pre-existing Ethereum wallet prior to onboarding with XMTP.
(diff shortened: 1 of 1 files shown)


xmtp_mls/IDENTITY.md
```diff
@@ -24,59 +24,143 @@ Using per-installation keys provides the following benefits:
...
+1. Publish a commit to remove nodes from the conversation that are not in the list of valid installations.
+1. Publish a commit to add nodes to the conversation that are in the list of valid installations and not already present.
+
+XMTP clients may perform performance optimizations, such as caching installation lists with a short TTL.
 
-**Installation management**
+To support revocations (Q2+ 2024), clients must additionally perform the following validation on receiving each payload in the group:
 
-At any time, the user may enumerate active installations by querying for all identity updates under the account. The user may identify each installation by the creation time as well as the installation key from the signing text.
+1. Query the server for identity updates on the sending wallet address and verify that the sending installation has not been revoked.
```
neekolas, line 139: I'm not sure this is going to be viable as part of the message processing flow. Any race condition here could lead to part of the group accepting a commit and part of a group rejecting it. Even if we are only checking whether the installation was revoked at the time the message was sent, and not the time it was read, it still opens the door to potential race conditions given that revocation state does not have the same total ordering properties as topic contents.

I propose we don't check for revocation when reading payloads and instead do it before synchronizing the group state. Anyone with privileges allowing them to remove users would check the network for any revoked members and submit a commit to remove the revoked installation from the group before processing any other intents.

How this interacts with group permissions is TBD. Maybe we want a carve-out allowing anyone to remove a revoked installation so that we don't have to wait for an admin to come online. Maybe we have each user go through all their groups and remove themselves as part of revocation.

richardhuaaa: My preference is to have total ordering between revocations and commits, which is a little more robust, but I concede that the current direction of decentralization work doesn't seem to be heading that way.

Updated the revocation approach in https://github.com/xmtp/libxmtp/pull/353/commits/a54dbad33ed3fad21565dc18381a65917159c0d0!

insipx (approved):

karthikbhargavan (commented):
These clarifications address all the security considerations we considered, so it looks great.  The only thing I would add is an explicit note that we the `installation_key` should only be used in MLS and not elsewhere. If it is, then we need to review those use cases to ensure signature non-collisions.

PR #355, opened 2023-12-01
Overhauled User Preferences

neekolas:
## Summary

- Radically simplifies PPPP encryption, removing ECIES and `SignedPayload`s altogether and opting for a much simpler flow
- Renamed the library to `xmtp_pppp` and made the methods less generic

The way the library works:

### Encryption
1. Generate a new key via HKDF from the user's identity key
2. Encrypt the message via AES-256-GCM using the derived encryption key and the user's public key as associated data. Convert to a `PrivatePreferencesPayload` protobuf containing the ciphertext, nonce, and salt.

### Decryption 

1. Decode the `PrivatePreferencesPayload`
2. Derive the encryption key via HKDF from the user's identity key
3. Decrypt the contents via AES-256-GCM using the public key as associated data

## TODO
- ~Fix all the bindings to use the new library~


xmtp_pppp/src/encryption.rs
```diff
@@ -0,0 +1,55 @@
+use xmtp_crypto::encryption::{decrypt, encrypt, hkdf};
+use xmtp_proto::xmtp::message_contents::{
+    ciphertext::{Aes256gcmHkdfsha256, Union},
+    Ciphertext,
+};
+
+const PRIVATE_PREFERENCES_ENCRYPTION_KEY_SALT: &[u8] = b"XMTP_PRIVATE_PREFERENCES_ENCRYPTION";
+
+fn derive_encryption_key(private_key: &[u8]) -> Result<[u8; 32], String> {
```
neekolas, line 9: Technically, this step is not necessary. There is already a HKDF as part of `xmtp_crypto`'s `encrypt` function. But it does give us some domain separation built in, and is relatively low cost.

xmtp_pppp/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+    private_preferences_payload::Version as PrivatePreferencesVersion, Ciphertext,
+    PrivatePreferencesPayload,
+};
+
+use crate::encryption::{decrypt_ciphertext, encrypt_to_ciphertext};
+
+pub fn encrypt_message(
+    private_key: &[u8], // secp256k1 private key
+    message: &[u8],     // any byte array
+    topic: String,      // the topic you are encrypting for
```
neekolas, line 15: I am taking the topic in as an argument and using it as associated data in the encryption. I could omit the associated data entirely, since we don't have anything great to use here, but it does explicitly make it impossible to copy messages from one topic to another. Given that the topic is tied to the private key, that's probably not a real concern.

richardhuaaa: It does mean we can't change/migrate the topic format in the future, as it's coupled directly to the encryption. Would it make sense to use the public key here, which would theoretically make rotations easier to support?

xmtp_pppp/src/encryption.rs
```diff
@@ -0,0 +1,55 @@
+use xmtp_crypto::encryption::{decrypt, encrypt, hkdf};
+use xmtp_proto::xmtp::message_contents::{
+    ciphertext::{Aes256gcmHkdfsha256, Union},
+    Ciphertext,
+};
+
+const PRIVATE_PREFERENCES_ENCRYPTION_KEY_SALT: &[u8] = b"XMTP_PRIVATE_PREFERENCES_ENCRYPTION";
```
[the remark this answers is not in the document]
neekolas: Based on [this post](https://soatok.blog/2021/11/17/understanding-hkdf/), it does seem to be better to use the info parameter. I'm OK with doing that, at the expense of consistency with our other HKDF uses.

neekolas:
> It does mean we can't change/migrate the topic format in the future, as it's coupled directly to the encryption. Would it make sense to use the public key here, which would theoretically make rotations easier to support?

I don't have super strong feelings here. They're both pretty tightly coupled to the keypair used for encryption (the topic is constructed with the hash of a key HKDF'd from your private key). Happy to update

neekolas:
@richardhuaaa updated to use the pub key as associated data instead of the topic

neekolas:
Note:

@nplasterer @nakajima @rygine. While this new method takes the same arguments as the old, I renamed all the methods and packages that included the term "ecies" since we are no longer using that type of encryption. That means all the client SDKs are going to need to update their imports once this goes in

nakajima (commented):
I feel like the name `pppp` isn't super descriptive for folks who haven't been talking about it for the past few months. I know it's a hell of a lot of typing but what would you think of renaming this to `personal_portable_private_preferences` instead?

Also super down for a different name as long as it's more descriptive than `pppp`.

nplasterer:
> Also super down for a different name as long as it's more descriptive than pppp.

This has been an issue I've had to explain several times with our topics in the sdk. At this point I'm not sure we can get away from it? Probably too late to rename the topics? If we can't rename the topic than I would just leave it. Otherwise I think just `private_preferences` portable and personal feel implied?

xmtp_proto/src/gen/xmtp.keystore_api.v1.rs
```diff
@@ -428,9 +515,42 @@ impl ErrorCode {
...
+    /// String value of the enum field names used in the ProtoBuf definition.
+    ///
+    /// The values are not transformed in any way and thus are considered stable
+    /// (if the ProtoBuf definition does not change) and safe for programmatic use.
+    pub fn as_str_name(&self) -> &'static str {
+        match self {
+            JobType::Unspecified => "JOB_TYPE_UNSPECIFIED",
+            JobType::RefreshV1 => "JOB_TYPE_REFRESH_V1",
+            JobType::RefreshV2 => "JOB_TYPE_REFRESH_V2",
+            JobType::RefreshPppp => "JOB_TYPE_REFRESH_PPPP",
```
nplasterer, line 537: Probably a change that needs to happen in the `proto` repo if we care to keep the naming consistent across everything?. cc @rygine pppp -> user_preferences

nplasterer (approved):
This looks good to me. I'm not sure how the WASM stuff works but I have the JNI and kotlin files created as well as a PR open in the swift bindings. https://github.com/xmtp/xmtp-rust-swift/pull/15

xmtp_proto/src/gen/xmtp.keystore_api.v1.rs
neekolas: That one only gets persisted locally, so not the end of the world if it changes post-launch. Only side effect is that it'll force a refresh of the topic from the beginning.

user_preferences_bindings_wasm/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@xmtp/ecies-bindings-wasm",
+  "name": "@xmtp/user_preferences-bindings-wasm",
```
rygine, line 2: nit: seems weird to use snake_case and kebab-case together

can we choose 1 or the other?

xmtp_proto/src/gen/xmtp.keystore_api.v1.rs
rygine: @neekolas is the PR that needs an update? https://github.com/xmtp/proto/pull/116

neekolas: We can also change that to User Preferences if we wanted to, but that's truly invisible to the end user and can be changed at any time. 

[This was the one I was thinking about](https://github.com/xmtp/proto/pull/114)

user_preferences_bindings_wasm/package.json
neekolas: Good catch. We should make all kebab for the package.json name.

nplasterer: Oops that was my bad! Nice catch.

PR #356, opened 2023-12-03
Initial Android MLS bindings

richardhuaaa:
Have extremely basic binding working in the Android example app (`create_client`), just putting this up first. Also got persistent DB storage working. Not quite ready to integrate yet - there's still more pieces to do here:

1. Get DB encryption to work (seems to be broken in general, not just with the bindings) - this PR disables encryption for now
2. Expose more API surface (and unit tests)
3. Android integration tests
4. Tidy up API surface and naming


richardhuaaa:
The example app with these changes is immediately crashing on launch with no error - will debug tomorrow

bindings_ffi/src/lib.rs
```diff
@@ -28,20 +32,16 @@ pub enum GenericError {
...
-impl From<String> for GenericError {
-    fn from(err: String) -> Self {
-        Self::Generic { err }
-    }
-}
-
-impl From<uniffi::UnexpectedUniFFICallbackError> for GenericError {
-    fn from(e: uniffi::UnexpectedUniFFICallbackError) -> Self {
-        Self::Generic { err: e.reason }
+impl<T: Error> From<T> for GenericError {
```
richardhuaaa, line 35: Lets us remove the `map_err` everywhere

bindings_ffi/README.md
```diff
@@ -38,19 +38,36 @@ You'll need to do the following one-time setup to run Kotlin tests:
...
+
 # Releasing new version
 
 Tag the commit you want to release with the appropriate version (e.g. 0.3.0-beta0).
 The Release github workflow will run the following jobs:
 
 - android
-    - downloads the `libxmtp-android.zip` build artifact
-    - make a release tagged the same way with the artifact attached
+
```
richardhuaaa, line 63: This is auto-format, can't remove the extra line. Doesn't render differently in markdown

bindings_ffi/src/lib.rs
```diff
@@ -69,41 +69,41 @@ pub async fn create_client(
...
         None => static_enc_key(),
     };
 
     let store = match db {
-        Some(path) => {
-            info!("Using persistent storage: {} ", path);
-            EncryptedMessageStore::new(StorageOption::Persistent(path), key)
-        }
+        // TODO support encryption
+        Some(path) => EncryptedMessageStore::new_unencrypted(StorageOption::Persistent(path)),
```
richardhuaaa, line 93: Disable DB encryption for now, seeing as it doesn't work correctly. Doesn't affect the bindings interface

insipx:
> Get DB encryption to work

What's wrong with DB Encryption?

richardhuaaa:
The migrations run successfully, but subsequent queries to the database fail with a generic error from sqlite3 (`File is not a database`) - typically means decryption failed. I've also verified myself that the encrypted database created from the process is not accessible using `sqlite3`, both using and not using the encryption key. Unencrypted database is working fine

xmtp_mls/src/storage/encrypted_store/db_connection.rs
```diff
@@ -12,26 +12,37 @@ enum RefOrValue<'a, T> {
...
 pub struct DbConnection<'a> {
-    wrapped_conn: RefCell<RefOrValue<'a, RawDbConnection>>,
+    wrapped_conn: Mutex<RefOrValue<'a, RawDbConnection>>,
 }
 
 impl<'a> DbConnection<'a> {
     pub(crate) fn new(conn: &'a mut RawDbConnection) -> Self {
         Self {
-            wrapped_conn: RefCell::new(RefOrValue::Ref(conn)),
+            wrapped_conn: Mutex::new(RefOrValue::Ref(conn)),
```
neekolas, line 21: Generally curious about this change

neekolas (approved):

neekolas (approved):

bindings_ffi/README.md
```diff
@@ -38,19 +38,36 @@ You'll need to do the following one-time setup to run Kotlin tests:
 
 If you want to skip the setup, you can also run `cargo test -- --skip kts` to only run Rust unit tests. CI will run all tests regardless.
 
+# Debugging
```
neekolas, line 41: We should also add some notes around the expected `docker-compose.yml` format to get a working local backend, and warn users that this won't work with the dev network just yet.

richardhuaaa: Done - added a note. Haven't updated the Android example app to use localhost, will do that in next PR

bindings_ffi/src/lib.rs
```diff
@@ -135,146 +135,157 @@ pub struct FfiConversations {
...
     }
 }
 
 #[derive(uniffi::Record)]
 pub struct FfiMessage {
-    pub id: String,
+    pub id: Vec<u8>,
     pub sent_at_ns: i64,
-    pub convo_id: String,
+    pub convo_id: Vec<u8>,
```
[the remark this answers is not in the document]
richardhuaaa: Note for others, we decided on `conversation`/`convo` for the methods in this SDK as opposed to `group`, because the former will be more appropriate when we eventually add 1:1 conversations to libxmtp

xmtp_mls/src/storage/encrypted_store/db_connection.rs
richardhuaaa: Updated the README with this detail:
We use Tokio as our [async runtime](https://rust-lang.github.io/async-book/08_ecosystem/00_chapter.html). Uniffi can use this runtime on async methods and objects using the annotation `#[uniffi::export(async_runtime = ‘tokio’)]`. Uniffi plumbs up an executor (scheduler) in the foreign language to the Tokio runtime in Rust. More details [here](https://github.com/mozilla/uniffi-rs/blob/734050dbf1493ca92963f29bd3df49bb92bf7fb2/uniffi_core/src/ffi/rustfuture.rs#L11-L18). By default, Tokio's [multi-thread scheduler](https://docs.rs/tokio/latest/tokio/runtime/index.html#multi-thread-scheduler) is used. This means that functions may resume execution in a different thread when an `await` is encountered. This in turn means that structs held across an `await` in libxmtp must be `Send` and `Sync`.

We could also consider switching to the single-thread executor, if uniffi allows us to specify it - I haven't investigated this yet.

PR #363, opened 2023-12-05
move add and remove members to use wallet addresses

### LINKED ISSUE #357: Move AddMembers and RemoveMembers intents to store the wallet address instead of key packages
https://github.com/xmtp/libxmtp/pull/343#discussion_r1406932396

insipx:
closes #357 

- slight refactor to `retry_async` to allow retrying functions that accept a mutable ref
- `remove add_members_by_installation_id` and `remove_members_by_installation_id`
- `wallet_address` stored on intent instead of KeyPackages
- key packages are fetched while getting publish intents
- simple validation of Ethereum addresses in `xmtp_cryptography` (Does not cover EIP-55)

I accidentally pulled some changes from #354 from proto, so this should be merged after that (just some changes to GroupMembership names/tests that need to be merged in)


xmtp_mls/src/groups/mod.rs
```diff
@@ -15,8 +14,6 @@ use openmls::{
 };
 use openmls_traits::OpenMlsProvider;
 use std::mem::discriminant;
-#[cfg(test)]
-use std::println as debug;
```
insipx, line 19: I removed this; I assume it was here because RUST_LOG=debug spits out a bunch of irrelevant stuff. Can always use `RUST_LOG=xmtp_mls=debug` to filter out any logs not part of the `xmtp_mls` crate. per-module filtering also works, `RUST_LOG=xmtp_mls::groups=debug`

dev/gen_protos.sh
```diff
@@ -5,7 +5,7 @@ if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
         exit 1
     fi
 fi
-if ! buf generate https://github.com/xmtp/proto.git#branch=nmolnar/mls-verification-service,subdir=proto; then
```
insipx, line 8: this is temp, will make a companion PR in proto once this is finished

neekolas: I also have a TODO item to get all of our MLS protos into `main`, which'll help clean this up

xmtp_mls/src/groups/mod.rs
neekolas: @richardhuaaa put this in, but I believe the intention was just to get the logs to emit when running unit tests

insipx: ah, i've been adding `crate::tests::setup` to unit tests if I need logs, should I go through and add that to each test you think? I've just been adding it as I go. Can also revert this if this way is easier

neekolas: In my perfect world, all unit tests would emit logs from our application automatically. And we could optionally set a command line flag to also get unit tests from libraries we use. Being able to set something per-file is second-best, and per-test is third.

insipx: That's fair -- I've added `ctor` to setup this logging before any tests run, it works without any manual calls now

neekolas:
I think something must have gotten miscommunicated in the goals here. We do need _some_ mechanism to add members by `installation_id`. This is necessary to add new devices to a conversation, and to support revocation.

What we're hoping to achieve here is that when you are adding or removing a member by wallet address we look up the full list of `installation_id`s to add/remove at the last possible moment to ensure no changes slip in between the intent being created and being fulfilled.

insipx:
> What we're hoping to achieve here is that when you are adding or removing a member by wallet address we look up the full list of installation_ids to add/remove at the last possible moment to ensure no changes slip in between the intent being created and being fulfilled.

So essentially we still want to get key_packages and the resulting installation_ids in `get_publish_intents`, but still want to add members by addr or installation id?

If so, we could TLS serialize an enum of WalletAddress or InstallationId that sits in the proto definition as bytes, and deserialize it in that fn to use the correct API method that gets the right KeyPackages. This way we can still add and remove members by address or installation ID

xmtp_cryptography/src/signature.rs
```diff
@@ -111,14 +112,58 @@ pub fn h160addr_to_string(bytes: H160) -> String {
...
+
+    if address.len() != 40 {
+        return false;
+    }
+
+    address.chars().all(|c| c.is_ascii_hexdigit())
+}
+
+/// Check if an ed25519 public signature key is valid.
+pub fn is_valid_ed25519_public_key<Bytes: AsRef<[u8]>>(public_key: Bytes) -> bool {
```
insipx, line 128: Not sure if completely necessary; however we had a test checking for invalid member registration and I didn't see how this would be a negative, since it would leave less errors for intent publishing

richardhuaaa: It's a nice touch, prevents confusion/misuse between removing installations/accounts

xmtp_mls/src/groups/mod.rs
richardhuaaa: @insipx, thank you for doing this - I couldn't figure out how to push all logs to print when tests fail. Does the `tracing_subscriber` automatically add all `log::` calls to the trace?

richardhuaaa (approved):
Apologies for the confusion! I could probably have talked through the issue more in advance just to make sure we were all on the same page. Thanks for tackling this, and for always refactoring and improving the code quality!

neekolas (approved):
This looks great. Thank you @insipx

PR #368, opened 2023-12-12
Introduce V2 client and add more missing bindings

nakajima:
Changes:

- Split `group.sync` out into its own function and remove it from find_messages and list_members

Questions:

- How should being removed be handled
- Seeing a fair amount of `database is locked` errors. Do we need some sort of operations queue or something?


nakajima:
Seeing this error in a client that was added to a group:

image

richardhuaaa:
Ahh, my guess is the client that got added is trying to read historical messages from before they were added, so these warnings/errors would make sense when trying to process them. Is the client still functioning normally? FYI @neekolas

neekolas:
Yes, those wrong epoch messages are expected. At some point we can optimize that out by not starting at the beginning when you are added to an existing conversation, but it's harmless. 

The locked database errors are more surprising to me. Shouldn't the connection pool prevent that?

neekolas (approved):

neekolas:
> How should being removed be handled

That's one of the items still on the todo list. OpenMLS knows that you've been removed, but we don't bubble it up into the application code at all.

nakajima:
Oh, so newly added clients don't get access to messages sent before they were added? If so then that makes sense.

nakajima:
I added a test to try to reproduce the database lock error: https://github.com/xmtp/xmtp-swift/blob/61266302dee5f210971f443107bf75abc137487a/Tests/XMTPTests/ClientTests.swift#L36-L58.

Sometimes I get a database is locked error, sometimes I'm getting back `group create error group create: Error accessing the key store.`, which seems to be coming from open-mls: https://github.com/openmls/openmls/blob/d249c76bb959d15cb465fcd35842c1921fe829a3/openmls/src/group/mls_group/errors.rs#L31. My hunch is that it's the same underlying problem.

richardhuaaa:
@nakajima, in the non-openmls case, is the only error message `database is locked`, or is there more context attached to it? That would help to pinpoint exactly which codepath it happened in.

At first I thought it was because there are multiple concurrent transactions from different connections, but it seems like that would manifest as SQLITE_BUSY, this seems to be SQLITE_LOCKED. https://www2.sqlite.org/cvstrac/wiki?p=DatabaseIsLocked

nakajima:
@richardhuaaa the full error is: `Failed to fetch key: DieselResult(DatabaseError(Unknown, "database is locked"))`

bindings_ffi/src/lib.rs
```diff
@@ -128,6 +152,317 @@ impl FfiXmtpClient {
...
+}
+
+#[derive(uniffi::Record)]
+pub struct FfiCursor {
+    pub digest: Vec<u8>,
+    pub sender_time_ns: u64,
+}
+
+#[derive(uniffi::Record)]
+pub struct FfiV2QueryRequest {
```
neekolas, line 196: Can we move all these V2 API calls into a separate file? This one is getting pretty huge and messy.

bindings_ffi/src/lib.rs
```diff
@@ -128,6 +152,317 @@ impl FfiXmtpClient {
...
+    }
+}
+
+#[derive(uniffi::Object)]
+pub struct FfiV2Client {
+    inner_client: Arc<xmtp_api_grpc::grpc_api_helper::Client>,
+}
+
+#[uniffi::export(async_runtime = "tokio")]
+impl FfiV2Client {
```
neekolas, line 429: We might want to call this a `FfiV2ApiClient` in case we ever want something like a full V2 client in Rust. Or at least for clarity

xmtp_mls/src/client.rs
```diff
@@ -198,6 +200,8 @@ where
     }
 
     pub async fn register_identity(&self) -> Result<(), ClientError> {
+        log::info!("registering identity");
+        // TODO: Mark key package as last_resort in creation
```
neekolas, line 204: Think we can remove this TODO. Must have snuck in with a merge. The TODO was done a few weeks ago

bindings_ffi/Cargo.toml
```diff
@@ -22,6 +22,8 @@ xmtp_api_grpc = { path = "../xmtp_api_grpc" }
 xmtp_v2 = { path = "../xmtp_v2" }
 futures = "0.3.28"
 tokio = { version = "1.28.1", features = ["macros"] }
+xmtp_proto = { path = "../xmtp_proto", features = ["proto_full"] }
```
neekolas, line 25: You should be able to get away with just setting `features = ["xmtp-message_api-v1"]`. That'll give it a smaller footprint.

neekolas (approved):

PR #369, opened 2023-12-13
Add support for validating commits

neekolas:
## Summary

- Moves some of the logic from the transcript messages to a new `ValidatedCommit` struct
- Adds some more validations for known bad commit types

### Currently rejects the following conditions
- Invalid credential of the sender
- Invalid credential of any new members
- Attempting to remove a member who is not present in the group
- Multiple proposals from different senders in the same commit
- Proposal contained in a commit that is not from a group member (including ExternalJoin)

## Notes

Hopefully this can act as a starting place to build further validations, such as group rules/permissions. The `ValidatedCommit` struct gives us a clean source of truth for what is contained in the commit, and who did it. We can then use that to apply the rules specified in the Immutable Group Metadata.

The blocking of "multiple proposals from different senders in the same commit" isn't strictly necessary to fulfill our business rules, but it sure simplifies things. Because we want to treat adding/removing installations for an existing user differently from adding/removing users entirely, it gets complicated to evaluate rules proposal-by-proposal. Instead, we aggregate all the proposals together so we can differentiate between actions on an installation or a user.

If we decide to support general purpose proposals down the line we will need to do a refactor to pre-group changes by sender and evaluate them at that level.


xmtp_mls/src/groups/validated_commit.rs
```diff
@@ -0,0 +1,459 @@
...
+                }
+            }
+            _ => return Err(CommitValidationError::ActorNotMember),
+        }
+    }
+
+    // Self updates don't produce any proposals I can see, so it will actually return
+    // None in that case.
+    // TODO: Figure out how to get the leaf index for self updates
+    Ok(leaf_index.map(|i| *i))
```
insipx, line 211:
```suggestion
    Ok(leaf_index.copied())
```

insipx (approved):
looks good!

xmtp_mls/src/groups/validated_commit.rs
insipx, line 211:
```suggestion
    Ok(leaf_index.copied())
```

PR #370, opened 2023-12-13
Fix DB issues

richardhuaaa:
Key learnings:
- The `sqlite3` command-line tool supports `pragma key`, but encrypts/decrypts in a way that is non-compliant with `sqlcipher`. There is a `sqlcipher` command-line tool that can be used to open sqlite databases encrypted via `sqlcipher` instead.
- We were unnecessarily opening a second connection when initializing the identity. The ephemeral store only allows one connection open at a time, which is why it was locking up when used via the bindings (and probably via CLI too).
- The sqlcipher `pragma key` initialization must be run on *every connection*, not just the first time the database is opened. If different connections write to the database with different encryption keys, the database becomes corrupted and can no longer be opened.

Added fixes for these, as well as unit tests and updating READMEs.

Also modified the bindings to create an unencrypted database when no encryption key is passed, rather than an encrypted database with a hardcoded encryption key. This should make development/debugging a little bit easier. Note that there isn't much of a security difference between using a hardcoded encryption key and no encryption key at all.

### DIFF
--- a/bindings_ffi/README.md
+++ b/bindings_ffi/README.md
@@ -53,8 +53,10 @@ There is no support for using breakpoints or a debugger with FFI currently. Meth
 1. Examine the database in your app
    1. Use logs to find the location of the Sqlite database and the database encryption key
    1. Find the database (for example, on Android emulator use Device File Explorer in Android Studio)
-   1. Copy the database to your local machine and open it on the command line using `sqlite3`
+   1. `brew install sqlcipher`
+   1. Copy the database to your local machine and open it on the command line using `sqlcipher <file>.db3`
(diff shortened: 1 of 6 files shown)


neekolas (approved):
Great work getting to the bottom. How was this issue not facing the Vmac version?

richardhuaaa:
@neekolas the vmac version did not open multiple simultaneous connections, at least in the limited testing situations we set it up in.

richardhuaaa:
@nakajima, heads up that if you are currently using the `encryption_key` parameter, any existing DB will get corrupted after syncing this change, as the libxmtp binary will actually start honoring it instead of ignoring it. Can either set the parameter to null, or start with a fresh DB

insipx (approved):

PR #374, opened 2023-12-14
Refactor credential text, add context and creation time

richardhuaaa:
Still to-do:
- Use `MlsCredential` wrapper proto
- Check/update validation logic, e.g. verifying installation public key matches leaf node
- Use `account_address` instead of `wallet_address` everywhere


neekolas:
@richardhuaaa can we get this merged today? I've removed the `xmtp` crate in another PR and fixed up a couple of lint warnings. Looks pretty good to go, although we might want to converge around the name `account_address`. I see a mix of `blockchain_address` and `wallet_address` in different places. Could be done as a subsequent PR.

xmtp_mls/IDENTITY.md
```diff
@@ -54,6 +54,7 @@ XMTP installations consist of a long-lived Ed25519 key-pair (the 'installation k
    XMTP: Grant Messaging Access
 
    Current Time: <ISO 8601 date and time with local UTC offset>
+   Account Address: <ethereum address>
    Installation ID: <hex(last_20_bytes(keccak256(Ed25519PublicKey)))>
```
richardhuaaa, line 58: @jhaaaa @saulmc - would love your feedback on the verbiage in the signature text. Q's:

- Should we add text explaining what is being granted, what the installation ID is used for, what kind of verification the user needs to do? We need the user to verify the 'current time' is correct, and that they are signing for the correct 'account address'. The installation ID is not a secret and they don't need to memorize it. The 'installation ID' and 'current time' are there so they can list installations in the future and recognize this one, revoking it if needed.
- Should we change the labels for clarity, for example `Granted Time` instead of `Current Time`, `Installation Token` or `Granted Token` etc instead of `Installation Id`?
- Ethereum address are formatted as `0x<random string>`, should installation Id's be formatted differently for clarity, e.g. `x:<Random String>`?

neekolas (approved):

xmtp_mls/IDENTITY.md
```diff
@@ -54,6 +54,7 @@ XMTP installations consist of a long-lived Ed25519 key-pair (the 'installation k
    XMTP: Grant Messaging Access
 
    Current Time: <ISO 8601 date and time with local UTC offset>
```
richardhuaaa, line 56: One thing I just realized, this leaks the user's timezone. Curious on other's thoughts here

neekolas: Good catch. Any harm in converting to UTC? Guess it makes it a little less human readable, but I can't imagine people are going to be reading these timestamps too closely

richardhuaaa: Then we can't really ask users to double-check the timestamp, and we're trusting the app to furnish a valid timestamp. I suppose it's okay if a malicious app forges a fake timestamp here, as it's only used for identification? I can change it for now

xmtp_mls/IDENTITY.md
richardhuaaa: Merging this to unblock others, but will keep track of this thread and add changes separately!

saulmc: Defer to @jhaaaa 's expertise on these questions!

PR #377, opened 2023-12-14
Add scaffolding for new policy framework

neekolas:
## Summary

- Adds in some basic structure for defining permissions and policies for evaluating proposals in commits
- Adds a few basic policies that can be enforced with the metadata we have right now
- Enables composition of simpler policies to form more complex policies

## Next up
- Serialization of the policies into something we can use in the metadata field
- Adding group creator/admin status to the `ValidatedCommit` object so it can be used in a policy
- Integrating immutable group metadata to determine group creator/admin status
- Enabling policy enforcement (initially with everything set to Allow)
- Adding some more logging and traceability to policy rejections. Devs should know why a commit was rejected

### Expected initial policies for V1

**Add Member**
- Actor is an installation with the same wallet address as the group creator

**Remove Member**
- Actor is an installation with the same wallet address as the group creator
OR
- The subject of the change has only one installation and that installation was revoked before the time of the commit

**Add Installation to existing member**
- Any group member can add an installation if the credential maps to the same account address as an existing member

**Remove installation from existing member**
- The subject installation has been revoked before the time of the commit

### DIFF
--- a/xmtp_mls/src/groups/group_permissions.rs
+++ b/xmtp_mls/src/groups/group_permissions.rs
@@ -1,0 +1,358 @@
+use super::validated_commit::{AggregatedMembershipChange, CommitParticipant, ValidatedCommit};
+
+pub trait Policy {
+    fn evaluate(&self, actor: &CommitParticipant, change: &AggregatedMembershipChange) -> bool;
+}
+
+#[derive(Clone, Copy, Debug)]
+#[allow(dead_code)]
+pub enum StandardPolicies {
+    Allow,
+    Deny,
+    // Allow if the change only applies to subject installations with the same account address as the actor
+    AllowSameMember,
+    // AllowIfActorAdmin, TODO: Enable this once we have admin roles
+    // AllowIfActorCreator, TODO: Enable this once we know who the creator is
+    // AllowIfSubjectRevoked, TODO: Enable this once we have revocation and have context on who is revoked
+}
+
+impl Policy for StandardPolicies {
+    fn evaluate(&self, actor: &CommitParticipant, change: &AggregatedMembershipChange) -> bool {
+        match self {
+            StandardPolicies::Allow => true,
+            StandardPolicies::Deny => false,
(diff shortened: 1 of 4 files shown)


neekolas:
One thing I realized pretty early on in working on this: if you want to describe our expected initial policy in config rather than procedural code, you actually need a pretty robust config language.

It makes me wonder whether we actually want to allow more permission models from the start. We will have already done 95% of the work to allow an "anyone can add/remove members" model and many other variations, so why hold it back from devs.

Having it configurable does make triaging issues and supporting devs more complicated, but we'd have to decide if that's worth the cost of holding back the feature.

richardhuaaa (approved):
This is a pretty powerful framework - I wonder if most of the expressiveness for adding/removing accounts can be captured just by controlling who the admins are (creator/all group members/some subset)? Happy to try it out though!

> Remove installation from existing member
> The actor is another installation with the same account address as the subject

I think we shouldn't allow this one, it should only happen if the installation is revoked. We shouldn't allow people to have different installation sets in different conversations, also the periodic installation checking from other members in the group will add this installation back anyway.


> Add Member
> Actor is an installation with the same wallet address as the group creator

Now that we have sophisticated epoch ordering and commit resolution, I wonder if it's actually simpler (and possibly better UX) just to allow everyone to add/remove members for the initial version?

neekolas:
> I think we shouldn't allow this one, it should only happen if the installation is revoked. We shouldn't allow people to have different installation sets in different conversations, also the periodic installation checking from other members in the group will add this installation back anyway.

That's a great point. I can omit

> Now that we have sophisticated epoch ordering and commit resolution, I wonder if it's actually simpler (and possibly better UX) just to allow everyone to add/remove members for the initial version?

It certainly makes sense for some use-cases. I know people were not super happy about that behaviour in Pat's earlier group chat take. And it makes things like token-gating bots impossible to build.

Given that we're likely going to have to live with whatever permissions/validation system we initially come up with for a while, I'm inclined to make the base level protocol more capable than we need today. The one thing we can change post-launch is which already-valid permission sets we encourage devs to use.

PR #385, opened 2023-12-18
Write transcript messages to DB

neekolas:
## Summary

- Integrates the work with validated commits to actually generate and save transcript messages upon successfully processing a commit from either yourself or another group member.


richardhuaaa (approved):
LGTM! Nice to see the pieces fitting together 👏

PR #387, opened 2023-12-19
wallet_address -> account_address

richardhuaaa:
Settling on consistent terminology as discussed. Depends on proto update https://github.com/xmtp/proto/pull/121

Note that this will break any existing DB's - my upcoming credential update will also break existing DB's.


neekolas (approved):
When the protos change we'll also need to update xmtp-node-go

PR #393, opened 2023-12-20
Create new Subscribe2 method

neekolas:
## Summary

- Adds support for the `Subscribe2` API endpoint, which allows callers to update their list of content topics without closing/re-opening the subscription
- Uses a new subscription return type. Instead of our home-brewed Subscription struct `subscribe2` returns a struct that fulfills `futures::Stream` + a new custom trait (`MutableApiSubscription`) that allows callers to update the content topics.

### DIFF
--- a/xmtp_api_grpc/src/grpc_api_helper.rs
+++ b/xmtp_api_grpc/src/grpc_api_helper.rs
@@ -340,20 +340,20 @@ impl XmtpApiSubscription for Subscription {
 type EnvelopeStream = Pin<Box<dyn Stream<Item = Result<Envelope, Error>> + Send>>;
 
 pub struct GrpcMutableSubscription {
-    inner: Abortable<EnvelopeStream>,
-    channel: futures::channel::mpsc::UnboundedSender<SubscribeRequest>,
+    envelope_stream: Abortable<EnvelopeStream>,
+    update_channel: futures::channel::mpsc::UnboundedSender<SubscribeRequest>,
     abort_handle: AbortHandle,
 }
 
 impl GrpcMutableSubscription {
     fn new(
-        inner: EnvelopeStream,
-        channel: futures::channel::mpsc::UnboundedSender<SubscribeRequest>,
+        envelope_stream: EnvelopeStream,
+        update_channel: futures::channel::mpsc::UnboundedSender<SubscribeRequest>,
     ) -> Self {
         let (abort_handle, abort_registration) = AbortHandle::new_pair();
         Self {
-            inner: Abortable::new(inner, abort_registration),
-            channel,
+            envelope_stream: Abortable::new(envelope_stream, abort_registration),
+            update_channel,
             abort_handle,
         }
     }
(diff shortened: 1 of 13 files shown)


nplasterer:
👀 This should make it easier for us to bring this to v2 and update swift. I think this would greatly improve the swift streaming. And may help fix this https://github.com/xmtp/xmtp-ios/issues/185

xmtp_proto/src/api_client.rs
```diff
@@ -86,11 +89,19 @@ pub trait XmtpApiSubscription {
...
+    async fn update(&mut self, req: SubscribeRequest) -> Result<(), Error>;
+    fn close(&self);
+}
+
 // Wasm futures don't have `Send` or `Sync` bounds.
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 pub trait XmtpApiClient {
     type Subscription: XmtpApiSubscription;
+    type MutableSubscription: MutableApiSubscription + Stream<Item = Result<Envelope, Error>>;
```
richardhuaaa, line 104: Might be cleaner/consistent to do:
```
pub trait MutableApiSubscription: Stream<Item = Result<Envelope, Error>> {
```

on L94 above, and

```
type MutableSubscription: MutableApiSubscription;
```

here. Then all implementers of `MutableApiSubscription` automatically must implement `Stream`

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -288,6 +337,56 @@ impl XmtpApiSubscription for Subscription {
     }
 }
 
+type EnvelopeStream = Pin<Box<dyn Stream<Item = Result<Envelope, Error>> + Send>>;
```
richardhuaaa, line 340: Out of curiosity, why is the `Pin` needed here?

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -288,6 +337,56 @@ impl XmtpApiSubscription for Subscription {
...
+
+pub struct GrpcMutableSubscription {
+    inner: Abortable<EnvelopeStream>,
+    channel: futures::channel::mpsc::UnboundedSender<SubscribeRequest>,
```
richardhuaaa, line 344: I wonder if these could be named more clearly, for example `envelopes` and `requests`, or `output` and `input`

richardhuaaa (approved):
Wow, this is very sophisticated, well done!

Do we still need the original `subscribe` endpoint?

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: That's fair. Let me update to some more thought out names

xmtp_proto/src/api_client.rs
neekolas: That's a great idea. Updated.

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: Abortable uses some sort of macro to ensure that the inner field is always pinned.

neekolas:
> Do we still need the original subscribe endpoint?

We won't use it in `xmtp_mls` (at least over GRPC), since the new endpoint works fine for streams that never need updates. But I'm leaving it in the grpc api helpers so that it can still be used by `xmtp-ios`.

PR #396, opened 2023-12-22
Use credential proto

richardhuaaa:
Use a wrapper `MlsCredential` proto instead of the `Eip191Association` proto for two reasons:

1. Include the `installation_public_key` inside of the credential, as suggested by Cryspen
2. Allow for other association types. In particular, we will be introducing an association type using the XMTP v2 identity key, for migration purposes.

There is some validation logic we will need to double-check, for example verifying that identity updates correspond to the correct wallet address, as well as installation syncing logic - will follow up on this after the break.

### DIFF
--- a/xmtp_mls/src/association.rs
+++ b/xmtp_mls/src/association.rs
@@ -279,7 +279,7 @@ impl AssociationText {
                 account_address,
                 installation_public_key,
                 iso8601_time,
-            } => gen_static_text_v1(account_address, installation_public_key, &iso8601_time),
+            } => gen_static_text_v1(account_address, installation_public_key, iso8601_time),
         }
     }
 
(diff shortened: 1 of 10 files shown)


xmtp_mls/src/association.rs
```diff
@@ -25,51 +29,138 @@ pub enum AssociationError {
...
+            association: Association::Eip191(association),
+        })
+    }
+
+    pub fn from_proto_validated(
+        proto: MlsCredentialProto,
+        expected_account_address: Option<&str>, // Must validate when fetching identity updates
+        expected_installation_public_key: Option<&[u8]>, // Must cross-reference against leaf node when relevant
+    ) -> Result<Self, AssociationError> {
+        let association = match proto.association.unwrap() {
```
neekolas, line 65: Should remove the unwrap and return the error here.

neekolas (approved):

PR #397, opened 2024-01-04
Allow streaming group messages

neekolas:
## Summary

- Adds support for `subscribe2` in `ApiClientWrapper`
- Convert stream of envelopes into a stream of `StoredGroupMessages`
- Handle streaming of group messages, whether sent by yourself or by others

## Notes

There are a few things about the implementation worth calling out, as they are subtle

1. We are not moving the `topic_refresh_state` when we are processing application messages. Streams may have been started after the last sync and may have been interrupted mid-sync so they are not considered safe to increment epochs. In the case of envelopes containing Commits, we will interrupt and roll back processing of the message and call `sync()` to load messages in the correct order from the server. This ensures that all commits are processed in the correct order.
2. We don't return the successfully processed messages from `process_message`, and instead rely on looking up the `StoredGroupMessage` based on the timestamp of the envelope. This is because messages can only be processed once, so if another thread has already synced the message (for example, after sending a commit) we will get an error if we attempt to process it a second time receiving it from the stream.

### DIFF
--- a/xmtp_api_grpc/src/lib.rs
+++ b/xmtp_api_grpc/src/lib.rs
@@ -187,7 +187,9 @@ mod tests {
             })
             .await
             .unwrap();
-
+        
+        // Sleep 100ms to ensure subscription is updated
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
         // Publish an envelope to the new topic
         let env_2 = test_envelope(topic_2.to_string());
         client
--- a/xmtp_api_grpc/src/lib.rs
+++ b/xmtp_api_grpc/src/lib.rs
@@ -187,7 +187,7 @@ mod tests {
             })
             .await
             .unwrap();
-        
+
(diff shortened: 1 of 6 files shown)


xmtp_api_grpc/src/lib.rs
```diff
@@ -188,6 +188,8 @@ mod tests {
             .await
             .unwrap();
 
+        // Sleep 100ms to ensure subscription is updated
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
neekolas, line 192: I noticed some flakiness in this test because of race conditions. This should resolve.

nplasterer (approved):
LGTM. First time I've seen tests in the same file as the code. Is that a Rust thing?

neekolas:
> Is that a Rust thing?

Sure is. You don't strictly have to, but it's a common practice

PR #399, opened 2024-01-05
Streaming bindings

neekolas:
## Summary

- Builds on top of #397 and #398
- Adds bindings for message streaming

Message streaming requires the caller to implement a class in Swift/Kotlin that meets the `FfiMessageCallback` interface defined in the UDL file. The `on_message` function will be called once per message.

## Notes

I tried _very_ hard to figure out a way to wrap a Stream type into something FFI friendly. The main issue is that our `group.stream()` method keeps a reference to `self` for the life of the stream, which in turn has a lifetimed reference to `Client`. This is required so that we can actually decrypt messages. The problem is that Uniffi doesn't play nicely with lifetimed references. We can't store a `MlsGroup` on any sort of FFI friendly struct, and so the reference always gets dropped before the stream is done. This makes the compiler unhappy.

Maybe someone smarter than me can figure it out (I've tried every permutation of boxes/arcs/mutexes I can think of), but I have to admit the callback interface might actually be easier in the end. At least we aren't exposing any sort of exotic types to Swift/Kotlin. The memory management should be simpler since the stream never leaves Rust-land.


nplasterer (approved):
LGTM. 👍 I like the evolution of Intro, Invite, Welcome. 😄

PR #407, opened 2024-01-06
Allow for serializing policies

neekolas:
## Summary

- Companion to https://github.com/xmtp/proto/pull/122
- Renames a bunch of types to better match the protos, and to allow for policies that don't involve group membership (for example, metadata policies)
- Removes use of generics and instead has all policies in a concrete enum. This makes deserialization a lot more straightforward.

## Notes

After some discussion with @richardhuaaa we've decided to not make add/remove installation policies configurable over the wire. Instead, I am hardcoding the policies into the `PolicySet`. The rules around adding/removing installations will still be handled via policies, but those policies cannot be changed without forking the repo. 

I also have a controversial take: the `remove_installation_policy` should be hardcoded to `deny`. Once we have a dedicated revocation proposal type (which will have its own policy), there is no valid use-case for removing an installation from anyone - even yourself.

Having the ability to remove your own installations can be abused by a malicious client to remove the user's other installations, leaving the the user unaware of the SPAM the malicious client is sending. It can also make it impossible for the user to send their own revocation message, so they have to wait for another group member to come online to kick out the revoked installation.

We might need to revisit this if we institute any sort of installation expiry, but I think that can be handled separately as well.


dev/gen_protos.sh
```diff
@@ -6,8 +6,7 @@ if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
     fi
 fi
 
-# Please always specify the exact commit to use for generation!
-if ! buf generate https://github.com/xmtp/proto.git#branch=main,ref=7be5c72,subdir=proto; then
+if ! buf generate https://github.com/xmtp/proto.git#branch=nm/add-policies,subdir=proto; then
```
neekolas, line 9: Will change this back to `main` once the other PR goes in.

xmtp_mls/src/groups/group_permissions.rs
```diff
@@ -1,162 +1,247 @@
...
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq)]
 #[allow(dead_code)]
-pub enum StandardPolicies {
+#[repr(u8)]
+pub enum BasePolicies {
     Allow,
     Deny,
     // Allow if the change only applies to subject installations with the same account address as the actor
     AllowSameMember,
```
neekolas, line 35: This is technically only useful for remove installation policies. We could make it error if you try and export one of these policies to a proto, so that it can only be used for one of the hardcoded policies. But that feels a little hacky. For add/remove members applying this policy will effectively be the same as `deny`, so it's probably safe to leave as-is.

xmtp_mls/src/groups/group_permissions.rs
```diff
@@ -1,162 +1,247 @@
...
 
+#[derive(Debug, Clone, PartialEq)]
 #[allow(dead_code)]
-impl<PolicyType> AndCondition<PolicyType>
-where
-    PolicyType: Policy,
-{
-    pub fn new(policies: Vec<PolicyType>) -> Self {
-        Self { policies }
+pub enum MembershipPolicies {
```
richardhuaaa, line 68: Not super important, but could `MembershipPolicies` be a trait on `BasePolicies`, `AndCondition` and `AnyCondition`, removing the need for this enum and the match arms?

richardhuaaa (approved):
Very elegant! LGTM

xmtp_mls/src/groups/group_permissions.rs
neekolas: That's how it started before this PR. Things got really nasty trying to convert the proto oneof value into one of several possible `dyn Policy` types.

PR #409, opened 2024-01-08
Assume all Key Packages are last resort.

Bren2010:


dev/gen_protos.sh
```diff
@@ -5,7 +5,9 @@ if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
         exit 1
     fi
 fi
-if ! buf generate https://github.com/xmtp/proto.git#branch=main,subdir=proto; then
+
+# Please always specify the exact commit to use for generation!
+if ! buf generate https://github.com/Bren2010/proto.git#branch=brendan/last-resort,ref=f50c3c99cd1,subdir=proto; then
```
neekolas, line 10: Let's get this into `xmtp/proto` main before merging.

xmtp_mls/src/identity.rs
```diff
@@ -91,11 +91,11 @@ impl Identity {
             None,
             None,
         );
-        // TODO: Set expiration
         let kp = KeyPackage::builder()
             .leaf_node_capabilities(capabilities)
             .leaf_node_extensions(leaf_node_extensions)
             .key_package_extensions(key_package_extensions)
+            .key_package_lifetime(Lifetime::new(30 * 86400))
```
neekolas, line 98: Not sure what the ideal lifetime is here, but probably closer to 6 months

neekolas (approved):

PR #411, opened 2024-01-09
Fix tests that occasionally hang

neekolas:
## Summary

Race conditions are causing some tests to occasionally hang. Adding in some sleeps to protect against this. I believe these issues are unique to the test environment, and should never happen in a normal event loop. But we should keep an eye on this as we are integrating into SDKs.


richardhuaaa (approved):
Any idea what the root cause is? Is there a race?

neekolas:
I haven't been able to trace it down precisely

PR #417, opened 2024-01-16
Use connection customizer

neekolas:
## Summary

I discovered that our connection pool (R2D2) has a more native way to apply operations to each of the underlying Sqlite connections. You can specify a struct that implements the `CustomizeConnection` trait when you are setting up the pool. The `on_acquire` function will then be called on each new connection created by the pool.

This seems preferable to applying it to the connections after they've already been pulled from the pool, since we don't know whether the connection is new or is being re-used.


neekolas:
While this seems like a good idea, when I combine it with `PRAGMA journal_mode = WAL` in any way I get something like 10 `database is locked` errors per unit test, which is weird. The tests actually pass, but it can't be a good sign. Trying to figure out how to resolve that before merging.

PR #425, opened 2024-01-18
Add legacy credential type

richardhuaaa:
This adds the legacy credential type. Support for generating these credentials will be added in the next PR.

- Apply the changes from https://github.com/xmtp/proto/pull/131
- Add ValidatedLegacySignedPublicKey type and tests
- Add LegacyCreateIdentityAssociation - cannot add tests until I have support for serializing into proto, will add in next PR
- Significant refactoring. We use separate classes in separate files for different associations. The strong typing ensures we can't mix them up (for example confusing grant messaging access with revoke messaging access).

Because the protos have changed, some steps are required to avoid errors for those who already have libxmtp running:
- Run `./dev/down && ./dev/up` from `libxmtp`. This rebuilds the validation service, and clears out any data stored on the server.
- Remove any existing databases, for example from the CLI and Android/iOS/flutter/RN


dev/gen_protos.sh
```diff
@@ -6,7 +6,7 @@ if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
     fi
 fi
 
-if ! buf generate https://github.com/xmtp/proto.git#branch=main,subdir=proto; then
+if ! buf generate https://github.com/xmtp/proto.git#branch=rich/legacy-keys-2,subdir=proto; then
```
richardhuaaa, line 9: Will switch back to main before merging

xmtp_mls/Cargo.toml
```diff
@@ -48,6 +48,7 @@ toml = "0.7.4"
 tracing = "0.1.37"
 xmtp_cryptography = { path = "../xmtp_cryptography" }
 xmtp_proto = { path = "../xmtp_proto", features = ["proto_full"] }
+xmtp_v2 = { path = "../xmtp_v2" }
```
neekolas, line 51: I think once we get to shaving fat off the WASM build we're going to want to extract the functions we need from this crate to avoid bloat. Or get really aggressive with features.

Problem for another day

xmtp_mls/src/credential/grant_messaging_access_association.rs
```diff
@@ -0,0 +1,291 @@
...
+};
+
+use super::AssociationError;
+
+#[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
+pub struct UnsignedGrantMessagingAccessData {
+    pub(crate) account_address: Address,
+    pub(crate) installation_public_key: Vec<u8>,
+    pub(crate) created_ns: u64,
+    iso8601_time: String,
```
neekolas, line 22: Do we actually need to store the iso8601 time on the struct. Seems like we can just generate it as-needed.

xmtp_mls/src/credential/validated_legacy_signed_public_key.rs
```diff
@@ -0,0 +1,147 @@
...
+#[cfg(test)]
+pub mod tests {
+    use crate::{assert_err, credential::AssociationError};
+
+    use super::ValidatedLegacySignedPublicKey;
+    use prost::Message;
+    use xmtp_proto::xmtp::message_contents::SignedPublicKey as LegacySignedPublicKeyProto;
+
+    #[tokio::test]
+    async fn validate_good_key() {
```
neekolas, line 109: Like the hardcoded text fixtures. These are going to be helpful for catching regressions

neekolas (approved):
Think this is looking good

xmtp_mls/src/credential/grant_messaging_access_association.rs
richardhuaaa: The act of generating the iso8601 time can [produce an error](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_timestamp), and it is better to validate the incoming data when creating the struct, rather than when it is eventually used to produce `text()`, hence we just generate it up-front

PR #426, opened 2024-01-18
Snor/mls api msgs

neekolas:


xmtp_mls/src/client.rs
```diff
@@ -373,42 +396,51 @@ where
...
-            .filter_map(|res| async {
-                match res.await {
-                    Ok(group) => Some(group),
-                    Err(err) => {
-                        log::error!("Error processing stream entry: {:?}", err);
-                        None
-                    }
-                }
-            });
+    // fn process_streamed_welcome(
```
neekolas, line 399: Going to add this back later. It's not consumed anywhere, so it's fine to omit

snormore (approved):

PR #428, opened 2024-01-19
chore: housekeeping to tidy up the lists of dependencies

tuddman:
This patch should be purely cosmetic and introduce **no** functional changes:

- Alphabetize the `Cargo.toml` dependency list(s)
- Removes some boilerplate when you `cargo init`
- Lint some inline comment widths

That's it.


neekolas (approved):
Would be nice if this was automatically applied. It looks like you can configure the "Even Better TOML" extension, which is what most of us use for TOML formatting with a [config file](https://taplo.tamasfe.dev/configuration/formatter-options.html) to do this.

neekolas:
Think your timing was very unlucky on this PR. It went up in the 5 minutes between a node-go change going live and the `libxmtp` change. Merging main should make everything pass.

PR #429, opened 2024-01-19
Revert to ubuntu-latest

neekolas:
## Summary

Switching to Warpbuild seems to work for everything except for this build and publish step, which only runs on `main`. Reverting this one until it gets resolved.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on:
       - main
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


tuddman (approved):
I'm normally for pinning versions of infra for CI, but only to prevent the sometimes-case of the underlying image updating and breaking some check(s) despite none of us changing _anything_.    The flipside is obviously that its easier to run on `latest` and get whatever updates come with _that_.

nakajima (approved):

PR #432, opened 2024-01-19
Use latest uniffi

neekolas:
## Summary

We've been locked to an old version of Uniffi for some time now. If we're going to upgrade to the latest stable version it's only going to get harder from here.

The biggest change is that `uniffi::Error` works differently now. Our old error type is incompatible with the new macro, and any sort of "generic error" doesn't work with the automatic translation of `std::Error` types (since `uniffi::Error`s are already `std::Error` and the `From` conversion can't work if both sides are already compatible). So, I now have a more normal error type which also gives us the ability to return more specific errors.

There may be other differences but that was the only one I noticed. I also moved the MLS stuff into its own file for tidiness, but that shouldn't matter to downstream users of the SDK.


neekolas:
Merge whenever you'd like @nakajima

PR #436, opened 2024-01-22
chore: finish ordering dependencies

tuddman:
This patch finishes the work started from https://github.com/xmtp/libxmtp/pull/428 and orders the deps I missed in that one.


neekolas (approved):
Happy to approve, but this feels like an uphill battle until there's some automation and/or linting. No one is going to be checking the ordering of dependencies as part of PR reviews, so it's only a matter of when this falls back out of order.

tuddman:
>  so it's only a matter of when this falls back out of order.

I totally agree that order can decay if not automated.  Especially if deps are added manually.  
Once it is sorted though, like they are now,  any additional `cargo add <new-dep>`  _should_ be inserted in the correct alphabetic place.

PR #437, opened 2024-01-22
Un-gitignore lock file

neekolas:
## Summary

For some reason the `bindings_ffi` Cargo.lock file was gitignored, causing inconsistent builds between different developer's machines. This adds it to Git to ensure consistency.


tuddman:
how about for `bindings_wasm` too?  https://github.com/xmtp/libxmtp/blob/main/bindings_wasm/.gitignore#L2

nplasterer (approved):

neekolas:
> how about for `bindings_wasm` too?  https://github.com/xmtp/libxmtp/blob/main/bindings_wasm/.gitignore#L2

Missed that one. Will add in a separate PR

PR #438, opened 2024-01-22
chore: confirmation that all deps are fully in order

tuddman:
## Summary

- With this patch all of the `Cargo.toml`s are all ordered. Maintaining dependency ordering should be straightforward + automatically handled by `cargo`, provided that any dependencies are added / removed respectively with:  `cargo add X` or `cargo remove Y`.
- If someone is using VSCode then `even-better-toml` is recommended to them as well.
- Adds a `.taplo.toml` file to the root dir to provide reordering keys rule.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,5 +3,8 @@
     "bindings_ffi/Cargo.toml",
     "examples/cli/Cargo.toml",
     "xmtp_api_grpc_gateway/Cargo.toml"
-  ]
+  ],
+  "[toml]": {
+    "editor.defaultFormatter": "tamasfe.even-better-toml"
+  }
 }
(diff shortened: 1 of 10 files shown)


neekolas:
I still don't see why we wouldn't update the vscode config to have a 
```json
  "[toml]": {
    "editor.defaultFormatter": "tamasfe.even-better-toml"
  }
```

and a .taplo.toml with something like this
```
[formatting]
reorder_keys = true
```

Then I wouldn't have to think about it ever

neekolas (approved):

.vscode/settings.json
```diff
@@ -3,5 +3,8 @@
     "bindings_ffi/Cargo.toml",
     "examples/cli/Cargo.toml",
     "xmtp_api_grpc_gateway/Cargo.toml"
-  ]
+  ],
+  "[toml]": {
+    "editor.defaultFormatter": "tamasfe.even-better-toml"
```
neekolas, line 8: We probably want to also make this a recommended extension.

Issue #423, opened 2024-01-17
Privacy for welcome messages


neekolas:
I thought I would add some background to this issue on privacy in the XMTP network for context.

The MLS Delivery Service for us is a fleet of XMTP nodes. Today, clients connect directly to a node from their client. In future, we expect the nodes to connect to a Gateway that may be run by third parties (for example, the developer of the app you are using). The gateway would be responsible for connecting to the XMTP nodes on behalf of all clients.

Today, messages are stored in a relational database on the nodes with the following schema:

```sql
CREATE TABLE group_messages (
    id BIGSERIAL PRIMARY KEY,
    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
snormore marked this conversation as resolved.
    group_id BYTEA NOT NULL,
    data BYTEA NOT NULL,
    group_id_data_hash BYTEA NOT NULL
);

CREATE INDEX idx_group_messages_group_id_created_at ON group_messages(group_id, created_at);
CREATE UNIQUE INDEX idx_group_messages_group_id_data_hash ON group_messages (group_id_data_hash);

CREATE TABLE welcome_messages (
    id BIGSERIAL PRIMARY KEY,
    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
    installation_key BYTEA NOT NULL, // This is the recipient's installation_id
    data BYTEA NOT NULL,
    installation_key_data_hash BYTEA NOT NULL
);

CREATE INDEX idx_welcome_messages_installation_key_created_at ON welcome_messages(installation_key, created_at);
CREATE UNIQUE INDEX idx_welcome_messages_group_key_data_hash ON welcome_messages (installation_key_data_hash);
```

A single copy of each group message is stored on the nodes irrespective of group size. Welcome messages are stored with a copy for each recipient, so that anyone can query for the full list of Welcome messages sent to them.

It's helpful to think of privacy separately for data at rest and data in transit.

## At rest

XMTP nodes can be queried freely by the public. For example, anyone can connect to the `StreamAllMessages` endpoint of the nodes and receive a copy of all new messages sent on the network. They can also query any group_id and receive all group messages, and query for all welcome messages sent to an installation_id. We treat any data stored at rest as public and try to minimize the amount of unencrypted metadata stored to the bare minimum required for the network to function.

[...]

PR #444, opened 2024-01-23
Support encrypted welcomes

neekolas:
## Summary

- Implements https://github.com/xmtp/libxmtp/issues/423
- Depends on https://github.com/xmtp/xmtp-node-go/pull/337 and https://github.com/xmtp/proto/pull/137
- Adds support for encrypting welcome messages using the `hpke_public_key`


dev/docker/docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   node:
-    image: xmtp/node-go:mls
+    image: xmtp/node-go:mls-dev
```
neekolas, line 3: Revert this before merging

dev/gen_protos.sh
```diff
@@ -6,7 +6,7 @@ if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
     fi
 fi
 
-if ! buf generate https://github.com/xmtp/proto.git#branch=main,subdir=proto; then
+if ! buf generate https://github.com/xmtp/proto.git#branch=nm/post-commit-action-updates,subdir=proto; then
```
neekolas, line 9: Revert this before merging

tuddman (approved):
While I'm not yet fully confident in understanding all of the subtleties in correctly implementing the welcome message encryption - there's nothing glaring that I can see.  But I gladly differ to others to give this a more thorough read-through before merging.

Without bike-shedding hard about `installation` V `installation_id` V `installation_key` - can you add a note explaining the decision / implications of going with one term vs. the other?

neekolas:
> Without bike-shedding hard about installation V installation_id V installation_key - can you add a note explaining the decision / implications of going with one term vs. the other?

I've switched anything new in this PR to use `installation_key`. There are still a bunch of places in `libxmtp` that reference `installation_id`. But now there are none over the wire at least.

The identifier for an installation is both an `id` (it's unique) and is a public key. So, there's a case to be made for calling it either `installation_id` or `installation_key`. I guess `installation_pub_key` was also an option. But the decision has been made, we're going with `installation_key`. There'll need to be more PRs to update references in other places.

PR #447, opened 2024-01-23
Use Warp MacOS runners

neekolas:
## Summary

- Warp now has faster/cheaper MacOS runners. Trying to use those to see if it improves build/release performance.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - nm/warp-macos-runners
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - nm/warp-macos-runners
```
neekolas, line 6: Will remove this before merging. Just need to confirm it works.

neekolas:
Confirmed this is 3-4X faster! Previously builds took ~30 minutes, this now takes ~8

tuddman (approved):

Issue #349, opened 2023-11-27
Prevent Welcome replays


richardhuaaa:
Concern around the security to keep track of the welcomes we've already recieved.

nplasterer:
With key rotation we shouldn't need this work.

PR #449, opened 2024-01-24
Add -> Remove -> Add flow

neekolas:
## Summary

This PR is trying to address an important situation.

- User A creates group
- User A adds User B to group
- User A removes User B from group
- Some time passes
- User A adds User B back to group

We would expect a few things here:
- User B should receive any messages sent between when they were first added and when they were removed
- User B shouldn't receive any messages sent during the time they were removed
- User B should be able to receive all messages after they were added back

This PR adds a test case for that scenario and fixes the group creation to enable it

## Note
[There is a separate issue regarding Welcome replays](https://github.com/xmtp/libxmtp/issues/349). That still needs to get resolved. A malicious user should not be able to replay a previously sent welcome message and corrupt another user's group state, or cause them to re-use key material that was meant to be only used one time.


xmtp_mls/src/client.rs
```diff
@@ -617,6 +619,63 @@ mod tests {
         assert_eq!(decrypted, to_encrypt);
     }
 
+    #[tokio::test]
+    async fn test_add_remove() {
```
tuddman, line 623:
```suggestion
    async fn test_add_remove_then_add_again() {
```

xmtp_mls/src/client.rs
```diff
@@ -617,6 +619,63 @@ mod tests {
...
+        amal_group
+            .remove_members_by_installation_id(vec![bola.installation_public_key()])
+            .await
+            .unwrap();
+        assert_eq!(amal_group.members().unwrap().len(), 1);
+
+        // See if Bola can see that they were added to the group
+        bola.sync_welcomes().await.unwrap();
+        let bola_groups = bola.find_groups(None, None, None, None).unwrap();
+        assert_eq!(bola_groups.len(), 1);
```
tuddman, line 646: ? - doesn't need to be answered here, but is there any mechanism for `bola` to see they've been removed from the group?
I read this as:
- bola was added to one (1) group;
- bola was then removed from said group;
- bola sees that they're in one (1) group ...?

tuddman (approved):

xmtp_mls/src/client.rs
neekolas: MLS has a way, but we aren't currently bubbling it up. That's coming next

PR #451, opened 2024-01-25
Fix signing flow for bindings with async signers

richardhuaaa:
We need a wallet signer from iOS/Android/etc that can be async. However, uniffi currently does not allow async callback interfaces to be passed into Rust: https://github.com/mozilla/uniffi-rs/issues/1729.

Instead, this PR restructures the libxmtp initialization process such that Rust can pass the signing text up towards the application, have it be signed externally, and then receive the signature as part of the next method call. The change in consumers looks as follows (pseudo-code), matching the changes in `bindings_ffi/src/mls.rs`.

**Before:**
```
struct Client {
  ...
}

client = await create_client(…, inbox_owner)
await client.register_identity()
```

**After:**
```
struct Client {
   ...
   text_to_sign() -> Option<String>
}

client = await create_client(…, account_address, legacy_identity_source, legacy_signed_private_key_proto)
if (text = client.text_to_sign()) {
   // Sign the text
}
await client.register_identity(signature_or_null)
```

Other notes:
- `legacy_identity_source` refers to how the XMTP v2 key was obtained (more description in the code). `legacy_signed_private_key_proto` is the serialized SignedPrivateKey from the v2 PrivateKeyBundle. This change is not related, however I've included it in this PR so that consumers don't need to perform multiple migrations.
- `register_identity()` is now idempotent - if a client has already registered before, calling it again will just no-op instead of erroring.
- We now also won't write the identity to the database unless the registration succeeds, so in the event of network failure during registration, the next run of the app will just create a new identity and then register it.

### DIFF
--- a/xmtp_mls/src/builder.rs
+++ b/xmtp_mls/src/builder.rs
@@ -40,6 +40,7 @@ pub enum ClientBuilderError {
 
 pub enum IdentityStrategy<Owner> {
     CreateIfNotFound(Owner),
+    CreateUnsignedIfNotFound(String),
     CachedOnly,
     #[cfg(test)]
     ExternalIdentity(Identity),
@@ -71,6 +72,15 @@ where
                 }
                 None => Ok(Identity::new(provider, &owner)?),
             },
+            IdentityStrategy::CreateUnsignedIfNotFound(account_address) => match identity_option {
+                Some(identity) => {
+                    if identity.account_address != account_address {
+                        return Err(ClientBuilderError::StoredIdentityMismatch);
+                    }
+                    Ok(identity)
+                }
+                None => Ok(Identity::new_unsigned(provider, account_address)?),
+            },
             #[cfg(test)]
             IdentityStrategy::ExternalIdentity(identity) => Ok(identity),
         }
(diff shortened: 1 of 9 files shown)


nakajima (approved):
Thanks for doing this

xmtp_mls/src/builder.rs
```diff
@@ -220,6 +236,7 @@ mod tests {
             .store(store_a)
             .build()
             .unwrap();
+        client_a.register_identity().await.unwrap(); // Persists the identity on registration
```
neekolas, line 239: Nice

neekolas (approved):
This is a really nice solution to a pretty ugly problem.

Just to be clear, the idea is that we remove `create_client` entirely, or that we add the `legacy_identity_source` and `legacy_signed_private_key_proto` to it?

If it's the latter, I'd suggest we rename these to `create_client_with_signer` and `create_client_without_signer` for clarity.

richardhuaaa:
I would suggest we eventually remove `create_client` entirely, because I think that using a blocking signing callback will cause issues for integrators. But happy to leave it in for now so that it doesn't immediately break people.

neekolas:
> I would suggest we eventually remove `create_client` entirely, because I think that using a blocking signing callback will cause issues for integrators. But happy to leave it in for now so that it doesn't immediately break people.

This is a nice enough interface I can get on board with that.

richardhuaaa:
Okay, so @nplasterer suggests that I just modify `create_client()` directly instead of making a separate `create_libxmtp_client()`, will do that and then land

PR #452, opened 2024-01-25
feat: get missing members and update installations

tuddman:
This patch is scoped to be only the implementation of the functions (+tests) to:
- get the missing members' installations, 
- and to update the group state with the members' installations.

Follow up will build on this to update the DB on some sort of schedule or conditions.


dev/test
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -eou pipefail
 
-cargo test
+cargo test "$@"
```
tuddman, line 4: I was getting non-deterministic failures when trying to run the tests with the base `dev/test`.   
The interesting thing was that it they were mostly all passing (75+ ✅ ) but any of the tests that were making DB calls were failing at different execution points.  Sometimes 5 would fail ❌ ; Sometime 10 would fail ❌.

Limiting the number of threads evidently allows the tests to pass with _some_ parallelism, but not too much.

With this addition I am able to locally run the tests successfully with `dev/test -- --test-threads=6`   <- oddly enough, doesn't need to be 1, but anything more than 6 and test failures would start.   Chalking this up to the speed and multi-core magic of the M3.

I

neekolas: Interesting. I haven't run into that myself, but it sounds like some of the issues @nakajima was facing with database locks.

Definitely something we should keep an eye on and try to find a root solution to it. This isn't a massively parallel app, and lock contention should be the least of our problems. So, something is not configured correctly.

xmtp_mls/src/groups/mod.rs
```diff
@@ -862,6 +864,91 @@ where
         Ok(())
     }
 
+    // If no changes, returns empty Vec
```
neekolas, line 867: Think we can remove these comments. I put them here mostly as a hint to some useful test cases.

neekolas (approved):
This is looking great @tuddman

xmtp_mls/src/groups/mod.rs
tuddman: removed.

PR #453, opened 2024-01-25
Group sync improvements

neekolas:
## Summary

This crosses off several todos in the group syncing protocol to make it more closely match the initial state machine designs.

1. When adding/removing members or doing a key update we now will retry the entire sync flow up to 3 times, waiting for the intent to be resolved. This helps handle cases where the intent cannot be successfully published until the receive flow has been run. This can happen when the MLS group has a pending commit, which will block any add/remove operations until the pending commit is resolved. This also helps handle cases where the local epoch is behind the remote epoch. Previously these issues would only be fixed the next time `sync()` was run
2. We now correctly return errors when adding/removing members fails
3. We now lowercase all wallet addresses that have been input from outside callers (for example in `add_members` or `remove_members`)
4. If an intent fails to be published through 3 successive iterations of sync, we move it to a state of "Error" which will prevent it from blocking future publishes
5. Moves all the synchronization code to a separate file, since `groups/mod.rs` was getting quite unwieldy

### DIFF
--- a/xmtp_mls/src/groups/sync.rs
+++ b/xmtp_mls/src/groups/sync.rs
@@ -2,6 +2,7 @@ use super::{intents::SendMessageIntentData, members::GroupMember, GroupError, Ml
 use crate::{
     api_client_wrapper::IdentityUpdate,
     codecs::ContentCodec,
+    configuration::MAX_INTENT_PUBLISH_ATTEMPTS,
     hpke::{encrypt_welcome, HpkeError},
     retry_async,
     storage::{
@@ -450,14 +451,6 @@ where
         let num_intents = intents.len();
 
         for intent in intents {
-            if intent.publish_attempts >= (crate::configuration::MAX_INTENT_PUBLISH_ATTEMPTS as i32)
-            {
-                log::error!("intent {} has reached max publish attempts", intent.id);
-                // TODO: Eventually clean up errored attempts
-                conn.set_group_intent_error(intent.id)?;
-                continue;
-            }
-
             let result = retry_async!(
                 Retry::default(),
                 (async {
@@ -468,7 +461,14 @@ where
 
             if let Err(err) = result {
                 log::error!("error getting publish intent data {:?}", err);
-                conn.increment_intent_publish_attempt_count(intent.id)?;
+                if (intent.publish_attempts + 1) as usize >= MAX_INTENT_PUBLISH_ATTEMPTS {
+                    log::error!("intent {} has reached max publish attempts", intent.id);
+                    // TODO: Eventually clean up errored attempts
+                    conn.set_group_intent_error(intent.id)?;
+                } else {
+                    conn.increment_intent_publish_attempt_count(intent.id)?;
+                }
+
                 return Err(err);
             }
 
--- a/xmtp_mls/src/groups/sync.rs
+++ b/xmtp_mls/src/groups/sync.rs
@@ -1,0 +1,689 @@
+use super::{intents::SendMessageIntentData, GroupError, MlsGroup};
+use crate::{
+    codecs::ContentCodec,
+    hpke::{encrypt_welcome, HpkeError},
+    retry_async,
+    storage::{
+        group_intent::{IntentKind, IntentState, ID},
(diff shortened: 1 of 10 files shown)


xmtp_mls/src/groups/sync.rs
```diff
@@ -0,0 +1,771 @@
...
+impl<'c, ApiClient> MlsGroup<'c, ApiClient>
+where
+    ApiClient: XmtpMlsClient,
+{
+    pub async fn sync(&self) -> Result<(), GroupError> {
+        let conn = &mut self.client.store.conn()?;
+        self.sync_with_conn(conn).await
+    }
+
+    pub(super) async fn sync_with_conn<'a>(
```
neekolas, line 70: `sync_with_conn` now continues even if there is a publishing error, so that the state can resolve itself by receiving messages from the network.

xmtp_mls/src/groups/sync.rs
```diff
@@ -0,0 +1,771 @@
...
+    }
+
+    /**
+     * Sync the group and wait for the intent to be deleted
+     * Group syncing may involve picking up messages unrelated to the intent, so simply checking for errors
+     * does not give a clear signal as to whether the intent was successfully completed or not.
+     *
+     * This method will retry up to `crate::configuration::MAX_GROUP_SYNC_RETRIES` times.
+     */
+    pub(super) async fn sync_until_intent_resolved<'a>(
```
neekolas, line 110: New method to sync up to 3X until an intent has been deleted

xmtp_mls/src/utils/address.rs
```diff
@@ -0,0 +1,28 @@
+use thiserror::Error;
+use xmtp_cryptography::signature::is_valid_ethereum_address;
+
+#[derive(Debug, Error)]
+pub enum AddressValidationError {
+    #[error("invalid addresses: {0:?}")]
+    InvalidAddresses(Vec<String>),
+}
+
+pub fn sanitize_evm_addresses(
```
neekolas, line 10: Moved this out of `groups/mod.rs` and made it lowercase the addresses after validation

xmtp_mls/src/groups/sync.rs
```diff
@@ -0,0 +1,771 @@
...
+                    last_err = Some(GroupError::Storage(err));
+                }
+            };
+            num_attempts += 1;
+        }
+
+        Err(last_err.unwrap_or(GroupError::Generic("failed to wait for intent".to_string())))
+    }
+
+    fn process_own_message(
```
neekolas, line 149: Everything below this has been moved over from `groups/mod.rs` unless otherwise noted

xmtp_mls/src/groups/sync.rs
```diff
@@ -0,0 +1,771 @@
...
+
+        if receive_errors.is_empty() {
+            Ok(())
+        } else {
+            debug!("Message processing errors: {:?}", receive_errors);
+            Err(GroupError::ReceiveErrors(receive_errors))
+        }
+    }
+
+    pub(super) async fn receive<'a>(&self, conn: &'a DbConnection<'a>) -> Result<(), GroupError> {
```
neekolas, line 387: I made `receive` take in a `DbConnection` so that we don't have to unnecessarily create multiple connections during syncing. I kinda wonder if this was the proximate source of our lock contention issues.

xmtp_mls/src/groups/sync.rs
```diff
@@ -0,0 +1,771 @@
...
+        for intent in intents {
+            let result = retry_async!(
+                Retry::default(),
+                (async {
+                    self.get_publish_intent_data(&provider, &mut openmls_group, &intent)
+                        .await
+                })
+            );
+
+            if let Err(err) = result {
```
neekolas, line 462: Either set the intent to Error status or increment the attempt count

neekolas: Previously it would continue to the next intent in case of error. Now it will return to make sure that intents get processed in order.

nplasterer (approved):

Issue #455, opened 2024-01-26
Set up Rust docs


neekolas:
See https://github.com/xmtp/libxmtp/pull/456

PR #456, opened 2024-01-26
Deploy Rust Docs to GH Pages

neekolas:
## Summary

Sets up RustDoc publishing to Github Pages. Adds some docs for `Client` (with many more to follow).

PR #457, opened 2024-01-26
Setup Rust Docs

tuddman:
Fixes https://github.com/xmtp/libxmtp/issues/455

### DIFF
--- a/dev/docs
+++ b/dev/docs
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -eou pipefail
+
+cargo doc --no-deps --workspace --document-private-items --open 
--- a/xmtp_mls/src/identity.rs
+++ b/xmtp_mls/src/identity.rs
@@ -132,7 +132,7 @@ impl Identity {
         }
 
         // Register the installation with the server
-        let kp = self.new_key_package(&provider)?;
+        let kp = self.new_key_package(provider)?;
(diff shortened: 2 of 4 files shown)


xmtp_mls/src/identity.rs
```diff
@@ -132,7 +132,7 @@ impl Identity {
         }
 
         // Register the installation with the server
-        let kp = self.new_key_package(&provider)?;
+        let kp = self.new_key_package(provider)?;
```
tuddman, line 135: removing this reference is not related to `rust doc` specifically; other than it no longer generates a lint warning.

dev/docs
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -eou pipefail
+
+cargo doc --no-deps --workspace --document-private-items --open
```
tuddman, line 4: @neekolas Open to suggestions for what flag option(s) you want to have here when building the docs.

neekolas: Can we drop this script in favour of the one in https://github.com/xmtp/libxmtp/pull/456, which is more optimized for GH pages?

tuddman:
Dupe of https://github.com/xmtp/libxmtp/pull/456

neekolas:
I'll copy over your fixes for the `sql_key_store` warnings

PR #460, opened 2024-01-26
Release swift bindings

cameronvoell:
Added a GH action for adding swift bindings to a GH Release. 

~~Adding the binary artifacts to a release would allow xmtp-ios to reference our libxmtp FFI bindings without needing to use libxmtp-swift as a "middle man" repo (release artifacts can be referenced directly from Swift Package Manager).~~ It turned out that we can not get rid of the `libxmtp-swift` repo because we still need the `Source` folder which is generated from `libxmtp`, and it is not possible to access that remotely from Swift Package Manager. However, if we hold on to `libxmtp-swift`, publishing the XCFramework as a release artifact in `libxmtp` does get us around having to use `git-lfs` in our `libxmtp-swift` which is important because it turns out Swift Package Manager does not work with `git-lfs` (although Cocoa Pods does fine with `git-lfs`, go figure!). 

A "pre-release" will be generated from this branch that will be tested against `xmtp-ios` in order to validate that this strategy works for getting our static binary `libxmtp` FFI into `xmtp-ios`.

The checksum in the Release description is used for specifying a binary target in our libxmtp-swift Package.swift file (for Swift Package Manager). See: https://github.com/xmtp/libxmtp-swift/pull/4/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677e

See Related PRs:
- https://github.com/xmtp/libxmtp/pull/433
- https://github.com/xmtp/libxmtp-swift/pull/4
- https://github.com/xmtp/xmtp-ios/pull/225

### DIFF
--- a/.github/workflows/release_swift_bindings.yml
+++ b/.github/workflows/release_swift_bindings.yml
@@ -1,11 +1,7 @@
 name: Release Swift Bindings
 
 on:
-  push:
-    branches:
-      - main
-      - nm/staticlib
-      - cv/release-swift-bindings
+  workflow_dispatch:
 
 jobs:
   build-macos:
@@ -106,7 +102,6 @@ jobs:
           mkdir -p Sources/LibXMTP
           mv build/swift/xmtpv3.swift Sources/LibXMTP/
           make framework
-          ls -la ../
           cp ../LICENSE ./LICENSE
           zip -r LibXMTPSwiftFFI.zip Sources LibXMTPRust.xcframework LICENSE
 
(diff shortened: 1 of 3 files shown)


neekolas:
How would this work with react-native's weird dependence on CocoaPods. My understanding is that `libxmtp-swift` is required for us to use libxmtp as a CocoaPod

cameronvoell:
> How would this work with react-native's weird dependence on CocoaPods. My understanding is that `libxmtp-swift` is required for us to use libxmtp as a CocoaPod, and CocoaPods are required to use `xmtp-ios` as a dependency in RN projects.

@neekolas This is no longer relevant because we can not get rid of `libxmtp-swift` after all (see description update). Publishing the XCFramework binaries as a release artifact in libxmtp does get us around a constraint of git-lfs not working with Swift Package Manager though.

.github/workflows/release_swift_bindings.yml
```diff
@@ -0,0 +1,138 @@
+name: Release Swift Bindings
+
+on:
+  workflow_dispatch:
```
cameronvoell, line 4: Updated the release swift bindings to just go on manual triggers for now.

*Note this can be triggered before this PR is merged to main using the `gh` cli:

```
gh workflow run workflow-name --ref branch-name
```

PR #461, opened 2024-01-29
feat: flutter v3: create client

dmccartney:
This is the first bit of implementing v3 `bindings_flutter`. See https://github.com/xmtp/xmtp-flutter/issues/112

This introduces a `createClient` that takes the owner and produces a consistent installation. This also includes all the build/CI scaffolding to bridge rust/dart and package the library for flutter.


nplasterer (approved):

richardhuaaa:
Just a heads up on https://github.com/xmtp/libxmtp/pull/451 - you may need to split `register_identity()` out rather than folding it into `create_identity()`

bindings_flutter/test/v3_test.dart
```diff
@@ -0,0 +1,71 @@
...
+        dbPath: dbPath,
+        encryptionKey: encryptionKey,
+        accountAddress: wallet.address.hex,
+      );
+      // ... it should require a signature...
+      var clientA = switch (createdA) {
+        CreatedClient_Ready() => throw StateError("Should require a signature"),
+        CreatedClient_RequiresSignature(field0: var req) =>
+          await req.sign(signature: await walletSign(req.textToSign)),
+      };
```
dmccartney, line 46: @richardhuaaa (in re [new signature flow](https://github.com/xmtp/libxmtp/pull/461#issuecomment-1918032981))

Here's how I exposed the flow (trying to keep it simple but still guide the app dev through how to use it properly):
```dart
var created = await createClient(...);
var client = switch (created) {
  CreatedClient_Ready(field0: var client) => client,
  CreatedClient_RequiresSignature(field0: var req) =>
    await req.sign(signature: await walletSign(req.textToSign)),
};
```
i.e. `createClient` returns a `CreatedClient` which is a sealed type that is either `Ready` or `RequiresSignature`. The `RequiresSignature` instance includes the `.textToSign` alongside a `.sign(...)` method that takes the `signature` and returns the ready `Client` upon success.

dmccartney: on the rust side that looks like this (check `libxmtp_api.rs`):

```rs

pub enum CreatedClient {
    Ready(Client),
    RequiresSignature(SignatureRequiredClient),
}

pub struct SignatureRequiredClient {
    pub text_to_sign: String,
    pub inner: Arc<InnerClient>,
}
```

Issue #463, opened 2024-01-30
Address must be lowercased


nplasterer:
Lets allow any case address to be passed

PR #466, opened 2024-01-31
Sanitize account addresses

richardhuaaa:
Fixes https://github.com/xmtp/libxmtp/issues/463

That should do it, but it is possible that I missed a spot. If there's anything left, please send me the error message and I'll fix it.


neekolas (approved):

nplasterer (approved):
I built this locally and can confirm all my tests pass now without having to be lowercased.

PR #469, opened 2024-02-01
Ensure correctly sorted messages

neekolas:
## Summary

I noticed that we don't explicitly apply a default sort to queries for messages. Somehow it is still coming out in the correct order because it happens to line up with the default sort of SQLite, but that isn't guaranteed.


richardhuaaa (approved):
Great catch

PR #474, opened 2024-02-02
Attempt at breaking streams

nplasterer:
WIP


richardhuaaa:
Okay, not 100%, but maybe you need to use a `callbackFlow` here? https://stackoverflow.com/questions/64623712/how-to-emit-data-from-an-asycnhronous-callback-using-kotlin-flow

When the app runs, the UI never shows - not 100% either here but I think it's because the `onCreate` is blocking on the `streamGroups().collect` because it is `runBlocking`, so we never get to rendering the UI. But I suspect testing the first `streamGroups()` call should still be fine even if the UX doesn't show

richardhuaaa:
Added some tweaks:

1. The DB directory was sometimes not being cleaned, which causes errors on the next run. Making it clean out the database if it exists during initialization to get rid of that error.
2. Added some logging to see why `onCreate()` is not completing - it seems `stream groups 2 done` is never logged, so that's what is not finishing

PR #477, opened 2024-02-02
build: create workflow for flutter release

dmccartney:


.github/workflows/bindings_flutter_release.yaml
```diff
@@ -0,0 +1,146 @@
+name: release bindings_flutter
+on:
+  workflow_dispatch:
```
dmccartney, line 3: This is manually triggered and publishes a `xmtp_bindings_flutter-v###` release that includes the binary artifacts for each platform.
Then we separately can publish to `pub.dev` with the build configs for each platform. Each platform's build configs know how to download the corresponding binary files from the github release on demand.

richardhuaaa (approved):

richardhuaaa:
Very nice!

PR #486, opened 2024-02-06
feat: Update DB when missing members was last checked

tuddman:
## Summary

- Adds the field `installation_list_last_checked` to the `StoredGroup` to track when the last (latest) installation list 'check' was performed.
- Adds function to update this value.
- Adds tests

---
#### unrelated, cosmetic change
- slight rewording of related tests
- ordering of some of the `use` modules


tuddman:
> LGTM!
> 
> Also linking https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/IDENTITY.md#installation-synchronization if you haven't seen it, feel free to update it if any changes are needed

Thanks!  There will be a follow up PR that sets the cadence and conditions when this value is _actually_ updated.  I will include a note in the `IDENTITY.md` as a part of that PR.

PR #489, opened 2024-02-07
Show stream is closed

neekolas:
## Summary

Callers are probably interested in whether a stream has been closed by the server. This PR exposes that information with a new `is_closed()` method.


nplasterer (approved):
I think this may aid in this issue here https://github.com/xmtp/xmtp-react-native/issues/108

PR #490, opened 2024-02-07
chore: organize all of the use statements

tuddman:
## Summary

No functional changes are introduced in this patch.

I noticed that some of the files were following the community convention of grouping + organizing the `use ...` statements according to ~ the following:

```
mod foo;

use std::{foo, ...};

use external_lib::{foo, ...};

use workspace_lib::{foo, ...};

use super::{foo, ...};
use crate::{foo, ...};
```

while others were not.

This patch organizes all of the `use` statements so they're all consistent and organized.  IDEs should now be able to auto-arrange and alphabetize any dependencies added in the future.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,1 @@
+imports_granularity = "Crate"


neekolas:
There's no way to lint for this community convention or auto-format in my editor?

I don't mind adding these blank lines as a one-time thing. It's harmless and agree it's a little nicer for readability. But it's another story to expect new outside contributors to follow a convention that isn't linted for or automatically applied. It also doesn't seem like the kind of thing we can expect reviewers to reliably look for, or a good use of finite review time.

tuddman:
> There's no way to lint for this community convention or auto-format in my editor?
> 
> I don't mind adding these blank lines as a one-time thing. It's harmless and agree it's a little nicer for readability. But it's another story to expect new outside contributors to follow a convention that isn't linted for or automatically applied. It also doesn't seem like the kind of thing we can expect reviewers to reliably look for, or a good use of finite review time.

I am absolutely not going for nor caring about perfection here.  If they're _mostly_ in order than its easy enough to maintain honestly. Its not something even I might notice or call out in a PR review.
Some 'set and forget' options we could do is:
- add `cargo fix` to the recommended dev toolchain,
- add `merge_imports = true` to the `rustfmt.toml` and `rustfmt` should auto-take-care-of-it.

neekolas:
> add imports_granularity = "Crate" to the rustfmt.toml and rustfmt should auto-take-care-of-it.

@tuddman Nice! I like my imports pretty, I just want to expend exactly 0 effort getting there.

neekolas (approved):

PR #492, opened 2024-02-07
feat: adds missing installations when group is synced

tuddman:
## Summary

Adds missing installations when a group `sync()` is performed.

Per [this comment](https://github.com/xmtp/libxmtp/pull/486#pullrequestreview-1865726287) from the previous PR, what, if anything, should be added to the IDENTITY.md doc about this?

### DIFF
--- a/xmtp_mls/src/groups/sync.rs
+++ b/xmtp_mls/src/groups/sync.rs
@@ -63,6 +63,11 @@ where
 {
     pub async fn sync(&self) -> Result<(), GroupError> {
         let conn = &mut self.client.store.conn()?;
+        
+        let provider = self.client.mls_provider(&conn);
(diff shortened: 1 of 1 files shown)


xmtp_mls/src/groups/sync.rs
```diff
@@ -63,6 +63,11 @@ where
 {
     pub async fn sync(&self) -> Result<(), GroupError> {
         let conn = &mut self.client.store.conn()?;
+        
+        let provider = self.client.mls_provider(&conn);
```
neekolas, line 67: Looks like we're double referencing here

neekolas (approved):
Looks good to me. We're going to want to add the "only check every N hours" part next, since this is a fairly expensive operation that will be a no-op something like 99% of the time.

PR #493, opened 2024-02-08
Add TLS connection config

neekolas:
## Summary

To better catch disconnects, we are adding back some of the timeouts we previously had in the V2 client. Should help with reliability in bad network conditions.

Hat tip for @richardhuaaa for noticing this was missing.


tuddman (approved):
👍

PR #494, opened 2024-02-08
Add crate for XPS operations

insipx:
Wanted to get some administrative tasks + skeleton out of the way before committing code

- add x86_64-unknown-linux-gnu to `rust-toolchain.toml`

I wasn't actually sure where this should live, but the way we discussed it at the offsite seemed like it would be in `libxmtp`.

this is the crate where we plan to impl `XmtpMlsClient` using [xps](https://github.com/xps-gateway)

~nothing will build yet until [reorg](https://github.com/xmtp/xps-gateway/pull/57) is merged into xps~

### DIFF
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,5 +1,5 @@
 [toolchain]
 channel = "stable"
 components = [ "rustc", "cargo", "clippy", "rustfmt", "rust-analyzer" ]
-targets = [ "wasm32-unknown-unknown" ]
+targets = [ "wasm32-unknown-unknown", "x86_64-unknown-linux-gnu"]
 profile = "default"
--- a/xps-client/Cargo.toml
+++ b/xps-client/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["web-programming::websocket", "web-programming::http-server", "web
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-lib_xps = { git = "https://github.com/xmtp/xps-gateway", branch = "main", package = "lib-xps" }
+lib-xps = { git = "https://github.com/xmtp/xps-gateway", branch = "main", package = "lib-xps" }
 xps-types = { git = "https://github.com/xmtp/xps-gateway", branch = "main", package = "xps-types" }
 
 xmtp_proto = { path = "../xmtp_proto" }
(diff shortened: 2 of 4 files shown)


jac18281828:
I agree this makes sense to me.  Thanks Andrew!

rust-toolchain
```diff
@@ -1,5 +1,5 @@
 [toolchain]
 channel = "stable"
 components = [ "rustc", "cargo", "clippy", "rustfmt", "rust-analyzer" ]
-targets = [ "wasm32-unknown-unknown" ]
+targets = [ "wasm32-unknown-unknown", "x86_64-unknown-linux-gnu"]
```
richardhuaaa, line 4: Noob question, why is this needed?

xps-client/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+homepage = "https://github.com/xmtp/libxmtp"
+repository = "https://github.com/xmtp/libxmtp"
+license = "MIT"
+keywords = ["ethereum", "did", "did:ethr", "jsonrpc"]
+categories = ["web-programming::websocket", "web-programming::http-server", "web-programming::http-client", "cryptography::cryptocurrencies"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+lib_xps = { git = "https://github.com/xmtp/xps-gateway", branch = "main", package = "lib-xps" }
```
richardhuaaa, line 16: Noob question, I don't see `lib-xps` in that repo, what is this package responsible for?

insipx: yeah it's part of the [reorg PR](https://github.com/xmtp/xps-gateway/pull/57), which splits the main.rs and lib.rs in the current `xps-gateway` into lib-xps and xps

rust-toolchain
insipx: I saw that in dev/up the target was being added

This change isn't necessary, but would remove the need for that

richardhuaaa: Ah, looks like it's for the validation service, guess we're building it for a docker container on `linux/amd64`

richardhuaaa (approved):

xps-client/Cargo.toml
insipx: jsonrpsee generates a client and server from the defined api 

As it is defined right now, we don't have the client or server feature gated, which is something we have to do before a proper release, so it will pull in both

I'll add the feature gates to the [reorg](https://github.com/xmtp/xps-gateway/pull/57) before it goes in

PR #496, opened 2024-02-09
feat: only update the installation list if its been more than 6 hours

tuddman:
## Summary

Adds a configuration setting, `UPDATE_INSTALLATION_LIST_INTERVAL_NS`, currently set to 6 hours (in nanoseconds. for easier calculations later),   and only adds missing installations if the duration or `delay` from the previous time updated has been more than 6 hours.  We can set this to whatever # of hours is desired, of course.
This is different from "update the list every X hours" <- which would require keeping a stateful time counter in-app somehow?  I wasn't sure where or how that should work.


neekolas:
Agree we don't need to be running some sort of cron schedule. This only needs to happen when you actually _do_ something with the group and it has been > N hours. 6 seems like a reasonable N to me

tuddman:
- I set the interval to 30 minutes.
- The tests pass locally.  not sure yet what's going on with the FFI / diesel / db locking test failure in CI.

xmtp_mls/src/groups/sync.rs
```diff
@@ -64,8 +64,13 @@ where
     pub async fn sync(&self) -> Result<(), GroupError> {
         let conn = &mut self.client.store.conn()?;
         let provider = self.client.mls_provider(&conn);
-        self.add_missing_installations(provider).await?;
-        self.update_latest_installation_list_timestamp(conn).await?;
+
+        let (last, latest) = self.update_latest_installation_list_timestamp(conn).await?;
```
richardhuaaa, line 68: We are essentially bumping the timestamp every time we sync, rather than only bumping the timestamp when we add missing installations. I think this is an issue - for example, if we sync every 15 minutes, then the timestamp will be bumped every time without ever triggering the missing installations to be added

xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -92,18 +92,33 @@ impl DbConnection<'_> {
...
+            let maybe_get_last_ts = dsl::groups
+                .find(&group_id)
+                .select(dsl::installation_list_last_checked)
+                .execute(conn)
+                .optional()?;
+
+            let now = crate::utils::time::now_ns();
+            let last_ts = match maybe_get_last_ts {
+                Some(ts) => ts as i64,
+                None => now,
```
richardhuaaa, line 111: I think `last_ts` should just be 0 if it was not previously set, that way we are guaranteed to check on the first time.

With this in mind, I think the DB schema can have `installation_list_last_checked` be a non-nullable value with a default of 0, so that we don't have to deal with an `Option` here. Also, I think we should return an error if the `group_id` doesn't exist in the DB, so that bugs can be caught here.

xmtp_mls/src/groups/sync.rs
tuddman: you are totally right.  fixed the logic.

xmtp_mls/src/storage/encrypted_store/group.rs
tuddman: Sure, I guess we don't need to care about an un-initialized state.
I made it a `NOT NULL` and initializes the value to `0`.  And fixed for the `NotFound` Err case.

richardhuaaa (approved):
Looks great! We should also do this check before message sends, maybe the same way we do it for syncs

tuddman:
> Looks great! We should also do this check before message sends, maybe the same way we do it for syncs

Sounds good.  I'll do it in a follow up PR - as I don't like to add new functionality to an existing PR once its been ✅ .

PR #502, opened 2024-02-10
Switch to ImmutableMetadata

neekolas:
## Summary

- Upgrades OpenMLS to the latest version, which includes a simplified version of Immutable Metadata
- Upgrades a few dependencies to the latest version, which will allow us to do the next upgrade of OpenMLS

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,10 +30,10 @@ futures = "0.3.30"
 futures-core = "0.3.30"
 hex = "0.4.3"
 log = "0.4.20"
-openmls = { git = "https://github.com/xmtp/openmls", branch = "main" }
-openmls_basic_credential = { git = "https://github.com/xmtp/openmls", branch = "main" }
-openmls_rust_crypto = { git = "https://github.com/xmtp/openmls", branch = "main" }
-openmls_traits = { git = "https://github.com/xmtp/openmls", branch = "main" }
+openmls = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
+openmls_basic_credential = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
+openmls_rust_crypto = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
+openmls_traits = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -30,10 +30,10 @@ futures = "0.3.30"
 futures-core = "0.3.30"
 hex = "0.4.3"
 log = "0.4.20"
-openmls = { git = "https://github.com/xmtp/openmls", branch = "main" }
-openmls_basic_credential = { git = "https://github.com/xmtp/openmls", branch = "main" }
-openmls_rust_crypto = { git = "https://github.com/xmtp/openmls", branch = "main" }
-openmls_traits = { git = "https://github.com/xmtp/openmls", branch = "main" }
+openmls = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
```
neekolas, line 33: I don't know about the rest of you, but I've been having weird issues with Rust Analyzer caching dependencies very aggressively. Even after running cargo update, it keeps the old version for type checking (but builds work fine when my code uses newer types). This persists even after running `cargo clean`. Pinning to a specific rev breaks the cache.

neekolas: Reloading the workspace and restarting the server also didn't help

xmtp_mls/src/groups/mod.rs
```diff
@@ -364,15 +364,9 @@ fn build_protected_metadata_extension(
...
-    let protected_metadata = ProtectedMetadata::new(
-        &identity.installation_keys,
-        identity.application_id(),
-        identity.credential()?,
-        identity.installation_keys.to_public_vec(),
-        metadata.try_into()?,
-    )?;
-
-    Ok(Extension::ProtectedMetadata(protected_metadata))
+    let protected_metadata = Metadata::new(metadata.try_into()?);
```
tuddman, line 367: nit about naming:   If its no longer instantiating a `ProtectedMetadata`, as that's being replaced with an `ImmutableMetadata`.

```suggestion
    let immutable_metadata = Metadata::new(metadata.try_into()?);
```

tuddman (approved):

Cargo.toml
neekolas: Woah. Maybe

PR #505, opened 2024-02-12
Make permissions configurable

neekolas:
## Summary

- Adds optional `permissions` parameter to group creation, allowing devs to configure groups where only the group creator can add/remove members
- Bubbles this up to the bindings so that SDK devs can add the new option.
- This will be a breaking change for iOS/Android since it adds a new parameter to `createGroup`

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -167,18 +167,35 @@ pub struct FfiConversations {
     inner_client: Arc<RustXmtpClient>,
 }
 
+#[derive(uniffi::Enum)]
+pub enum GroupPermissions {
+    EveryoneIsAdmin,
+    GroupCreatorIsAdmin,
+}
+
 #[uniffi::export(async_runtime = "tokio")]
 impl FfiConversations {
     pub async fn create_group(
         &self,
         account_addresses: Vec<String>,
+        permissions: Option<GroupPermissions>,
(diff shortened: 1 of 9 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -167,18 +167,35 @@ pub struct FfiConversations {
...
 
-        let convo = self.inner_client.create_group()?;
+        let group_permissions = match permissions {
+            Some(GroupPermissions::EveryoneIsAdmin) => {
+                Some(xmtp_mls::groups::PreconfiguredPolicies::EveryoneIsAdmin)
+            }
+            Some(GroupPermissions::GroupCreatorIsAdmin) => {
+                Some(xmtp_mls::groups::PreconfiguredPolicies::GroupCreatorIsAdmin)
+            }
+            _ => None,
```
richardhuaaa, line 195: Should we default this to `EveryoneIsAdmin`? Or is it intended to mean nobody is an admin here? If so, should we add that as an explicit option rather than the default behavior?

richardhuaaa (approved):

bindings_ffi/src/mls.rs
neekolas: Ya. It's a little confusing. `None` in practice becomes `EverybodyIsAdmin` because we have that implemented using the `Default` trait, which gets used when this passes back None.

PR #506, opened 2024-02-12
feat: adds installation list to maybe update when send_message

tuddman:
## Summary

follow up to https://github.com/xmtp/libxmtp/pull/496#issuecomment-1939261167

This patch extends:  checking for the last time the installation list was updated, and updates it if enough time (30 min) has passed,  and adds that process when `send_message()` is called as well as `sync()`.

### DIFF
--- a/xmtp_mls/src/groups/sync.rs
+++ b/xmtp_mls/src/groups/sync.rs
@@ -63,15 +63,8 @@ where
 {
     pub async fn sync(&self) -> Result<(), GroupError> {
         let conn = &mut self.client.store.conn()?;
-        let provider = self.client.mls_provider(conn);
 
-        let now = crate::utils::time::now_ns();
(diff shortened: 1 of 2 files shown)


xmtp_mls/src/groups/sync.rs
```diff
@@ -711,22 +720,6 @@ where
...
-        Ok(last_ts)
-    }
-
-    pub(super) async fn update_latest_installation_list_timestamp(
-        &self,
-        conn: &DbConnection<'_>,
-    ) -> Result<(), GroupError> {
-        let _result = conn.update_installation_list_time_checked(self.group_id.clone())?;
-        Ok(())
-    }
```
tuddman, line 728: Removed, as this was adding needless indirection to get to the DB fields.

neekolas (approved):

PR #507, opened 2024-02-12
prototype integration of contact ops

insipx:
I'm taking a shot at integrating the `xps-gateway` with libxmtp via a external crate `xps-client`

There are definitely some pain points, and things I've learned could be better. But here are some notes (along with some (`NOTE`'s) in the code

- we undo the serialization into an MLS credential in order to access the address, this could be better
-- Can we push TLS serialization/grpc proto-ification into the `api_client` and pass the keypackage un-serialized into the `api-client`? This removes some boilerplate from the d14n integration 
- libxmtp would need an interface to sign attribute payloads with `InboxOwner`/InboxOwner just has to require an implementation of `RegistrySignerExt`
-- signature payload may have to occur within the SDK's (https://viem.sh/docs/actions/wallet/signTypedData.html)
--- the SDKS need to know how to sign DIDRegistry setAttribute typed-data to pass into libxmtp. i think that is how the signing happens now, except that we are signing a message instead of typed data 
- crates are currently circular i.e highly coupled (xmtp_mls depends on xps-client, xps-client depends on xmtp_mls). Pushing serialization/deserialialization into `api_client` also solves this issue
- rename (`xps-gateway`): fetch_key_packages to get_identity_updates. 
-- In XPS we named our identity updates `fetch_key_packages`, but this is actually `get_identity_updates` in libxmtp
- the integration currently requires a dependency on `DIDRegistry` in `lib_didethresolver` b/c it fetches the nonce for signing purposes. We need to expose a `nonce` endpoint for the contract via xps-gateway/didethresolver.

I purposefully tried doing this w/o touching any libxmtp code to see what would happen. So far it seems reasonable sans what I said above. I still need to test this out with the CLI, and things can definitely be made cleaner overall with some changes in XPS and libxmtp

Also, I didn't see a method for revocation. Is this somewhere else? @neekolas @richardhuaaa 

TODO:
* [x] - grant_installation/register_installation
* [x] - get_identity_updates
* [ ] - feature flags
* [x] - instantiate client somewhere (XMTP CLI Example)


xps-client/src/lib.rs
```diff
@@ -1,14 +1,222 @@
...
+    #[error("Registry Error {0}")]
+    Registry(#[from] lib_didethresolver::error::RegistrySignerError<Provider<Http>>),
+}
+
+pub struct XmtpXpsClient<LegacyClient> {
+    /// Client for the XPS API
+    client: WsClient,
+    /// This is the current mls client to fill in non-d14n functionality
+    legacy_client: LegacyClient,
+    owner: LocalWallet,
```
insipx, line 57: This should be removed in favor of a trait, i.e `RegistrySignerExt`e

PR #512, opened 2024-02-13
feat: detect is_active group in bindings_flutter

dmccartney:


bindings_flutter/src/libxmtp_api.rs
```diff
@@ -137,6 +137,13 @@ impl Client {
         });
     }
 
+    pub async fn is_active_group(&self, group_id: Vec<u8>) -> Result<bool, XmtpError> {
+        self.inner.sync_welcomes().await?;
```
richardhuaaa, line 141: Wondering if this is needed, here and in the other methods? I don't think the caller would be able to get a group_id without syncing the group it belongs to. Also, in bindings_ffi, the integrator can control when the syncing happens, via FfiConversations.sync()

bindings_flutter/src/libxmtp_api.rs
```diff
@@ -137,6 +137,13 @@ impl Client {
...
+        let group = self.inner.group(group_id)?;
+        group.sync().await?; // TODO: consider an explicit sync method
```
richardhuaaa, line 143: Yep, would recommend this be done explicitly, just like in bindings_ffi, so that the integrator can control when network requests are made, although doesn't have to block this PR

richardhuaaa (approved):

bindings_flutter/src/libxmtp_api.rs
dmccartney: Aye, I'm kinda using this as a mini exploration of a `sync()`-less API surface.
My plan has been to bring `bindings_flutter` to mirror `bindings_ffi` by adding `.sync()` and removing the auto-sync calls. But my gut says that the sync call requirements are gonna be a confusing source of bugs. And once they get bit, apps will just end up calling .sync overly aggressively anyways.

So I've started mulling another pattern -- "default-auto-sync": methods will everywhere "call sync for you" by default but let you pass a flag to prevent auto-syncing. I.e. let app devs opt into the manual-sync complexity if they know what they're doing.

bindings_flutter/src/libxmtp_api.rs
dmccartney: My tests hit this when clients B and C made `is_active_group` calls against a `groupId` the test held onto from client A's creation call. But you're right, in a real situation they would only have access to the `groupId` because they got it from an earlier sync'd list/stream call. I'll ✂️

bindings_flutter/src/libxmtp_api.rs
richardhuaaa: That's a great suggestion - we are now planning to do this for iOS/Android too, albeit at the SDK layer

neekolas: I think this makes sense. Most people will want fresh results, and it's only once you start getting into pagination that you might want to turn it off.

PR #525, opened 2024-02-21
Sync after sending

neekolas:
## Summary

For the dev alpha, we should be focusing on reliability over performance. As we move to beta we can start to optimize for perf.

This undoes an early decision to not sync after sending. It will make sending a message a little slower, but will surface errors earlier and ensure your local group state is more in sync.

We can always revert this change later to optimize for performance, maybe with some additional work on optimistic updates to the local DB.

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -231,15 +231,13 @@ where
             .await?;
 
         let intent_data: Vec<u8> = SendMessageIntentData::new(message.to_vec()).into();
-        let intent =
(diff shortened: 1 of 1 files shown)


richardhuaaa:
Can we call this in a background thread, so there's no performance impact?

I was actually wondering if we can call the whole intent publishing flow (or at least starting from whenever there's a network request) in a background thread, so that optimistic message UI can come through quicker for implementers

neekolas:
Going to leave this in draft for now, and maybe our other approach to optimistic sending is going to be enough.

PR #526, opened 2024-02-21
[1/2] stream_all_messages: bindings and streaming messages from an unchanging group list

richardhuaaa:
Putting this up early, so that work on propagating this to SDK's can begin in parallel. This implements `stream_all_messages()`, but it doesn't pick up messages from any groups that get created after the method was called.

Some of the setup for picking up messages from *new* groups is in this PR, but that functionality won't be complete until the next PR (along with some cleanup). At that point, only the libxmtp binary will need to be updated - the binding interface won't be any different from the one in this PR.

Other changes:
1. Make a file called `subscriptions.rs` and move `process_streamed_welcome` and `stream_conversations` to it - `client.rs` was becoming a monster file.
2. Move the logic for streaming conversations with a callback from the `stream()` method in the bindings into a `stream_conversations_with_callback` method in xmtp_mls. The logic for running an infinite listener loop in a background thread, as well as closing the stream, will also be needed for the groups stream in `stream_all_messages`.

### DIFF
--- a/xmtp_mls/src/subscriptions.rs
+++ b/xmtp_mls/src/subscriptions.rs
@@ -1,6 +1,15 @@
-use std::{collections::HashMap, pin::Pin};
+use std::{
+    collections::HashMap,
+    pin::Pin,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc, Mutex,
+    },
+};
 
 use futures::{Stream, StreamExt};
+use tokio::sync::oneshot::{self, Sender};
+// use xmtp_api_grpc::grpc_api_helper::Client as ApiClient;
 use xmtp_proto::{api_client::XmtpMlsClient, xmtp::mls::api::v1::WelcomeMessage};
 
 use crate::{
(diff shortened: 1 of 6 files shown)


xmtp_mls/src/subscriptions.rs
```diff
@@ -0,0 +1,278 @@
...
+                    match res {
+                        Ok(envelope) => {
+                            let group_id = extract_group_id(&envelope)?;
+                            let stored_group = &group_id_to_info_clone
+                                .get(&group_id)
+                                .ok_or(ClientError::StreamInconsistency(
+                                    "Received message for a non-subscribed group".to_string(),
+                                ))?
+                                .group;
+                            MlsGroup::new(
```
neekolas, line 181: Smart way to work around the lifetime issue

richardhuaaa:
Ah - I just realized that for the next PR, if the server delivers messages out-of-order, the in-memory cursor approach here will still have gaps when tearing down and re-opening the messages stream on a new group. Have some ideas but will address in next PR (could start separate streams for new groups, or search DB for 'gaps' in messages, or strategically re-sync).

nplasterer (approved):

richardhuaaa:
Landing so that @nplasterer can begin integrating into SDK's - if you have any feedback @neekolas I will include it in the next PR

PR #532, opened 2024-02-22
CLI Release Process

neekolas:
## Summary

- Adds a release pipeline for the CLI so that we can have binaries built for linux and MacOS that can be easily downloaded
- Updates the CLI to support registering to the network with a seed phrase. This should make it easier to incorporate with existing wallets and the V2 SDK, since you can share the wallet and register with both
- Adds support for prettier logging to the CLI and fixes up json logging with the `--json` option
- Adds support for adding/removing multiple members at one time

## Notes

One side effect of this is that every merge to main will trigger a release, which creates a tag and triggers the `Release` workflow that runs on `tags: *`. That doesn't seem so bad, but we can always filter down which release tags we want to run the Workflow from if it is a total of compute.

### DIFF
--- a/examples/cli/README.md
+++ b/examples/cli/README.md
@@ -40,7 +40,7 @@ Use the CLI to send a [double ratchet message](https://github.com/xmtp/libxmtp/b
 6. Add user 2 to the group
 
    ```bash
-   ./xli.sh --db user1.db3 add-group-members --account-addresses $USER_2_ACCOUNT_ADDRESS $GROUP_ID
+   ./xli.sh --db user1.db3 add-group-members $GROUP_ID --account-addresses $USER_2_ACCOUNT_ADDRESS
    ```
 
 7. Send a message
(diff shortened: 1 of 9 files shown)


examples/cli/README.md
```diff
@@ -40,7 +40,7 @@ Use the CLI to send a [double ratchet message](https://github.com/xmtp/libxmtp/b
 6. Add user 2 to the group
 
    ```bash
-   ./xli.sh --db user1.db3 add-group-member $GROUP_ID $USER_2_ACCOUNT_ADDRESS
```
richardhuaaa, line 43: How will multiple account addresses be distinguished from the group ID here? Might want to clarify

richardhuaaa (approved):

examples/cli/README.md
neekolas: Good eye. I forgot to hit save on this file, the order is actually backwards. The $GROUP_ID belongs first

PR #533, opened 2024-02-22
[2/n] stream_all_messages that detects messages in new groups

richardhuaaa:
This implements `stream_all_messages` at the same reliability as V2 XMTP streams - in other words, best-effort reliability that may occasionally lead to dropped messages. Improved reliability is coming in future PR's.

Note: Due to `conversations.stream()` not picking up conversations created by yourself, this will still be missing messages from newly created conversations of your own, although it will pick up messages from your existing conversations. This will also be addressed next.

## Implementation notes
- **Stream hierarchy**: A groups stream that holds onto a messages stream via an Arc. When a new group is received, the groups stream will close the messages stream it is holding onto, and create a new one with the new list of groups. External callers are listening to messages from the messages stream.
- **Merging streams**: The messages stream should be swapped out under the hood without impacting external callers. To do this, instead of having `stream_all_messages` return a stream, it uses a callback interface. The callback from the external caller can then be moved across to the new stream invisibly to the caller. If a stream interface is needed in the future, I believe it can be wrapped around the callback interface.
    - To compose callbacks across both the groups stream and messages stream, I moved the existing callback-based streaming methods out of the bindings and into `xmtp_mls` so that they could be used as building blocks.
- **Closing streams**: When the groups stream is closed, it will close the messages stream it is holding onto before closing itself.
- **Multi-threaded client**: Anywhere a callback interface is used, a thread is spawned to listen to the stream. We need a `Client` that can live as long as the new thread, so `Client` must be wrapped in `Arc`. Therefore, all of the callback interfaces take in `Arc<Client>` instead of `self`.

## Upcoming work
- Most of the callback interfaces feature unwraps() from within the spawned listener thread. We can remove these unwraps() to improve reliability, or at least detect streaming errors better.
- Make `conversations.stream()` stream conversations initiated by yourself, which is currently missing
- We can update the `cursor` passed to the server when establishing each stream, to smooth over gaps between when one stream is closed and another is opened. Note that although this might improve reliability slightly, it cannot be guaranteed unless the server is guaranteed to deliver the messages in a consistent order.
- We can open separate message streams for new groups, without closing existing streams on old groups, up to a limited number of concurrent streams. This improves reliability as there is no gap in the streaming for the old groups.
- We can cut down on unnecessary clones and Arcs/Mutexes

### DIFF
--- a/xmtp_mls/src/groups/subscriptions.rs
+++ b/xmtp_mls/src/groups/subscriptions.rs
@@ -1,12 +1,15 @@
+use std::collections::HashMap;
 use std::pin::Pin;
+use std::sync::Arc;
 
-use futures::{Stream, StreamExt};
+use futures::Stream;
 
 use xmtp_proto::{api_client::XmtpMlsClient, xmtp::mls::api::v1::GroupMessage};
 
 use super::{extract_message_v1, GroupError, MlsGroup};
-use crate::api_client_wrapper::GroupFilter;
 use crate::storage::group_message::StoredGroupMessage;
+use crate::subscriptions::{MessagesStreamInfo, StreamCloser};
+use crate::Client;
 
 impl<'c, ApiClient> MlsGroup<'c, ApiClient>
 where
@@ -44,35 +47,35 @@ where
     pub async fn stream(
         &'c self,
     ) -> Result<Pin<Box<dyn Stream<Item = StoredGroupMessage> + 'c + Send>>, GroupError> {
-        let last_cursor = 0;
-
-        let subscription = self
+        Ok(self
             .client
-            .api_client
-            .subscribe_group_messages(vec![GroupFilter::new(
+            .stream_messages(HashMap::from([(
                 self.group_id.clone(),
-                Some(last_cursor as u64),
-            )])
-            .await?;
-        let stream = subscription
-            .map(|res| async {
-                match res {
-                    Ok(envelope) => self.process_stream_entry(envelope).await,
-                    Err(err) => Err(GroupError::Api(err)),
-                }
-            })
-            .filter_map(move |res| async {
-                match res.await {
-                    Ok(Some(message)) => Some(message),
-                    Ok(None) => None,
-                    Err(err) => {
-                        log::error!("Error processing stream entry: {:?}", err);
-                        None
-                    }
-                }
-            });
+                MessagesStreamInfo {
+                    convo_created_at_ns: self.created_at_ns,
+                    cursor: 0,
+                },
+            )]))
+            .await?)
+    }
 
-        Ok(Box::pin(stream))
+    pub async fn stream_with_callback(
+        client: Arc<Client<ApiClient>>,
+        group_id: Vec<u8>,
+        created_at_ns: i64,
+        mut callback: impl FnMut(StoredGroupMessage) + Send + 'static,
+    ) -> Result<StreamCloser, GroupError> {
+        Ok(Client::<ApiClient>::stream_messages_with_callback(
+            client,
+            HashMap::from([(
+                group_id,
+                MessagesStreamInfo {
+                    convo_created_at_ns: created_at_ns,
+                    cursor: 0,
+                },
+            )]),
+            move |message| callback(message),
+        )?)
     }
 }
 
--- a/xmtp_mls/src/groups/subscriptions.rs
+++ b/xmtp_mls/src/groups/subscriptions.rs
@@ -30,9 +30,13 @@ where
                 "Processing message in process_stream_entry {}",
                 self.client.account_address()
             );
-            self.process_message(&mut openmls_group, &provider, &msgv1, false)
-                .map_err(GroupError::ReceiveError)
+            let res = self
+                .process_message(&mut openmls_group, &provider, &msgv1, false)
+                .map_err(GroupError::ReceiveError);
+            log::info!("Got process message result {:?}", res);
+            res
         });
+        log::info!("Got process_result {:?}", process_result);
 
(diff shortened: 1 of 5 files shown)


xmtp_proto/src/api_client.rs
```diff
@@ -132,7 +132,7 @@ pub type WelcomeMessageStream = Pin<Box<dyn Stream<Item = Result<WelcomeMessage,
 // Wasm futures don't have `Send` or `Sync` bounds.
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-pub trait XmtpMlsClient: Send + Sync {
```
richardhuaaa, line 135: Why I think this is needed:
1. A struct which holds references generally has a generic lifetime parameter attached to it. Downstream users of the struct then have to handle the lifetime, seeing as the reference may expire while the struct is still being held. A struct which holds no references is considered `'static`.
2. Likewise, any downstream users of any trait will have to deal with lifetimes, because you cannot know if the struct that implements the trait holds any references.
3. To handle subscriptions, we need to be able to pass an `Arc<Client>` across threads. Anything passed across to `thread::spawn` must have a `'static` lifetime because it cannot be determined how long the thread will live for.
4. `Client` holds a member that implements the `XmtpMlsClient` trait. Therefore, `Client` is only `'static` if the implementer of `XmtpMlsClient` is.

Here, we specify that the implementer of `XmtpMlsClient` is always `'static` - that is, it is a struct that either holds no references, or only holds references with `'static` lifetime.

It's possible to only specify `'static` bounds on `XmtpMlsClient` wherever streaming is used, and leave open the option of having an `XmtpMlsClient` which is not `'static` and forgo streaming. I chose this way because it is less duplicative elsewhere and I don't see any need for a non-static `XmtpMlsClient`.

xmtp_mls/src/groups/subscriptions.rs
```diff
@@ -44,35 +47,35 @@ where
...
+                },
+            )]))
+            .await?)
+    }
+
+    pub async fn stream_with_callback(
+        client: Arc<Client<ApiClient>>,
+        group_id: Vec<u8>,
+        created_at_ns: i64,
+        mut callback: impl FnMut(StoredGroupMessage) + Send + 'static,
```
neekolas, line 66: These callbacks are meant to be short-lived and discarded when the stream closes. Won't the `static tag actually keep them around forever?

xmtp_proto/src/api_client.rs
neekolas: Static lifetime doesn't seem that bad here, given you only have one `XmtpMlsClient` for each client instance. Even in the bindings we keep around the client for a long time. It's not like you are going to have thousands of these floating around over time, except maybe in some weird bot edge cases.

richardhuaaa: To be clear, I don't think it's about how long `XmtpMlsClient` lives for, because it is a struct and not a reference. What I'm saying is all structs are automatically `'static` unless they hold a reference that is not `'static`. What I believe I'm doing here is enforcing that whatever implements `XmtpMlsClient` doesn't hold non-static references

xmtp_mls/src/groups/subscriptions.rs
richardhuaaa: The same reasoning as above applies here. My understanding:

1. Adding `'static` to a declaration doesn't functionally change how long it lives for. It is just a compile-time check that whatever is passed in will actually live that long.
2. All structs are `'static` unless they hold a non-static reference
3. I'm adding `'static` to an implementer of the `FnMut` trait here, which is a closure that is struct-like (passed by value, not by reference). What I'm doing is enforcing that the closure that an external caller provides here doesn't hold any non-static references.

I hope this understanding is correct, would love it if @tuddman or @insipx could double-check this!

tuddman: `'static` by definition is guaranteed to live for as long as the lifetime of the program is resident in memory.  If something is passed in, it will remain there, forever.  
Point 4 is the unfortunately important one here.  Crossing thread boundaries relinquishes borrowing control.  The alt is to wrap it in an `Arc<T>` which exists, in part,  for this very reason.

richardhuaaa: The main point I'm trying to convey here is I don't think this is sacrificing anything. By default, an `FnMut` passed to a foreign thread is already `'static`, because it is owned by the foreign thread and guaranteed to live as long as the thread does. The only way in which it would not be static is if the closure held onto non-static references from the surrounding environment. We're simply prohibiting that - anything the closure holds onto either has to be moved, wrapped in an Arc, or `'static` to begin with.

richardhuaaa: Here's a thread explaining my current understanding: https://users.rust-lang.org/t/static-lifetime-with-owned-object/50637/4

> Lifetimes don't apply to types that don't borrow anything. This makes self-contained types appear to satisfy any lifetime, but in fact they ignore all lifetimes.
>
> T: 'static will allow a type like String. It's not because String lives for as long as 'static (it could be dropped at any time), but because it's not borrowing from any scope, so there's no lifetime limiting its usage.

neekolas: This is very useful context. Maybe my original comment is based on a poor understanding of how lifetimes apply to things that don't borrow.

neekolas (approved):

richardhuaaa:
This was auto-merged by accident - main issue is one unit test that is flaky in CI and a bunch of debugging logs. Reverted it for now, and will re-land tomorrow

PR #534, opened 2024-02-22
fix: updating protos required matching prost, pbjson updates too

tuddman:
## Summary

When we updated the proto repo yesterday with https://github.com/xmtp/proto/pull/150, in order to implement the changes that that PR brought about, I ran `cd xmtp_proto/ && ../dev/gen_protos.sh` to pull those changes and generate new rust bindings.  

However, that introduced a **score** of compilation errors.  

AFAICT, the root cause of that was/is that by using [buf generate ...](https://github.com/neoeinstein/protoc-gen-prost), unless otherwise specified uses the latest version of [prost & prost-types](https://github.com/neoeinstein/protoc-gen-prost/blob/main/protoc-gen-prost/Cargo.toml#L22-L23) which are incompatible with trying to compile with _our_ declared `prost` & `prost-types` version pins.
 
Using different tools / libraries to a) generate, and then b) compile rust-bindings for proto files, uses different versions, which causes issues.
 
 I don't have an immediate proposed solution on how to best handle this going forward; but updating to the latest dependency version(s) resolves the issues.


neekolas:
Buf does let you pin to a specific version of the remote plugin. We could pin to an older version if the upgrades are proving troublesome.

I'm just happy the generator is getting any updates at all. I was worried the thing was dead and we were going to have to DIY something to get upgrades.

xmtp_proto/src/gen/xmtp.message_api.v1.tonic.rs
```diff
@@ -249,7 +249,7 @@ pub mod message_api_server {
             request: tonic::Request<super::PublishRequest>,
         ) -> std::result::Result<tonic::Response<super::PublishResponse>, tonic::Status>;
         /// Server streaming response type for the Subscribe method.
-        type SubscribeStream: futures_core::Stream<
+        type SubscribeStream: tonic::codegen::tokio_stream::Stream<
```
tuddman, line 252: This here, not compiling, with the error that `tokio_stream` can't be found in `tonic::codegen::*` was the tell that something wasn't adding up with the versions.

tonic moved from `futures` to `tokio_stream` for `Stream` in `0.10.0`

neekolas (approved):
Thanks for pushing through this.

Do you mind also pinning our generator in the buf.gen.yaml to the current version, so the next time the maintainer wakes up from a year-long nap on releases we don't get caught by surprise?

PR #536, opened 2024-02-23
Fix failing grpc test

richardhuaaa:
This test seems to be blocking a few PR's from landing.

It's failing now because the server behavior is slightly different [on this line](https://github.com/xmtp/xmtp-node-go/blob/aded2d716bca5950de9491817e8a262c3d3af82f/pkg/api/message/v1/service.go#L396) after [this PR](https://github.com/xmtp/xmtp-node-go/pull/354/files)


neekolas (approved):
I just fixed the same thing in my CLI branch. Mine is set to auto-merge, so maybe we just wait for it to finish

PR #537, opened 2024-02-23
fix: prefer installations over installation_list

tuddman:
## Summary

Small PR.  Re-reading the naming for keeping and updating an `installation_list`, IMO it reads easier to simply use the plural `installations`

So,   `s/installation_list/installations`


xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
```diff
@@ -5,8 +5,8 @@ CREATE TABLE groups (
     "created_at_ns" BIGINT NOT NULL,
     -- Enum of GROUP_MEMBERSHIP_STATE
     "membership_state" INT NOT NULL,
-    -- Last time the installation list was checked for the purpose of seeing if any are missing
-    "installation_list_last_checked" BIGINT NOT NULL
+    -- Last time the installations were checked for the purpose of seeing if any are missing
+    "installations_last_checked" BIGINT NOT NULL
```
neekolas, line 9: I think one consequence of being in "alpha" (even with our many warnings of breaking changes) is that we should start adding new migrations rather than updating old ones.

neekolas (commented):
Generally in favour of this change, but we should do this in a way that isn't going to require people to wipe their local DB

xmtp_mls/migrations/2023-10-29-205333_state_machine_init/up.sql
tuddman: updated to use a new migration in https://github.com/xmtp/libxmtp/pull/537/commits/c13abe42cee0d88d5dfca1511d40f61b8f36bb9e

tuddman: True true. + The buidl week has officially begun.  No need to break early adopters if we can avoid it.

neekolas (approved):

PR #538, opened 2024-02-23
fix: mostly clippy annotation warnings

tuddman:
## Summary

Friday evening housekeeping.  Fixes some clippy annotation warnings and other noise in our GH actions.

As much as I'd like to also quiet the annotation warnings from the proto gen ones, I'm going to leave the auto-generated code alone and leave it to the upstream generator to someday 🤞 correctly annotate generated code.


neekolas (approved):
Thanks for fixing these

PR #539, opened 2024-02-24
Remove ESM export

rygine:
the ESM export does not seem to work in practice and the conversion to ESM from CJS is very brittle. since CJS can be imported from ESM, this PR removes the ESM export and makes the module CJS by default when importing from node.


neekolas (approved):
If you don't think we need ESM support here, fine by me. Will be less brittle this way. I guess the bundler in xmtp-js will just take care of it anyways.

PR #541, opened 2024-02-26
Delete expired KeyPackage, rotate KeyPackages

Bren2010:
We still need to rotate the KeyPackage if there haven't been any Welcome messages processed in the last X months, but I'm not sure the best way to implement that check?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,10 +30,10 @@ futures = "0.3.30"
 futures-core = "0.3.30"
 hex = "0.4.3"
 log = "0.4.20"
-openmls = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
-openmls_basic_credential = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
-openmls_rust_crypto = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
-openmls_traits = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
+openmls = { git = "https://github.com/Bren2010/openmls", rev = "3992d09" }
+openmls_basic_credential = { git = "https://github.com/Bren2010/openmls", rev = "3992d09" }
+openmls_rust_crypto = { git = "https://github.com/Bren2010/openmls", rev = "3992d09" }
+openmls_traits = { git = "https://github.com/Bren2010/openmls", rev = "3992d09" }
 prost = "^0.12"
 prost-types = "^0.12"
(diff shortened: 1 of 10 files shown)


richardhuaaa (approved):
Looks great! How do we plan to delete the key material from the old key package on the client once the new package is uploaded? Is the idea to pipe an expiry into MLS for the old package at the time that the new one is created?

richardhuaaa:
For the CI issue, just run `cargo fmt` at the root of the repo and commit the change!

Cargo.toml
```diff
@@ -30,10 +30,10 @@ futures = "0.3.30"
...
 log = "0.4.20"
-openmls = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
-openmls_basic_credential = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
-openmls_rust_crypto = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
-openmls_traits = { git = "https://github.com/xmtp/openmls", rev = "0da7dcb" }
+openmls = { git = "https://github.com/Bren2010/openmls", rev = "3992d09" }
+openmls_basic_credential = { git = "https://github.com/Bren2010/openmls", rev = "3992d09" }
+openmls_rust_crypto = { git = "https://github.com/Bren2010/openmls", rev = "3992d09" }
+openmls_traits = { git = "https://github.com/Bren2010/openmls", rev = "3992d09" }
 prost = "^0.12"
```
tuddman, line 37: Shouldn't a PR be put up against https://github.com/xmtp/openmls with any necessary changes here?  
(Rather than changing the base repo on which we're using. )
Without more context here for why this change is necessary, I would advise on holding off on merging this until:
- It's clear here why we should switch to using Brendan's fork, or
- we wait for the changes to land in `xmtp/openmls` and we retain that dependency.

cc @neekolas @richardhuaaa

tuddman (changes requested):

Bren2010:
> Looks great! How do we plan to delete the key material from the old key package on the client once the new package is uploaded? Is the idea to pipe an expiry into MLS for the old package at the time that the new one is created?

My idea was just to just leave all KeyPackages in storage until they expire, based on the thought that they might still be used.

Cargo.toml
Bren2010: There is a PR here: https://github.com/xmtp/openmls/pull/20

The Cargo.toml here was just changed temporarily

Issue #529, opened 2024-02-21
Optimistically store message in libxmtp DB on message send


richardhuaaa:
On message send, libxmtp does a sophisticated publishing flow that may take multiple network round trips to resolve. Integrators should be able to optimistically render that message in their UI before the send has been confirmed to be published. To enable that, we can write the message to our database as soon as the send is requested, with a "state" field of UNPUBLISHED, and update it after publishing to PUBLISHED. That way, integrators can just query the messages, and decide whether or not they want to render the UNPUBLISHED messages.

The message ID currently depends on a server timestamp, and therefore can only be determined once the message is published. In order to optimistically write the message to the database, we need to modify how the message ID is determined, and use a client timestamp instead. The client timestamp is not included on the message payload anywhere, so others won't be able to determine the message ID until we add it onto the payload.

1. Write an envelope proto for the encrypted MLS payload that wraps encoded content. Add a field to the envelope called ‘sender_idempotency_key’ or similar.
2. (Breaking change) Run `dev/gen_protos.sh` in libxmtp and implement the proto (serialization on message send and deserialization on message receive), including setting the new field on message send to the sender's client timestamp. Once landed, notify everyone that it is a breaking change.
3. Update how message_id is generated to use the new field - previously used `the group_id, hash of the decrypted message contents, and the server timestamp` - now we can replace the server timestamp with `(sender_account_address, sender_client_timestamp)`. Add “state” field to group_messages table (default it to PUBLISHED in the SQL migration, but make it a required field in the code going forward). Insert an optimistic row on message send with state UNPUBLISHED, and update it to PUBLISHED after commit publishing flow.

Other thoughts:
- The `send()` function won't return for integrators until the full publish flow is complete. In future, we could consider having it return early and do the rest in a background thread. Not sure if this matters or not/how multi-threaded integrators are already.
- The sent message won't be returned via message streams until it is published and then read back from the server. Unsure if this matters or not.
- Integrators can't subscribe to changes on message state. I assume this is too sophisticated for now, and they could also determine that the state has transitioned to PUBLISHED once they read it back from the stream.

neekolas:
I have a [PR up for the envelope](https://github.com/xmtp/proto/pull/150) (not sure about the naming, open to suggestions).

I also added some ideas of where we should use this, and where we should store the two fields separately. Feel free to weigh in here on usage. Think we want to get that clear so that we don't have callers expecting one thing and getting another.

insipx:
[...]

PR #544, opened 2024-02-29
Optimistically store message in libxmtp DB on message send 

tuddman:
## Summary

See https://github.com/xmtp/libxmtp/issues/529 for details

---

still todo:  There's one failing unit test that I haven't worked out yet.


xmtp_mls/src/groups/sync.rs
```diff
@@ -424,6 +460,7 @@ where
             let group_id = self.group_id.as_slice();
             let message_id =
                 get_message_id(encoded_payload_bytes.as_slice(), group_id, timestamp_ns);
```
neekolas, line 462: This is still calling the deprecated `get_message_id`. Isn't that going to cause duplicate messages? If not, maybe we should include an explanatory comment

neekolas (approved):

xmtp_mls/src/groups/sync.rs
tuddman: AFAIK the alternative method for generating message ids via `calculate_message_id()` is used for application messages, and not here for  transript messages.   I'll be the first to say it's a bit jank to have two different methods for generating ids.  But these transcript messages won't have idempotency keys to use as part of the message id generation.  Unless we wanted to say that all messages of all types get an idempotency key.  Thoughts on that?

neekolas: That’s fair, it’s a bit of a weird case. I’d be ok with giving transcript messages an idempotency key (maybe the commit message timestamp) just for consistency

xmtp_mls/src/groups/mod.rs
```diff
@@ -233,36 +242,79 @@ where
...
+            &self.group_id,
+            message,
+            &self.client.account_address(),
+            &now.to_string(),
+        );
+        let group_message = StoredGroupMessage {
+            id: message_id,
+            group_id: self.group_id.clone(),
+            decrypted_message_bytes: message.to_vec(),
+            sent_at_ns: now,
```
richardhuaaa, line 284: We are mixing client and server timestamps with this column, which might cause some confusion on clients where they are not aligned. But I'm okay with deferring that problem, this has the benefit that optimistic messages are automatically sorted for now. Perhaps we can leave a comment explaining this, in case we run into it down the line?

xmtp_mls/src/groups/sync.rs
```diff
@@ -213,16 +219,25 @@ where
...
+                    (idempotency_key, content)
+                } else {
+                    return Err(MessageProcessingError::InvalidPayload);
+                };
+
+                let message_id =
+                    calculate_message_id(group_id, &message, &self.client.account_address(), &key);
+
+                conn.set_delivery_status_to_published(&message_id)?;
+                conn.set_sent_at_ts(message_id, envelope_timestamp_ns)?;
```
richardhuaaa, line 240: Could we merge these, so that there are fewer DB operations, and it's impossible to call one without the other, maybe by adding `delivery_timestamp_ns` as an argument to `set_delivery_status_to_published`?

xmtp_mls/src/storage/encrypted_store/group_message.rs
```diff
@@ -24,12 +27,14 @@ pub struct StoredGroupMessage {
...
     /// Time in nanoseconds the message was sent.
     pub sent_at_ns: i64,
-    /// Group Message Kind Enum
+    /// Group Message Kind Enum: 1 = Application, 2 = MembershipChange
     pub kind: GroupMessageKind,
     /// The ID of the App Installation this message was sent from.
     pub sender_installation_id: Vec<u8>,
     /// Network wallet address of the Sender
     pub sender_account_address: String,
+    /// Enum: 1 = Published, 2 = Unpublished
```
richardhuaaa, line 36: Another nit, it might be better to not document the number value in all of these comments, so there is a single source of truth in the enum itself, in the event more variants are added or existing variants are updated.

xmtp_mls/src/storage/encrypted_store/group_message.rs
```diff
@@ -130,6 +167,41 @@ impl DbConnection<'_> {
...
+        let res = self.raw_query(|conn| {
+            diesel::update(dsl::group_messages)
+                .filter(dsl::id.eq(msg_id.as_ref()))
+                .set(dsl::delivery_status.eq(DeliveryStatus::Published))
+                .execute(conn)
+        });
+
+        match res {
+            Ok(n) => Ok(n),
+            Err(_e) => Err(StorageError::NotFound),
```
richardhuaaa, line 184: You can verify via unit test, but I don't think this method and the one below will error if nothing was found. Under the hood the sql is just doing `update WHERE`. There are some other examples in the codebase, but the `Ok` variant of `res` should return the number of updates, you just need to verify that it's equal to 1.

xmtp_mls/src/utils/id.rs
```diff
@@ -17,3 +16,18 @@ pub fn get_message_id(
...
+    group_id: &[u8],
+    decrypted_message_bytes: &[u8],
+    sender_account_address: &str,
+    idempotency_key: &str,
+) -> Vec<u8> {
+    let mut id_vec = Vec::new();
+    id_vec.extend_from_slice(group_id);
+    id_vec.extend_from_slice(decrypted_message_bytes);
+    id_vec.extend_from_slice(sender_account_address.as_bytes());
+    id_vec.extend_from_slice(idempotency_key.as_bytes());
```
richardhuaaa, line 31: `decrypted_message_bytes` and `idempotency_key` are not a fixed length, so I wonder if collisions are possible by manipulating the message and idempotency_key.  The old `get_message_id()` gets around this problem by putting `decrypted_message_bytes` at the end. Maybe we can also put `decrypted_message_bytes` at the end, and do a `sha256` on the `idempotency_key` so it's always the same length?

richardhuaaa (commented):
Great work, this was a tricky change!

As this is a breaking change to the wire format, once this lands we might also need to reset the dev network and/or notify those who are currently using or building on this

xmtp_mls/src/storage/encrypted_store/group_message.rs
tuddman: you are absolutely right.  I looked at for 10 seconds before realizing we don't need the `match res {...}` block at all.  Replaced with a `?` above, which should bubble up the correct variant of `StorageError`, if any.

xmtp_mls/src/storage/encrypted_store/group_message.rs
tuddman: removed.  Folks will have to jump to the `DeliveryStatus` itself to find the truth.

xmtp_mls/src/groups/sync.rs
tuddman: I collapsed these into one DB call that updates both fields 👍

xmtp_mls/README.md
```diff
@@ -35,6 +35,8 @@ CREATE TABLE group_messages (
     -- Could remove this if we added a table mapping installation_ids to wallet addresses
     "sender_installation_id" BLOB NOT NULL,
     "sender_account_address" TEXT NOT NULL,
+    -- Enum: 1 = 'published' or 2 = 'unpublished'
```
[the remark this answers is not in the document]
tuddman: These are now sorted according to the expected sequence of events:  `1 = Unpublished; 2 = Published`

xmtp_mls/src/groups/mod.rs
tuddman: Agree.  Its a side-effect of the main crux of this PR to store application messages.  I'll do a follow up PR to reconcile and unify these.

xmtp_mls/src/utils/id.rs
tuddman: I put the message bytes as the last extended piece.
Clarifying ?:   are you suggesting:
`sha256(group_id, sender_account_address, sha256(idempotency_key), decrypted_message_bytes)`

richardhuaaa: Yep! Curious what @neekolas thinks

neekolas: Still not sure how manipulating the idempotency key or message becomes an attack vector tbh. The sender account address and group id are part of the hash, so you can't replay anyone else's messages and get the same ID. And in the original the two are separated by the sender account address which can't be manipulated, so they won't bleed into one another.

Given this is all getting hashed at the end, we could also just put in separators between the fields if we thought this was a risk. Should be more performant than doing an extra `sha256`.

richardhuaaa: Separators make sense, but we'd have to make sure the character set for the separator cannot be present in any of the fields. To explain more, imagine sender A has address abcdefg, and sender b has address bcdefgh. Sender A sends "hello", which will create an `id_vec` that contains "<group_id>helloabcdefg<idempotency_key>". Sender B can deliberately cause a collision by sending "helloa", and massage their idempotency key to match sender A.

Obviously difficult to do in practice, but if a small tweak makes edge cases like this impossible, I think it's worth making the security easier to reason about.

neekolas: Because we have the sender address in the middle, anything that isn’t a hex character should be a valid separator.

xmtp_mls/src/groups/sync.rs
tuddman: We are now using the same function, `calculate_message_id()` throughout.  I thought that would introduce breaking changes unrelated to this work, but apparently not.

PR #552, opened 2024-03-15
fix: streaming disconnect after delay

nplasterer:
After a lot of reading and trial error I found that the tests will pass if we remove this line.
I even bumped the test up to a 300 second delay to make sure it still passes and confirmed that it does.

### DIFF
--- a/xmtp_api_grpc/src/grpc_api_helper.rs
+++ b/xmtp_api_grpc/src/grpc_api_helper.rs
@@ -33,7 +33,6 @@ async fn create_tls_channel(address: String) -> Result<Channel, Error> {
         .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?
         .keep_alive_while_idle(true)
         .connect_timeout(Duration::from_secs(5))
-        .http2_keep_alive_interval(Duration::from_secs(10))
         .keep_alive_timeout(Duration::from_secs(25))
         .tls_config(ClientTlsConfig::new())
         .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?
--- a/xmtp_api_grpc/src/lib.rs
+++ b/xmtp_api_grpc/src/lib.rs
@@ -249,7 +249,7 @@ mod tests {
     #[tokio::test]
     async fn long_lived_subscribe_test() {
         let auth_token = get_auth_token();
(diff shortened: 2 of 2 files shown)


xmtp_api_grpc/src/lib.rs
```diff
@@ -278,7 +278,7 @@ mod tests {
                 panic!("Message 1 Error: {}", err);
             }
 
-            tokio::time::sleep(std::time::Duration::from_secs(15)).await;
+            tokio::time::sleep(std::time::Duration::from_secs(50)).await;
```
nplasterer, line 281: I bumped this up to 300 and it still passed with this change.

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -33,7 +33,6 @@ async fn create_tls_channel(address: String) -> Result<Channel, Error> {
         .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?
         .keep_alive_while_idle(true)
         .connect_timeout(Duration::from_secs(5))
-        .http2_keep_alive_interval(Duration::from_secs(10))
```
nplasterer, line 36: I bumped this all the way down to 1 and it still wouldn't pass. Removing it seems to be the only way to make it pass. Do we need this?

neekolas: I thought the default of 20 seconds would kick in if we removed this setting, which might break this some time later than 50 seconds. 

Maybe I'm wrong about the defaults

nplasterer: I set it to 300 seconds and it still didn't break. I can try even longer.

nplasterer: Okay just set the delay to 1000 seconds which is like 16 minutes. And it still passed. So I think this is safe.

neekolas (approved):

alexrisch (approved):

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: There's a risk that this is going to break the disconnect detection on mobile, but that feels less bad than not streaming at all.

Issue #503, opened 2024-02-12
Enforce group size limit


neekolas:

alexrisch:
@peterferguson This will be included in Discourse and we will be capturing them in Group Chat docs 😄

PR #554, opened 2024-03-19
feat: Add Limit on Groups in Add Member method

alexrisch:
Added limit in config
Added check to determine if adding memebers would add too many users


xmtp_mls/src/groups/mod.rs
```diff
@@ -272,6 +272,15 @@ where
         account_addresses_to_add: Vec<String>,
     ) -> Result<(), GroupError> {
         let account_addresses = sanitize_evm_addresses(account_addresses_to_add)?;
+        // get current number of users in group
```
alexrisch, line 275: I'd love to add a test, is there a better way to add one other than creating 250 clients and adding them?

xmtp_mls/src/groups/mod.rs
```diff
@@ -272,6 +272,15 @@ where
         account_addresses_to_add: Vec<String>,
     ) -> Result<(), GroupError> {
```
alexrisch, line 273: I only added this to add_members, since that's what's exposed to sdks, and add_members_by_installation_id is used during syncing the group members
I didn't want a sync to perpetually fail if somehow more than the limit were able to be added

richardhuaaa: That makes a lot of sense, good thinking. Enforcing the 250 limit on users rather than installations is also easier to explain to developers and end users

xmtp_mls/src/groups/mod.rs
richardhuaaa: You could investigate how to mock the `members()` method or creating a `members_count()` method to mock, but is the reason you don't want to use the real thing performance?

I'm suspecting the performance won't be that bad if you batch add 250 members, as opposed to adding them sequentially, because I believe that's still a single commit.

richardhuaaa (approved):
Mentioning explicitly, it's possible there's a race condition where your group state is not up-to-date, and therefore you add members over the limit - that's fine IMO as it's not a hard limit beyond which things will break.

LGTM, just needs a unit test!

xmtp_mls/src/groups/mod.rs
```diff
@@ -272,6 +272,15 @@ where
...
         let account_addresses = sanitize_evm_addresses(account_addresses_to_add)?;
+        // get current number of users in group
+        let member_count = self.members().unwrap().len();
```
richardhuaaa, line 276: Can you use `?` here rather than `unwrap()`, to forward the error rather than panicking?

xmtp_mls/src/groups/mod.rs
```diff
@@ -272,6 +272,15 @@ where
...
+        if member_count + account_addresses.len() > MAX_GROUP_SIZE as usize {
+            return Err(GroupError::Generic(format!(
```
richardhuaaa, line 278: It probably makes sense to add a variant here rather than using `Generic`, seeing as this is the type of thing that might need to be programmatically detected and handled.

xmtp_mls/src/groups/mod.rs
alexrisch: I might do one for mocking and another for creating the clients
Just wasn't sure what performance hit would be like for a fairly simplistic test

richardhuaaa (approved):
Perfect!

alexrisch:
Closes https://github.com/xmtp/libxmtp/issues/503

PR #558, opened 2024-03-20
Update xmtp/openmls to latest crate HEAD

zombieobject:


neekolas:
I would expect there to be updates to the root Cargo.toml pointing to different `rev` values in this PR

xmtp_mls/src/hpke.rs
```diff
@@ -49,7 +49,7 @@ pub fn decrypt_welcome(
         .read::<HpkePrivateKey>(hpke_public_key)
         .ok_or(HpkeError::KeyNotFound)?;
 
-    let ciphertext = HpkeCiphertext::tls_deserialize_exact(ciphertext)?;
+    let ciphertext = HpkeCiphertext::tls_deserialize_exact(ciphertext).unwrap();
```
neekolas, line 52: We're trying to keep this codebase unwrap free. If the `?` doesn't work anymore it's probably because you need a new error variant that supports the error type from `tls_deserialize_exact`

neekolas:
Looking at [this commit](https://github.com/xmtp/openmls/commit/8392840391531087d1a5ad89b7f34fad2fd4805f) all the `credential.identity()` errors make sense. They changed the API so that now you have to convert the credential into a `BasicCredential` using `into()` or `try_into()` and then you'll get access to the `identity` method.

xmtp_mls/src/identity.rs
```diff
@@ -137,13 +136,13 @@ impl Identity {
...
             let credential =
-                OpenMlsCredential::new(credential_proto.encode_to_vec(), CredentialType::Basic)?;
+            OpenMlsCredential::new(CredentialType::Basic, credential_proto.encode_to_vec());
             self.set_credential(credential)?;
         }
 
         // Register the installation with the server
         let kp = self.new_key_package(provider)?;
-        let kp_bytes = kp.tls_serialize_detached()?;
+        let kp_bytes = kp.tls_serialize_detached().unwrap();
```
neekolas, line 145: This should also go back to the `?`

xmtp_mls/src/client.rs
```diff
@@ -106,9 +108,9 @@ pub enum MessageProcessingError {
     #[error("storage error: {0}")]
     Storage(#[from] crate::storage::StorageError),
     #[error("tls deserialization: {0}")]
-    TlsDeserialization(#[from] tls_codec::Error),
+    TlsDeserialization(#[from] openmls::prelude::Error),
```
neekolas, line 111: If this is now a more general MLS error type, maybe we need to rename our mapped type. I'm assuming `openmls::prelude::Error` now includes more things than `TlsDeserialization` errors, although correct me if I'm wrong.

xmtp_mls/src/groups/sync.rs
```diff
@@ -517,7 +522,8 @@ where
                     intent_data.message.as_slice(),
                 )?;
 
-                let msg_bytes = msg.tls_serialize_detached()?;
+                let msg_bytes = msg.tls_serialize_detached()
+                    .map_err(|e| GroupError::TlsSerialization(e))?;
```
neekolas, line 526: If you have the right `#[from]` on `GroupError` it should magically convert the error types for you with the `?` and you won't need `map_err`

xmtp_mls/src/identity.rs
```diff
@@ -137,13 +137,14 @@ impl Identity {
...
-                OpenMlsCredential::new(credential_proto.encode_to_vec(), CredentialType::Basic)?;
+            OpenMlsCredential::new(CredentialType::Basic, credential_proto.encode_to_vec());
             self.set_credential(credential)?;
         }
 
         // Register the installation with the server
         let kp = self.new_key_package(provider)?;
-        let kp_bytes = kp.tls_serialize_detached()?;
+        let kp_bytes = kp.tls_serialize_detached()
+            .map_err(|e| IdentityError::TlsSerialization(e))?;
```
neekolas, line 147: Same as above for the `#[from]`. You should be able to automatically convert these

Issue #589, opened 2024-03-25
Scaffold credential verifier trait


neekolas:
We are going to need an exposed interface in the new [Identity crate](#588) that allows for the verification of credentials.

We should finalize the trait definition [listed in the XIP](https://www.notion.so/xmtplabs/Team-A-XIP-Off-chain-85e6a4e228a34828989ea02b7dbcf3a3) and make sure it properly solves our needs

PR #591, opened 2024-03-25
Add `xmtp_xid`

### LINKED ISSUE #588: Scaffold new identity crate in libxmtp

insipx:
- Adds a new crate for `xmtp_xid` closes #588 
- did some general cleanup around using workspace dependencies in `xmtp_mls` and related crates

not really sure how much we need here so I just copied a very minimum of the `Identity` struct from `xmtp_mls`

We probably want to reduce dependency on `xmtp_mls` as much as possible so we don't have a circular dependency, which implies also importing `GrantMessagingAccessAssociation`. This may make more sense to be apart of #589 & the trait. We can potentially do something to the effect of:
- decouple the 'association text' and EOA signing stuff
- refactor messaging associations to be just EOA signature credential
- implement new trait from XIP for these structs
This will be clearer for me once I play around with the trait in 589 a bit


xmtp_xid/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+use openmls_basic_credential::SignatureKeyPair;
+use xmtp_mls::{
+    configuration::CIPHERSUITE, credential::UnsignedGrantMessagingAccessData, types::Address,
+    utils::time::now_ns,
+};
+
+use crate::error::IdentityError;
+
+pub struct Identity {
+    pub(crate) account_address: Address,
```
neekolas, line 15: Need to ignore dead code for now

37ng (commented):
Suggest to rename `xmtp_xid` to `xmtp_id` since x represents xmtp

neekolas:
Sounds reasonable to me. It's weird enough that we are prefixing all of our stuff with the company name, we don't need to do it twice

Issue #527, opened 2024-02-21
Push notifications handling guide


neekolas:
We need to provide some resources for devs around push notification handling. We also need to expose methods for actually taking in the push notifications when they are received by the notification extension.

## Topics

Welcome topics have the format:
```ts
`/xmtp/mls/1/w-${hexEncode(installationId)}/proto`
```

Group message topics have the format:
```ts
`/xmtp/mls/1/g-${hexEncode(groupId)}/proto`
```

Clients should listen to these topics and then decode the payloads.

## Payload types
 
- Welcome messages are protobuf messages encoded with [`mlsv1.WelcomeMessage`](https://github.com/xmtp/proto/blob/main/proto/mls/api/v1/mls.proto#L115)
- Group messages are protobuf messages encoded with [`mlsv1.GroupMessage`](https://github.com/xmtp/proto/blob/main/proto/mls/api/v1/mls.proto#L145)

## What to do with messages
- Once payloads have been decoded from the push notification callers can pass them to `process_streamed_welcome_message` and `process_streamed_group_message` in our SDKs (these functions are not exposed outside of Rust).
```[tasklist]
### Tasks
- [x] Expose FFI interface for processing welcome and group messages to libxmtp
- [x] Add ability for callers to pass in Welcome messages to iOS SDK for processing
- [x] Add ability for callers to pass in Group messages to iOS SDK for processing
- [x] Add ability for callers to pass in Welcome messages to Android SDK for processing
- [x] Add ability for callers to pass in Group messages to Android SDK for processing
- [x] Add ability in RN
```

nplasterer:
Would be nice to have a example running in RN for this but going to push this down in priority in favor of the HMAC work for now. PR started here: https://github.com/xmtp/xmtp-react-native/pull/347

PR #593, opened 2024-03-26
Process streamed group message

nplasterer:
Part of https://github.com/xmtp/libxmtp/issues/527

Exposes a new method on groups called `process_streamed_group_message` which takes in a byte array of [mlsv1.GroupMessage](https://github.com/xmtp/proto/blob/main/proto/mls/api/v1/mls.proto#L145) and then returns a `GroupMessage` then the bindings call `into()` to convert it to a `FfiMessage`

Exposes a new method called `process_streamed_welcome_message` which takes in a byte array of [mlsv1.WelcomeMessage](https://github.com/xmtp/proto/blob/main/proto/mls/api/v1/mls.proto#L115) and then returns a `WelcomeMessage` then the bindings call `into()` to convert it to a `FfiGroup`

### DIFF
--- a/xmtp_mls/src/groups/subscriptions.rs
+++ b/xmtp_mls/src/groups/subscriptions.rs
@@ -48,12 +48,12 @@ where
     pub async fn process_streamed_group_message(
         &self, 
         envelope_bytes: Vec<u8>
-    ) -> Result<Option<StoredGroupMessage>, GroupError> {
+    ) -> Result<StoredGroupMessage, GroupError> {
         let envelope = GroupMessage::decode(envelope_bytes.as_slice())
             .map_err(|e| GroupError::GroupMessageNotFound(e.to_string()))?;
         
         let message = self.process_stream_entry(envelope).await.map_err(|e| e)?;
-        Ok(message)
+        Ok(message.unwrap())
     }
 
     pub async fn stream(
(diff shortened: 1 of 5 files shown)


xmtp_mls/src/groups/subscriptions.rs
```diff
@@ -44,6 +45,17 @@ where
...
     }
 
+    pub async fn process_streamed_group_message(
+        &self, 
+        envelope_bytes: Vec<u8>
+    ) -> Result<StoredGroupMessage, GroupError> {
+        let envelope = GroupMessage::decode(envelope_bytes.as_slice())
+            .map_err(|e| GroupError::GroupMessageNotFound(e.to_string()))?;
+        
+        let message = self.process_stream_entry(envelope).await.map_err(|e| e)?;
```
neekolas, line 55: You should be able to drop the `map_err` here

neekolas (approved):
Think this looks good, aside from the two lint warnings.

xmtp_mls/src/groups/subscriptions.rs
```diff
@@ -44,6 +45,17 @@ where
         Ok(new_message)
     }
 
+    pub async fn process_streamed_group_message(
+        &self, 
+        envelope_bytes: Vec<u8>
+    ) -> Result<StoredGroupMessage, GroupError> {
+        let envelope = GroupMessage::decode(envelope_bytes.as_slice())
+            .map_err(|e| GroupError::GroupMessageNotFound(e.to_string()))?;
```
richardhuaaa, line 53: Is this the right error type?

richardhuaaa (approved):

PR #595, opened 2024-03-26
Message History - 2/N

tuddman:
## Summary

More work towards supporting Message History.

This PR includes 
- `proto` changes 
- stubs for handling message history messages.
- `sync` group creation function
- a separate filtered `find_sync_groups` method to only find the `sync` group types.
- adds initital tests


xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -82,6 +99,14 @@ impl DbConnection<'_> {
         Ok(self.raw_query(|conn| query.load(conn))?)
     }
 
+    /// Return only the [`Purpose::Sync`] groups
+    pub fn find_sync_groups(&self) -> Result<Vec<StoredGroup>, StorageError> {
```
neekolas, line 103: I guess there isn't really a reason to ever look for all sync and conversation groups at the same time, right?

neekolas (approved):

xmtp_mls/src/groups/sync.rs
```diff
@@ -262,38 +267,38 @@ where
...
+                            sender_installation_id,
+                            sender_account_address,
+                            delivery_status: DeliveryStatus::Published,
+                        }
+                        .store(provider.conn())?
+                    }
+                    Some(Content::V2(_)) => {
+                        todo!()
+                    }
+                    None => return Err(MessageProcessingError::InvalidPayload),
```
richardhuaaa, line 300: Double-checking, are the changes in this file purely refactoring, or is there a functional effect too?

xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -49,9 +50,25 @@ impl StoredGroup {
...
+            membership_state,
+            installations_last_checked: 0,
+            purpose: Purpose::Sync,
+        }
+    }
 }
 
 impl DbConnection<'_> {
+    /// Return  regular [`Purpose::Conversation`] groups with additional filters
     pub fn find_groups(
```
richardhuaaa, line 72: We need to add the code change to filter out the sync groups here (or filter to Conversation groups) right?

richardhuaaa (approved):

xmtp_mls/src/storage/encrypted_store/group.rs
tuddman: right, AFAICT.  I couldn't think of a scnerio where we'd want to. 
And we explicitly want to keep `listGroups()` as it is, for only returning "normal (conversation)" groups.

If we should find a need later to look for all groups at the same time, we'll add it if/when that need arises.

xmtp_mls/src/groups/sync.rs
tuddman: you're correct - it is purely a refactoring (and stubbing support for the `V2` match arm).
I saw that I was needlessly pulling out the `key` and `message` just to use a few lines later.   That is now self-contained within the match arm.

xmtp_mls/src/storage/encrypted_store/group.rs
tuddman: yep, that's already happening from my [last PR](https://github.com/xmtp/libxmtp/pull/553/files#diff-a7388e26da76ba3ef25a9251cfca5a9b66a99e7d91e3d041c6dce54e4618feeaR80) 😉

PR #596, opened 2024-03-26
Update to Latest OpenMLS

zombieobject:


xmtp_mls/src/groups/members.rs
```diff
@@ -37,8 +38,9 @@ pub fn aggregate_member_list(openmls_group: &OpenMlsGroup) -> Result<Vec<GroupMe
     let member_map: HashMap<String, GroupMember> = openmls_group
         .members()
         .filter_map(|member| {
+            let basic_credential = BasicCredential::try_from(&member.credential).ok()?;
```
neekolas, line 41: This will convert the Result to an Option, which will suppress the original error message when used with the `?`. You're better off ditching the `ok` and handling the error

neekolas (approved):
This is looking great. Thanks for powering through all these changes

PR #602, opened 2024-03-28
scaffold verifier trait

### LINKED ISSUE #589: Scaffold credential verifier trait

insipx:
closes #589


xmtp_id/src/credential_verifier.rs
```diff
@@ -0,0 +1,99 @@
...
+        for credential in credentials_to_verify {
+            results.push(Self::verify_credential(credential));
+        }
+
+        futures::future::join_all(results).await
+    }
+}
+
+#[async_trait::async_trait]
+impl CredentialVerifier for Credential {
```
insipx, line 80: I feel like it would be better if we could somehow do `impl CredentialVerifier for GrantMessagingAccess` and `impl CredentialVerifier for LegacyCreateIdentity` in order to further decouple legacy verification (v2) and current v3 verification. Then, we could potentially introduce another trait, `Credential` which would replace the enum and maybe clean things up elsewhere as well.

The problem is we don't know the credential type before decoding, making the implementation more complicated and possibly not worth it.

In the interest of not blocking @37ng and moving forward with SCW support going to leave that problem for another day. Eventually V2 will be phased out anyway. The `CredentialVerifier` trait works well with what we have so far

neekolas: I can see how that'd be tricky. We'd need to have two steps of extraction I guess. 

This seems fine for now and we can always refactor further down the line if we think it makes sense.

xmtp_id/src/credential_verifier.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[async_trait::async_trait]
+pub trait CredentialVerifier {
+    async fn verify_credential(request: VerificationRequest) -> VerificationResult;
+    async fn batch_verify_credentials(
+        credentials_to_verify: Vec<VerificationRequest>,
+    ) -> Vec<VerificationResult> {
+        let mut results = Vec::new();
+        for credential in credentials_to_verify {
+            results.push(Self::verify_credential(credential));
```
neekolas, line 72: So, the semantics here are that if any credential fails to verify the whole batch fails? Maybe that's fine, but we might have to rethink it when we start to integrate into real workflows.

neekolas (approved):

xmtp_id/src/credential_verifier.rs
insipx: Actually no, the results are returned as a vec for each verification future. Once the batch_verify completes the caller can sort through the Results for `Ok`'s and `Err`'s. 

[join_all](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.16/futures/future/fn.join_all.html)

PR #603, opened 2024-03-29
Mutable Metadata via new UpdateMetadata intent kind

### LINKED ISSUE #548: Mutable Metadata (name a group)
## Summary

We would like to have a metadata extension type that can be modified after group creation. This is possible using the Unknown Group Context Extension type, which accepts a byte array as the only field. It is then left up to the application to handle the formatting of the data and any commit validations that should be applied when it is modified.

Some known requirements here:
- We need to have some sort of standard format for this (likely a proto message)
- We would like to have some fields standardized, such as group name
- We need to have some sort of permissions model specifying who can modify this information. Each commit must be checked against these permissions to ensure it is valid

## Potential use-cases
- Storing a list of "admins" that can then be referenced in other group permission policies
- Group name and other metadata (header image? description?)
- Application-specific data. Possibly namespaced per user, so that anyone has some part of the group metadata that is writeable

cameronvoell:
Closes #548

Relies on:
- https://github.com/xmtp/proto/pull/158
- https://github.com/xmtp/openmls/pull/22


cameronvoell:
> Will likely break this out into several smaller PRs once I get it working, but saving all the work in this one PR while I get some feedback for now.

Closing this in favor of:
- https://github.com/xmtp/libxmtp/pull/643
- https://github.com/xmtp/libxmtp/pull/644

PR #604, opened 2024-03-29
Push notification follow ups

nplasterer:
Just returns one installationId instead of the whole list.
Adds tests for processing bytes.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -138,11 +138,8 @@ impl FfiXmtpClient {
         Ok(results)
     }
 
-    pub async fn installation_ids(&self) -> Result<Vec<Vec<u8>>, GenericError> {
-        let address = self.inner_client.account_address();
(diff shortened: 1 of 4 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -138,11 +138,8 @@ impl FfiXmtpClient {
...
     }
 
-    pub async fn installation_ids(&self) -> Result<Vec<Vec<u8>>, GenericError> {
-        let address = self.inner_client.account_address();
-        let installations = self.inner_client.get_all_active_installation_ids(vec![address]).await?;
-
-        Ok(installations)
+    pub fn installation_id(&self) -> Vec<u8> {
+        self.inner_client.installation_public_key()
     }
```
nplasterer, line 143: Per Nicks feedback in the original PR probably don't need the whole list just the one for the device of this client.

neekolas (approved):

PR #607, opened 2024-03-30
fix: dont even have todo. err instead.

tuddman:
## Summary

The existence of `Content::V2(_)` as a match arm was causing problems and crashing on the `todo!()`

This patch changes those `todo`s to `Err` instead.


nplasterer (approved):
Really wild that iOS was hitting this. I'll have to dig deeper but atleast it shouldn't crash. 🙏

tuddman:
> Really wild that iOS was hitting this. I'll have to dig deeper but atleast it shouldn't crash. 🙏

Sure.  Sorry that happened.  There's something to learn here about _how_ and _why_ that was happening.

PR #609, opened 2024-04-02
ERC-1271 support and testing a partner wallet app.

### LINKED ISSUE #608: Initial Smart Wallet support - partner wallet app
As first step of #590 that supports [Coinbase Smart Wallet](https://github.com/coinbase/smart-wallet)

37ng:
Adds ERC-1271 support and tests for [open sourced coinbase smart wallet](https://github.com/coinbase/smart-wallet).

closes #608


xmtp_id/src/verifier.rs
```diff
@@ -0,0 +1,47 @@
+use ethers::types::{Address, BlockId, Signature};
+use ethers::providers::{Http, Middleware, Provider};
+use std::convert::TryFrom;
+use std::sync::Arc;
+use crate::bindings::coinbase_smart_wallet::CoinbaseSmartWallet;
```
insipx, line 5: Where are the bindings being generated? Is this a manual generation or is there a script somewhere?

in `didethresolver`, we used an `abigen!` macro tied to just the top-level `DIDRegistry` smart contract (despite DIDRegistry pulling in a bunch of dependencies), which was enough to let us use the DIDRegistry struct and call into the smart contract from rust code. 

Is that not possible here? I would rather avoid adding a bunch of bindings we won't use

xmtp_id/src/verifier.rs
```diff
@@ -0,0 +1,47 @@
...
+use anyhow::Error;
+
+const BASE_SEPOLIA_RPC_URL: &str = "https://sepolia.base.org";
+const EIP1271_MAGIC_VALUE: [u8; 4] = [0x16, 0x26, 0xba, 0x7e];
+
+pub struct ERC1271Verifier {
```
insipx, line 11: I'm not sure if this is possible, but could we just import the ERC1271 Interface Smart contract, and verify the signature with that? In this case, that might support all ERC1271 signatures and not just Coinbase Smart Wallet.

Otherwise, I'm in favor of renaming ERC-1271 to `CoinbaseSmartWallet`, and implementing the CredentialVerifier trait on that. This could also mean we would have to support each smart wallet individually

xmtp_id/src/verifier.rs
37ng: The bindings are generated by `forge find` in the coinbase smart wallet repo. I didn't use `abigen!` because it is not supported well by my editor but I have no problems using it.

xmtp_id/src/verifier.rs
37ng: Each ERC-1271 wallet(Gnosis, Coinbase and others I don't know of) has its own verification logic so an ERC-1271 interface doesn't suffice and we will need to support each SCW individually. 
Renaming makes sense to me.

insipx: Do we need to worry about generating the signatures, or will that be something the SDK  does?

AFAIU every smart wallet will implement `is_valid_signature`, so even if the signature generation or signing is different, signature verification will always be that function which coincides with `ERC-1271`. The only difference will be the address that `is_valid_signature` will be called on since each SCW will implement its own verification, but it must be according to ERC-1271

37ng: Yeah that makes sense. 

I was trying to figure out the whole lifecycle from message to signature but here I think you are right we just need to put a simple ERC-1271 `is_valid_signature` interface here and all additional logic should be used in the tests.

neekolas: Seconding that. At verification time we want this to be generic for any valid ERC-1271. The only things that should be implementation-specific are test-cases, where we would want to create specific Smart Wallets to test against.

37ng (commented):
.

xmtp_id/src/verifier.rs
37ng: tysm!My beef was mainly because I can't checkout the abigen'ed code in my editor.

Don't want to clutter the codebase either, using abigen! now.

insipx: ah ok, yeah you can use `cargo doc --document-private-items` to inspect the abigen!'ed code

xmtp_id/src/erc1271_verifier.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub fn new(url: &str) -> Self {
+        let provider =  Arc::new(Provider::<Http>::try_from(url).unwrap());
+        Self { provider }
+    }
+    
+    /// Verifies an ERC-1271(https://eips.ethereum.org/EIPS/eip-1271) signature.
+    ///
+    /// # Arguments
+    ///
+    /// * `wallet_address` - Address of the ERC1271 wallet.
```
neekolas, line 30: We might want to call this `contract_address` instead, just so it doesn't cause confusion

xmtp_id/src/erc1271_verifier.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// * `hash`, `signature` - Inputs to ERC-1271, used for signer verification.
+    pub async fn is_valid_signature<M: Middleware>(&self, wallet_address: Address, block_number: BlockNumber, hash: [u8; 32], signature: Bytes) -> Result<bool, Error> {
+        let erc1271 = ERC1271::new(wallet_address, self.provider.clone());
+
+        let res: [u8; 4] = erc1271
+            .is_valid_signature(hash, signature)
+            .block(block_number)
+            .call()
+            .await?
+            .into();
```
neekolas, line 41: Don't think you need the `into()` here

neekolas (approved):

37ng:
Added integration tests for signature verification, owner revocation, time travel logic.
Please give another review

neekolas (approved):
This is looking good

tuddman:
Adding here that running our test suite locally breaks;  added `foundry` and its bundled chain tool, `Anvil` to the dependency list.  For anyone that hits the 

```
failures:
    erc1271_verifier::tests::test_coinbase_smart_wallet
```

you need to install [foundry](https://book.getfoundry.sh/) with:

```
# get the installer
curl -L https://foundry.paradigm.xyz | bash

# `source` your shell environment. e.g
source ~/.zshenv

# install foundry  <- this will also install anvil as well as some other chain tools.
foundryup 
```

then run the tests again, and they should pass.

37ng:
@tuddman Thanks! I wonder can we easily add `foundry` in the dev/up to automate this? I tried the scripts here but didn't work.

PR #614, opened 2024-04-04
Who Added Me

zombieobject:
Save the wallet address of who added a user to a group.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -220,7 +220,7 @@ impl FfiConversations {
             inner_client: self.inner_client.clone(),
             group_id: convo.group_id,
             created_at_ns: convo.created_at_ns,
-            added_by_address: Some(self.inner_client.account_address()),
+            added_by_address: convo.added_by_address,
         });
 
         Ok(out)


bindings_ffi/src/mls.rs
```diff
@@ -352,13 +388,18 @@ impl FfiGroup {
...
         );
 
         group.sync().await?;
 
         Ok(())
     }
 
+    pub fn added_by_address(&self) -> Option<String> {
+        self.added_by.as_ref().map(|member| member.account_address.clone())
+    }
```
nplasterer, line 401: Lets return the member here instead. I think it might be something like this

```suggestion
    pub fn added_by(&self) -> Option<FfiMember> {
        self.added_by.into()
    }
```

bindings_ffi/src/mls.rs
```diff
@@ -367,6 +408,7 @@ impl FfiGroup {
             self.inner_client.as_ref(),
             self.group_id.clone(),
             self.created_at_ns,
+            self.added_by_address(),
```
nplasterer, line 411: Why do you need to add this to every new group instance?

bindings_ffi/src/mls.rs
```diff
@@ -318,6 +320,39 @@ pub struct FfiGroup {
     inner_client: Arc<RustXmtpClient>,
     group_id: Vec<u8>,
     created_at_ns: i64,
+    added_by: Option<FfiGroupMember>,
+}
+
+pub fn create_ffi_group(
```
nplasterer, line 326: You shouldn't be creating a new function. It should just be added to the existing function https://github.com/xmtp/libxmtp/blob/main/bindings_ffi/src/mls.rs#L195-L227

nplasterer (changes requested):

xmtp_mls/src/groups/mod.rs
```diff
@@ -203,9 +218,20 @@ where
...
+            added_by_address.clone(),
+        );
+
         stored_group.store(provider.conn())?;
-        Ok(Self::new(client, group_id, stored_group.created_at_ns))
+        Ok(Self::new(
+            client,
+            group_id,
+            stored_group.created_at_ns,
+            added_by_address,
```
nplasterer, line 233: Ethan and I paired and got on the same page. My initial hesitation was that I didn't understand why we needed to set the address on the `MlsGroup::new` and also on the `StoredGroup:new` seems like we should just be able to set it on the stored group and then fetch it from the stored group? Came to the conclusion that we don't do anything with the storedGroup in the bindings so this seems like the best way given our current architecture cc @richardhuaaa if I'm missing some context.

bindings_ffi/src/mls.rs
nplasterer: Ethan and I talked about this and he's just going to remove this function in favor of the impl in the struct

bindings_ffi/src/mls.rs
nplasterer: Got extra clarity on Zoom and posted an addition comment outside of the bindings.

bindings_ffi/src/mls.rs
nplasterer: Per conversation we're going to roll this back this commit `3c3d9f363400fb35a75165c362773cdd2ecaf3d6`

xmtp_mls/src/groups/mod.rs
richardhuaaa: @nplasterer it's a good point. Today, we're fetching the StoredGroup, and then manually caching the values onto the MlsGroup via the constructor so that we don't need to re-fetch it again. The drawback is that the constructor for MlsGroup is getting more complicated over time. Could be refactored one of two ways:

1. Don't cache the values onto the MlsGroup. This is simpler code, but tradeoff is you need to re-fetch from the DB every time you need a property like `added_by`.

2. Instead of manually copying the values over to the MlsGroup, have the MlsGroup hold a copy of StoredGroup inside it so there's no duplication. That's probably the best of both worlds.

bindings_ffi/src/mls.rs
```diff
@@ -216,11 +216,11 @@ impl FfiConversations {
         if !account_addresses.is_empty() {
             convo.add_members(account_addresses).await?;
         }
-
         let out = Arc::new(FfiGroup {
             inner_client: self.inner_client.clone(),
             group_id: convo.group_id,
             created_at_ns: convo.created_at_ns,
+            added_by_address: Some(self.inner_client.account_address()),
```
richardhuaaa, line 223: We should read this from `convo` so that the value is guaranteed to be consistent with `convo`

richardhuaaa (approved):
LGTM. I have a slight preference for making the property and DB schema non-optional, and just eating the breaking change now so that devs don't have to deal with a nullable `added_by`. But leaving it up to you

nplasterer:
> just eating the breaking change now so that devs don't have to deal with a nullable added_by

I think that makes sense. Better to break things now than later. 👍

xmtp_mls/migrations/2024-04-08-180113_group_added_by_address/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE groups
+ADD COLUMN added_by_address TEXT
```
nplasterer, line 2: @zombieobject I think @richardhuaaa was suggesting that this is NOT NULL do you mind doing that in a follow up PR?

xmtp_mls/src/groups/mod.rs
nplasterer: I like 2. do you think this will be hard to do later/a breaking change? Just seeing how we should prioritize that work. My gut is saying we should probably do it now before we filter this through the sdks.

PR #615, opened 2024-04-04
interface for determining scw or eoa

### LINKED ISSUE #599: Interface for detecting whether a Signer is a regular wallet or a smart wallet

insipx:
There is no fool-proof way to determine whether an address is a SCW or EOA. The best we have is to check the code-size on-chain. If the smart contract wallet has not yet been deployed, then there is no way of knowing just from the address whether an account is a SCW or an EOA.

This just adds a very simple trait, `WalletIdentity` for SDKs to implement in order to indicate whether an identity is a Smart Wallet or not.

I couldn't decide whether this should live on `InboxOwner` or not. It seems to make sense for it to live on that trait, but I also didn't want to prematurely add a change into `xmtp_mls` that inbox's would have to implement w/o Identity being ready as a whole yet.

I also found some unused dependencies, will have to look closer for those in reviews

closes #599


neekolas (approved):
I actually think this works great as a standalone utility we can offer devs. Especially since it's not clear how much we'll be using `LocalWallet` in a world with Signature Requests

PR #617, opened 2024-04-04
Association History Processing (Part 1)

neekolas:
## Summary

- Adds new associations module
- Adds the struct for storing the immutable `AssociationState` (will make more sense after you look at the next PR
- Adds a `MemberIdentifier` struct used to hold either wallet addresses or installation public keys.


neekolas:
* **#617** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

xmtp_id/src/associations/test_utils.rs
```diff
@@ -0,0 +1,21 @@
...
+    let v: String = rand::thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(32)
+        .map(char::from)
+        .collect();
+
+    v
+}
+
+pub fn rand_u64() -> u64 {
```
neekolas, line 13: These are unused for now, but will be once the second PR in the stack goes in

xmtp_id/src/associations/hashes.rs
```diff
@@ -0,0 +1,12 @@
+use sha2::{Digest, Sha256};
+
+pub fn sha256_string(input: String) -> String {
```
insipx, line 3: What's the benefit to making this a string vs passing around some bytes?

insipx (approved):
Looks good!

xmtp_id/src/associations/hashes.rs
insipx: We could maybe make a new type, `Xid` that wraps a `Vec<u8>` and has a `Display` impl that outputs the `String` impl. This is probably a tad more efficient, but mostly it would be more obvious in-code when we have an XID vs just a String

neekolas: This is only used for the `xid` right now. I'd like the `xid` to be a string, since it makes it easier to handle at the platform SDK level. And there isn't much value in keeping it as bytes, since it doesn't represent a public key or anything like that.

insipx: That makes sense

neekolas: I'd be OK with a type called XID that wraps a string to get that clarity. I just know that we're already annoying the platform SDK teams having `installation_id` be bytes. That one has a good reason (it's a public key), but this one is harder to justify.

37ng (approved):

neekolas:
### Merge activity

* **Apr 5, 8:03 PM EDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/617).

PR #620, opened 2024-04-05
Add Unsigned Identity Updates

neekolas:
## tl;dr

Adds a new `UnsignedIdentityUpdate` struct and unsigned variations of all the identity actions.

## Why

We need this in two places:

1. When deserializing protos to build an `IdentityUpdate`, we need to get the full signature text before we can properly build the signatures on each IdentityAction
2. When creating a new `IdentityUpdate` the signature text requires the full set of actions in the update. So, we need to create the full `IdentityUpdate` in some form, gather the necessary signatures, and then turn it into the final format.

PR #618, opened 2024-04-04
Association History Processing (Part 2)

neekolas:
## Summary

- Adds the actual association log and verifications
- Adds a bunch of tests

## TODO

- More tests coming
- Verify all logic in the XIP is identical to the implementation
- Serialization and deserialization
- Actual signature types

PR #633, opened 2024-04-07
IdentityUpdate serialization

neekolas:
## tl;dr

- Adds serialization and deserialization for IdentityUpdate
- Adds a missing `inbox_id` field to a few of the action types
- Scaffolds out the `Signature` types. Still need real verification of signatures and recovery addresses.
- Rename the `IdentityUpdateBuilder` to `SignatureRequestBuilder`
- Make current with final version of protos (moved the `inbox_id` to the `IdentityUpdate` instead of being on each action)


neekolas:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/633?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#633** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/633?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#632** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/632?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#620** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/620?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#618** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/618?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#617** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/617?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

dev/gen_protos.sh
```diff
@@ -6,7 +6,7 @@ if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
     fi
 fi
 
-if ! buf generate https://github.com/xmtp/proto.git#branch=main,subdir=proto; then
+if ! buf generate https://github.com/xmtp/proto.git#branch=nm/prototype-identity-apis,subdir=proto; then
```
neekolas, line 9: We can update this to point back to `main` once that PR merges

xmtp_id/src/associations/association_log.rs
```diff
@@ -103,6 +109,7 @@ impl IdentityAction for AddAssociation {
     ) -> Result<AssociationState, AssociationError> {
         let existing_state = maybe_existing_state.ok_or(AssociationError::NotCreated)?;
         self.replay_check(&existing_state)?;
+        ensure_matching_inbox_id(&self.inbox_id, existing_state.inbox_id())?;
```
richardhuaaa, line 112: Is there a scenario where the wallet is revoked from the current inbox ID, but associated with a different inbox ID, as part of a single batch signature? One of the two will need to be 'skipped over', right?

neekolas: I don't think we want to allow batches that span multiple inboxes. It makes the security model and programming model a lot more complicated, for what's a pretty rare case. Suddenly any update can interfere with any other update. You'd need global total ordering instead of per inbox. And it's not clear what state to apply each action to in advance. 

They can just do two signatures.

neekolas: For the server, I do think we are going to need a top-level `inbox_id` field on the `IdentityUpdate`. Without it, the server is going to have a hard time figuring out what previous state to apply the update to. The way the protos are structured today, the server would need to inspect the individual actions to find the `inbox_id`. 

I'm hoping on the Go side we can be pretty blind to the contents of these Actions and just forward them over to Rust-land for all processing.

xmtp_id/src/associations/signature.rs
```diff
@@ -52,3 +59,182 @@ impl Clone for Box<dyn Signature> {
...
+        }
+    }
+}
+
+impl Signature for LegacyDelegatedSignature {
+    fn recover_signer(&self) -> Result<MemberIdentifier, SignatureError> {
+        // TODO: Two steps needed here:
+        // 1. Verify the RecoverableEcdsaSignature and make sure it recovers to the public key specified in the `signed_public_key`
+        // 2. Verify the wallet signature on the `signed_public_key`
+        // Return the wallet address
```
richardhuaaa, line 216: In case it's helpful, for your next PR - there should be an implementation of this already that you can copy/paste/refactor, same with the todo's above

xmtp_id/src/associations/unsigned_actions.rs
```diff
@@ -14,58 +19,52 @@ pub struct UnsignedCreateInbox {
 
 impl SignatureTextCreator for UnsignedCreateInbox {
     fn signature_text(&self) -> String {
-        format!(
-            // TODO: Finalize text
-            "Create Inbox: {}",
-            generate_inbox_id(&self.account_address, &self.nonce)
-        )
+        format!("- Create inbox\n(Owner: {})", self.account_address)
```
richardhuaaa, line 22:
```suggestion
        format!("- Create inbox\n  (Owner: {})", self.account_address)
```
Small nit to get alignment - added two whitespaces

xmtp_id/src/associations/unsigned_actions.rs
```diff
@@ -14,58 +19,52 @@ pub struct UnsignedCreateInbox {
...
 
 impl SignatureTextCreator for UnsignedAddAssociation {
     fn signature_text(&self) -> String {
+        let (target_kind, id_kind) = get_terms_for_kind(self.new_member_identifier.kind());
         format!(
-            // TODO: Finalize text
-            "Add {} to Inbox {}",
-            self.new_member_identifier, self.inbox_id
+            "- Grant messaging access to {target_kind}\n({id_kind}: {})",
+            self.new_member_identifier
```
richardhuaaa, line 36: I think the change made to `RevokeAssociation` needs to be applied here

xmtp_id/src/associations/unsigned_actions.rs
```diff
@@ -14,58 +19,52 @@ pub struct UnsignedCreateInbox {
...
-            // TODO: Finalize text
-            "Remove {} from Inbox {}",
-            self.revoked_member, self.inbox_id
-        )
+        let (_, id_kind) = get_terms_for_kind(self.revoked_member.kind());
+        let prefix = match self.revoked_member.kind() {
+            MemberKind::Address => "Revoke messaging access from app",
+            MemberKind::Installation => "Unlink address from inbox",
+        };
+        format!("- {prefix}\n({id_kind}: {})", self.revoked_member)
```
richardhuaaa, line 53:
```suggestion
        format!("- {prefix}\n  ({id_kind}: {})", self.revoked_member)
```

xmtp_id/src/associations/unsigned_actions.rs
```diff
@@ -14,58 +19,52 @@ pub struct UnsignedCreateInbox {
...
     pub new_recovery_address: String,
 }
 
 impl SignatureTextCreator for UnsignedChangeRecoveryAddress {
     fn signature_text(&self) -> String {
         format!(
             // TODO: Finalize text
-            "Change Recovery Address for Inbox {} to {}",
-            self.inbox_id, self.new_recovery_address
+            "- Change inbox recovery address\n(Address: {})",
```
richardhuaaa, line 66:
```suggestion
            "- Change inbox recovery address\n  (Address: {})",
```

xmtp_id/src/associations/unsigned_actions.rs
```diff
@@ -104,63 +121,67 @@ impl SignatureTextCreator for UnsignedIdentityUpdate {
             .map(|action| action.signature_text())
             .collect::<Vec<String>>();
         format!(
-            "I authorize the following actions on XMTP:\n\n{}\n\nAuthorized at: {}",
+            "{HEADER}\n\nInbox ID: {}\nCurrent time: {}\n\n{}\n\n{FOOTER}",
+            self.inbox_id,
+            ns_date_to_iso_8601(self.client_timestamp_ns),
             all_signatures.join("\n\n"),
```
richardhuaaa, line 127:
```suggestion
            all_signatures.join("\n"),
```

xmtp_id/src/associations/unsigned_actions.rs
```diff
@@ -104,63 +121,67 @@ impl SignatureTextCreator for UnsignedIdentityUpdate {
...
-            // TODO: Pretty up date
-            self.client_timestamp_ns
         )
     }
 }
 
+fn ns_date_to_iso_8601(ns_date: u64) -> String {
+    let date = DateTime::from_timestamp_nanos(ns_date as i64);
+    date.to_string()
```
richardhuaaa, line 134: This isn't actually ISO8601, it looks like this: `2015-05-15 00:00:00 UTC`

I actually like it more though because it's way more human readable, I can update the XIP to match. But we should rename this method in that case

xmtp_id/src/associations/unsigned_actions.rs
```diff
@@ -104,63 +121,67 @@ impl SignatureTextCreator for UnsignedIdentityUpdate {
...
                 UnsignedAction::RevokeAssociation(revoke_association.clone()),
                 UnsignedAction::ChangeRecoveryAddress(change_recovery_address.clone()),
             ],
         };
-
+        let now_timestamp = ns_date_to_iso_8601(client_timestamp_ns);
         let signature_text = identity_update.signature_text();
-        let expected_text = format!("I authorize the following actions on XMTP:\n\nCreate Inbox: {}\n\nAdd {} to Inbox {}\n\nRemove {} from Inbox {}\n\nChange Recovery Address for Inbox {} to {}\n\nAuthorized at: {}",
-        inbox_id,
+        let expected_text = format!("{HEADER}\n\nInbox ID: {inbox_id}\nCurrent time: {now_timestamp}\n\n- Create inbox\n(Owner: {})\n\n- Grant messaging access to inbox\n(Address: {})\n\n- Revoke messaging access from app\n(Address: {})\n\n- Change inbox recovery address\n(Address: {})\n\n{FOOTER}",
```
richardhuaaa, line 180: Could we hardcode the actual text here? You can just print the text from one test run and then paste it here. That way we can:
(1) Visually inspect the result to see if it makes sense
(2) Protect ourselves if something changes, for example the implementation of `DateTime.to_string()` changes to a different format. That would irrevocably screw up many accounts

xmtp_id/src/associations/unsigned_actions.rs
richardhuaaa: Also, would it make sense to round the nanoseconds to seconds, before feeding it into `DateTime`? Looks like it's sometimes rendered with decimal places, e.g. `2000-01-12 01:02:03.004 UTC`, which might cause some clients to fail validation

xmtp_id/src/associations/signature.rs
37ng: ethers crate supports this and I'm working on [this issue](https://github.com/xmtp/libxmtp/issues/636).

btw, if anyone knows how to convert bytes(Vec<u8>, [u8] etc.) to/from different signature types(ethers, k256 etc.), that would be so helpful!

neekolas: I'd suggest writing a `TryFrom<...>` implementation for _our_ types, and if their libraries already have `TryFrom` implemented you can use that internally. That way you don't have to worry about constraints around implementing traits for foreign types.

xmtp_id/src/associations/unsigned_actions.rs
neekolas: Missed that one

xmtp_id/src/associations/unsigned_actions.rs
neekolas: OK. I've fixed the ISO formatting and added rounding to seconds.

neekolas:
@richardhuaaa I've updated to match the spacing in your comments and added hardcoded text in the tests

richardhuaaa (approved):
Thanks for handling my nits! I preferred the original time format you had to ISO8601, but that's fine

neekolas:
### Merge activity

* **Apr 10, 8:47 PM EDT**: @neekolas merged this pull request with [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/633).

PR #637, opened 2024-04-08
fix: fixes some clippy lints

tuddman:


insipx (approved):

insipx:
don't forget to `fmt`

37ng (approved):
there are a few more "the borrowed expression implements the required traits". can these be easily fixed as well?

tuddman:
> there are a few more "the borrowed expression implements the required traits". can these be easily fixed as well?

Yes, they could be easily fixed.  However,  they are from auto-generated code that will come just come back whenever we update the protos.  I intentionally just left those, despite their annoying persistence.

PR #640, opened 2024-04-09
Fix clippy on generated code & add a deny for warnings

insipx:
This gets rid of the clippy warnings on code generated for protos by allowing all on an intermediate module declaration, as well as makes it a requirement for clippy to successfully finish w/o warnings for the Dockerfile build


tuddman (approved):
nice!  should've done this yesterday.  
Thanks for finishing it up :)

insipx:
For some reason I didn't consider this strategy yesterday but remembered while working on an unrelated PR -- we're finally going to have a clean PR review view 🥳

PR #643, opened 2024-04-10
Mutable Metadata group name Part 1

cameronvoell:
Part 1 for https://github.com/xmtp/libxmtp/issues/548

We would like to have a metadata extension type that can be modified after group creation. This is possible using the Unknown Group Context Extension type, which accepts a byte array as the only field. See https://github.com/xmtp/libxmtp/issues/548 for more info.

Related PR's:

- Part 2: https://github.com/xmtp/libxmtp/pull/644

- [x] Update openmls reference in `Cargo.toml` to commit on `xmtp/openmls` `main` branch
- [x] Update dev/gen_protos.sh to point to `xmtp-proto` repo `main` branch


xmtp_mls/src/groups/intents.rs
```diff
@@ -222,6 +224,48 @@ impl From<RemoveMembersIntentData> for Vec<u8> {
...
+pub struct UpdateMetadataIntentData {
+    pub group_name: String,
+}
+
+impl UpdateMetadataIntentData {
+    pub fn new(group_name: String) -> Self {
+        Self { group_name }
+    }
+
+    pub(crate) fn to_bytes(&self) -> Vec<u8> {
```
insipx, line 237: Should we just unify all the conversion methods in the `TryFrom` to avoid confusion; what's the benefit of having the method on `UpdateMetadataIntentData`?

insipx (approved):
Looks good overall!, just a nit wondering why we are separating out the `TryFrom`/`TryInto` between `impl` on a struct and in the trait implementation?

xmtp_mls/src/groups/group_permissions.rs
```diff
@@ -231,19 +231,22 @@ pub struct PolicySet {
     pub remove_member_policy: MembershipPolicies,
     pub add_installation_policy: MembershipPolicies,
     pub remove_installation_policy: MembershipPolicies,
+    pub update_group_name_policy: MembershipPolicies,
```
neekolas, line 234: I kinda think we should have a new type of policy for metadata. Some of the `MembershipPolicies` aren't applicable here (for example, `AllowSameMember`) and it makes easier to be misused having that as an option.

What if we had a different type, like `MetadataPolicy` that can be more tailored towards updating metadata.

cameronvoell: Good point, ill make that update 👍

cameronvoell: @neekolas updating to include a new `MetadataPolicy` type actually adds a lot of cascading updates, since it touches all the validation logic. I was going to tackle validation logic in part 3, so how do we feel about leaving MetadataPolicy to the part 3 PR?

xmtp_mls/src/groups/intents.rs
cameronvoell: Updated [here](d91fe5b52db8f62f4aa20fb77cd3a6b87b64829b). I think this is what you were suggesting, but let me know if you meant something else.

xmtp_mls/src/groups/group_mutable_metadata.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl GroupMutableMetadata {
+    pub fn new(group_name: String) -> Self {
+        Self {
+            group_name
+        }
+    }
+
+    pub(crate) fn from_proto(
```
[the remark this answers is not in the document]
cameronvoell: I definitely copied from `xmtp_mls/src/groups/group_metadata.rs`, but I agree with your thinking (if I'm understanding correctly).

Updated here => d91fe5b52db8f62f4aa20fb77cd3a6b87b64829b Let me know if you meant something else.

cameronvoell:
> Looks good overall!, just a nit wondering why we are separating out the `TryFrom`/`TryInto` between `impl` on a struct and in the trait implementation?

Thanks for the feedback @insipx . I liked you suggestion to put all conversion logic in TryFrom functions and avoiding having redundant functions unused anywhere. I created an issue here => https://github.com/xmtp/libxmtp/issues/657 for cleaning up other instances of this pattern, because I think there are quite a few, but it is probably best to keep outside the scope of this PR. Let me know if you think I misunderstood your suggestion 🙇

PR #644, opened 2024-04-10
Mutable Metadata group name Part 2

cameronvoell:
We would like to have a metadata extension type that can be modified after group creation. This is possible using the Unknown Group Context Extension type, which accepts a byte array as the only field. See https://github.com/xmtp/libxmtp/issues/548 for more info.

Related PR's:

- Part 1: https://github.com/xmtp/libxmtp/pull/643

- [x] Update openmls reference in `Cargo.toml` to commit on `xmtp/openmls` `main` branch
- [x] Update dev/gen_protos.sh to point to `xmtp-proto` repo `main` branch


insipx:
Something cool that nick found was using `graphite` to manage prs in parts like this, found it super useful so far in case you're interested, [graphite](https://graphite.dev/)

xmtp_mls/src/groups/mod.rs
```diff
@@ -378,6 +382,19 @@ where
         self.sync_until_intent_resolved(conn, intent.id).await
     }
 
+    pub async fn update_group_metadata(&self, group_name: String) -> Result<(), GroupError> {
```
neekolas, line 385: If all this function does is update the group name, maybe we should rename it to be more specific. Otherwise we'll have a breaking change when we add a second metadata field.

xmtp_mls/src/groups/mod.rs
```diff
@@ -464,13 +489,57 @@ fn build_protected_metadata_extension(
...
+pub fn build_mutable_metadata_extension(
+    group_name: String,
+) -> Result<Extension, GroupError> {
+    let mutable_metadata: GroupMutableMetadataV1 = GroupMutableMetadataV1 {
+        group_name,
+    };
+
+    let unknown_gc_extension = UnknownExtension(mutable_metadata.encode_to_vec());
+
+    // TODO: Where should the constant hex value live?
```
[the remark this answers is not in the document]
cameronvoell: good call - added here: https://github.com/xmtp/libxmtp/pull/644/files#diff-80e1437765f38906f066439ad9805b99205fb55f992bceb400a9d091628b6942

xmtp_mls/src/groups/mod.rs
cameronvoell: nice catch 🙏 - updated here: f2f1a966a0be2c8b8b1d4ee8450a86a6f201abd4

neekolas (approved):

PR #652, opened 2024-04-11
Message History 3/N

tuddman:
## Summary

- Stubs out the request and response objects and first pass at fns that use them.
- Includes the proto updates for supporting `MessageHistoryRequest` and `MessageHistoryResponse`
- one function name change:  `add_idempotency_key()` -> `into_envelope()`. to better reflect what the fn returns


xmtp_mls/src/groups/mod.rs
```diff
@@ -186,7 +186,7 @@ where
         let protected_metadata = build_protected_metadata_extension(
             &client.identity,
             permissions
-                .unwrap_or(PreconfiguredPolicies::default())
+                .unwrap_or_default()
```
neekolas, line 189: Did not know that existed. Nice

neekolas (approved):

tuddman:
tests are failing; fix is in https://github.com/xmtp/libxmtp/pull/654

PR #654, opened 2024-04-12
Fix group update

tuddman:
## Summary

After issue-ing `dev/gen-proto.sh` ,
Tests were failing with 

```
error[E0063]: missing field `update_group_name_policy` in initializer of `xmtp_proto::xmtp::mls::message_contents::PolicySet`
   --> xmtp_mls/src/groups/group_permissions.rs:295:12
    |
295 |         Ok(PolicySetProto {
    |            ^^^^^^^^^^^^^^ missing `update_group_name_policy`
```

This patch fixes the acute issue of the missing field, but idk if there is more work to do here to have it _correctly_ implemented?


xmtp_mls/src/groups/group_permissions.rs
```diff
@@ -324,14 +327,6 @@ impl PolicySet {
     }
 }
 
-fn default_add_installation_policy() -> MembershipPolicies {
-    MembershipPolicies::allow()
-}
-
-fn default_remove_installation_policy() -> MembershipPolicies {
-    MembershipPolicies::deny()
-}
```
tuddman, line 333: These seemed like a bit of needless indirection to only be called in the `new()` constructor above.

xmtp_mls/src/groups/group_permissions.rs
```diff
@@ -242,8 +243,9 @@ impl PolicySet {
         Self {
             add_member_policy,
             remove_member_policy,
-            add_installation_policy: default_add_installation_policy(),
-            remove_installation_policy: default_remove_installation_policy(),
+            add_installation_policy: MembershipPolicies::allow(),
+            remove_installation_policy: MembershipPolicies::deny(),
+            update_group_name_policy: MembershipPolicies::allow()
```
tuddman, line 248: @cameronvoell  What should this `update_group_name_policy` be default set to?

cc @neekolas

tuddman:
Closing.  The sames fixes were brought in https://github.com/xmtp/libxmtp/pull/643

PR #656, opened 2024-04-12
Get an Association State from a list of IdentityUpdates in Validation Service

### LINKED ISSUE #622: Expose API in MLS Validation Service for processing identity updates
The MLS Validation Service must provide an API that takes in a list of `IdentityUpdates` and returns the final AssociationState (or a relevant error).
```[tasklist]
### Tasks
- [x] Update protos for MLS validation service to include new `GetAssociationState` method
- [x] Implement method to get a final AssociationState
- [x] Define Protobuf serialization for AssociationState
```

insipx:
closes #622

needs to be merged after https://github.com/xmtp/proto/pull/160


dev/build_validation_service_local
```diff
@@ -8,9 +8,9 @@ if [ ! -x "$(command -v x86_64-linux-gnu-gcc)" ] && [ "$(uname)" = "Darwin" ]; t
     brew install x86_64-unknown-linux-gnu    
 fi
 
-rustup target add x86_64-unknown-linux-gnu
+# rustup target add x86_64-unknown-linux-gnu
```
neekolas, line 11: Won't we need this if the user's machine doesn't have that target available?

mls_validation_service/src/handlers.rs
```diff
@@ -81,6 +83,25 @@ impl ValidationApi for ValidationService {
...
+        &self,
+        request: Request<GetAssociationStateRequest>,
+    ) -> Result<Response<GetAssociationStateResponse>, Status> {
+        let updates = request
+            .into_inner()
+            .updates
+            .into_iter()
+            .map(IdentityUpdate::from_proto)
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| Status::new(Code::Cancelled, e.to_string()))?;
```
neekolas, line 97: Canceled doesn't feel like the right GRPC status code for this error. What about `Code:: InvalidArgument`

xmtp_id/src/associations/test_utils.rs
```diff
@@ -68,6 +79,35 @@ impl Signature for MockSignature {
     }
 
     fn to_proto(&self) -> SignatureProto {
-        SignatureProto { signature: None }
+        match self.signature_kind {
```
neekolas, line 82: Nice

xmtp_id/src/associations/association_log.rs
```diff
@@ -332,6 +338,22 @@ impl IdentityUpdate {
     }
 }
 
+impl TryFrom<IdentityUpdate> for IdentityUpdateProto {
```
neekolas, line 341: Thanks for adding these

mls_validation_service/src/handlers.rs
```diff
@@ -81,6 +83,25 @@ impl ValidationApi for ValidationService {
...
+        let state = associations::get_state(updates)
+            .map_err(|e| Status::new(Code::Cancelled, e.to_string()))?;
```
neekolas, line 99: This could be a `PermissionDenied` code or `InvalidArgument`

neekolas (approved):

neekolas:
I think you'll need to merge the protos for CI to pass, since `main` in `xmtp/proto` is ahead of your branch.

xmtp_id/src/associations/serialization.rs
```diff
@@ -219,6 +232,28 @@ fn from_signature_kind_proto(
     })
 }
 
+fn from_association_state_proto(
```
37ng, line 235: can this be a `TryFrom`?

xmtp_id/src/associations/serialization.rs
```diff
@@ -171,6 +177,13 @@ fn from_member_identifier_kind_proto(proto: MemberIdentifierKindProto) -> Member
     }
 }
 
+fn from_member_proto(member: MemberProto) -> Result<Member, DeserializationError> {
```
37ng, line 180: can this be a `TryFrom`?

xmtp_id/src/associations/serialization.rs
```diff
@@ -289,6 +324,13 @@ fn to_identity_action_proto(action: &Action) -> IdentityActionProto {
     }
 }
 
+fn to_member_proto(member: Member) -> MemberProto {
```
37ng, line 327: can this be `From`?

xmtp_id/src/associations/serialization.rs
```diff
@@ -300,6 +342,30 @@ fn to_member_identifier_proto(member_identifier: MemberIdentifier) -> MemberIden
     }
 }
 
+fn to_association_state_proto(association_state: AssociationState) -> AssociationStateProto {
```
37ng, line 345: can this be a `impl From<AssociationState> for AssociationStateProto`?

insipx: Yeah this is implemented below

The reason I split this up into functions is to follow the precedent already set in the serialization file. although I would prefer to only use `TryFrom` and `From` more generally. I implemented `From`/`TryFrom` in places where my code used it elsewhere

I think we should create a ticket for converting these standalone functions into their own `From`/`TryFrom` unless there is a good reason for keeping them as standalone. It's an easier task but would be nice not to make a different ticket too noisy. 

@neekolas @richardhuaaa WDYT of converting these into trait implementations?

mls_validation_service/src/handlers.rs
```diff
@@ -81,6 +86,48 @@ impl ValidationApi for ValidationService {
...
+            let new_state = associations::get_state(&new_updates).map_err(to_status)?;
+            return Ok(Response::new(GetAssociationStateResponse {
+                association_state: Some(new_state.clone().into()),
+                state_diff: Some(AssociationStateDiff::from(new_state).into()),
+            }));
+        }
+
+        let old_state = associations::get_state(&old_updates).map_err(to_status)?;
+        old_updates.append(new_updates.as_mut());
+        let new_state = associations::get_state(old_updates).map_err(to_status)?;
```
neekolas, line 123: `xmtp_id` has an `apply_update` method specifically designed for this case, so you don't need to re-run the initial updates.

insipx: nice! I'll add this in

xmtp_id/src/associations/serialization.rs
neekolas: One consideration I had when designing it this way is the likelihood that we are going to need to pass in some sort of `ERC1271SignatureVerifier` in to these functions so that we have everything we need to verify the signatures on the `IdentityUpdate`.

That struct is going to need to hold the RPC URL etc and can't just be a singleton.

neekolas (approved):
If @richardhuaaa is happy I'm happy. This looks good to me

mls_validation_service/src/handlers.rs
```diff
@@ -81,6 +86,51 @@ impl ValidationApi for ValidationService {
...
+        let conv_proto =
+            |updates: Vec<IdentityUpdateProto>| -> Result<Vec<IdentityUpdate>, Status> {
+                updates
+                    .into_iter()
+                    .map(IdentityUpdate::from_proto)
+                    .collect::<Result<Vec<_>, _>>()
+                    .map_err(|e| Status::new(Code::InvalidArgument, e.to_string()))
+            };
+
+        let to_status = |e: AssociationError| Status::new(Code::InvalidArgument, e.to_string());
```
neekolas, line 108: Seems like this could live outside the function, since it doesn't require any context

xmtp_id/src/associations/serialization.rs
37ng: We can have a mapping of CAIP-10 to RPC URL and keep it in libxmtp, so that we don't need to de/serialize them.

insipx: Opened a new [pr](https://github.com/xmtp/libxmtp/pull/672) to change some of these standalones into the trait where possible

mls_validation_service/src/handlers.rs
insipx: removed both closures

richardhuaaa (approved):
Sorry for the slow review, this looks great! Should be consumable by the server as-is

Issue #347, opened 2023-11-27
When joining a group from a welcome, members should update their keys before sending the first message


neekolas:

richardhuaaa:
Broader validation checklist: https://www.rfc-editor.org/rfc/rfc9420.html#name-credential-validation

lfzkoala:
I think this issue has become key updating not only joining a group from welcome, but also key updating before every new group intent that I did in https://github.com/xmtp/libxmtp/pull/658 and mentioned in https://xmtp-labs.slack.com/archives/C05GBHEAGPP/p1713302830615839

PR #658, opened 2024-04-15
Adding Key Update Before Every New GroupIntent

lfzkoala:
trying to solve this issue - https://github.com/xmtp/libxmtp/issues/347

This is what I'm doing

Follow the readme file ([xmtp_mls/migrations](https://github.com/xmtp/libxmtp/tree/main/xmtp_mls/migrations))  to create a migration for SQL database

Install the CLI onto local system (one-time)
cargo install diesel_cli --no-default-features --features sqlite

Create migration SQL
diesel migration generate key_update

Edit the up.sql and down.sql files created
Put in up.sql
``
ALTER TABLE key_update
ADD COLUMN rotated_at_ns BIGINT NOT NULL DEFAULT 0
``
Leave down.sql blank

Generate application code
cargo run --bin update-schema
This updates the generated schema.rs file. This updates models.rs to reference it and encrypted_store/mod.rs to define queries against the model.

Add a new item called `rotated_at_ns` in the StoredGroup struct in storage/encrypted_store/group.rs

Write a function called `get_rotated_time_checked`, which is similar to the function `get_installations_time_checked`, in order to fetch the `rotated_at_ns` time stamp. 

Then in the `create_from_welcome` function, set the `rotated_at_ns = 0` when creating the StoredGroup. 

Write an async function called `pre_intent_hook()`, which does the following
- Rotate the key if checking the `rotated_at_ns == 0`
- Call `maybe_update_installations()`
Before every `GroupIntent::New(...)`, call the `pre_intent_hook()` function. 

Add a boolean flag as an input of the `maybe_update_installations` function, set the flag to be true to skip the hook when it is called inside from the `pre_intent_hook()` function. This is intended to avoid infinite loops and make sure it is not called before the key update intent.


xmtp_mls/src/groups/mod.rs
```diff
@@ -223,11 +223,11 @@ where
...
         let stored_group = provider.conn().insert_or_ignore_group(to_store)?;
 
-        Ok(Self::new(
-            client,
-            stored_group.id,
-            stored_group.created_at_ns,
-        ))
+        let xmtp_group = Self::new(client, stored_group.id, stored_group.created_at_ns);
+
+        let _ = xmtp_group.queue_key_update();
```
neekolas, line 228: Once this intent is queued, it will cause the key update to be sent the next time `sync()` is run on the group. That will happen even if the member is purely a "listener" to the group.

We really only need to perform this key update for members who are active and sending messages, and so we would like to trigger this action on first message send so that we don't overwhelm the group with unnecessary key updates.

lfzkoala: I was trying to use the existing `key_update` function which is async, but noticed that the `create_from_welcome` function is sync. That's why I duplicate it to the `queue_key_update` function which is sync (thanks to @richardhuaaa for help) 

Based on your comments, it seems there are still serious concerns and I do think it becomes hard to write test for `queue_key_update` because many other functions are written in async way. I'm wondering what's the best approach to do this without causing other issues? I'm thinking changing all the calling chain to async, but it would be very complicated.

lfzkoala: It seems I got the point. The key update should be done once the group has the first message sent. Let me think how to do that

neekolas: We really want to do this when the individual installation is about to send their first message to the group.

richardhuaaa: From Daniel and I's discussion, here is how we can do it now:

```
Add column ‘rotated_at_ns’ to groups table
When creating a group from welcome, set ‘rotated_at_ns’ to 0
When sending a message:
  - If ‘rotated_at_ns’ is 0, rotate the key, and then set ‘rotated_at_ns’ to now()
```

xmtp_mls/src/groups/mod.rs
```diff
@@ -522,8 +541,44 @@ where
...
+
+    /// Checking the last key rotation time before rotating the key. 
+    pub async fn pre_intent_hook(&self) -> Result<(), GroupError> {
+        let conn = &self.client.store.conn()?;
+        let last_rotated_time = conn.get_rotated_time_checked(self.group_id.clone());
+        
+        match last_rotated_time {
+            Ok(rotated_time) => {
+                if rotated_time == 0 {
+                    let _ = self.queue_key_update();
```
richardhuaaa, line 572: The `rotated_time` is not currently updated after the key update is queued. That means the key update will be queued every single time, rather than only on an interval. I wonder if it makes sense to make a method `maybe_queue_key_update` similar to `maybe_update_installations`, that both checks the `rotated_time` and updates it if a key update is queued.

It might make sense to also add a unit test to catch this error

xmtp_mls/src/groups/mod.rs
```diff
@@ -522,8 +541,44 @@ where
...
+                    let update_interval = Some(5_000_000);
+                    self.maybe_update_installations(conn, update_interval).await?;
```
richardhuaaa, line 574: `maybe_update_installations` should not depend on `rotated_time` - it uses its own timestamp to determine whether an update is needed

xmtp_mls/src/groups/mod.rs
```diff
@@ -522,8 +541,44 @@ where
...
+                }
+            }
+            Err(err) => {
+                // Handle the error appropriately
+                log::error!("database error fetching rotated time {:?}", err);
```
richardhuaaa, line 579: You can probably just bubble the error up instead of using a match statement and then logging the error.

let last_rotated_time = conn.get_rotated_time_checked(self.group_id.clone())**?**;

xmtp_mls/src/groups/mod.rs
```diff
@@ -522,8 +541,44 @@ where
...
+        intent.store(conn)?;
+
+        self.sync_with_conn(conn).await
+    }
+
+    // Update the installation's leaf key in a synchronous way.
+    pub fn queue_key_update(&self) -> Result<(), GroupError> {
+        let conn = &mut self.client.store.conn()?;
+        
+        self.pre_intent_hook();
```
richardhuaaa, line 557: There's an infinite loop here because `queue_key_update` and `pre_intent_hook` both call each other.

What we can do is *not* call `pre_intent_hook` from `queue_key_update`. That way, `queue_key_update` can only be reached from `pre_intent_hook` but not the other way around. Then you can remove the `pub` from the signature of `queue_key_update`.

Also, I'm not sure if we need both `queue_key_update` and `key_update`, it seems like only one of the two is sufficient.

richardhuaaa: It might make sense to write a unit test that reproduces this, before fixing it, to make sure that the fix works

richardhuaaa: > Also, I'm not sure if we need both queue_key_update and key_update, it seems like only one of the two is sufficient.

I think it might make sense to keep `key_update` and remove `queue_key_update`, seeing as you no longer have the 'calling async method from inside sync method' issue. You don't need to update the key inside of `create_from_welcome` anymore, because the key update will happen on the next message send.

You can call `key_update` from inside of `pre_intent_hook`, but you'll need to remove `pre_intent_hook` from inside of `key_update` too

richardhuaaa:
Idea for test 1:

Client a makes a group with client B
Client B creates it from welcome
Verify no new payloads on client A
Call pre_intent_hook on client B
Verify client A receives a key rotation payload
Call pre_intent_hook on client B again
Verify client A receives nothing new
Register a new client for the same user as client A
Call pre_intent_hook on client A
Verify client B receives both a key rotation for client A and a payload to add the new installation

Idea for test 2:
Client a makes a group with client B
Client b creates it from welcome
Verify no new payloads on client A
Client b sends a message to client A
Verify client A receives a key rotation and then the message

xmtp_mls/src/groups/mod.rs
```diff
@@ -485,6 +484,7 @@ impl MlsGroup {
         if inbox_id_map.len() != account_addresses.len() {
             let found_addresses: HashSet<&String> = inbox_id_map.keys().collect();
             let to_add_hashset = HashSet::from_iter(account_addresses.iter());
+            self.pre_intent_hook(client).await?;
```
lfzkoala, line 487: @richardhuaaa I merged the new changes which use `inbox_id` instead and it seems we don't create new group intent in the `add_members` function - see [this commit ](https://github.com/xmtp/libxmtp/pull/658/commits/4a317b4d93b1013db101500f16d0d8eb42ac6474#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3L480)for a more clear comparison of what changed.

The original code was applying the `pre_intent_hook` function before every new group intent created, but now since it seems there is no new group intent created anymore and we directly add members by inbox id, my question was that I'm not sure if we still need to apply `pre_intent_hook` to update the key anymore (?)

If we still need, where is the best place to apply the key update?

xmtp_mls/src/groups/mod.rs
```diff
@@ -521,16 +530,7 @@ impl MlsGroup {
...
-        let intent_data: Vec<u8> = RemoveMembersIntentData::new(account_addresses.into()).into();
-
-        self.pre_intent_hook(client).await?;
-
-        let intent = conn.insert_group_intent(NewGroupIntent::new(
-            IntentKind::RemoveMembers,
-            self.group_id.clone(),
-            intent_data,
-        ))?;
+        let inbox_id_map = client.api_client.get_inbox_ids(account_addresses).await?;
```
lfzkoala: @richardhuaaa Also the same question here - we don't have new group intent anymore, if I should apply the `pre_intent_hook` and where is the best place to apply it?

xmtp_mls/src/groups/mod.rs
richardhuaaa: So there must be an intent created somewhere, because intents are the only way that we can produce commits, and you can't add a member to a group without creating a commit.

It may have just been moved deeper down into the call tree, looks like it might be here? https://github.com/xmtp/libxmtp/blob/187173415812fc85977d6534df7d06d8b393918e/xmtp_mls/src/groups/mod.rs#L505

Maybe you can move your pre_intent_hook there? 

In general, it might make sense to search for all instances of `Intent::new` or `IntentData::new` to see all cases where we are creating intents.

xmtp_mls/src/groups/mod.rs
richardhuaaa: See above answer!

richardhuaaa:
Please see #1108

Issue #655, opened 2024-04-12
Bug: iOS SDK is throwing "API error: setup error: transport error" for unknown reasons


bwcDvorak:
### Describe the bug

Partner iOS app is having their prod clients return `LibXMTP.GenericError.ApiError(message: "API error: setup error: transport error")` in Sentry from time to time. 

Seems to happen in the notification extension when they try to instantiate the XMTP Client in Swift to decode the notification. Appears to be related to this line:
`let client = try await Client.from(bundle: privateKeyBundle, options: .init(api: .init(env: xmtpEnv)))`
or the one after where `client.register(codec)` gets called

### Expected behavior

expected behavior is to not trigger the transport errors

PR #659, opened 2024-04-15
API error: setup error: transport error

nplasterer:
Mostly adds better error messaging so that we can pinpoint where in the setup this is coming from.

Sort of related:
I think I have a better answer for why removing http2 fixed this issue https://github.com/xmtp/libxmtp/pull/552

From a lot of internet digging I came across this implementation https://github.com/graphprotocol/graph-node/blob/449aed99d1488939596a1839e5dd2c8415d4a6a5/graph/src/firehose/endpoints.rs#L179C1-L195

With a handy code comment 

```
        // Note on the connection window size: We run multiple block streams on a same connection,
        // and a problematic subgraph with a stalled block stream might consume the entire window
        // capacity for its http2 stream and never release it. If there are enough stalled block
        // streams to consume all the capacity on the http2 connection, then _all_ subgraphs using
        // this same http2 connection will stall. At a default stream window size of 2^16, setting
        // the connection window size to the maximum of 2^31 allows for 2^15 streams without any
        // contention, which is effectively unlimited for normal graph node operation.
        //
        // Note: Do not set `http2_keep_alive_interval` or `http2_adaptive_window`, as these will
        // send ping frames, and many cloud load balancers will drop connections that frequently
        // send pings.
```

Not sure how much of that implementation we need. But tests seem to pass even with a long 16 minute delay the connection is still not lost. I added a lot of comments explaining each item. If code review is hard with the comments on just toggle the comments off. 

Hopefully fixes: https://github.com/xmtp/libxmtp/issues/655

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -949,6 +949,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 5)]
+    #[ignore]
     async fn test_conversation_streaming() {
         let amal = new_test_client().await;
         let bola = new_test_client().await;
--- a/xmtp_api_grpc/src/grpc_api_helper.rs
+++ b/xmtp_api_grpc/src/grpc_api_helper.rs
@@ -31,7 +31,7 @@ use xmtp_proto::{
 
 async fn create_tls_channel(address: String) -> Result<Channel, Error> {
     let channel = Channel::from_shared(address)
-        .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?
+        .map_err(|e| Error::new(ErrorKind::SetupCreateChannelError).with(e))?
         .initial_connection_window_size(Some((1 << 31) - 1))
         .keep_alive_while_idle(true)
         .connect_timeout(Duration::from_secs(10))
@@ -39,10 +39,10 @@ async fn create_tls_channel(address: String) -> Result<Channel, Error> {
         .timeout(Duration::from_secs(120))
         .keep_alive_timeout(Duration::from_secs(25))
         .tls_config(ClientTlsConfig::new())
-        .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?
+        .map_err(|e| Error::new(ErrorKind::SetupTLSConfigError).with(e))?
         .connect()
         .await
-        .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?;
+        .map_err(|e| Error::new(ErrorKind::SetupConnectionError).with(e))?;
 
     Ok(channel)
 }
@@ -74,10 +74,10 @@ impl Client {
             })
         } else {
             let channel = Channel::from_shared(host)
-                .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?
+                .map_err(|e| Error::new(ErrorKind::SetupCreateChannelError).with(e))?
                 .connect()
                 .await
-                .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?;
+                .map_err(|e| Error::new(ErrorKind::SetupConnectionError).with(e))?;
 
             let client = MessageApiClient::new(channel.clone());
             let mls_client = ProtoMlsApiClient::new(channel.clone());
--- a/xmtp_api_grpc/src/grpc_api_helper.rs
+++ b/xmtp_api_grpc/src/grpc_api_helper.rs
@@ -32,11 +32,29 @@ use xmtp_proto::{
 async fn create_tls_channel(address: String) -> Result<Channel, Error> {
     let channel = Channel::from_shared(address)
         .map_err(|e| Error::new(ErrorKind::SetupCreateChannelError).with(e))?
+        // Purpose: This setting controls the size of the initial connection-level flow control window for HTTP/2, which is the underlying protocol for gRPC.
+        // Functionality: Flow control in HTTP/2 manages how much data can be in flight on the network. Setting the initial connection window size to (1 << 31) - 1 (the maximum possible value for a 32-bit integer, which is 2,147,483,647 bytes) essentially allows the client to receive a very large amount of data from the server before needing to acknowledge receipt and permit more data to be sent. This can be particularly useful in high-latency networks or when transferring large amounts of data.
+        // Impact: Increasing the window size can improve throughput by allowing more data to be in transit at a time, but it may also increase memory usage and can potentially lead to inefficient use of bandwidth if the network is unreliable.
         .initial_connection_window_size(Some((1 << 31) - 1))
+        // Purpose: Configures whether the client should send keep-alive pings to the server when the connection is idle.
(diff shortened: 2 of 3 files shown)


xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -32,8 +32,11 @@ use xmtp_proto::{
 async fn create_tls_channel(address: String) -> Result<Channel, Error> {
     let channel = Channel::from_shared(address)
         .map_err(|e| Error::new(ErrorKind::SetupError).with(e))?
+        .initial_connection_window_size(Some((1 << 31) - 1))
```
nplasterer, line 35: **Purpose:** This setting controls the size of the initial connection-level flow control window for HTTP/2, which is the underlying protocol for gRPC.

**Functionality:** Flow control in HTTP/2 manages how much data can be in flight on the network. Setting the initial connection window size to (1 << 31) - 1 (the maximum possible value for a 32-bit integer, which is 2,147,483,647 bytes) essentially allows the client to receive a very large amount of data from the server before needing to acknowledge receipt and permit more data to be sent. This can be particularly useful in high-latency networks or when transferring large amounts of data.

**Impact:** Increasing the window size can improve throughput by allowing more data to be in transit at a time, but it may also increase memory usage and can potentially lead to inefficient use of bandwidth if the network is unreliable.

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -32,8 +32,11 @@ use xmtp_proto::{
...
         .keep_alive_while_idle(true)
```
nplasterer, line 36: **Purpose:** Configures whether the client should send keep-alive pings to the server when the connection is idle.

**Functionality:** When set to true, this option ensures that periodic pings are sent on an idle connection to keep it alive and detect if the server is still responsive.

**Impact:** This helps maintain active connections, particularly through NATs, load balancers, and other middleboxes that might drop idle connections. It helps ensure that the connection is promptly usable when new requests need to be sent.

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -32,8 +32,11 @@ use xmtp_proto::{
...
-        .connect_timeout(Duration::from_secs(5))
+        .connect_timeout(Duration::from_secs(10))
```
nplasterer, line 37: **Purpose:** Sets the maximum amount of time the client will wait for a connection to be established.

**Functionality:** If a connection cannot be established within the specified duration, the attempt is aborted and an error is returned.

**Impact:** This setting prevents the client from waiting indefinitely for a connection to be established, which is crucial in scenarios where rapid failure detection is necessary to maintain responsiveness or to quickly fallback to alternative services or retry logic.

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -32,8 +32,11 @@ use xmtp_proto::{
...
+        .tcp_keepalive(Some(Duration::from_secs(15)))
```
nplasterer, line 38: **Purpose:** Configures the TCP keep-alive interval for the socket connection.

**Functionality:** This setting tells the operating system to send TCP keep-alive probes periodically when no data has been transferred over the connection within the specified interval.

**Impact:** Similar to the gRPC-level keep-alive, this helps keep the connection alive at the TCP layer and detect broken connections. It's particularly useful for detecting half-open connections and ensuring that resources are not wasted on unresponsive peers.

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -32,8 +32,11 @@ use xmtp_proto::{
...
+        .timeout(Duration::from_secs(120))
```
nplasterer, line 39: **Purpose:** Sets a maximum duration for the client to wait for a response to a request.

**Functionality:** If a response is not received within the specified timeout, the request is canceled and an error is returned.

**Impact:** This is critical for bounding the wait time for operations, which can enhance the predictability and reliability of client interactions by avoiding indefinitely hanging requests.

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -32,8 +32,11 @@ use xmtp_proto::{
...
         .keep_alive_timeout(Duration::from_secs(25))
```
nplasterer, line 40: **Purpose:** Specifies how long the client will wait for a response to a keep-alive ping before considering the connection dead.

**Functionality:** If a ping response is not received within this duration, the connection is presumed to be lost and is closed.

**Impact:** This setting is crucial for quickly detecting unresponsive connections and freeing up resources associated with them. It ensures that the client has up-to-date information on the status of connections and can react accordingly.

tuddman (approved):
connection tuning is always more art than science.  I'll trust these settings (in seconds) have some basis for improving the reliability of the connection and mitigate some of the dropped connections we've seen.

xmtp_api_grpc/src/grpc_api_helper.rs
nplasterer: @neekolas any strong feelings for or against this one? This one feels like the biggest change but should help with latency issues.

xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -31,15 +31,18 @@ use xmtp_proto::{
 
 async fn create_tls_channel(address: String) -> Result<Channel, Error> {
     let channel = Channel::from_shared(address)
```
nplasterer, line 33: Now I'm not sure these changes are going to fix it cause it looks like the transport error is only thrown from this call

`https://github.com/hyperium/tonic/pull/920/files#diff-2dc4a5ebbcd9a8a198e55baa6958f271c1df257a9e3d6ae9c70295a4df7773deR77`

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: Should we drop these descriptions in as code comments? They are going to be very helpful for future maintenance

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: This seems worth a shot. I'm not quite sure how it'll work with GRPC streams, but it's a good sign that The Graph has been using this setting in production and it hasn't blown up.

neekolas:
Love all the details around why these particular settings matter. Where did you find the info in all these comments?

neekolas (approved):

xmtp_api_grpc/src/grpc_api_helper.rs
neekolas: Based on my read of that code comment, it'll only come into play for clients that are connecting to a _lot_ of different streams at once. But it's worth a try.

bindings_ffi/src/mls.rs
```diff
@@ -949,6 +949,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 5)]
+    #[ignore]
```
nplasterer, line 952: @richardhuaaa marked this as ignored for now.

PR #660, opened 2024-04-15
More message history definitions

tuddman:
## Summary

Small patch which further defines the message history `request` and `reply`.  Updates the protos to match the changes introduced in https://github.com/xmtp/proto/pull/165


xmtp_mls/src/groups/message_history.rs
```diff
@@ -16,49 +19,79 @@ where
...
+}
+
+fn new_pin() -> String {
+    let mut rng = rand::thread_rng();
+    let pin: u32 = rng.gen_range(0..10000);
+    format!("{:04}", pin)
+}
+
+fn new_request_id() -> String {
+    Alphanumeric.sample_string(&mut rand::thread_rng(), 24)
```
richardhuaaa, line 78: How will this request_id be used?

richardhuaaa (approved):

xmtp_mls/src/groups/message_history.rs
tuddman: Thinking was that a responder would offer a reply to a specifc msg-history request.
I had thought that using the `idempotency_key` for this could work, but that's serving a different purpose at least for regular messages.

and also its in the original XIP.

We _may_ not need this value in the end; as it may be sufficient that: given any msg-history request that gets sent, and any msg-history reply would have passed through the PIN check to get sent, and  so the requesting installation-id can assume _any_ msg-history reply should be pulled down.

PR #663, opened 2024-04-16
closed - Mutable Metadata group name Part 3

cameronvoell:
This PR adds the Validation that group names obey the `update_group_name_policy` set upon creating a group.

See https://github.com/xmtp/libxmtp/issues/548 for more info.

Previous PR's:

- Part 1: https://github.com/xmtp/libxmtp/pull/643
- Part 2: https://github.com/xmtp/libxmtp/pull/644

PR's related to this change:

- xmtp openmls: https://github.com/xmtp/openmls/pull/23
- proto: https://github.com/xmtp/proto/pull/161

- [ ] Update openmls reference in `Cargo.toml` to commit on `xmtp/openmls` `main` branch
- [ ] Update dev/gen_protos.sh to point to `xmtp-proto` repo `main` branch


xmtp_mls/src/groups/group_permissions.rs
```diff
@@ -6,11 +6,215 @@
...
+    MembershipPolicy as MembershipPolicyProto, MetadataPolicy as MetadataPolicyProto,
+    PolicySet as PolicySetProto,
 };
 
 use super::validated_commit::{AggregatedMembershipChange, CommitParticipant, ValidatedCommit};
 
+// A trait for policies that can update Metadata for the group
+
+pub trait MetadataPolicy: std::fmt::Debug {
+    fn evaluate(&self, actor: &CommitParticipant, change: bool) -> bool;
```
neekolas, line 22: I would recommend having a struct for the change, which includes the new value (and possibly the previous value).

It also seems like if there is no change, we should not even need to evaluate the policy.

The simple boolean works fine when the policy is just "can I change the group name", but it's easy to imagine more complicated policies being required. For example, if the value is an array we may want to evaluate whether items have been added or removed, not just if there is an equality.

cameronvoell:
this PR branch was made off of branch from https://github.com/xmtp/libxmtp/pull/644 which got squash merged to libxmtp main, so I'm cherry-picking changes in this PR to a new branch here, so the PR shows the correct new commits to be merged to main:
- https://github.com/xmtp/libxmtp/pull/677

xmtp_mls/src/groups/group_permissions.rs
cameronvoell: @neekolas commit history in this PR was off since it was made off a branch that was squash merged to main (I'll try graphite on my next multi PR change)

I cherry-picked changes to branch here => https://github.com/xmtp/libxmtp/pull/677, and I believe I addressed your concerns in this commit https://github.com/xmtp/libxmtp/pull/677/commits/f6ae8d8430cab499a95e3c0438def42c703aa4b2

Issue #516, opened 2024-02-15
Message send state


richardhuaaa:
Expose optimistic message sends.

Per conversation/request from a partner on April 10, when we do this work, let's make sure
`await group.send` returns the group id (it currently returns the message ID)

nmalzieu:
Hey @richardhuaaa , it's the opposite :) It currently returns the group id but it should return the message id (as we already know the group id!)

richardhuaaa:
Context: https://github.com/xmtp/libxmtp/issues/529

richardhuaaa:
After https://github.com/xmtp/libxmtp/pull/544/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f, optimistic message sends will be returned when listing messages in a conversation, with delivery status UNPUBLISHED.

- We might need a way to get the delivery status of messages
- We could return the message ID from a message send, but unsure if it still matters here - the integrator doesn't need to manually create the optimistic message, it will be in the libxmtp DB
- The optimistic message will be returned when listing messages, but won't be returned on subscriptions until they are updated to PUBLISHED. Does this matter?

richardhuaaa:
Naomi has a good Q: if the message send fails after 3 retries, do we delete the optimistic message?

richardhuaaa:
Maybe in addition to UNPUBLISHED and PUBLISHED, we should have a FAILED status

richardhuaaa:
Maybe add option to filter published/unpublished message

nmalzieu:
> After https://github.com/xmtp/libxmtp/pull/544/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f, optimistic messages will be inserted into the libxmtp DB and returned to the integrator when listing messages in a conversation, with delivery status UNPUBLISHED.
> 
> * We might need a way to get the delivery status of messages
> * We could return the message ID from a message send, but unsure if it still matters here - the integrator doesn't need to manually create the optimistic message, it will be in the libxmtp DB
> * The optimistic message will be returned when listing messages, but won't be returned on subscriptions until they are updated to PUBLISHED. Does this matter?

Getting back the message id from `await group.send()` would still be useful, because as you said, the subscriptions won't return the message, but we need a way to display it immediatly, and for that we need its unique id (so that when it's published and we get it back through subscriptions, we recognize it and don't show duplicates in the UI)

PR #669, opened 2024-04-17
Message send state

nplasterer:
Part of https://github.com/xmtp/libxmtp/issues/516

- [x] Add enum field of failed to delivery status for when unpublished never publishes
- [x] Expose delivery status
- [x] Expose delivery status filtering on find_messages
- [x] Expose message id on send

In a follow up PR we need to figure out how to set the delivery status to failed after 3 unsuccessful retries to publish. @tuddman maybe you have some ideas on how to do that?

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -338,7 +338,7 @@ pub struct FfiListMessagesOptions {
 
 #[uniffi::export(async_runtime = "tokio")]
 impl FfiGroup {
-    pub async fn send(&self, content_bytes: Vec<u8>) -> Result<(), GenericError> {
+    pub async fn send(&self, content_bytes: Vec<u8>) -> Result<Vec<u8>, GenericError> {
         let group = MlsGroup::new(
             self.inner_client.as_ref(),
             self.group_id.clone(),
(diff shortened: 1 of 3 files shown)


tuddman:
> maybe you have some ideas on how to do that?
Keeping a counter of attempts is pretty straightforward; in-memory or even stored-on-disk,  Exponential back-off, too, if we want to get fancy.
I'm keen to learn in what condition(s) publishing could or would fail, and see if we can ensure _that_ never happens.
(basically making it more correct over hardening it from being wrong)

nplasterer:
In the follow up PR we should do the following

1. Mark the optimistic message as FAILED whenever its intent is marked as failed (set_group_intent_error)
2. Make sure that every possible [error case in this method](https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/src/groups/sync.rs#L156) results in the corresponding intent either being marked as TO_PUBLISH or FAILED
3. At some point in the future revisit whether there is a possibility of an infinite loop of intent retries. Maybe we should call increment_intent_publish_attempt_count() wherever we reset an intent to TO_PUBLISH, and just set a really high retry limit, like 10?

For (2), maybe the most fool-proof way is to mark all MessageProcessingErrors as either retryable or non-retryable, and have the caller of the method catch all errors and mark the intent as TO_PUBLISH for retryable errors and FAILED for non-retryable errors

richardhuaaa (approved):
Looks great

PR #672, opened 2024-04-18
Use `From` in xmtp_id serialization

insipx:
Continuation of comments in #622 .
- converts standalone serialization fns to use `From`
- removes some implicit cloning with `From<&T>` in favor of moving owned values.


neekolas (approved):
Thanks for doing this

PR #673, opened 2024-04-18
signature verification

### LINKED ISSUE #636: Implement signature verification for all known signature types
In [this PR](https://github.com/xmtp/libxmtp/pull/633/files#diff-89b6a0b4851dce77ffc6a6e49fbb747fbb5397d1948a36991c108087fb0ac684) we have a bunch of new signature types introduced, but the `recover_signer` method is stubbed out for all of them. 

We'll need real `recover_signer` methods for all signature types before we can have IdentityUpdate verification working end to end.

Feel free to update the struct and error types as-needed to make this happen.
```[tasklist]
### Tasks
- [ ] Move stuff that belongs to `xmtp_cryptography` to it
- [ ] working version
- [ ] remove xmtp_mls dependencies
- [ ] refactor `allowed_signature_for_kind` into xmtp_id::Signature
- [ ] CAIP-10 struct
```

37ng:
closes #636

# tl:dr

- impl `recover_signer` and make it async with A LOT OF `await`, `async`.
- moved stuff from `xmtp_mls` to `xmtp_id`, `xmtp_cryptography`. (fwiw, `xmtp_cryptography` can be completely moved to `xmtp_id` and be deleted)

# TODO

inject `chain_rpc_url` into ERC1271Signature.


37ng:
* **#673** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

xmtp_id/src/associations/signature.rs
```diff
@@ -60,181 +102,253 @@ impl Clone for Box<dyn Signature> {
...
-                signature: self.bytes(),
             })),
         }
     }
+
+    fn verify(&self) -> Result<(), SignatureError> {
+        let verifier = ERC1271Verifier::new(CHAIN_RPC_PROVIDER);
+
+        // wrap the async function with tokio Runtime
+        let runtime = Runtime::new().unwrap();
```
neekolas, line 222: FWIW, I don't mind adjusting the trait so that this method can be async

xmtp_id/src/associations/signature.rs
```diff
@@ -60,181 +102,253 @@ impl Clone for Box<dyn Signature> {
...
-        self.signature_bytes.clone()
+    fn to_bytes(&self) -> Vec<u8> {
+        self.bytes.clone()
     }
 
     fn to_proto(&self) -> SignatureProto {
         SignatureProto {
             signature: Some(SignatureKindProto::Erc1271(Erc1271SignatureProto {
-                contract_address: self.contract_address.clone(),
+                message: self.message.clone(),
```
neekolas, line 209: We should not be including the message into the serialized signature.

Part of how we get the security guarantees out of this system is that we re-generate the signature text based on the actions in the `IdentityUpdate`. If a malicious client can include a custom message here, then you could re-use a signature that was signed for another purpose or from another inbox.

xmtp_id/src/associations/signature.rs
```diff
@@ -60,181 +102,253 @@ impl Clone for Box<dyn Signature> {
...
-        self.signature_bytes.clone()
+    fn to_bytes(&self) -> Vec<u8> {
+        self.bytes.to_vec()
     }
 
     fn to_proto(&self) -> SignatureProto {
         SignatureProto {
             signature: Some(SignatureKindProto::Erc191(RecoverableEcdsaSignatureProto {
-                bytes: self.bytes(),
+                message: self.message.clone(),
```
neekolas, line 144: Same as the comment below. The `message` cannot be trusted to come from the payload. It needs to be generated by the client doing the verifying.

xmtp_id/src/associations/signature.rs
```diff
@@ -60,181 +102,253 @@ impl Clone for Box<dyn Signature> {
...
+                chain: self.chain.into(),
+                contract_address: self.contract_address.to_string(),
                 block_number: self.block_number,
-                signature: self.bytes(),
             })),
         }
     }
+
+    fn verify(&self) -> Result<(), SignatureError> {
+        let verifier = ERC1271Verifier::new(CHAIN_RPC_PROVIDER);
```
neekolas, line 219: This is fine as a hack for now, but we are going to need devs to provide their own `CHAIN_RPC_PROVIDER` value as part of client creation. These RPC providers cost money and most dapps already have an account with one that they would like to use.

xmtp_id/src/associations/association_log.rs
```diff
@@ -69,35 +65,34 @@ impl IdentityAction for CreateInbox {
...
     fn signatures(&self) -> Vec<Vec<u8>> {
-        vec![self.initial_address_signature.bytes()]
+        vec![self.signature.to_bytes()]
     }
 }
 
 /// AddAssociation Action
 pub struct AddAssociation {
     pub new_member_signature: Box<dyn Signature>,
-    pub new_member_identifier: MemberIdentifier,
```
neekolas, line 100: We need this to get the signature text, which we need to generate before we can verify the signature

xmtp_id/Cargo.toml
```diff
@@ -22,9 +24,11 @@ rand.workspace = true
 serde.workspace = true
 sha2 = "0.10.8"
 thiserror.workspace = true
+tokio.workspace = true
 tracing.workspace = true
 xmtp_cryptography.workspace = true
 xmtp_proto = { workspace = true, features = ["proto_full"] }
+xmtp_mls.workspace = true
```
neekolas, line 31: Anything you need from `xmtp_mls` should get copied in to this crate, since we don't want a circular dependency.

xmtp_id/src/associations/signature.rs
neekolas: Ohhh, I see. We can't do that because this is all happening inside `new(...)` and that gets run from inside synchronous methods.

xmtp_id/src/associations/signature.rs
```diff
@@ -60,181 +102,253 @@ impl Clone for Box<dyn Signature> {
...
+    ) -> Result<Self, SignatureError> {
+        let sig = Erc1271Signature {
+            message,
+            bytes,
+            chain,
             contract_address,
             block_number,
-        }
+        };
+        sig.verify()?;
```
neekolas, line 188: I do like this change to doing all the verification as part of `new`. Maybe we want to rename the types to `ValidatedSignature` or something to make it more obvious to callers that these are safe.

xmtp_id/src/associations/member.rs
```diff
@@ -17,7 +19,7 @@ impl std::fmt::Display for MemberKind {
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum MemberIdentifier {
     Address(String),
-    Installation(Vec<u8>),
+    Installation([u8; 32]),
```
neekolas, line 22: Nice change

xmtp_id/src/associations/signature.rs
37ng: Super helpful, thank you

xmtp_id/src/associations/signature.rs
neekolas: I was thinking about this a bit more. I'm not sure verifying the signatures as part of `new()` is a good idea here.

- This change puts the validation logic in two places. Some of it happens during `apply_update` and some of it happens during deserialization. That makes errors harder to handle and manage. For example, if we want to allow some Identity Updates verification failures to just cause the update to be skipped instead of aborting the entire process, we would have to do that in two places.
- It means it's not possible to deserialize an identity update without also verifying the signatures. That should probably be optional and up to the developer.
- We're now doing network calls to verify 1271 signatures inside `TryFrom` statements, which is pretty unintuitive. We have to spin up a new Runtime to do async work inside a synchronous process, which we'll have to be super careful about since it can lead to deadlocks or weird performance characteristics.

Maybe a better pattern is to have two signature structs. An `UnverifiedSignature` and a `VerifiedSignature`. Converting from an unverified -> verified could be an async process to allow for 1271 network calls. We can then make `update_state` async and do it as part of that process. The previous `recover_signer()` method would also still work, if we made it async.

37ng: The main reason I wanted to verify the signature during `new()` was I didn't want a signature to be verified multiple time during its lifetime since it's expensive especially for 1271.

What you said makes sense, I think I'll keep it simple by sticking to the original design that has one `Signature` trait with `recover_signer()` and have all functionalities done. I can think about `UnverifiedSignature` and `VerifiedSignature` after that.

neekolas: The good news is that the code path that actually looks at these signatures is pretty narrow, so we can be careful to only call `recover_signer` once and then re-use that value.

But you're right, we do only want to do it once.

xmtp_id/src/associations/mod.rs
```diff
@@ -17,26 +17,23 @@ pub use self::signature::{Signature, SignatureError, SignatureKind};
...
-        },
-    )?;
-
-    new_state.ok_or(AssociationError::NotCreated)
+    let mut state = None;
+    for update in updates.as_ref().iter() {
+        state = Some(update.update_state(state).await?);
+    }
+
+    state.ok_or(AssociationError::NotCreated)
```
37ng, line 36: I refactored `try_fold` to a for loop, can probably do sth similar to `Future:all`.

xmtp_id/src/associations/signature.rs
```diff
@@ -238,3 +296,284 @@ impl Signature for LegacyDelegatedSignature {
...
+            89, 162, 128, 96, 96, 120, 131, 17, 70, 38, 231, 2, 27, 91, 29, 66, 110, 128, 140, 1,
+            42, 217, 185, 2, 181, 208, 100, 143, 143, 219, 159, 174, 1, 233, 191, 16, 1,
+        ];
+        let mut proto = LegacySignedPublicKeyProto::decode(proto_bytes.as_slice()).unwrap();
+        proto.key_bytes[0] += 1; // Corrupt the serialized key data
+        assert!(matches!(
+            ValidatedLegacySignedPublicKey::try_from(proto),
+            Err(super::SignatureError::Invalid)
+        ));
+    }
```
37ng, line 487: line 311 to 487 is moved from `xmtp_mls`.

37ng (commented):
ValidatedLegacySignedPublicKey is validated during deserialization. Do we want to align it with other signatures where validation happens at `recover_signer`.

37ng:
forgot to change proto, auch

xmtp_id/src/associations/signature.rs
```diff
@@ -98,35 +124,51 @@ impl Signature for RecoverableEcdsaSignature {
...
 impl Signature for Erc1271Signature {
-    fn recover_signer(&self) -> Result<MemberIdentifier, SignatureError> {
-        // TODO: Verify signature first
-        Ok(self.contract_address.clone().into())
+    async fn recover_signer(&self) -> Result<MemberIdentifier, SignatureError> {
+        let verifier = crate::erc1271_verifier::ERC1271Verifier::new(self.chain_rpc_url.clone());
+        // let runtime = Runtime::new().unwrap();
+        let is_valid = verifier
+            .is_valid_signature(
+                Address::from_slice(self.contract_address.as_bytes()), // TODO: `from_slice` will panic when input is not 20 bytes
```
neekolas, line 161: I'd expect the contract address to be a CAIP-10 address, at least according to the XIP.

xmtp_id/src/associations/signature.rs
```diff
@@ -182,24 +233,24 @@ impl Signature for InstallationKeySignature {
...
     legacy_key_signature: RecoverableEcdsaSignature,
-    signed_public_key: xmtp_proto::xmtp::message_contents::SignedPublicKey,
+    signed_public_key: ValidatedLegacySignedPublicKey,
 }
 
 impl LegacyDelegatedSignature {
     pub fn new(
         legacy_key_signature: RecoverableEcdsaSignature,
-        signed_public_key: xmtp_proto::xmtp::message_contents::SignedPublicKey,
+        signed_public_key: ValidatedLegacySignedPublicKey,
```
neekolas, line 253: We can handle in a subsequent PR, but I do think it would be cleaner if these were unvalidated to start and all validation happened in `recover_signer` like the other signature types.

xmtp_id/src/associations/signature.rs
37ng: Here are my understanding of the flow:
- the platform SDK calls `getRequiredSignatures` and will need to pass in RPC URLs if required signatures contains an ERC1271 signature.
- We can do a sanity check that if the RPC URLs match the chain id specified in the CAIP-10.
- Use the RPC URL later on in the `recover_signer`.

I plan to break down the CAIP-10 string into pieces during deserialization .

WDYT?

neekolas: > the platform SDK calls getRequiredSignatures and will need to pass in RPC URLs if required signatures contains an ERC1271 signature.

I think the platform SDK is going to need to pass in a RPC URL at the time of client creation. Clients should be in control of their RPC provider, and the signer cannot be trusted to provide a valid `rpc_url` as part of signature generation (what if I created a malicious RPC provider that validated signatures which were not actually valid on the blockchain).

We can hack it for now by hardcoding a RPC URL as a constant in our code. 

> I plan to break down the CAIP-10 string into pieces during deserialization .

Sure, that sounds fine

neekolas (approved):

xmtp_id/src/associations/signature.rs
37ng: I think wallet apps will probably provide those RPC URLs after signing. Otherwise each inbox app will have to provide a set of node RPC URLs for all the chains where the wallets reside.

37ng: PR #682 adds the CAIP-10 support

PR #679, opened 2024-04-19
Get association state diff

neekolas:


neekolas:
* **#679** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/679?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

xmtp_mls/src/identity_updates.rs
```diff
@@ -75,6 +78,33 @@ where
 
         Ok(get_state(updates)?)
     }
+
+    pub fn get_association_state_diff<InboxId: AsRef<str>>(
+        &self,
+        conn: &'a DbConnection<'a>,
+        inbox_id: String,
+        starting_sequence_id: Option<i64>,
+        ending_sequence_id: Option<i64>,
```
richardhuaaa, line 87: The latest iteration of the protos don't use sequence_id, but a list of old and new updates: https://github.com/xmtp/proto/pull/160/files

Not using sequence ID means that the validation service can work with `IdentityUpdate` instead of `IdentityUpdateLog`, and it means validation can be done on the client and/or before insertion into the server database. Your call here!

xmtp_mls/src/identity_updates.rs
```diff
@@ -75,6 +78,33 @@ where
...
+
+        let incremental_updates = conn
+            .get_identity_updates(inbox_id, starting_sequence_id, ending_sequence_id)?
+            .into_iter()
+            .map(|update| update.try_into())
+            .collect::<Result<Vec<IdentityUpdate>, AssociationError>>()?;
+
+        let final_state = incremental_updates
+            .into_iter()
+            .try_fold(initial_state.clone(), |state, update| {
```
richardhuaaa, line 102: Will this clone on every step of the iteration?

xmtp_mls/src/identity_updates.rs
```diff
@@ -75,6 +78,33 @@ where
...
+    pub fn get_association_state_diff<InboxId: AsRef<str>>(
+        &self,
+        conn: &'a DbConnection<'a>,
+        inbox_id: String,
+        starting_sequence_id: Option<i64>,
+        ending_sequence_id: Option<i64>,
+    ) -> Result<AssociationStateDiff, ClientError> {
+        let initial_state = self.get_association_state(conn, &inbox_id, starting_sequence_id)?;
+        if starting_sequence_id.is_none() {
+            return Ok(AssociationStateDiff::from(initial_state));
```
37ng, line 91: I find this semantically confusing to `From` a state into a state diff. Does it make sense inline the logic here and remove `From`?
```
AssociationStateDiff {
            new_members: state.members.keys().cloned().collect(),
            removed_members: vec![],
        }
```

xmtp_id/src/associations/state.rs
```diff
@@ -127,6 +127,15 @@ impl AssociationState {
...
 }
 
+impl From<AssociationState> for AssociationStateDiff {
+    fn from(state: AssociationState) -> Self {
+        AssociationStateDiff {
+            new_members: state.members.keys().cloned().collect(),
+            removed_members: vec![],
+        }
+    }
+}
```
richardhuaaa, line 137: When would we need this one?

richardhuaaa (approved):

xmtp_mls/src/identity_updates.rs
richardhuaaa: Another option could be to reverse the operands in `initial_state.diff(&final_state)` to `final-state.diff(&initial_state)`, and allow `initial_state` to be null or empty

xmtp_mls/src/identity_updates.rs
neekolas: This is for the client's need to generate a diff when evaluating commits. There we need to search the SQLite database for ranges of `sequence_id`s before we can get the Identity Updates

xmtp_mls/src/identity_updates.rs
neekolas: Ya, it is a bit of a stretch. If it's confusing I can just inline it or make a separate function

xmtp_mls/src/identity_updates.rs
neekolas: It will. `apply_update` treats the existing state as immutable so that you can discard updates if they fail part-way.

I think in at least some cases (one way associations come to mind) we are going to want to move past invalid updates rather than failing the entire process.

xmtp_id/src/associations/state.rs
neekolas: I actually borrowed it from Andrew's PR. You need it when you want to work with a diff, but there is only one entry in the list.

I do think @37ng is right that it's kinda confusing to use a `From` to do this.

neekolas: The validation service is using it for the same purpose

insipx (approved):

xmtp_id/src/associations/state.rs
insipx: `as_diff` definitely clearer. Just needed a way to make a `Default` diff in case we only had one update for a diff.

37ng (approved):

PR #680, opened 2024-04-19
Allow message history sync

tuddman:
## Summary

More plumbing to allow the history to be synced.


xmtp_mls/src/builder.rs
```diff
@@ -218,7 +218,13 @@ where
             .identity_strategy
             .initialize_identity(&api_client_wrapper, &store)
             .await?;
-        Ok(Client::new(api_client_wrapper, network, identity, store))
+        let new_client = Client::new(api_client_wrapper, network, identity, store);
+        // scott: I don't know where else this could go...
+        new_client
+            .create_sync_group()
+            .map_err(ClientBuilderError::StorageError)?;
```
tuddman, line 225: I'm not super happy with puting client functionality in the builder itself.
We could have a `build()` fn that returns a client, and also have a  `build_with_sync_group()`.  wdyt?

neekolas (approved):

PR #684, opened 2024-04-24
Return a canMessage address bool hashmap

nplasterer:
Requested here: https://github.com/xmtp/xmtp-react-native/issues/351 the can message function is a lot more useful to integrators if it's clear what addresses are not on the network.


tuddman (approved):
👍

PR #685, opened 2024-04-24
Validation happens in `recover_signer()` instead of in deserialization

37ng:
# Summary

1. Move validation logic into `recover_signer()`
2. Move de/serialization logic to signature.rs

Resolves this [comment](https://github.com/xmtp/libxmtp/pull/673#discussion_r1576390550)

# TODO
I'm still not fond of using a proto in the Signature struct and think a "unvalidated" Signature struct that directly convert into/from Proto would be nicer.
Also, seems like the we can simply `ValidatedLegacySignedPublicKey` by flattening 4 nested layers to 2. I'm happy to take on this in the future but now would like to focus on demo-related work first.

### DIFF
--- a/xmtp_id/src/associations/serialization.rs
+++ b/xmtp_id/src/associations/serialization.rs
@@ -5,7 +5,7 @@ use super::{
     member::Member,
     signature::{
         AccountId, Erc1271Signature, InstallationKeySignature, LegacyDelegatedSignature,
-        RecoverableEcdsaSignature,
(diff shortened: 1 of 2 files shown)


neekolas (approved):

neekolas:
> I'm still not fond of using a proto in the Signature struct and think a "unvalidated" Signature struct that directly convert into/from Proto would be nicer.

We can tackle this whenever we have time. Agree it'd be nicer

PR #688, opened 2024-04-29
IdentityStrategy

neekolas:
## tl;dr

- Starts scaffolding the new Identity struct to work with `inbox_id`

## Missing pieces

- Everything to do with `IdentityStrategy`
- We will need to somehow get the nonce as well as the `inbox_id` from the platform SDK
- Database model to persist the new identity


neekolas:
* **#688** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/688?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

xmtp_mls/src/identity/xmtp_id/identity.rs
```diff
@@ -0,0 +1,240 @@
...
+        account_address: String,
+        api_client: ApiClientWrapper<ApiClient>,
+    ) -> Result<Self, IdentityError> {
+        let signature_keys = SignatureKeyPair::new(CIPHERSUITE.signature_algorithm())?;
+        let installation_public_key = signature_keys.public();
+        let member_identifier: MemberIdentifier = account_address.into();
+
+        let builder = SignatureRequestBuilder::new(inbox_id);
+
+        if !Self::is_existing_inbox(inbox_id, api_client).await? {
```
neekolas, line 75: Maybe we want to do this check as part of the IdentityStrategy, and just pass in the result. I guess it'll depend on whether we need this information in the strategy's business logic too. We'd only want to do this expensive call once.

xmtp_mls/src/identity/xmtp_id/identity.rs
```diff
@@ -0,0 +1,240 @@
...
+        let installation_key_sig = InstallationKeySignature::new(
+            signature_text.clone(),
+            sig.to_vec(),
+            verifying_key.as_bytes().to_vec(),
+        );
+
+        Ok(installation_key_sig)
+    }
+
+    async fn sign_with_legacy_key(
```
neekolas, line 198: This probably doesn't need to live on `Identity`. Could just be a plain function

37ng:
replaced by #688

37ng:
@neekolas My thoughts around `IdentityStrategy` is documented [here](https://github.com/xmtp/libxmtp/pull/688/files#diff-4a05f3ef4fc1fd2c7641e3c7a937420268e93d4fddbf85670cd96b8ba6d8d135R22-R48).

Simply put, `IdentityStrategy` only cover two cases - create a new inbox if `address` doesn't have one, or use the `address`-associated inbox. Other inbox <> address related actions should happen elsewhere.

37ng:
Haven't resolved comments, but I think it's reviewable now.

xmtp_mls/src/identity/xmtp_id/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    ///
+    /// inbox_id = libxmtp.get_inbox_id(wallet_address)
+    /// if !inbox_id {
+    ///     inbox_id = libxmtp.generate_inbox_id(wallet_address, 0)
+    ///     if use_legacy_key() {
+    ///         pass // We will create inbox & add association in one call in `IdentityStrategy.initialize_identity(api_client, store)`.
+    ///     } else {
+    ///         inbox_id = libxmtp.generate_inbox_id(wallet_address, random_nonce)
+    ///         libxmtp.create_new_inbox_with_user_signature(inbox_id, wallet_address)
+    ///         TODO: Similar to use_legacy_key(), we can include this in one call so that user doesn't have to sign twice.
```
37ng, line 34: @neekolas what do you think about this TODO?

neekolas: Definitely fine to leave as a TODO. We can discuss in our sync tomorrow.

xmtp_mls/src/identity/xmtp_id/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+                stored_identity.ok_or(ClientBuilderError::RequiredIdentityNotFound)
+            }
+            IdentityStrategy::CreateIfNotFound(member) => {
+                let identity = match member {
+                    Member::Address(address) => {
+                        let inbox_ids = api_client.get_inbox_ids(vec![address.clone()]).await?;
+                        let inbox_id = inbox_ids
+                            .get(&address)
+                            .ok_or(ClientBuilderError::UncoveredCase)?;
+                        if stored_identity.is_some() {
```
neekolas, line 80: This might read better as a match statement

neekolas (commented):
I can't approve since technically it's my own PR, but this looks good. Some lint warnings that will need to be fixed, but otherwise feels solid.

One thing we should think about for future improvement is what happens in cases where the platform SDK _has_ a legacy key and uses `Member::LegacyKey` but we don't want to actually use it (because the inbox id is not 0 or the inbox has already been created and this is a second installation). Can we make the platform SDKs job easier by just falling back to the `Member::Address` flow in that case?

37ng:
> One thing we should think about for future improvement is what happens in cases where the platform SDK has a legacy key and uses Member::LegacyKey but we don't want to actually use it (because the inbox id is not 0 or the inbox has already been created and this is a second installation). Can we make the platform SDKs job easier by just falling back to the Member::Address flow in that case?

Absolutely. It makes sense to aggregate those two, and respect wallet address over legacy key when they conflict.

neekolas:
@37ng Feel free to approve your own PR, since I can't approve as the "author".

xmtp_mls/src/identity/xmtp_id/identity.rs
37ng: removed `is_existing_inbox` altogether and refactored the `create_to_be_signed`, `create_from_legacy`. 
Will create a quick PR aggregate both functions so that we have LegacyKey -> Address fallback flow, as you suggested.

37ng (approved):

PR #689, opened 2024-04-29
Validate Credential

### LINKED ISSUE #675: Validate credentials with inbox_id
A method that takes in a credential and installation public key and can evaluate whether it's valid or not.

This would be used when validating new welcome messages, and when reading key packages from the network before adding new members.

insipx:
closes #675 

- [x] validate_inbox_ids
    - validate that a `Memberidentifier` belongs to an `inbox_id`
- [x] validate_inbox_id_key_packages  
    - verify MLS key packages and pull out the MLS Credential (InboxID) and installation public key
- [x] e2e test inbox_id_key_packages 
- [x] e2e test validating whether member identifier belongs to a set of inbox id identities


dev/gen_protos.sh
```diff
@@ -1,13 +1,13 @@
 pushd xmtp_proto > /dev/null
-if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
-    if ! cargo install protoc-gen-prost-crate; then
-        echo "Failed to install protoc-gen-prost-crate"
-        exit 1
-    fi
-fi
+#if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
```
neekolas, line 2: Ya. I guess we can delete this code

xmtp_mls/src/verified_key_package_v2.rs
```diff
@@ -0,0 +1,76 @@
...
+    MlsValidation(#[from] KeyPackageVerifyError),
+    #[error("invalid lifetime")]
+    InvalidLifetime,
+    #[error("wrong credential type")]
+    WrongCredentialType(#[from] BasicCredentialError),
+    #[error(transparent)]
+    Decode(#[from] DecodeError),
+}
+
+pub struct VerifiedKeyPackageV2 {
```
neekolas, line 29: The old `VerifiedKeyPackage` actually checked the signatures in the credential, so you knew you could trust it.

I wonder if we want to either

A) Rename this so that it's clear users still need to check the identity update history for the inbox id to make sure it's valid
B) Add those checks in here and take a reference to the client to access the DB/API

Not a blocker. Just food for thought.

mls_validation_service/src/handlers.rs
```diff
@@ -341,4 +653,126 @@ mod tests {
...
+        assert_eq!(
+            first_response.error_message,
+            "XMTP Key Package failed mls validation: The leaf node signature is not valid."
+        );
+        assert_eq!(first_response.credential, None);
+        assert_eq!(first_response.installation_public_key, Vec::<u8>::new());
+    }
+
+    #[tokio::test]
+    async fn test_validate_inbox_ids_happy_path() {
```
neekolas, line 736: These tests are great

xmtp_id/src/associations/association_log.rs
```diff
@@ -77,6 +77,10 @@ impl IdentityAction for CreateInbox {
 
         let account_address = self.account_address.clone();
         let recovered_signer = self.initial_address_signature.recover_signer().await?;
+        println!(
```
neekolas: We can drop this println

neekolas (approved):

xmtp_mls/src/verified_key_package_v2.rs
insipx: hm,

That's a good catch. I think we should probably go with B), I think A might still lead to users glossing over that. If users _dont_ have a use for a key package that's actually invalid w.r.t InboxId, I think we should definitely include this check, despite added complexity with a Client

37ng (approved):

xmtp_mls/src/verified_key_package_v2.rs
neekolas: Maybe it's a bit of both A and B.

This is kinda a half-verified key package. The MLS checks have passed, but not our Inbox ID authentication. That's still valuable for places like the MLS Validation Service, where you need to parse the key package and then later verify the inbox ID stuff.

So, maybe we have some sort of `ParsedKeyPackage` and `VerifiedKeyPackage` or whatever you want to name it. Then you can move from raw -> parsed -> verified, or stop in the middle.

PR #691, opened 2024-04-30
Message History:  Request and Reply structs

tuddman:
## Summary

- Introduces `HistoryRequest` and `HistoryReply` structs.
- Moves the initialization of the `sync_group` to be self contained in the `message_history` mod.


xmtp_mls/src/groups/message_history.rs
```diff
@@ -130,4 +210,13 @@ mod tests {
 
         let _ = amal_b.send_message_history_request().await;
     }
+
+    #[tokio::test]
+    async fn test_prepare_messages_to_send() {
+        let wallet = generate_local_wallet();
+        let amal_a = ClientBuilder::new_test_client(&wallet).await;
+        // let group_a = amal_a.create_group(None);
+        let messages_result = amal_a.prepare_messages_to_send().await;
```
neekolas, line 219: This test doesn't do a whole lot. Would be good to send a message and make sure that it gets added to the bundle

neekolas (approved):

xmtp_mls/src/groups/message_history.rs
```diff
@@ -43,27 +60,78 @@ where
...
+    fn from(req: HistoryRequest) -> Self {
+        MessageHistoryRequest {
+            pin_code: req.pin_code,
+            request_id: req.request_id,
+        }
+    }
+}
+
+struct HistoryReply {
+    request_id: String,
```
neekolas, line 109: Doesn't this struct need to have some sort of encryption keys in it?

tuddman: Yep.  I was going to put that change in the proto first, and once that's in, make the corresponding addition here, too.

neekolas: 👍

tuddman: https://github.com/xmtp/proto/pull/179

xmtp_mls/src/groups/message_history.rs
```diff
@@ -43,27 +60,78 @@ where
         };
         Ok(())
     }
+
+    #[allow(dead_code)]
+    pub(crate) async fn prepare_messages_to_send(
```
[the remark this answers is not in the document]
neekolas: Maybe `prepare_messages_to_sync`

xmtp_mls/src/groups/message_history.rs
tuddman: The next PR will actually be the roundtrip-ing.. I'll for sure be adding a number of tests around that.  This test is really a stub for that, at this point.

PR #692, opened 2024-04-30
Migrate to new OpenMLS storage API

nplasterer:
Using new storageAPI defined here: https://github.com/openmls/openmls/pull/1565


xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -22,60 +31,951 @@ impl<'a> SqlKeyStore<'a> {
...
-    type Error = StorageError;
-
-    /// Store a value `v` that implements the [`MlsEntity`] trait for
-    /// serialization for ID `k`.
-    ///
-    /// Returns an error if storing fails.
-    fn store<V: MlsEntity>(&self, k: &[u8], v: &V) -> Result<(), Self::Error> {
-        self.conn()
-            .insert_or_update_key_store_entry(k.to_vec(), db_serialize(v)?)?;
+    pub fn write<const VERSION: u16>(
```
franziskuskiefer, line 35: Right, so the easiest transition is to keep the generic read/write/delete functions in some way and use them in all the other functions.
I'm sure you'll want to do something more intelligent later though.

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -22,60 +31,951 @@ impl<'a> SqlKeyStore<'a> {
...
+    ) -> Result<(), Self::Error> {
+        let mut values = self.values.write().unwrap();
+        let key = build_key::<CURRENT_VERSION, &GroupId>(CONFIRMATION_TAG_LABEL, group_id);
+        let value = serde_json::to_vec(&confirmation_tag).unwrap();
+
+        values.insert(key, value);
+        Ok(())
+    }
+
+    fn write_signature_key_pair<
```
franziskuskiefer, line 284: I'm not sure if you're actually using the signature key functions. You may want to skip the unused functions to make sure they're not used.

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -22,60 +31,951 @@ impl<'a> SqlKeyStore<'a> {
...
+        let key =
+            build_key::<CURRENT_VERSION, &SignaturePublicKey>(SIGNATURE_KEY_PAIR_LABEL, public_key);
+
+        let value = values.get(&key).unwrap();
+        let value = serde_json::from_slice(value).unwrap();
+
+        Ok(value)
+    }
+
+    fn write_key_package<
```
franziskuskiefer, line 415: Here is a place where I'm pretty sure you want to do more.
Because you only used last resort key pairs, you need to be able to address them later to delete them.
So you could store the hash references separately in order to have a way to iterate over them and potentially delte them.

franziskuskiefer (commented):
Looks good. I think the main questions are about where you may want to do more in order to be able to handle the store well (see the key package comment for an example).

xmtp_mls/src/storage/sql_key_store.rs
nplasterer: Return not implemented errors.

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -8,12 +11,18 @@ use super::{
...
 use crate::{Delete, Fetch};
 
-/// CRUD Operations for an [`OpenMlsKeyStore`]
-#[derive(Debug)]
 pub struct SqlKeyStore<'a> {
     conn: &'a DbConnection<'a>,
 }
 
+#[derive(QueryableByName)]
+#[table_name = "storage"]
```
neekolas, line 19: Is there a migration where this table actually gets created?

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -22,60 +32,923 @@ impl<'a> SqlKeyStore<'a> {
...
+    }
+
+    pub fn append<const VERSION: u16>(
+        &mut self,
+        label: &[u8],
+        key: &[u8],
+        value: &[u8],
+    ) -> Result<(), <Self as StorageProvider<CURRENT_VERSION>>::Error> {
+        let storage_key = build_key_from_vec::<VERSION>(label, key.to_vec());
+        let query = "INSERT INTO storage (storage_key, version, data) VALUES (?, ?, JSON_ARRAY_APPEND(COALESCE((SELECT data FROM storage WHERE storage_key = ? AND version = ?), '[]'), '$', ?)) ON DUPLICATE KEY UPDATE data = VALUES(data)";
```
neekolas, line 63: Does SQLite actually support `ON DUPLICATE KEY`?

I thought [it didn't](https://stackoverflow.com/questions/23622504/on-duplicate-key-not-working-in-sqlite). But they do have an alternative

I'm also not sure if [SQLite supports `JSON_ARRAY_APPEND`](https://www.sqlite.org/json1.html)

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -22,60 +32,923 @@ impl<'a> SqlKeyStore<'a> {
...
+        self.conn.raw_query(|conn| {
+            sql_query(query)
+                .bind::<diesel::sql_types::Binary, _>(&storage_key)
+                .bind::<diesel::sql_types::Text, _>(&VERSION.to_string())
```
neekolas, line 67: Why not make `version` a number?

xmtp_mls/src/storage/sql_key_store.rs
neekolas: I don't think reading and then writing is the end of the world if there is no alternative to `json_array_append`

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -22,60 +32,923 @@ impl<'a> SqlKeyStore<'a> {
...
+    }
+
+    pub fn remove_item<const VERSION: u16>(
+        &mut self,
+        label: &[u8],
+        key: &[u8],
+        value: &[u8],
+    ) -> Result<(), <Self as StorageProvider<CURRENT_VERSION>>::Error> {
+        let storage_key = build_key_from_vec::<VERSION>(label, key.to_vec());
+        let query = "UPDATE storage SET data = JSON_REMOVE(data, JSON_UNQUOTE(JSON_SEARCH(data, 'one', ?))) WHERE storage_key = ? AND version = ?";
```
neekolas, line 83: I don't think JSON_SEARCH is a thing in sqlite either

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -22,60 +32,923 @@ impl<'a> SqlKeyStore<'a> {
...
+        });
+
+        Ok(())
+    }
+
+    pub fn read<const VERSION: u16, V: Entity<VERSION>>(
+        &self,
+        label: &[u8],
+        key: &[u8],
+    ) -> Result<Option<V>, <Self as StorageProvider<CURRENT_VERSION>>::Error> {
```
neekolas, line 99: It feels a little off that we do deserialization here, using a generic, but the `write` method just takes in `&[u8]`.

If we left serialization out of the base keystore methods it would also unlock us using the much more efficient `tls_codec` where it's available.

xmtp_mls/src/storage/sql_key_store.rs
neekolas: That's an example, right? I'd imagine we are using the `SignatureKeyPair` methods, since that one seems pretty important.

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -1,81 +1,1008 @@
...
+        let key = serde_json::to_vec(group_id)?;
+        self.delete::<CURRENT_VERSION>(OWN_LEAF_NODES_LABEL, &key)
+    }
+
+    fn aad<GroupId: traits::GroupId<CURRENT_VERSION>>(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<Vec<u8>, Self::Error> {
+        let key = serde_json::to_vec(group_id)?;
+        self.read_list(AAD_LABEL, &key)
```
neekolas, line 887: I know this is how it's done in the MemoryStorage, but it's a little weird to me that we are using the single `write` for aad but `read_list`.

Is there meant to be one `aad` or many per key? Maybe it actually all works because everything is a JSON array under the hood.

nplasterer: I agree something doesn't seem to checkout here. 🤔 I think it should just be read not read_list.. @franziskuskiefer do you have thoughts here?

franziskuskiefer: Yeah, this looks like it should really be a `read`. Good catch. I'll check the memory store in openmls.

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -1,87 +1,1052 @@
...
+        self.delete::<CURRENT_VERSION>(EPOCH_KEY_PAIRS_LABEL, &key)
+    }
+
+    fn clear_proposal_queue<GroupId: traits::GroupId<CURRENT_VERSION>>(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<(), Self::Error> {
+        // I think this key is incorrect and should be let key = serde_json::to_vec(&(group_id, &proposal_ref))?;
+        let key = build_key::<CURRENT_VERSION, &GroupId>(QUEUED_PROPOSAL_LABEL, group_id);
+        let _ = self.delete::<CURRENT_VERSION>(QUEUED_PROPOSAL_LABEL, &key);
```
nplasterer, line 843: I think we need the `proposal_ref` here. Was this just over looked? Or is this supposed to be a way to clear all of the proposal queue?

franziskuskiefer: This needs a small change in the API I think. I made a PR on OpenMLS to fix this and also add a test for the proposals. https://github.com/openmls/openmls/pull/1579
wdyt?

nplasterer:
I think this is getting really close. Just have 41 test failures to work through. But they all seem related to data storage of some sort... This PR has merge conflicts any time anything merges into main. Would love to get this merged ASAP.

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -94,12 +1051,161 @@ mod tests {
...
+
+    //     // key_store
+    //     //     .clear_own_leaf_nodes::<GroupId>(&group_id)
+    //     //     .unwrap();
+
+    //     // assert!(key_store
+    //     //     .own_leaf_nodes::<GroupId, LeafNode>(&group_id)
+    //     //     .unwrap()
+    //     //     .is_empty());
+    // }
```
nplasterer, line 1210: @franziskuskiefer I can't seem to get the aad or the proposal ref tests working. What am I missing here? Mind helping me out?

franziskuskiefer (commented):
There are still a bunch of unwraps in here and some cleanup (like pulling out common code) that may be nice to do. But other than that it looks good to me.
Ther serialization is really easy to mess up so maybe it's worth having tests for all functions to ensure a baseline where read/write/delete works.

xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -96,14 +1136,179 @@ mod tests {
...
+        key_store
+            .delete_signature_key_pair::<StorageId>(&public_key)
+            .unwrap();
+
+        assert!(key_store
+            .signature_key_pair::<StorageId, SignatureKeyPair>(&public_key)
+            .unwrap()
+            .is_none());
+    }
+
```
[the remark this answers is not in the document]
nplasterer: @neekolas can you confirm this?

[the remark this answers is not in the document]
neekolas: We don't explicitly access it from your store. We have our own database table the stores each user's signature keys and we access from there.

xmtp_mls/src/groups/sync.rs
```diff
@@ -219,7 +219,8 @@ impl MlsGroup {
                     openmls_group.merge_pending_commit(&provider)
                 {
                     log::error!("error merging commit: {}", err);
-                    openmls_group.clear_pending_commit();
+                    let _ = openmls_group.clear_pending_commit(provider.storage());
```
insipx, line 222: how bad would it be if we error here and don't clear the commit? Previously `clear_pending_commit` wouldn't interact with storage, but now it does a write

xmtp_mls/src/hpke.rs
```diff
@@ -44,19 +47,35 @@ pub fn decrypt_welcome(
...
-        private_key.to_vec().as_slice(),
-        WELCOME_HPKE_LABEL,
-        &[],
-        &ciphertext,
-        CIPHERSUITE,
-        &RustCrypto::default(),
-    )?)
+    let hash_ref: Option<KeyPackageRef> = match provider.storage().read(
+        KEY_PACKAGE_REFERENCES,
+        &hpke_public_key.tls_serialize_detached().unwrap(),
```
insipx, line 54: should remove this unwrap

xmtp_mls/src/identity/v3/legacy.rs
```diff
@@ -236,7 +239,27 @@ impl Identity {
...
             )?;
 
-        Ok(kp)
+        // Store the hash reference, keyed with the public init key.
+        // This is needed to get to the private key when decrypting welcome messages.
+        let public_init_key = kp.key_package().hpke_init_key().tls_serialize_detached()?;
+
+        // Serialize the hash reference
+        let hash_ref =
+            match serde_json::to_vec(&kp.key_package().hash_ref(provider.crypto()).unwrap()) {
```
insipx, line 248: unwrap needs removal

insipx (commented):
So close! Found a couple more unwraps that need to be removed

xmtp_mls/src/groups/sync.rs
insipx: If the write fails somehow we wouldn't know and the pending commit wouldn't be cleared, I think we want to bubble the error with `?` here?

nplasterer: Hmmmm... I'm not sure maybe @richardhuaaa might know?

neekolas (approved):
Heroic effort to get this across the line.

I look forward to banging my head against the wall with merge conflicts next week

nplasterer:
🫡 I will be at your service to help untangle those conflicts.

PR #698, opened 2024-05-01
Create validated commit for MLS

neekolas:
## tl;dr

- Adds the commit validation logic for `xmtp_id` commits
- This is an implementation of the [MLS Commit Validation section of XIP-46](https://community.xmtp.org/t/xip-46-multi-wallet-identity/639#mls-commit-validation-29) and the code should be tested and verified against the design there

## Notable changes from the spec
- We aren't actually using a `UpdateGroupMembership` proposal and instead are just storing the `GroupMembership` as an `UnknownExtension` and the proposal that updates it is a regular `GroupContextExtensions` proposal.

## TODO
- We will need an updated set of `group_permissions` compatible with the new `ValidatedCommit`
- We are going to need some more pieces in place before we can write tests. Without the Identity struct, the group and commits would have to be painfully hand-crafted to work here.

### DIFF
--- a/xmtp_mls/src/groups/validated_commit_v2.rs
+++ b/xmtp_mls/src/groups/validated_commit_v2.rs
@@ -1,0 +1,478 @@
+use std::collections::HashSet;
+
+use openmls::{
+    credentials::{errors::BasicCredentialError, BasicCredential, Credential as OpenMlsCredential},
+    extensions::{Extension, UnknownExtension},
+    group::{GroupContext, MlsGroup as OpenMlsGroup, StagedCommit},
+    messages::proposals::Proposal,
+    prelude::{LeafNodeIndex, Sender},
+    treesync::LeafNode,
+};
+use prost::Message;
+use thiserror::Error;
+#[cfg(doc)]
+use xmtp_id::associations::AssociationState;
+use xmtp_proto::{
+    api_client::{XmtpIdentityClient, XmtpMlsClient},
+    xmtp::identity::MlsCredential,
+};
+
+use crate::{
+    configuration::GROUP_MEMBERSHIP_EXTENSION_ID,
+    identity_updates::{InstallationDiff, InstallationDiffError},
+    storage::db_connection::DbConnection,
+    Client,
+};
+
+use super::{
+    group_membership::{GroupMembership, MembershipDiff},
+    group_metadata::{extract_group_metadata, GroupMetadataError},
+};
+
+#[derive(Debug, Error)]
+pub enum CommitValidationError {
+    #[error("Actor could not be found")]
+    ActorCouldNotBeFound,
+    // Subject of the proposal has an invalid credential
+    #[error("Inbox validation failed for {0}")]
+    InboxValidationFailed(String),
+    // Not used yet, but seems obvious enough to include now
+    #[error("Insufficient permissions")]
+    InsufficientPermissions,
+    // TODO: We will need to relax this once we support external joins
+    #[error("Actor not a member of the group")]
+    ActorNotMember,
+    #[error("Subject not a member of the group")]
+    SubjectDoesNotExist,
+    // Current behaviour is to error out if a Commit includes proposals from multiple actors
+    // TODO: We should relax this once we support self remove
+    #[error("Multiple actors in commit")]
+    MultipleActors,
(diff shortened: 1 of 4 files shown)


neekolas:
* **#698** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/698?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

xmtp_mls/src/groups/validated_commit_v2.rs
```diff
@@ -0,0 +1,458 @@
...
+    pub inbox_id: String,
+    pub is_creator: bool,
+    // TODO: add is_admin support
+    // pub is_admin: bool,
+}
+
+/**
+ * A [`ValidatedCommit`] is a summary of changes coming from a MLS commit, after all of our validation rules have been applied
+ *
+ * Commit Validation Rules:
```
neekolas, line 100: I tried to boil the validations in XIP-46 down to these 5 rules. They don't include the rules around permissions (who is actually allowed to add/remove members or mutate metadata), which will be handled separately.

xmtp_mls/src/groups/validated_commit_v2.rs
```diff
@@ -0,0 +1,465 @@
...
+        client: &'client Client<ApiClient>,
+    ) -> Result<Self, CommitValidationError> {
+        // Get the actor who created the commit.
+        // Because we don't allow for multiple actors in a commit, this will error if two proposals come from different authors.
+        let actor = extract_actor(staged_commit, openmls_group)?;
+
+        // Get the expected diff of installations added and removed based on the difference between the current
+        // group membership and the new group membership.
+        // Also gets back the added and removed inbox ids from the expected diff
+        let (new_group_membership, expected_installation_diff, inboxes_added, inboxes_removed) =
```
insipx, line 131: Does it make sense to create a struct for this? Could still get the individual variables with struct desugaring

xmtp_mls/src/groups/validated_commit_v2.rs
```diff
@@ -0,0 +1,465 @@
...
+    existing_group_context: &GroupContext,
+    new_group_context: &GroupContext,
+) -> Result<(GroupMembership, InstallationDiff, Vec<Inbox>, Vec<Inbox>), CommitValidationError> {
+    let old_group_membership = extract_group_membership(existing_group_context)?;
+    let new_group_membership = extract_group_membership(new_group_context)?;
+    let membership_diff = old_group_membership.diff(&new_group_membership);
+    let added_inboxes = membership_diff
+        .added_inboxes
+        .iter()
+        .map(|inbox_id| Inbox {
```
insipx, line 265: might be worth `impl From<String> for Inbox` and then this becomes `.iter().map(Into::into).collect<Vec<Inbox>>()`

insipx (approved):
Good job! Complicated validation checking but it's well organized and the ways it can fail are clear to me

xmtp_mls/src/groups/validated_commit_v2.rs
```diff
@@ -0,0 +1,465 @@
...
+            extract_expected_diff(
+                conn,
+                client,
+                openmls_group.export_group_context(),
+                staged_commit.group_context(),
+            )
+            .await?;
+
+        // Get the installations actually added and removed in the commit
+        let (added_installations, removed_installations, mut credentials_to_verify) =
```
insipx, line 141: same here, I think maybe adding more structs would clear up some function signatures

37ng (approved):

xmtp_mls/src/groups/validated_commit_v2.rs
neekolas: In the next PR I add the `is_creator` flag, which requires some more data than just the Inbox ID

xmtp_mls/src/groups/validated_commit_v2.rs
neekolas: Ya. You're probably right. I'll change that up

neekolas:
### Merge activity

* **May 1, 5:59 PM PDT**: @neekolas merged this pull request with [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/698).

PR #702, opened 2024-05-02
Message History: Extend sync group to other installations

tuddman:
## Summary

- updates `ConversationType` to include `Sync`
- `create_from_welcome` now handles creating `sync_group`
- adds `new_key()` for generating the "secret input" for encrypting and signing a message bundle;  although after a conversation with @neekolas this will be changed/upgraded to more specifically use an explicit encryption library we're already using.


neekolas (approved):

tuddman:
CI should pass once https://github.com/xmtp/libxmtp/pull/703 is in.

PR #703, opened 2024-05-02
fix: allow get_then_check and other fixes other lint

tuddman:
## Summary

with `1.78.0` [released today](https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html), our CI pulls `latest` and runs against the new lints, which include this one:

https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_get_then_check

which fails to compile if you accept the lint change of using a unary `!` on an `Option<T>` instead of `is_none()` which always worked exactly as it reads.

 #rust-aint-perfect

### DIFF
--- a/xmtp_mls/src/groups/group_membership.rs
+++ b/xmtp_mls/src/groups/group_membership.rs
@@ -26,7 +26,6 @@ impl GroupMembership {
         self.members.get(inbox_id.as_ref())
     }
 
-    #[allow(clippy::unnecessary_get_then_check)]
     pub fn diff<'inbox_id>(
         &'inbox_id self,
         new_group_membership: &'inbox_id Self,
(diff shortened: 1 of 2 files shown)


xmtp_mls/src/groups/group_membership.rs
```diff
@@ -26,6 +26,7 @@ impl GroupMembership {
         self.members.get(inbox_id.as_ref())
     }
 
+    #[allow(clippy::unnecessary_get_then_check)]
     pub fn diff<'inbox_id>(
         &'inbox_id self,
         new_group_membership: &'inbox_id Self,
```
neekolas, line 32: <i class='graphite__hidden'>[Re: line 54]<br/><br/></i>

You mind just switching this to `contains_key`? Then we don't need the clippy allow<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/703?utm_source=unchanged-line-comment">Graphite</a>.</span>

neekolas (approved):

xmtp_mls/src/groups/group_membership.rs
tuddman: Switched 👍

PR #699, opened 2024-05-02
Scaffolding to start add/remove members using GroupMembership

neekolas:
## tl;dr

- Adds a new intent data type `UpdateGroupMembership`. This will eventually replace both `AddMembers` and `RemoveMembers` intents
- Adds the `creator_inbox_id` to the Immutable Metadata for a group so that we can use it in permissions
- Attaches the `GroupMembership` extension to all new groups

PR #708, opened 2024-05-03
Message History: Better key generation method

tuddman:
## Summary

Updates key generation definition to match what was changed in  https://github.com/xmtp/proto/pull/182


xmtp_mls/src/groups/group_metadata.rs
```diff
@@ -67,6 +67,7 @@ impl GroupMetadata {
     pub(crate) fn to_proto(&self) -> Result<GroupMetadataProto, GroupMetadataError> {
         let conversation_type: ConversationTypeProto = self.conversation_type.clone().into();
         Ok(GroupMetadataProto {
+            creator_inbox_id: "".to_string(),
```
tuddman, line 70: temporary placeholder, as https://github.com/xmtp/proto/commit/1ea946c9ef79197b32e7bc88696e6423ecf3b7b6  introduced `creator_inbox_id`

neekolas: My apologies. [This PR](https://github.com/xmtp/libxmtp/pull/699) is just waiting for reviews

neekolas: Let's merge this one first and I'll deal with the conflicts.

neekolas (approved):

xmtp_mls/src/groups/message_history.rs
```diff
@@ -118,50 +117,64 @@ impl HistoryReply {
...
-    let mut key = [0u8; 32];
-    let rng = rand::thread_rng();
-    rng.sample_iter(&Alphanumeric)
-        .take(32)
-        .enumerate()
-        .for_each(|(i, c)| key[i] = c);
-    key
+impl HistoryKeyType {
+    #[allow(dead_code)]
+    fn new_chacha20_poly1305_key() -> Self {
```
neekolas, line 152: In hindsight I should have suggested `chacha20_poly1305_key_material` since it's not the key itself but the inputs, but whatever. Renaming proto fields is non-breaking, and if it's really a bother we can always change it later.

tuddman: Agree. `chacha20_poly1305_key_material` is more accurate to what this field is for. 
I suspect there may be at least one additional change to the proto before those whole feature set is finished, so I'll pick it up in the next change...

PR #709, opened 2024-05-03
Patch openssl-sys crate to v0.9.92

zombieobject:
A regression in openssl-sys exists where libatomic is dynamically linked 
for i686-linux-android targets. https://github.com/sfackler/rust-openssl/issues/2163

The result is dynamically linked symbols are included in the binary header link table:

```
         U __atomic_fetch_or_8
         U __atomic_is_lock_free
         U __atomic_load
```

The presence of these symbols breaks Android CI tests for i686 targets with the following error:

```
java.lang.UnsatisfiedLinkError: Unable to load library 'uniffi_xmtpv3':
	dlopen failed: cannot locate symbol "__atomic_is_lock_free" referenced by "/data/app/org.xmtp.android.library.test-Cq_BL0nCyvMEKUk7x28gzw==/base.apk!/lib/x86/libuniffi_xmtpv3.so"...
```

This is fixed in the openssl-sys fork at https://github.com/xmtp/rust-openssl on the branch clone-v0.9.92, which is pinned
to that version. Version 0.9.92 statically links libatomic for i686 targets. Once this is addressed upstream we can remove the patch.

A patch is required as openssl-sys is a dependency crate. A fork is required  because it is not possible to patch to an older version of the same source.

### DIFF
--- a/bindings_ffi/Cargo.toml
+++ b/bindings_ffi/Cargo.toml
@@ -20,6 +20,15 @@ xmtp_proto = { path = "../xmtp_proto", features = ["proto_full", "grpc"] }
 xmtp_user_preferences = { path = "../xmtp_user_preferences" }
 xmtp_v2 = { path = "../xmtp_v2" }
 
+# NOTE: A regression in openssl-sys exists where libatomic is dynamically linked 
+# for i686-linux-android targets. https://github.com/sfackler/rust-openssl/issues/2163
+#
+# This is fixed in the openssl-sys fork at
+# https://github.com/xmtp/rust-openssl on the branch clone-v0.9.92, which is pinned
+# to that version. Once this is addressed upstream we can remove the patch.
+[patch.crates-io]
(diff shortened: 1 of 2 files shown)


bindings_ffi/Cargo.toml
```diff
@@ -20,6 +20,15 @@ xmtp_proto = { path = "../xmtp_proto", features = ["proto_full", "grpc"] }
 xmtp_user_preferences = { path = "../xmtp_user_preferences" }
 xmtp_v2 = { path = "../xmtp_v2" }
 
+# NOTE: A regression in openssl-sys exists where libatomic is dynamically linked 
+# for i686-linux-android targets. https://github.com/sfackler/rust-openssl/issues/2163
+#
+# This is fixed in the openssl-sys fork at
+# https://github.com/xmtp/rust-openssl on the branch clone-v0.9.92, which is pinned
+# to that version. Once this is addressed upstream we can remove the patch.
```
nplasterer, line 28: Do you have a PR to fix this upstream by any chance we could link here?

nplasterer (approved):

PR #713, opened 2024-05-07
remove option from map

37ng:
Use `V` instead of `Option<V>` because HashMap has built-in  Option<V> when returning a value.


37ng:
* **#713** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/713?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

neekolas (approved):
I think there was a reason I did it this way, but maybe it's not relevant

37ng:
> I think there was a reason I did it this way, but maybe it's not relevant

Might be that `get_inbox_ids` returns Options and directly putting Options in the HashMap will add an extra layer of Option.

PR #720, opened 2024-05-07
feat: adds pin check and MessageHistoryError

tuddman:
## Summary

- adds an initial PIN checking mechanism + test.
- adds a `thiserror` style `MessageHistoryError` to handle message history specific errors that may occur.

### DIFF
--- a/xmtp_mls/src/groups/message_history.rs
+++ b/xmtp_mls/src/groups/message_history.rs
@@ -3,6 +3,7 @@ use rand::{
     distributions::{Alphanumeric, DistString},
     Rng, RngCore,
 };
+use thiserror::Error;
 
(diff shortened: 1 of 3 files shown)


xmtp_mls/src/groups/message_history.rs
```diff
@@ -94,6 +105,41 @@ where
...
+        let requests = conn.get_group_messages(
+            sync_group_id,
+            None,
+            None,
+            Some(GroupMessageKind::Application),
+            None,
+            None,
+        )?;
+        let request = requests.into_iter().find(|msg| {
+            let msg_bytes = &msg.decrypted_message_bytes;
```
neekolas, line 126: Neat hack to avoid decoding the proto

neekolas (approved):

PR #721, opened 2024-05-08
aggregate Identity creation

37ng:
# Summary

aggregated Identity creation logic to make platform SDK's job easier. i.e. merged `create_from_legacy` and `create_to_be_signed` into one function.

Instead of requiring platform SDK to choose what it wants, we just let them pass in `address` and `Option<legacy_key>` and take care of the rest.

### DIFF
--- a/xmtp_mls/src/identity/xmtp_id/identity.rs
+++ b/xmtp_mls/src/identity/xmtp_id/identity.rs
@@ -1,7 +1,6 @@
 use std::array::TryFromSliceError;
 
 use ed25519_dalek::SigningKey;
-use ethers::signers::{LocalWallet, WalletError};
 use openmls::{
     credentials::{errors::BasicCredentialError, BasicCredential},
     prelude::Credential as OpenMlsCredential,
@@ -32,7 +31,6 @@ use xmtp_v2::k256_helper;
 use crate::{
     api::{ApiClientWrapper, WrappedApiError},
     configuration::CIPHERSUITE,
-    InboxOwner,
 };
 
 #[derive(Debug, Error)]
@@ -49,8 +47,6 @@ pub enum IdentityError {
     BasicCredential(#[from] BasicCredentialError),
     #[error("Legacy key re-use")]
     LegacyKeyReuse,
-    #[error("legacy key does not match address")]
-    LegacyKeyMismatch,
     #[error("Installation key {0}")]
     InstallationKey(String),
     #[error("Malformed legacy key: {0}")]
@@ -59,8 +55,6 @@ pub enum IdentityError {
     LegacySignature(String),
     #[error(transparent)]
     Crypto(#[from] CryptoError),
-    #[error(transparent)]
-    WalletError(#[from] WalletError),
 }
 
 #[derive(Debug, Clone)]
@@ -102,12 +96,6 @@ impl Identity {
                 return Err(IdentityError::LegacyKeyReuse);
             }
 
-            // sanity check if address matches the one derived from legacy_signed_private_key
-            let legacy_key_address = legacy_key_to_address(legacy_signed_private_key.clone())?;
-            if address != legacy_key_address {
-                return Err(IdentityError::LegacyKeyMismatch);
-            }
-
             // 2. has legacy key, no inbox -> create an inbox and sign the installation key with the legacy key.
             let nonce = 0;
             let inbox_id = generate_inbox_id(&address, &nonce);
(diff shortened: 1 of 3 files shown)


37ng:
* **#721** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/721?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

xmtp_mls/src/identity/xmtp_id/identity.rs
```diff
@@ -71,112 +71,129 @@ impl Identity {
...
+        let member_identifier: MemberIdentifier = address.clone().into();
 
-        let mut signature_request = builder
-            .add_association(installation_public_key.to_vec().into(), member_identifier)
-            .build();
+        // if an inbox is not associated, we will create a new one
+        // and sign the installation key using legacy key or wallet.
+        if inbox_id.is_none() {
+            if let Some(legacy_signed_private_key) = legacy_signed_private_key {
+                // TODO: check if address matches legacy_signed_private_key
```
37ng, line 98: is this TODO needed at all? Can we assume that `address` and `legacy_signed_private_key` passed in from platform SDK always match?

neekolas: They should always match in practice, but let's do the check just in case

xmtp_mls/src/identity/xmtp_id/identity.rs
```diff
@@ -71,112 +71,129 @@ impl Identity {
...
+            }
+        } else {
+            // If an inbox is associated, we just need to associate the installation key
+            // Only wallet is allowed to sign the installation key, legacy key is not
+            let nonce = rand::random::<u64>();
+            let inbox_id = generate_inbox_id(&address, &nonce);
+            let builder = SignatureRequestBuilder::new(inbox_id.clone());
+            let mut signature_request = builder
+                .add_association(installation_public_key.to_vec().into(), member_identifier)
+                .build();
```
neekolas, line 175: I'm not sure I understand this bit of code.

If there is no existing `inbox_id` and we are randomly generating the nonce, we would need to create the inbox with the random nonce. Can't just do an `add_association` to the inbox that hasn't been created yet.

37ng: this `else {}` part is for existing `inbox_id`.

I also find this code structure confusing, & tedious, will make this DRYer.

37ng:
- Added the logic to assert that `address` matches `legacy_signed_private_key`.
- Less LoC and more concise for `Identity::new()`

37ng:
Think I made a mistake that removed fallback flow... Will mark this as ready for review once it's fixed.

xmtp_mls/src/identity/xmtp_id/identity.rs
```diff
@@ -71,112 +77,127 @@ impl Identity {
...
-        let identity_update = signature_request.build_identity_update()?;
+            let mut signature_request = builder
+                .add_association(installation_public_key.to_vec().into(), member_identifier)
+                .build();
 
-        api_client.publish_identity_update(identity_update).await?;
+            // We can pre-sign the request with an installation key signature, since we have access to the key
+            signature_request
+                .add_signature(Box::new(
+                    sign_with_installation_key(
```
insipx, line 184: It's a bit confusing signing with installation key in libxmtp today. OpenMls offers its own signing procedure with BasicCredential that is currently used in v3 (where any additional signing complexity like a prehash is abstracted), in identity we are using our own signing procedure with dalek, and then again ethers offers another signing procedure similar to dalek but more functions to call.

I'm wondering if it's worth abstracting these signing methods into a central place, with conversions between differnt types. Like an `InstallationKey` with just a `sign` method. I could convert a [u8; 32] to this, a INstallatonKeyProto, and a BasicCredential to this one struct. This could make signing code more concise and easier to follow, as well as potentially reduce any errors we get by screwing up the prehash stuff.

I think a standalone function is OK for now, but i worry about the discoverability of it, esp as it's tucked away deep in an identity module

xmtp_mls/src/identity/xmtp_id/identity.rs
```diff
@@ -71,112 +77,127 @@ impl Identity {
...
-            inbox_id: inbox_id.clone(),
-            installation_keys: signature_keys,
-            credential: create_credential(inbox_id)?,
-            signature_request: None,
-        };
+            let identity = Self {
+                inbox_id: inbox_id.clone(),
+                installation_keys: signature_keys,
+                credential: create_credential(inbox_id.clone())?,
+                signature_request: Some(signature_request),
```
insipx, line 196: Can we get rid of the Option<SignatureRequest>?

Or, when would we have a None here

Also, when this signature request is resolved will we:

1. Update this identity struct with the `Resolved` signature 
2. resolve to a new Identity type like `SignedIdentity` or otherwise


Not sure which one i would prefer. with 1. we would have to probably have todeal with checking `signed.is_some()` if an `Identity` is signed, but with `SignedIdentity` we would have the added complexity of another struct

Maybe there's a better design too?

xmtp_mls/src/identity/xmtp_id/identity.rs
37ng: I totally agree, we could have a central place to handle cryptography / primitives conversion / utils so that other parts of code won't need to deal with the complexity.
It would be a fairly big clean up task, hope we get to prioritize that soon.

xmtp_mls/src/identity/xmtp_id/identity.rs
37ng: > Or, when would we have a None here
After SDK created an `Identity` instance, it might ask user wallet to sign the signature_request. After signing, this will be None. It's similar to 1.

neekolas (approved):

xmtp_mls/src/identity/xmtp_id/identity.rs
neekolas: ➕

xmtp_mls/src/identity/xmtp_id/identity.rs
neekolas: Also in the case that you are using the SDK for the second+ time (have already done the signing) the `signature_request` will always be None

Issue #664, opened 2024-04-16
Mutable admin permissions 


nplasterer:
## Task Tracking:

1. [x] Protos update - PR: https://github.com/xmtp/proto/pull/175
2. [x] libxmtp updates - PR: https://github.com/xmtp/libxmtp/pull/724
   1. [x] Move Permissions from Immutable Metadata to GroupMutablePermissions Extension
   2. [x] Add new admin list and super admin list to Mutable Metadata Extension
   3. [x] Add new default permissions set "Admins only"
   4. [x] Add functions for reading Admin List, Super Admin List, and each permission policy
   5. [x] Add validation functions that correctly evaluate admin/super admin policies
   6. [x] Add functions for add/remove admin+ super admin
   7. [x] Add function for updating permissions
   8. [x] add tests showing admin only permissions applied to add/remove group member, update metadata
   9. [x] Add tests showing superadmin permissions for add/remove admin. 
   10. [x] Add a test showing how to add a new Permission Policy to an existing group
   11. [ ] BONUS - Add a test showing how to upgrade to a new Permissions Extension
3. [ ] SDK Updates
   1. [x]  Android
       1. [x] Functions for reading admin list, super admin, permission policies
       2. [x] add/remove admin function and super_admin function
       3. [x] update permission function
       4. [x] Add tests
   5. [x] iOS
       1. [x] Functions for reading admin list, super admin, permission policies
       2. [x] add/remove admin function and super_admin function
       3. [x] update permission function
       4. [x] Add tests
   7. [ ] React Native
       1. [x] Functions for reading admin list, super admin, permission policies
       2. [x] add/remove admin function and super_admin function
       3. [x] update permission function
       4. [x] Add tests
       5. [ ] Example UI

## Background on Groups Permissions
Currently in groups beta we have the following **Permissions Policies**:

1. add member policy
2. remove member policy
3. update metadata policy

Each of those policies can have one of the following **Permission Options**:

A. unspecified
B. allow
C. deny
D. creator only

## Mutable Admin Permissions Update Specification
After the Mutable admin permissions change we will have the following  new **Permission Policies**:

4 add admin policy
5. remove admin policy
6. update permissions policy

We will also add the following new **Permission Options**:

E. allow if admin or super admin
F. allow if super admin

To enable the two new permission options the following will be added to Mutable medata proto:

```
message GroupMutableMetadataV1 {
...
  AccountAddresses admin_list = 2;
  AccountAddresses super_admin_list = 3;
}

[...]

PR #724, opened 2024-05-08
Added new Permissions Extension

cameronvoell:
Part of #664 

Related proto PR: https://github.com/xmtp/proto/pull/175

Breaking up Permission Updates into multiple PRs. 

This PR incorporates Proto updates from  https://github.com/xmtp/proto/pull/175 which add a new Permissions Extension, and moves existing permissions from immutable data to this new unknown extension. We also update the Permission Preconfigured Policies from `EveryoneIsAdmin` and `GroupCreatorOnly` to `AllMembers` and `AdminsOnly`. This update is currently invisible to the SDKs. Since the Group Creator is the only admin and super admin starting out, nothing changes for the SDKs until we add the methods for updating the admin list in the follow up PR, which will be linked here. 

Before Merging:
- [ ] Update gen_protos.sh back to pointing to `main` branch


xmtp_mls/src/groups/group_mutable_metadata.rs
```diff
@@ -110,7 +134,27 @@ impl TryFrom<GroupMutableMetadataProto> for GroupMutableMetadata {
...
+            Some(inboxes) => {
+                super_admin_list = inboxes.inbox_ids;
+            }
+            None => return Err(GroupMutableMetadataError::MissingMetadataField),
+        }
+        Ok(Self::new(
+            value.attributes.clone(),
+            admin_list,
+            super_admin_list,
+        ))
```
37ng, line 157: Can we do this?
```suggestion
        Ok(Self::new(
            value.attributes.clone(),
            value.admin_list.ok_or(Err(GroupMutableMetadataError::MissingMetadataField))?,
            value. super_admin_list.ok_or(Err(GroupMutableMetadataError::MissingMetadataField)),
        ))
```

cameronvoell: Thanks @37ng  - that pattern was eluding me for some reason

cameronvoell: got a more concise version working in a commit here: e08e027ff89e35ba71dee10c8a1e11d25e8f3e07

dev/gen_protos.sh
```diff
@@ -6,7 +6,7 @@ if ! cargo install --list | grep "protoc-gen-prost-crate" > /dev/null; then
     fi
 fi
 
-if ! buf generate https://github.com/xmtp/proto.git#branch=main,subdir=proto; then
+if ! buf generate https://github.com/xmtp/proto.git#branch=cv/policy-set-mutable,subdir=proto; then
```
neekolas, line 9: You'll need to point this back to main before merging

neekolas (approved):

neekolas (commented):
I'm going to put an hour on your calendar for Wednesday and we are going to try and verify that I am not butchering this work when I merge it into the identity stuff

PR #727, opened 2024-05-09
Refactor + Simplify Lifetime mgmt for SqlKeyStore & MlsGroup 

insipx:
This is a fairly large refactor PR, but it should significantly improve the ergonomics around MlsGroup management & Client, and we should run into significantly fewer lifetime errors. Additinoally, this aims to simplify binding creation.

- Introduce `XmtpApi` supertrait
- Decouple `Client` from `MLSGroup`.`MlsGroup` fns now accept a `Client` argument
- remove generics and lifetimes from `MlsGroup`
- Remove generics and lifetimes from `DbConnection`
- use `impl OpenMlsProvider` trait more often
- remove lots and lots of lifetimes
- change `transaction` to use diesel `AnsiTransactionManager` and remove DbConnection mutable reference. All transactions require owned connection to start. DbConnection no longer holds a Mutex<&mut Connection> but a Mutex<Connection>, allowing us to remove lifetimes everywhere.

notes:

- there are a few methods that really need the client. Most methods accept a client and pass i through. is there a way where we can minimize passing around of a client?
- at the end of some codepaths, we don't make use of client but instead of `ApiClientWrapper` using `client.api_client`. Can we pass api_client_wrapper instead?
- MlsGroup can be thought of as an abstraction on Client, yet Client <-> MlsGroup have a circular relationship where MlsGroup is created from Client, and the Mls Group Uses Client:

Client - creates -> MlsGroup -> uses -> Client

## Next Steps

We can choose to take this refactor further and create a stateless client that can be plugged into any MlsGroup. Currently we need to be careful to use the right client in tests, i.e:

```rust
let client = ClientBuilder::new_test_client(&generate_local_wallet()).await;
        let client_2 = ClientBuilder::new_test_client(&generate_local_wallet()).await;
        let group = client.create_group(None).expect("create group");

        group
            .add_members_by_installation_id(vec![client_2.installation_public_key()], &client)
            .await
            .unwrap();

```

This is because `Client` has context attached to it like identity etc.

A good next step would be taking a closer look at the `Client` methods MLS group actually needs. I would suspect that we would be able to further decouple the methods which require a context like identity vs just methods like `find_groups` which acts totally on context/arguments passed into it. Although this would take some time to accomplish correctly and the structure as it is is completely ok for now

We could introduce another struct `Xmtp` that doles out MlsGroups withs Clients to create a nice high-level API without the Client <-> MlsGroup relationship that currently exists. 

The bindings kindof already do this (the part where there's another struct exposing a higher-level api), and we can use that client wherever we need to allowing the bindings api to remain unchanged


xmtp_mls/src/lib.rs
```diff
@@ -21,6 +21,12 @@ mod xmtp_openmls_provider;
 pub use client::{Client, Network};
 use storage::StorageError;
 use xmtp_cryptography::signature::{RecoverableSignature, SignatureError};
+use xmtp_proto::api_client::{XmtpIdentityClient, XmtpMlsClient};
+
+/// XMTP Api Super Trait
+/// Implements all Trait Network APIs for convenience.
+pub trait XmtpApi: XmtpMlsClient + XmtpIdentityClient {}
+impl<T> XmtpApi for T where T: XmtpMlsClient + XmtpIdentityClient {}
```
neekolas, line 29: Great call

xmtp_mls/src/xmtp_openmls_provider.rs
```diff
@@ -4,25 +4,68 @@ use openmls_traits::OpenMlsProvider;
 use crate::storage::{db_connection::DbConnection, sql_key_store::SqlKeyStore};
 
 #[derive(Debug)]
-pub struct XmtpOpenMlsProvider<'a> {
+pub struct XmtpOpenMlsProvider {
+    crypto: RustCrypto,
+    key_store: SqlKeyStore<'static>,
+}
+
+pub struct XmtpOpenMlsProviderRef<'a> {
```
richardhuaaa, line 12: Out of curiosity, why do we need this separate type?

insipx: I'm still working through the compile errors, so I might do away with this in the end. But, it 's useful when we want to work with `OpenMlsProvider` and only have a `&DbConnection` (maybe we're deep in a codepath).

insipx: Ultimately removed it

bindings_ffi/jniLibs/arm64-v8a/libuniffi_xmtpv3.so
neekolas, line 1: I think we'll want to rebuild these and commit them. Pretty sure we initially allowed them in the repo so that it would speed up compiles for the bindings.

.vscode/settings.json
```diff
@@ -1,9 +1,25 @@
...
+    },
+    "linkedProjects": [
+      "bindings_ffi/Cargo.toml",
+      "examples/cli/Cargo.toml",
+      "xmtp_api_grpc_gateway/Cargo.toml"
+    ],
+    "procMacro" : {
+      "enable": true,
+      "attributes.enable": true,
+      "ignored": {
```
neekolas, line 14: Curious what issue this fixes

insipx: Can remove, but for me with rust-analyzer, I got alot of 'unresolved-proc-macro`, for well-known macros like async-trait and others. This tells rust-analyzer to ignore them and avoid polluting the error-namespace. I think @37ng also got similar errors iirc but may be conflating it with different lsp errors.

xmtp_id/src/associations/mod.rs
```diff
@@ -423,7 +420,6 @@ mod tests {
                 SignatureKind::InstallationKey,
                 None,
             ),
-            ..Default::default()
```
neekolas, line 426: Good find. I swear, at some point these defaults were necessary

bindings_ffi/jniLibs/arm64-v8a/libuniffi_xmtpv3.so
insipx: ah good catch, I tried rebuilding these but it errored, didn't realize that it deleted the libs and _then_ errored

bindings_ffi/src/mls.rs
```diff
@@ -713,32 +727,32 @@ mod tests {
...
     struct RustStreamCallback {
-        num_messages: Arc<Mutex<u32>>,
+        num_messages: Arc<AtomicU32>,
     }
 
     impl RustStreamCallback {
         pub fn new() -> Self {
             Self {
-                num_messages: Arc::new(Mutex::new(0)),
+                num_messages: Arc::new(AtomicU32::new(0)),
```
insipx, line 736: I switched out the Mutex here b/c `AtomicU32` makes more sense IMO but also because the `stream_all_messages` test kept erroring and I was troubleshooting the problem. I think it turned out the test was not waiting long enough for the message to be received, so increasing the timeout fixed it. Although maybe @richardhuaaa knows more in that area.

it might be good to try and use `std::sync::Barrier` here or the async variant on crates.io `async-barrier` instead of tokio::sleep, but might be tricky with the callback + async

bindings_ffi/src/mls.rs
```diff
@@ -335,24 +337,25 @@ pub struct FfiListMessagesOptions {
 impl FfiGroup {
     pub async fn send(&self, content_bytes: Vec<u8>) -> Result<Vec<u8>, GenericError> {
         let group = MlsGroup::new(
```
neekolas, line 339: With these changes, we can't just store an `Arc<MlsGroup>` on the `FfiGroup` and remove this silliness around having to re-create the group on each method call?

That was purely there because of lifetime issues with Uniffi

xmtp_mls/src/lib.rs
neekolas, line 29: This is just objectively nicer than before

xmtp_proto/src/api_client.rs
```diff
@@ -143,6 +143,7 @@ pub type WelcomeMessageStream = Pin<Box<dyn Stream<Item = Result<WelcomeMessage,
 // Wasm futures don't have `Send` or `Sync` bounds.
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+// TODO INSIPX: do we need this 'static bond?
 pub trait XmtpMlsClient: Send + Sync + 'static {
```
neekolas, line 147: I believe the static bound came in for streaming, but maybe we don't need it anymore.

bindings_ffi/src/mls.rs
insipx: Good call, that's definitely worth doing now that we don't have to deal with the lifetimes

37ng (approved):
great to see that we have Client decoupled from MlsGroup. LGTM given compiler is happy with those one liner changes resulting from the decoupling, and all tests pass.

neekolas (approved):

.vscode/settings.json
neekolas: If it's helping your Rust analyzer and doesn't have any adverse side effects, happy to leave it as-is.

bindings_ffi/src/mls.rs
neekolas: Can be a separate PR

xmtp_proto/src/api_client.rs
insipx: confirmed we actually still need it

richardhuaaa (approved):
This is fantastic! So many improvements compressed into one PR. I like the refactoring of the transaction management to remove transaction lifetimes, and the `XmtpApi` trait. Definitely much better than it was before.

The `Context` added here pulls out identity and DB from `Client` so that generics can be avoided on `MlsGroup`, but it means that we still need to pass `Client` as an argument everywhere that networking is used. I'm wondering if we could have kept networking/storage/identity together - by passing `Client` as an argument everywhere without the new `Context` type, or adding the ApiClient onto `Context` and passing that everywhere. No immediate action needed though, this is definitely still an improvement.

Also left some nits, but there's nothing major there - if changes are not trivial, feel free to land as I know others are blocked.

xmtp_mls/src/xmtp_openmls_provider.rs
```diff
@@ -4,28 +4,59 @@ use openmls_traits::OpenMlsProvider;
 use crate::storage::{db_connection::DbConnection, sql_key_store::SqlKeyStore};
 
 #[derive(Debug)]
-pub struct XmtpOpenMlsProvider<'a> {
+pub struct XmtpOpenMlsProvider {
     crypto: RustCrypto,
-    key_store: SqlKeyStore<'a>,
+    key_store: SqlKeyStore<'static>,
```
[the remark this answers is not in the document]
insipx: good catch, it can & easy change to make

xmtp_mls/src/xmtp_openmls_provider.rs
```diff
@@ -4,28 +4,59 @@ use openmls_traits::OpenMlsProvider;
...
+    pub(crate) fn conn(&self) -> DbConnection {
         self.key_store.conn()
     }
+
+    pub(crate) fn conn_ref(&self) -> &DbConnection {
+        self.key_store.conn_ref()
+    }
+}
+
+impl OpenMlsProvider for XmtpOpenMlsProvider {
```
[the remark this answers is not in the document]
insipx: It does make it easier since the reference and owned version can be used interchangeably, owned & ref variants do need their own impls generally since they can differ in small ways

bindings_ffi/src/mls.rs
```diff
@@ -214,7 +214,9 @@ impl FfiConversations {
 
         let convo = self.inner_client.create_group(group_permissions)?;
         if !account_addresses.is_empty() {
-            convo.add_members(account_addresses).await?;
+            convo
```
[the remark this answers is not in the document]
neekolas: I agree. Let's do that as a separate follow-up after all the identity stuff goes in to main

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -165,15 +165,31 @@ impl EncryptedMessageStore {
...
+                    PoolTransactionManager::<AnsiTransactionManager>::commit_transaction(&mut *conn)
+                })?;
+                Ok(value)
+            }
+            Err(err) => match conn.raw_query(|conn| {
+                PoolTransactionManager::<AnsiTransactionManager>::rollback_transaction(&mut *conn)
+            }) {
+                Ok(()) => Err(err),
+                Err(Error::BrokenTransactionManager) => Err(err),
+                Err(rollback) => Err(rollback.into()),
```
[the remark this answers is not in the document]
insipx: `AnsiTransactionManager` will try to salvage the rollback before it makes it here, the behavior would be the same as in the normal `transaction` fn on SqliteConnection. We would get a [`RollbackOnCommit`] or another error. Depending on the error the connection may be completely broken or the transaction manager marked as broken broken

[Errors](https://docs.rs/diesel/latest/diesel/result/enum.Error.html) 

[Description of what happens](https://docs.rs/diesel/latest/diesel/connection/trait.Connection.html#method.transaction)

can add some comments in a follow-up

xmtp_mls/src/client.rs
```diff
@@ -165,73 +162,122 @@ impl From<&str> for ClientError {
...
-    ApiClient: XmtpMlsClient + XmtpIdentityClient,
+    ApiClient: XmtpApi,
 {
     /// Create a new client with the given network, identity, and store.
     /// It is expected that most users will use the [`ClientBuilder`](crate::builder::ClientBuilder) instead of instantiating
     /// a client directly.
     pub fn new(
         api_client: ApiClientWrapper<ApiClient>,
-        network: Network,
+        _network: Network,
```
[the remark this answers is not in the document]
insipx: fixed, for some reason i felt like this was more important than it was

PR #728, opened 2024-05-09
[identity-release] new identity in client

37ng:
# Summary

1. Replaced the old `Identity` struct, primarily using `inbox_id` instead of `account_address`.
2. Ignored all failing tests, ~50.


xmtp_mls/src/groups/mod.rs
```diff
@@ -214,7 +215,7 @@ where
         client: &'c Client<ApiClient>,
         membership_state: GroupMembershipState,
         permissions: Option<PreconfiguredPolicies>,
-        added_by_address: String,
+        added_by_address: &InboxId,
```
neekolas, line 218: Do you mind putting some TODOs on these cases where we are using inbox IDs as addresses, just so we can make sure to get to them all

37ng: For sure, should've done that in the first place

xmtp_mls/src/utils/id.rs
```diff
@@ -8,13 +10,13 @@ pub fn serialize_group_id(group_id: &[u8]) -> String {
 pub fn calculate_message_id(
     group_id: &[u8],
     decrypted_message_bytes: &[u8],
-    sender_account_address: &str,
+    inbox_id: InboxId,
```
neekolas, line 13: In another PR I actually just remove the sender identity from the calculation altogether. I don't think we need it for uniqueness, since we have the `idempotency_key`

xmtp_mls/src/groups/mod.rs
37ng: renamed this one to `inbox_id`, and commented a few `TODO` elsewhere

37ng:
I think CI should all pass except `test_is_smart_wallet`. The test is calling `https://eth.llamarpc.com` to get code size of a given address. I guess the website is down or rate limiting.

xmtp_mls/src/identity/mod.rs
```diff
@@ -1,2 +0,0 @@
-pub mod v3;
-pub mod xmtp_id;
```
37ng, line 2: remove multiple layers of this identity mod including legacy Identity and use a single file mod instead.

neekolas (approved):

PR #730, opened 2024-05-10
feat: stubs wasm bindings for nodejs

tuddman:
## Summary

Stubs in some skeleton code to build, test and run the WASM bindings from Javascript.


richardhuaaa (approved):

bindings_wasm/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+// use std::collectioashMap;
```
richardhuaaa, line 1: Maybe just remove this?

tuddman: gone. It may come back on Monday - actually implemented.

PR #733, opened 2024-05-14
Cache association state

richardhuaaa:
This caches the association state for a given inbox_id and sequence_id, allowing us to skip recomputing it/issuing network requests for smart wallet associations.


xmtp_mls/migrations/2024-05-11-004236_cache_association_state/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE association_state (
+    "inbox_id" TEXT NOT NULL,
+    "sequence_id" BIGINT NOT NULL,
+    "state" BLOB NOT NULL,
```
richardhuaaa, line 4: In the future we could consider removing unused cached state. I'm not sure if it's safe to assume that only the latest sequence_id needs to be cached. We could consider adding the client time here, and expiring old entries

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -216,12 +216,13 @@ macro_rules! impl_fetch {
             type Key = $key;
             fn fetch(&self, key: &Self::Key) -> Result<Option<$model>, $crate::StorageError> {
                 use $crate::storage::encrypted_store::schema::$table::dsl::*;
-                Ok(self.raw_query(|conn| $table.find(key).first(conn).optional())?)
+                Ok(self.raw_query(|conn| $table.find(key.clone()).first(conn).optional())?)
```
richardhuaaa, line 219: The `clone` is needed to support fetching by composite primary key. In this case, `find()` expects a tuple `(key1, key2)`, not a reference `&(key1, key2)` and will return a type error otherwise

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -243,6 +244,29 @@ macro_rules! impl_store {
...
+            fn store_or_ignore(
+                &self,
+                into: &$crate::storage::encrypted_store::db_connection::DbConnection<'_>,
+            ) -> Result<(), $crate::StorageError> {
+                let result = into.raw_query(|conn| {
+                    diesel::insert_into($table::table)
+                        .values(self)
+                        .execute(conn)
+                });
+                $crate::storage::ignore_unique_violation(result)
```
richardhuaaa, line 264: Note - specifically ignoring unique constraint violations. Other approaches (for example, sqlite's INSERT OR IGNORE, ON CONFLICT, etc) will throw away non-null constraint errors, which risk us losing DB writes without realizing it. There are other places in the codebase where we do this, not sure if those need to be changed.

xmtp_mls/src/identity_updates.rs
```diff
@@ -115,28 +116,37 @@ where
...
+        let inbox_id = inbox_id.as_ref();
         let updates = conn.get_identity_updates(inbox_id, None, to_sequence_id)?;
-        let last_update = updates.last();
-        if last_update.is_none() {
+        let last_sequence_id = updates
+            .last()
+            .ok_or::<ClientError>(AssociationError::MissingIdentityUpdate.into())?
+            .sequence_id;
+        if to_sequence_id.is_some() && to_sequence_id != Some(last_sequence_id) {
+            // Is this right?
```
richardhuaaa, line 126: I tidied this a little bit, but the logic is still the same. I'm unsure of the assumptions behind the original logic - shouldn't we be tolerant if, for example, the storage is *ahead* of the requested sequence ID?

Also, we are putting some invisible requirements on the caller of this method to make sure that the requested `to_sequence_id` is in the DB, and to call `load_identity_updates` if not. Should we just refactor to call `load_identity_updates` natively within this method if needed?

xmtp_mls/src/lib.rs
```diff
@@ -54,16 +59,32 @@ pub trait Delete<Model> {
...
     #[ctor::ctor]
+    // Capture traces in a variable that can be checked in tests, as well as outputting them to stdout on test failure
+    #[traced_test]
     fn setup() {
-        INIT.call_once(|| {
-            tracing_subscriber::fmt::init();
-        })
+        // Capture logs (e.g. log::info!()) as traces too
+        let _ = tracing_log::LogTracer::init();
+    }
```
richardhuaaa, line 71: Heads up @insipx - this allows tests to capture the logs and check them, while still outputting them to stdout

xmtp_mls/src/storage/encrypted_store/mod.rs
neekolas: That's a very good point. We should fix those other call sites as well

xmtp_mls/migrations/2024-05-11-004236_cache_association_state/up.sql
neekolas: We almost need some sort of reference counter. It's only desirable to delete when no conversations reference that sequence ID.

xmtp_mls/src/identity_updates.rs
neekolas: > shouldn't we be tolerant if, for example, the storage is ahead of the requested sequence ID?

If the storage is ahead, those entries shouldn't be returned because we are providing a `to_sequence_id` in the call to `conn.get_identity_updates`.

> Also, we are putting some invisible requirements on the caller of this method to make sure that the requested to_sequence_id is in the DB, and to call load_identity_updates if not

I'm sure there are cases for a combined `get_association_state` + `load_identity_updates`. That would tell you what an inboxes state is right now.

I was primarily thinking about commit processing when I wrote this.

For commit processing, we actually want to load all the updates for all changed inboxes in a batch operation to save trips to the network, and then we want to get diffs for each inbox between the old and new state, which involves getting the association state at one or two different sequence IDs (depending on whether someone is a new addition to the group or an update).

xmtp_mls/migrations/2024-05-11-004236_cache_association_state/up.sql
37ng: Is this correct - If no new members of an inbox_id get added to the group, then using the association_state with old sequence id of that inbox_id is fine.

neekolas: It's valid for a group to stay on an older sequence_id. But we do try and get groups up-to-date.

Group members periodically do checks to see if any members have newer identity updates, and will publish the new `sequence_id` to the group if they find any. This happens any time you add or remove members, as well as periodically when calling sync.

xmtp_mls/src/lib.rs
insipx: sounds awesome!

neekolas (approved):

xmtp_mls/migrations/2024-05-11-004236_cache_association_state/up.sql
richardhuaaa: I'm wondering if we can do some kind of combined heuristic - we clean up caches if they are more than a day old and a newer sequence ID is cached. If so, I can address in a follow-up PR

xmtp_mls/src/storage/encrypted_store/mod.rs
richardhuaaa: Will fix in next PR

xmtp_mls/src/identity_updates.rs
richardhuaaa: > If the storage is ahead, those entries shouldn't be returned because we are providing a to_sequence_id in the call to conn.get_identity_updates

D'oh! That's obvious in hindsight.

> For commit processing, we actually want to load all the updates for all changed inboxes in a batch operation to save trips to the network, and then we want to get diffs for each inbox between the old and new state, which involves getting the association state at one or two different sequence IDs

Ahh I see. I'm wondering if it makes sense to pass the identity updates into the function, instead of separately fetching it in the function and validating against `to_sequence_id`, seeing as we're hitting the DB for `filter_inbox_ids_needing_updates()` anyway. But this is pretty minor at this point, thanks for explaining!

PR #734, opened 2024-05-14
Wasm bindings client

nplasterer:
The building blocks for the node sdk

Closes https://github.com/xmtp/libxmtp/issues/565

### DIFF
--- a/bindings_ffi/src/inbox_owner.rs
+++ b/bindings_ffi/src/inbox_owner.rs
@@ -1,3 +1,5 @@
+use xmtp_cryptography::signature::{RecoverableSignature, SignatureError};
+
 // TODO proper error handling
 #[derive(Debug, thiserror::Error)]
 pub enum SigningError {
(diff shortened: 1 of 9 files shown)


nplasterer:
**current status:**

- it compiles (with cargo build ) yay,
- not so much with cargo build --target "wasm32-unknown-unknown" which is what actually matters. boo.
- you'll get the same errors with wasm-pack --target nodejs or npm run build  

**what we need to do:**

- resolve the "lower level" dependency gaps that we take for granted as having access to for an x86 build; and import | implement those parts.
- in parallel, strip back or stub some of the function calls and return JsValue stubbed results instead, so we can at least move forward with fleshing out the consuming JS functions side.
- identify what functions are sending JS (objects | streams of objects) -> Rust; and start to handle those as well...

richardhuaaa:
Makes sense, the wasm/node integration was always going to be a little bit hairier than it first looked. Out of curiosity, what were the lower level dependency gaps mentioned here? And what kind of errors are coming up with the wasm builds?

tuddman:
The gaps & errors are around missing feature-sets & functions in the indirect dependencies.  e.g

```
error[E0432]: unresolved import `crate::sys::IoSourceState`
  --> ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mio-0.8.10/src/io_source.rs:12:5
   |
12 | use crate::sys::IoSourceState;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ no `IoSourceState` in `sys`

error[E0432]: unresolved import `crate::sys::tcp`
  --> ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mio-0.8.10/src/net/tcp/listener.rs:15:17
   |
15 | use crate::sys::tcp::{bind, listen, new_for_addr};
   |                 ^^^ could not find `tcp` in `sys`
```

and a littany of others that are mostly pretty similar.

richardhuaaa:
Not sure if helpful/you've already seen this, but this seems like a good guide with sections for both web and Node bindings: https://www.nhatcher.com/post/rust-in-anger/ . There's sections on conditional compilation and using napi-rs vs wasm-bindgen too

PR #737, opened 2024-05-15
Implement `inbox_sequence_id` method in client

37ng:
# tl;dr

Refactor `load_identity_updates` method into a function so that it can be used during client creation.
Add `sequence_id` field in `Identity` struct.
Some renaming according to [rust style guidelines](https://doc.rust-lang.org/1.0.0/style/README.html)

### DIFF
--- a/xmtp_mls/src/builder.rs
+++ b/xmtp_mls/src/builder.rs
@@ -7,6 +7,7 @@ use crate::{
     api::ApiClientWrapper,
     client::Client,
     identity::{Identity, IdentityStrategy},
+    identity_updates::load_identity_updates,
     retry::Retry,
     storage::EncryptedMessageStore,
     StorageError, XmtpApi,
@@ -19,6 +20,8 @@ pub enum ClientBuilderError {
 
     #[error("Missing parameter: {parameter}")]
     MissingParameter { parameter: &'static str },
+    #[error(transparent)]
+    ClientError(#[from] crate::client::ClientError),
 
     // #[error("Failed to serialize/deserialize state for persistence: {source}")]
     // SerializationError { source: serde_json::Error },
@@ -89,13 +92,25 @@ where
             .take()
             .ok_or(ClientBuilderError::MissingParameter { parameter: "store" })?;
         debug!("Initializing identity");
-        let identity = self
+        let mut identity = self
             .identity_strategy
             .initialize_identity(&api_client_wrapper, &store)
             .await?;
-        let new_client = Client::new(api_client_wrapper, identity, store);
 
-        Ok(new_client)
+        // get sequence_id from identity updates
+        let updates = load_identity_updates(
+            &api_client_wrapper,
+            &store.conn()?,
+            vec![identity.clone().inbox_id],
+        )
+        .await?;
+
+        let sequence_id = updates
+            .get(&identity.inbox_id)
+            .and_then(|updates| updates.last())
+            .map_or(0, |update| update.sequence_id);
+        identity.set_sequence_id(sequence_id);
+        Ok(Client::new(api_client_wrapper, identity, store))
     }
 }
 
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -32,7 +32,7 @@ use crate::{
         PreconfiguredPolicies,
     },
     identity::Identity,
-    identity_updates::IdentityUpdateError,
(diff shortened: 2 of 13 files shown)


xmtp_mls/src/groups/mod.rs
```diff
@@ -207,34 +207,42 @@ impl MlsGroup {
...
+            &client.context.identity,
             permissions.unwrap_or_default().to_policy_set(),
             Purpose::Conversation,
         )?;
         let mutable_metadata = build_mutable_metadata_extension_default()?;
         let group_membership = build_starting_group_membership_extension(
-            context.inbox_id(),
-            context.inbox_latest_sequence_id(),
+            client.context.inbox_id(),
+            client.inbox_latest_sequence_id().await?,
```
neekolas, line 229: What if we did this on client creation instead of on every group create? 

If it's out of date, some other client will quickly pick it up. And your client will as soon as you add a new member.

37ng: I can add a `sequence_id` field and use that for `inbox_latest_sequence_id()`.

37ng: > some other client will quickly pick it up

Is `other client` referring to those from the same user's device?

neekolas: Any other member of the group, including your other devices.

37ng: Done

xmtp_mls/src/identity.rs
```diff
@@ -132,6 +132,7 @@ pub enum IdentityError {
 #[derive(Debug, Clone)]
 pub struct Identity {
     pub(crate) inbox_id: InboxId,
+    pub(crate) sequence_id: u64,
```
neekolas, line 135: If we're storing it in the database as well, maybe we should just use that as the source of truth. What would happen if the DB says you are on `sequence_id` 3 and the struct says you are on `sequence_id` 2

37ng: It's populated in client creation and not getting set after and reading from db is a bit expensive.

If we want the source of truth in stored in db, would `installation_keys` be stored in db as well?

neekolas: The `installation_keys` are never going to change. But any time you publish an `IdentityUpdate` to the network the `sequence_id` will need to be updated.

We could update the `Identity` object stored on the client every time you publish an `IdentityUpdate` and not worry about the DB if we had it in a mutex.

I just worry about two sources of truth when the data can change over the life of the client.

37ng: > We could update the Identity object stored on the client every time you publish an IdentityUpdate and not worry about the DB if we had it in a mutex.

Will do

neekolas: One of the remaining TODOs in this work is that we should really call `load_identity_updates` after `apply_signature_request` so that the local DB is always up to date when you make changes. Maybe you can do that as well.

xmtp_mls/src/client.rs
```diff
@@ -343,8 +344,9 @@ where
     ) -> Result<(), ClientError> {
         log::info!("registering identity");
         let connection = self.store().conn()?;
-        let provider = self.mls_provider(connection);
+        let provider = self.mls_provider(connection.clone());
         self.apply_signature_request(signature_request).await?;
+        load_identity_updates(&self.api_client, &connection, vec![self.inbox_id()]).await?;
```
37ng, line 349: calling `load_identity_updates` after `apply_signature_request`.

insipx (approved):

xmtp_mls/src/builder.rs
```diff
@@ -93,10 +96,30 @@ where
...
+        let updates = load_identity_updates(
+            &api_client_wrapper,
+            &store.conn()?,
+            vec![identity.clone().inbox_id],
+        )
+        .await?;
+        let inbox_update_vec = updates.get(&identity.inbox_id);
+        if let Some(inbox_updates) = inbox_update_vec {
+            if !inbox_updates.is_empty() {
+                let sequence_id = inbox_updates.last().unwrap().sequence_id;
```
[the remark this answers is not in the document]
37ng: Oh thanks!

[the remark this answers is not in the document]
37ng: TIL `and_then` & `map_or`

PR #739, opened 2024-05-15
Identity DB Schema

neekolas:
## tl;dr

- Updates the database schema to use `inbox_id` instead of `account_address` everywhere
- Updates all the call-sites to use the new fields


xmtp_mls/migrations/2024-05-15-145138_new_schema/up.sql
```diff
@@ -0,0 +1,104 @@
+-- Your SQL goes here
+CREATE TABLE openmls_key_store(
+    key_bytes BLOB PRIMARY KEY NOT NULL,
+    value_bytes BLOB NOT NULL
+);
+
+CREATE TABLE identity_inbox(
```
37ng, line 7: The naming of `identity_inbox` was to avoid conflicts with legacy identity, can be `identity` now
```suggestion
CREATE TABLE identity(
```

xmtp_mls/migrations/2024-05-15-145138_new_schema/up.sql
```diff
@@ -0,0 +1,104 @@
...
+    "inbox_id" text NOT NULL,
```
37ng, line 8: Should we add sequence id to identity? 

```suggestion
    "inbox_id" text NOT NULL,
    "sequence_id" BIGINT NOT NULL,
```

37ng (approved):
some nits for `identity_inbox` table

xmtp_mls/migrations/2024-05-15-145138_new_schema/up.sql
neekolas: Probably, but lets do that as a separate PR

xmtp_mls/migrations/2024-05-15-145138_new_schema/up.sql
neekolas: Will do

PR #746, opened 2024-05-16
Adding and removing group members

neekolas:
## tl;dr

- Adds support for adding and removing group members using `GroupMembership`
- Makes `add_missing_members` work with the new membership primitives

## Next up

- I stripped out `group.members()` from this PR. Will be updating that next


xmtp_mls/src/groups/mod.rs
```diff
@@ -446,73 +453,85 @@ impl MlsGroup {
...
     ) -> Result<(), GroupError>
     where
         ApiClient: XmtpApi,
     {
         let account_addresses = sanitize_evm_addresses(account_addresses_to_add)?;
+        let inbox_id_map = client.api_client.get_inbox_ids(account_addresses).await?;
         // get current number of users in group
         let member_count = self.members()?.len();
-        if member_count + account_addresses.len() > MAX_GROUP_SIZE as usize {
+        if member_count + inbox_id_map.len() > MAX_GROUP_SIZE as usize {
```
37ng, line 474: Could multiple addresses in `account_addresses` link to the same inbox id so that we over-calculate the group size?

xmtp_mls/src/groups/sync.rs
```diff
@@ -817,95 +766,106 @@ impl MlsGroup {
...
+        // TODO:nm don't bother updating the removed members
+        let mut inbox_ids = existing_group_membership.inbox_ids();
+        inbox_ids.extend(inbox_ids_to_add);
+        let conn = provider.conn_ref();
+        // Load any missing updates from the network
+        client.load_identity_updates(conn, &inbox_ids).await?;
+
+        let latest_sequence_id_map = conn.get_latest_sequence_id(&inbox_ids)?;
+
+        // Get a list of all inbox IDs that have increased for the group
```
37ng, line 837:
```suggestion
        // Get a list of all inbox IDs that have increased sequence_id for the group
```

37ng (approved):

xmtp_mls/src/groups/mod.rs
neekolas: Only until we update `group.members()` to be based on inbox ID. Coming very soon

PR #748, opened 2024-05-16
bindings_ffi: new client & signatureRequest.

37ng:
# summary

- Implements `signature_request` & `register_identity` on Client.
- Introduces `FfiSignatureRequest` and methods `add_ecdsa_signature`, `add_erc1271_signature`, `missing_address_signatures` etc.
- `Erc1271Signature::new_with_rpc` to make platform SDK life easier by only taking `chain_rpc_url` without `chain_id`, `block_number` for Erc1271Signature

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -12,8 +12,7 @@ use tokio::sync::oneshot::Sender;
 use xmtp_api_grpc::grpc_api_helper::Client as TonicApiClient;
 use xmtp_id::{
     associations::{
-        builder::SignatureRequest, AccountId, Erc1271Signature, MemberIdentifier,
-        RecoverableEcdsaSignature,
+        builder::SignatureRequest, Erc1271Signature, MemberIdentifier, RecoverableEcdsaSignature,
     },
     InboxId,
 };
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -121,7 +121,7 @@ pub async fn create_client(
 
 #[derive(uniffi::Object)]
 pub struct FfiSignatureRequest {
-    // Using `tokio::sync::Mutex`bc MutexGuard cannot be sent between threads.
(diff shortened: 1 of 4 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -114,6 +119,73 @@ pub async fn create_client(
...
+        inner.add_signature(Box::new(erc1271_siganture)).await?;
+        Ok(())
+    }
+
+    pub async fn signature_text(&self) -> Result<String, GenericError> {
+        Ok(self.inner.lock().await.signature_text())
+    }
+
+    /// missing signatures that are from [MemberKind::Address]
+    pub async fn missing_address_signatures(
```
neekolas, line 167: You'd probably be better off just converting these addresses to strings rather than returning a `FfiMemberIdentifier` type.

Right now `FfiMemberIdentifier` doesn't even do anything, so there's no way to get the address out of it. But I think the only thing the caller _would_ want to do is to get the address as a string.

xmtp_id/src/associations/signature.rs
```diff
@@ -171,6 +184,27 @@ impl Erc1271Signature {
...
+    /// This could be used by platform SDK who only needs to provide the RPC URL and account address.
+    pub async fn new_with_rpc(
+        signature_text: String,
+        signature_bytes: Vec<u8>,
+        account_address: String,
+        chain_rpc_url: String,
+    ) -> Result<Self, SignatureError> {
+        let provider = Provider::<Http>::try_from(&chain_rpc_url)?;
+        let block_number = provider.get_block_number().await?;
+        let chain_id = provider.get_chainid().await?;
```
neekolas, line 198: Oooh. This is way nicer than what I suggested earlier.

Love this idea.

neekolas (approved):

bindings_ffi/src/mls.rs
37ng: No, I just didn't think of converting them to strings. Will do what you suggested

PR #749, opened 2024-05-17
Update group.members()

neekolas:
## tl;dr

- Updates the `group.members()` method to load from the StoredAssociationState table

## Notes

Still a bit of a hack. This approach assumes we have a `StoredAssociationState` for each member at their current `sequence_id`. The next iteration on top of this will build the association state in cases where the inbox is missing from the cache.

In pretty much all cases we should have it cached but we shouldn't need to rely on that being true.

### DIFF
--- a/xmtp_mls/src/groups/group_membership.rs
+++ b/xmtp_mls/src/groups/group_membership.rs
@@ -1,3 +1,4 @@
+use openmls::extensions::Extensions;
 use prost::{DecodeError, Message};
 use std::collections::HashMap;
 use xmtp_proto::xmtp::mls::message_contents::GroupMembership as GroupMembershipProto;
--- a/xmtp_mls/src/groups/group_membership.rs
+++ b/xmtp_mls/src/groups/group_membership.rs
@@ -1,4 +1,3 @@
-use openmls::extensions::Extensions;
 use prost::{DecodeError, Message};
 use std::collections::HashMap;
 use xmtp_proto::xmtp::mls::message_contents::GroupMembership as GroupMembershipProto;
--- a/xmtp_mls/src/groups/members.rs
+++ b/xmtp_mls/src/groups/members.rs
@@ -1,16 +1,15 @@
-use std::collections::HashMap;
+use xmtp_id::InboxId;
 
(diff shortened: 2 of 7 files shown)


xmtp_mls/src/groups/group_membership.rs
```diff
@@ -1,3 +1,4 @@
+use openmls::extensions::Extensions;
```
insipx:
```suggestion
```

xmtp_mls/src/storage/encrypted_store/association_state.rs
```diff
@@ -62,4 +68,86 @@ impl StoredAssociationState {
             })
             .transpose()
     }
+
+    pub fn batch_read_from_cache(
```
37ng, line 72: nit: this could be a `DbConnection` method

37ng (approved):

xmtp_mls/src/storage/encrypted_store/association_state.rs
neekolas: Agreed. Rich put the other `StoredAssociationState` methods on the Struct itself so I decided to follow his pattern. But I agree we should make all three methods on the DbConnection to match the rest of the storage modules. Can do as a followup

xmtp_mls/src/groups/members.rs
```diff
@@ -24,39 +23,32 @@ impl MlsGroup {
...
-                    .or_insert(GroupMember {
-                        account_address,
-                        installation_ids: vec![signature_key],
-                    });
-                acc
-            },
-        );
+        let conn = provider.conn_ref();
+        let association_state_map = StoredAssociationState::batch_read_from_cache(conn, &requests)?;
+        // TODO: Figure out what to do with missing members from the local DB. Do we go to the network? Load from identity updates?
```
insipx, line 39: Should we make this into an additional ticket?

insipx (approved):

xmtp_mls/src/groups/members.rs
neekolas: https://github.com/xmtp/libxmtp/issues/750

PR #755, opened 2024-05-20
Identity E2E integration

neekolas:
## tl;dr

- Gets identity working end to end, with tests (pretty much) all uncommented
- Updates bindings to have all required methods
- Removes a bunch of TODOs in the code

## Follow-ups

There are a few changes I had to make to get everything passing that need some follow-up.
- https://github.com/xmtp/xmtp-node-go/issues/391
- https://github.com/xmtp/libxmtp/issues/760
- https://github.com/xmtp/libxmtp/issues/761
- https://github.com/xmtp/libxmtp/issues/750
- https://github.com/xmtp/libxmtp/issues/762
- https://github.com/xmtp/libxmtp/issues/763

There are also many places that need better test coverage.


xmtp_mls/src/groups/mod.rs
```diff
@@ -468,13 +467,25 @@ impl MlsGroup {
...
+            .api_client
+            .get_inbox_ids(account_addresses.clone())
+            .await?;
         // get current number of users in group
         let member_count = self.members()?.len();
         if member_count + inbox_id_map.len() > MAX_GROUP_SIZE as usize {
             return Err(GroupError::UserLimitExceeded);
         }
 
+        if inbox_id_map.len() != account_addresses.len() {
```
neekolas, line 480: The way I am doing this validation is not very good. I have to jump through some hoops to get the actual list of missing addresses.

xmtp_mls/src/groups/mod.rs
```diff
@@ -490,6 +501,13 @@ impl MlsGroup {
             .get_membership_update_intent(client, &provider, inbox_ids, vec![])
             .await?;
 
+        // TODO:nm this isn't the best test for whether the request is valid
```
neekolas, line 504: Will fix this to do a more thorough check. This is the wrong way to do it.

xmtp_mls/src/storage/encrypted_store/identity_update.rs
```diff
@@ -85,6 +85,17 @@ impl DbConnection {
         })?)
     }
 
+    pub fn get_latest_sequence_id_for_inbox(&self, inbox_id: &str) -> Result<i64, StorageError> {
```
neekolas, line 88: @37ng I realized I gave you some bad advice last week on the sequence_id stuff.

Keeping the in-memory value up to date is a lot of hassle, so I switched to just using the latest value in the database.

Apologies for the churn.

37ng: This makes total sense! I was wondering if we could just fetch sequence_id from Identity Updates table since it's the the source of truth, but that slipped off my radar.

bindings_ffi/src/mls.rs
```diff
@@ -1144,6 +1144,8 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 5)]
+    // This one is flaky for me. Passes reliably locally and fails on CI
```
insipx, line 1147: yeah this is the one that fails because of the timeout (found during the refactor). It passes locally since CI is just slower. using `async_barrier` for sync points instead of `tokio::sleep` should fix it.

not sure if a ticket yet but should be, will make it if not

37ng (approved):
This is looking great! Left a few questions

bindings_ffi/src/mls.rs
```diff
@@ -778,6 +886,19 @@ mod tests {
...
+    async fn register_client(inbox_owner: &LocalWalletInboxOwner, client: &FfiXmtpClient) {
+        let signature_request = client.signature_request().unwrap();
+        signature_request
+            .add_ecdsa_signature(
+                inbox_owner
+                    .sign(signature_request.signature_text().await.unwrap())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
```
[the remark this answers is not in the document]
neekolas: Yes, that's right. 

Hopefully in most cases we can just use the legacy keys and so there would be no `signature_request`

bindings_ffi/src/mls.rs
neekolas: I don't think this one has a ticket. You mind making one

xmtp_mls/src/client.rs
```diff
@@ -240,8 +236,8 @@ where
     }
 
     /// Get sequence id, may not be consistent with the backend
-    pub fn inbox_sequence_id(&self) -> u64 {
-        self.context.inbox_sequence_id()
+    pub fn inbox_sequence_id(&self, conn: &DbConnection) -> Result<i64, StorageError> {
+        self.context.inbox_sequence_id(conn)
```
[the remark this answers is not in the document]
neekolas: Passing in a conn lets you use an existing transaction, instead of creating a new connection

bindings_ffi/src/mls.rs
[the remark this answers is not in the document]
37ng: How about new users, do they have to sign `signature_request`?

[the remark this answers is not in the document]
neekolas: At launch, the platform SDKs are going to create a V2 identity before creating the V3 SDK. So even new users will have a legacy key.

Once we completely kill V2, new users will have to sign a `signature_request`. But they won't have the 2 V2 signatures, so it's actually one less signature than today.

PR #757, opened 2024-05-20
Support ERC-6492

### LINKED ISSUE #756: Implement ERC-6492 verifier

37ng:
Implemented `erc6492_is_valid_signature` function that does an eth_call that deploys a `ValidateSigOffchain` contract, during its construction it deploys a `UniversalSigValidator` and validates the `address, hash, signature` inputs.

closes #756


insipx (approved):

xmtp_id/src/scw_verifiers.rs
```diff
@@ -56,6 +65,29 @@ impl ERC1271Verifier {
...
+    /// * `signature` - Could be encoded smart wallet signature or raw ECDSA signature.
+    pub async fn erc6492_is_valid_sig(
+        &self,
+        signer: Address,
+        block_number: Option<BlockNumber>,
+        hash: [u8; 32],
+        signature: Bytes,
+    ) -> Result<bool, VerifierError> {
+        let erc6492 = ERC6492::new(signer, self.provider.clone());
+        Ok(erc6492
```
neekolas, line 85: Where does the universal verifier contract get referenced here? Don't we need that as a prerequisite to this call working?

In Viem, for example, it ephemerally deploys the verifier contract bytecode before calling the contract.

37ng: I was using on chain validation where we expect a deployed `UniversalSigValidator` and call its `isValidSig` function.

Viem uses the off chain validation where it deploys `ValidateSigOffchain` contract that uses `UniversalSigValidator::isValidSig` logic in the constructor. 

Validation logic for both is identical but I figured we can only use on-chain solution given that off-chain consumes way more gas.

neekolas: Does gas matter? It reverts before any transaction goes through, so nothing actually gets spent.

The advantage of the offchain is that it works across all the chains, even ones where the validator contract hasn't been deployed yet.

37ng: Thanks, I didn't know eth_call can do verification like that - I just made the changes per your suggestion.

37ng:
Made the change to use the offchain validation, run it through Ethereum mainnet and it worked. Adding more tests...

xmtp_id/src/scw_verifier.rs
```diff
@@ -1,17 +1,27 @@
...
...e01b600052604160045260246000fd5b60005b83811015620004d5578181015183820152602001620004bb565b50506000910152565b600082601f830112620004f057600080fd5b81516001600160401b03808211156200050d576200050d620004a2565b604051601f8301601f19908116603f01168101908282118183101715620005385762000538620004a2565b816040528381528660208588010111156200055257600080fd5b62000565846020830160208901620004b8565b9695505050505050565b6000806000606084860312156200058557600080fd5b8351620005928162000489565b6020850151604086015191945092506001600160401b03811115620005b657600080fd5b620005c486828701620004de565b9150509250925092565b600080600060608486031215620005e457600080fd5b8351620005f18162000489565b60208501519093506001600160401b03808211156200060f57600080fd5b6200061d87838801620004de565b935060408601519150808211156200063457600080fd5b50620005c486828701620004de565b6000825162000657818460208701620004b8565b9190910192915050565b828152604060208201526000825180604084015262000688816060850160208701620004b8565b601f01601f1916919091016060019392505050565b600060208284031215620006b057600080fd5b81516001600160e01b0319811681146200044a57600080fd5b634e487b7160e01b600052603260045260246000fdfe5369676e617475726556616c696461746f72237265636f7665725369676e6572";
```
neekolas, line 24: Can we add a link to wherever this bytecode came from as a comment

xmtp_id/src/scw_verifier.rs
```diff
@@ -56,6 +66,69 @@ impl ERC1271Verifier {
...
+        let tokens = &[
+            Token::Address(signer),
+            Token::Bytes(hash.to_vec()),
+            Token::Bytes(signature.to_vec()),
+        ];
+        let data = constructor.encode_input(code, tokens)?;
+
+        let tx: TypedTransaction = TransactionRequest::new()
+            .from(
+                // Since this is a eth_call, we need to make sure the `from` address has balance to complete the simulation
```
neekolas, line 116: Very useful comment.

The signature validator code doesn't take `message.sender` into account anywhere?

xmtp_id/src/scw_verifier.rs
```diff
@@ -39,7 +49,7 @@ impl ERC1271Verifier {
     /// * `contract_address` - Address of the ERC1271 wallet.
     /// * `block_number` - Block number to verify the signature at.
     /// * `hash`, `signature` - Inputs to ERC-1271, used for signer verification.
-    pub async fn is_valid_signature(
+    pub async fn erc1271_is_valid_signature(
```
neekolas, line 52: Once we have tests for `erc6492_is_valid_signature` this function will go away, right?

37ng: I believe so, erc6492 is a superset of erc1271.

xmtp_id/src/scw_verifier.rs
37ng: Nope, the `message.sender` deploys the validation contract with some arguments for the actual validation.

xmtp_id/src/scw_verifier.rs
37ng: Will do. It's copied from [viem's test](https://github.com/wevm/viem/blob/6d6d092c6cacb229bb2696261ae9dbe742c26647/src/constants/contracts.ts#L1) so I don't know what the source solidity code looks like.

xmtp_id/src/scw_verifier.rs
37ng: erc6492 also covers general ECDSA signature recovery, so we might end up aggregating `add_ecdsa_signature` & `add_erc1271_signature` into one `add_erc6492_signature`.

neekolas: It's true. It would work. But then we have to go to the network to validate every signature, which is going to be a lot slower.

We do also have a bunch of additional fields for SCW signatures (like block height) that don't really make sense for EOA sigs.

37ng: Makes sense, `add_ecdsa_signature` is way faster

xmtp_id/src/scw_verifier.rs
neekolas: I did a Github search for some of the bytecode and it led me to [this repo](https://github.com/AmbireTech/signature-validator)

neekolas, line 24: Also, there's a typo in the const name `VALIDATRE_SIG_OFF_CHAIN_BYTECODE`

37ng: Thanks! Look like Viem uses Ambire wallet's 6492 contracts for testing.

37ng:
Added tests that supersede Viem's test coverage.

Will do full support(bindings, replacement of ERC1271 for ERC6492) in a follow-up PR.

xmtp_id/src/scw_verifier.rs
```diff
@@ -265,4 +332,112 @@ pub mod tests {
...
+                .erc6492_is_valid_signature(
+                    smart_wallet_address,
+                    H256::random().into(),
+                    &signature,
+                    None,
+                )
+                .await
+                .unwrap());
+
+            // TODO: Testing ERC-6492 signatures with undeployed ERC-1271.
```
37ng, line 397: It's a little bit tricky to generate signatures for undeployed ERC-1271. Though I consider it's fine to left a TODO because Viem doesn't test this either and it's recommended to be used by Coinbase.

neekolas (approved):

PR #758, opened 2024-05-20
add flag to called v3 methods for key package validation

insipx:


xmtp_mls/src/api/mls.rs
```diff
@@ -155,7 +155,11 @@ where
         Ok(out)
     }
 
-    pub async fn register_installation(&self, key_package: Vec<u8>) -> Result<Vec<u8>, ApiError> {
+    pub async fn register_installation(
+        &self,
+        key_package: Vec<u8>,
+        is_inbox_id_credential: bool,
```
37ng, line 161: Do you mind adding a doc that explains when `is_inbox_id_credential` is true / false? 
my understanding is that it's always true when registering a new installtion key to an inbox.

neekolas (approved):

xmtp_mls/src/api/mls.rs
insipx: it is always true for an inbox, it's only false with the old v3 stuff. Once we land v3 it will all go away

Issue #732, opened 2024-05-14
Bug: Crashes when DB is shared with notification extension


neekolas:
### Describe the bug

Reported by Noe from Converse

## Context

The Converse iOS app is actually made of 2 processes that need to share data: the app itself, and the [[Notification Extension](https://developer.apple.com/documentation/usernotifications/modifying-content-in-newly-delivered-notifications)](https://developer.apple.com/documentation/usernotifications/modifying-content-in-newly-delivered-notifications).

By default, on iOS, these 2 processes are considered as 2 apps, with sandboxed data.

To share data between those 2, we need them to be part of the same [[App Group](https://developer.apple.com/documentation/xcode/configuring-app-groups)](https://developer.apple.com/documentation/xcode/configuring-app-groups). App in the same App Group have access to a “Shared Container”, a place to store shared data.

## Sharing data

The goal for the app extension is to decrypt incoming notifications. For that, the app and the extension need to share some data.

In V2 (i.e. our current app) :

- from the app, we export conversations topic data
- we store it (encrypted) using MMKV into the [[shared container](https://developer.apple.com/documentation/foundation/filemanager/1412643-containerurl)](https://developer.apple.com/documentation/foundation/filemanager/1412643-containerurl) accessible from the app & the extension
- when a notification is received, the notification extension is triggered
- it loads the topic data from shared MMKV, and imports it into an XMTP Client instance
- it decodes the notification and displays it

In V3, it seems that there is no export topic / import topic feature, and that everything is stored inside an encrypted SQLite database managed by libxmtp. So for the notification extension to be able to decode V3 notifications, we might need the SQLite database to be stored inside the Shared Container.

## The issue

It is pretty hard to share an SQLite database between two processes on iOS.

We have experienced it firsthand because we currently already have our own SQLite database stored inside the shared container.

If you use Converse from Tesflight, you might have seen this kind of screen sometimes **when not on the app:**

![Untitled](https://prod-files-secure.s3.us-west-2.amazonaws.com/d9969d96-94d1-4638-bf41-198d19c2f74e/01017c01-40c8-4694-9e9d-176b649cbcc2/Untitled.png)

All the crash logs from those show it is a  `0xDEAD10CC` exception.

The reason why the app crashes while you’re not using it is because it is killed by iOS.

[...]

PR #759, opened 2024-05-20
Add ability to drop database lock

nplasterer:
Should fix https://github.com/xmtp/libxmtp/issues/732

Exposes a function to drop the database lock that can be called in iOS when the app is detected in the background.

Should reconnect the database the next time a call is made.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -145,14 +145,12 @@ impl FfiXmtpClient {
         self.inner_client.installation_public_key()
     }
 
-    pub fn release_db_connection(&self) {
-        self.inner_client.release_db_connection();
+    pub fn release_db_connection(&self) -> Result<(), GenericError> {
+        Ok(self.inner_client.release_db_connection()?)
     }
 
     pub async fn db_reconnect(&self) -> Result<(), GenericError> {
-        let _ = self.inner_client.reconnect_db();
-
-        Ok(())
+        Ok(self.inner_client.reconnect_db()?)
     }
 }
 
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -9,7 +9,6 @@ use std::sync::{
     Arc, Mutex,
 };
 use tokio::sync::oneshot::Sender;
-use uniffi::deps::once_cell::sync::Lazy;
 use xmtp_api_grpc::grpc_api_helper::Client as TonicApiClient;
 use xmtp_mls::groups::group_metadata::ConversationType;
 use xmtp_mls::groups::group_metadata::GroupMetadata;
@@ -146,15 +145,14 @@ impl FfiXmtpClient {
         self.inner_client.installation_public_key()
     }
 
-    pub async fn release_db_connection(&self) -> Result<(), GenericError> {
-        self.inner_client.release_connection();
-        Ok(())
+    pub fn release_db_connection(&self) {
+        self.inner_client.release_db_connection();
     }
 
     pub async fn db_reconnect(&self) -> Result<(), GenericError> {
-        let results = self.inner_client.reconnect_db();
+        let _ = self.inner_client.reconnect_db();
 
-        Ok(results)
+        Ok(())
     }
 }
 
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -9,13 +9,13 @@ use std::sync::{
     Arc, Mutex,
 };
 use tokio::sync::oneshot::Sender;
+use uniffi::deps::once_cell::sync::Lazy;
 use xmtp_api_grpc::grpc_api_helper::Client as TonicApiClient;
 use xmtp_mls::groups::group_metadata::ConversationType;
 use xmtp_mls::groups::group_metadata::GroupMetadata;
 use xmtp_mls::groups::group_permissions::GroupMutablePermissions;
 use xmtp_mls::groups::PreconfiguredPolicies;
 use xmtp_mls::identity::v3::{IdentityStrategy, LegacyIdentity};
-use xmtp_mls::storage::StorageError;
 use xmtp_mls::{
     builder::ClientBuilder,
     client::Client as MlsClient,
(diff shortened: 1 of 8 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -144,6 +168,25 @@ impl FfiXmtpClient {
...
         self.inner_client.installation_public_key()
     }
+
+    pub async fn release_db_connection(&self) -> Result<(), GenericError> {
+        if let Some(store) = STORE_MANAGER.lock().unwrap().get_store() {
+            store.lock().unwrap().release_connection()?;
+        }
+
+        Ok(())
+    }
```
nplasterer, line 178: Should I set a flag in here like `needs_to_be_reconnected?` and then check that through out the code and reconnect only if that is set? Or do you think it would be sufficient to tell clients to call reconnect themselves?

nplasterer:
@insipx mind helping me with this? Not sure if this is a bad idea but need a way to drop the database lock and then reconnect next time the app gets foregrounded.

bindings_ffi/src/mls.rs
```diff
@@ -28,6 +29,26 @@ use xmtp_mls::{
 
 pub type RustXmtpClient = MlsClient<TonicApiClient>;
 
+struct StoreManager {
+    store: Option<Arc<Mutex<EncryptedMessageStore>>>,
```
insipx, line 33: We probably don't want so many Mutexes, too many of these will lead to deadlocks that will be difficult to debug. Diesel Connections/Pools already have internal locking mechanisms, so it shouldn't be necessary

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -198,6 +198,35 @@ impl EncryptedMessageStore {
         crypto_utils::rng().fill_bytes(&mut key[..]);
         key
     }
+
+    pub fn release_connection(&self) -> Result<(), StorageError> {
+        let conn = self.raw_conn()?;
+
+        // Explicitly drop the connection to release the lock
+        drop(conn);
```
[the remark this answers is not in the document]
insipx: Looking through the [ConnectionManager](https://docs.rs/diesel/latest/diesel/r2d2/struct.ConnectionManager.html) Documentation, we can safely hold onto this and just drop + re-create the pool whenever we need to. The connection manager does not hold any connections to the database itself, but it allows creating them.

So, we can refactor the `EncryptedMessageStore` api a little bit to make this possible. 

- wrap `Pool` in an Option
- add a field to `EncryptedMessageStore` for `ConnectionManager`, so that we don't have to read from disk for the database file
- in `release_connection`, use [Option::take](https://doc.rust-lang.org/std/option/enum.Option.html#method.take) to replace the `Pool` with `None`
- `reconnect` stays largely the same, except you can use the already-stored `ConnectionManager` on `EncryptedStore`


You probably need to bubble this API up to `Client`/`client.rs` and then you can get rid of the global `STORE_MANAGER`

this should probably work

happy to pair with you on this if you want!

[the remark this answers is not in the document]
nplasterer: I'd love the pair.

bindings_ffi/src/mls.rs
```diff
@@ -144,6 +144,16 @@ impl FfiXmtpClient {
...
+
+    pub fn release_db_connection(&self) {
+        self.inner_client.release_db_connection();
+    }
+
+    pub async fn db_reconnect(&self) -> Result<(), GenericError> {
+        let _ = self.inner_client.reconnect_db();
+
+        Ok(())
+    }
```
nplasterer, line 156: Everything will error if they don't reconnect the database on foreground of the app. Should we be doing this reconnect for them if a release db connection gets called store a flag thats called `needs_reconnect` then when they try to do any other functions we reconnect for them? Or should we just let them manage the reconnection themselves? Is there some sort of place we could check that flag and do the reconnection for them easily?

nplasterer: Since this is just for an iOS bug we could also only expose this in the iOS sdk limiting peoples access to these functions unless explicitly needed.

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -133,15 +136,15 @@ impl EncryptedMessageStore {
     fn raw_conn(
         &self,
     ) -> Result<PooledConnection<ConnectionManager<SqliteConnection>>, StorageError> {
-        let mut conn = self
+        let pool = self
             .pool
-            .get()
+            .read()
             .map_err(|e| StorageError::Pool(e.to_string()))?;
```
insipx, line 142: It would be better practice to create a new error variant specific to locking, just in case a codepath upstream wants to handle poisons, as well as makes debugging more precise. Maybe like a `Lock(#[from] PoisonError)` variant on `StorageError` makes sense

insipx, line 142: It would be better practice to create a new error variant specific to locking, just in case a codepath upstream wants to handle poisons, as well as makes debugging more precise. Maybe like a Lock(#[from] PoisonError) variant on StorageError makes sense

bindings_ffi/src/mls.rs
insipx: +1 for hiding the complexity. If our users don't have a need for managing whether connections are active or not I think we should limit it's use as much as possible. Seems like a pretty huge footgun if someone mishandles it

nplasterer: Is there a flag like delicate or dangerous API in Rust to discourage people from using it?

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -198,6 +201,40 @@ impl EncryptedMessageStore {
         crypto_utils::rng().fill_bytes(&mut key[..]);
         key
     }
+
+    pub fn release_connection(&self) {
+        if let Ok(pool) = self.pool.write() {
+            let _conn = pool.get().expect("Failed to get a connection");
+            // Connection will be released when `_conn` goes out of scope
+        } else {
+            eprintln!("Failed to acquire write lock on the connection pool.");
```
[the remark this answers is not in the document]
nplasterer: Do we want to propagate this error? If they don't have a write lock then they don't have a connection to drop which means it's already been dropped which is a good thing in this scenario right?

insipx (approved):
Great job!

PR #767, opened 2024-05-22
Legacy key issues

neekolas:
## Summary

I've found issues creating a client with a legacy key. I've created a new test case that reproduces an issue. The legacy key I'm using was exported from xmtp-js.

1. We are validating that the result of `legacy_key_to_address` matches the address provided. But `legacy_key_to_address` returns the XMTP V2 key converted to an eth address, when it should be returning the _wallet that signed the XMTP v2 key_. I have a fix for that.
2. Even after that fix, I am seeing `Identity(SignatureRequestBuilder(Signature(Invalid)))` errors coming from my test case

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -21,6 +21,7 @@ use xmtp_mls::groups::group_metadata::ConversationType;
 use xmtp_mls::groups::group_metadata::GroupMetadata;
 use xmtp_mls::groups::group_permissions::GroupMutablePermissions;
 use xmtp_mls::groups::PreconfiguredPolicies;
+use xmtp_mls::identity::sign_with_legacy_key;
 use xmtp_mls::identity::IdentityStrategy;
 use xmtp_mls::retry::Retry;
 use xmtp_mls::{
@@ -162,10 +163,11 @@ impl FfiSignatureRequest {
         let mut inner = self.inner.lock().await;
         let signature_text = inner.signature_text();
         inner
-            .add_signature(Box::new(LegacyDelegatedSignature::new_with_bytes(
-                signature_text,
-                signature_bytes,
-            )))
+            .add_signature(Box::new(
+                sign_with_legacy_key(signature_text, signature_bytes)
+                    .await
+                    .unwrap(),
+            ))
             .await?;
 
         Ok(())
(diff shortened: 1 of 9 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -163,10 +163,11 @@ impl FfiSignatureRequest {
         let mut inner = self.inner.lock().await;
         let signature_text = inner.signature_text();
         inner
-            .add_signature(Box::new(RecoverableEcdsaSignature::new(
-                signature_text,
-                signature_bytes,
-            )))
+            .add_signature(Box::new(
```
neekolas, line 166: This is the problem. We don't need to `sign_with_legacy_key` here. The code you wrote in the `IdentityStrategy` will sign with the legacy key automatically, and there will be no signature request at all in that case. This method is for cases where we _need_ a wallet signature.

xmtp_id/src/associations/signature.rs
```diff
@@ -314,25 +320,32 @@ impl LegacyDelegatedSignature {
...
+    //         sign_with_legacy_key(signature_text, signature_bytes).unwrap();
+    // }
 }
 
 #[async_trait]
 impl Signature for LegacyDelegatedSignature {
     async fn recover_signer(&self) -> Result<MemberIdentifier, SignatureError> {
         // 1. Verify the RecoverableEcdsaSignature
-        let legacy_signer = self.legacy_key_signature.recover_signer().await?;
+        // let legacy_signer = self.legacy_key_signature.recover_signer().await?;
```
neekolas, line 337: We still need to verify the signature from the private key. The only validation this function does is check which wallet signed the public key. But we also need to make sure that the signature of the signature text is valid.

xmtp_mls/src/builder.rs
```diff
@@ -179,6 +179,27 @@ mod tests {
         assert!(!client.installation_public_key().is_empty());
     }
 
+    #[tokio::test]
+    async fn legacy_keys() {
```
neekolas, line 183: There is a quirk with this test case. It uses the legacy key the first time it is run, but after it has been run once on the local DB you will get a different result in calls to `get_inbox_ids` for that account address. That's because the first test run will register `0x0bD00B21aF9a2D538103c3AAf95Cb507f8AF1B28` with the network. If the address is already registered, the IdentityStrategy won't use the legacy key and will prompt for a wallet signature.

We should probably replace this with a dynamically generated legacy key so that we don't get confusing results in local testing where you might not drop the DB after each test run.

xmtp_id/src/associations/signature.rs
37ng: #775 should be fixing this.

PR #770, opened 2024-05-22
Async Transactions

insipx:
- adds `async_transaction` fn to `EncryptedMessageStore` which allows running transactions in async context
- refactors some errors on `StorageError` to remove the `String` in place of real error types. This should make backtraces better.

### DIFF
--- a/xmtp_mls/src/storage/encrypted_store/mod.rs
+++ b/xmtp_mls/src/storage/encrypted_store/mod.rs
@@ -93,16 +93,15 @@ impl EncryptedMessageStore {
         enc_key: Option<EncryptionKey>,
     ) -> Result<Self, StorageError> {
         log::info!("Setting up DB connection pool");
-        let pool = match opts {
-            StorageOption::Ephemeral => Pool::builder()
-                .max_size(1)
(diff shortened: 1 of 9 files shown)


xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -198,6 +194,41 @@ impl EncryptedMessageStore {
         crypto_utils::rng().fill_bytes(&mut key[..]);
         key
     }
+
+    pub async fn transaction_async<T, F, E, Fut>(&self, fun: F) -> Result<T, E>
+    where
+        F: FnOnce(XmtpOpenMlsProvider) -> Fut,
```
insipx, line 200: We can make this `FnOnce(&XmtpOpenMlsProvider)` by using a `BoxFuture` and be consistent with `fn transaction` but it seems unnecessary to box something if we don't have to.

There are some fussy alternatives using traits and stuff but until rust has real `async_closures` the lifetime issue of making the future live as long as the closure arguments w/o having to box the future is a limitation of rust

37ng (approved):

PR #771, opened 2024-05-22
feat: adds ability to load messages into file, encrypt, and decrypt

tuddman:
## Summary

- `Serialize` StoredGroupMessages into JSON to write to a `.jsonl` file
- Encrypts that file via AES-GCM
- Includes functions to decrypt that file

Follow up PR(s) will leverage using these functions to flow the files through the [message-history-server.](https://github.com/xmtp/xmtp-message-history-server)

### DIFF
--- a/xmtp_mls/src/groups/message_history.rs
+++ b/xmtp_mls/src/groups/message_history.rs
@@ -1,3 +1,12 @@
+use std::fs::File;
+use std::io::{Read, Write};
+use std::path::PathBuf;
+
+use aes_gcm::aead::generic_array::GenericArray;
+use aes_gcm::{
+    aead::{Aead, KeyInit},
+    Aes256Gcm,
+};
 use prost::Message;
 use rand::{
     distributions::{Alphanumeric, DistString},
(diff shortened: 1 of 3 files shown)


xmtp_mls/src/groups/message_history.rs
```diff
@@ -181,6 +199,74 @@ where
     }
 }
 
+#[allow(dead_code)]
```
insipx, line 202: who is calling these functions? Is it the message history server directly?

tuddman: These functions will be called (in the next PR) by this library itself, as part of the message-history request-response cycle.
No - the message history server is just a dumb file server:  takes encrypted files in; sends encrypted files out, without ever knowing nor having any abilities to see(decrypt) the contents.

I think a flow diagram for how this feature-set works is in order.  Our partners will likely have the same question(s).

xmtp_mls/src/groups/message_history.rs
```diff
@@ -31,12 +40,21 @@ use crate::{
...
     PinMismatch,
+    #[error("IO error: {0}")]
+    IO(#[from] std::io::Error),
+    #[error("JSON Serialization error: {0}")]
+    Serialization(#[from] serde_json::Error),
+    #[error("AES-GCM Encryption error")]
+    AesGcm(#[from] aes_gcm::Error),
 }
 
 impl<'c, ApiClient> Client<ApiClient>
```
insipx, line 60: We should be able to drop the lifetime here now

insipx (approved):
Looks good!

PR #775, opened 2024-05-23
Fix legacy key recovery.

37ng:
Resolves this [comment](https://github.com/xmtp/libxmtp/pull/767#discussion_r1610894707).

Derive address from legacy public key(instead of recovering the signature) and make sure it matches the address of the wallet that signed the legacy key.

Use ethers::LocalWallet to sign the message so that `EIP-191` is abstracted away, which is the reason current signature cannot be recovered properly.

Moved `sign_with_legacy_key` from xmtp_mls to xmtp_id so that we can use it in tests without introducing circular dependencies.


xmtp_id/src/associations/signature.rs
```diff
@@ -363,6 +379,37 @@ impl Signature for LegacyDelegatedSignature {
...
+        LegacySignedPrivateKeyProto::decode(legacy_signed_private_key.as_slice())?;
+    let signed_private_key::Union::Secp256k1(secp256k1) = legacy_signed_private_key_proto
+        .union
+        .ok_or(SignatureError::MalformedLegacyKey(
+            "Missing secp256k1.union field".to_string(),
+        ))?;
+    let legacy_private_key = secp256k1.bytes;
+    let wallet: LocalWallet = hex::encode(legacy_private_key).parse::<LocalWallet>()?;
+    let signature = wallet.sign_message(signature_text.clone()).await?;
+
```
richardhuaaa, line 397: This is the old logic:

```
    let legacy_private_key = secp256k1.bytes;
    let (mut delegating_signature, recovery_id) = k256_helper::sign_sha256(
        &legacy_private_key, // secret_key
        // TODO: Verify this will create a verifiable signature
        signature_text.as_bytes(), // message
    )
    .map_err(IdentityError::LegacySignature)?;
    delegating_signature.push(recovery_id); // TODO: normalize recovery ID if necessary
 ```
  Out of curiosity, what is the reason we need to change this to construct a `LocalWallet` out of the `legacy_private_key` here?

File: xmtp_mls/src/builder.rs [hunk 2]
```diff
@@ -183,7 +183,7 @@ mod tests {
     async fn legacy_keys() {
         // This test is supposed to be run with a fresh state where account_address has not been registered.
         // Subsequent runs will use a different code path.
-        let account_address = "0x0bD00B21aF9a2D538103c3AAf95Cb507f8AF1B28".to_lowercase();
+        let account_address = "0x0bd00b21af9a2d538103c3aaf95cb507f8af1b28";
```
richardhuaaa, line 186: Will this test fail if we leave it with a mix of upper and lower case? It might almost be better to test it without normalizing the address

xmtp_id/src/associations/signature.rs
```diff
@@ -531,36 +579,18 @@ mod tests {
...
+        let account_address = "0x0bd00b21af9a2d538103c3aaf95cb507f8af1b28".to_string();
+        let legacy_signed_private_key = "0880bdb7a8b3f6ede81712220a20ad528ea38ce005268c4fb13832cfed13c2b2219a378e9099e48a38a30d66ef991a96010a4c08aaa8e6f5f9311a430a41047fd90688ca39237c2899281cdf2756f9648f93767f91c0e0f74aed7e3d3a8425e9eaa9fa161341c64aa1c782d004ff37ffedc887549ead4a40f18d1179df9dff124612440a403c2cb2338fb98bfe5f6850af11f6a7e97a04350fc9d37877060f8d18e8f66de31c77b3504c93cf6a47017ea700a48625c4159e3f7e75b52ff4ea23bc13db77371001".to_string();
+        let legacy_signature = sign_with_legacy_key(
+            signature_text,
+            hex::decode(legacy_signed_private_key).unwrap(),
+        )
+        .await
+        .unwrap();
+        let expected = MemberIdentifier::Address(account_address);
+        let actual = legacy_signature.recover_signer().await.unwrap();
```
richardhuaaa, line 593: This tests the happy path, would it make sense to add a test for the case where signature validation would fail? I.e. a test that would have failed with the code from before this PR, that the code in this PR fixes?

xmtp_id/src/associations/signature.rs
37ng: Because `LocalWallet` wraps the signature_text with EIP-191 text while old logic doesn't. EIP-191 is implemented by evm wallet like Metamask. If it's not use, we won't be able to recover the signature to the signer address properly.

We can add the EIP-191 to the old logic, but I feel it's better to abstract away EIP-191 from our codebase.

37ng: [ethers crate source code](https://docs.rs/ethers-signers/2.0.14/src/ethers_signers/wallet/mod.rs.html#85) to add EIP-191 during signing, note the `hash_message`.

xmtp_id/src/associations/signature.rs
37ng: Good callout, will do

xmtp_mls/src/builder.rs
37ng: Yes the current use of strings for account addresses is a bit hacky and I'm planning to replace all account addresses with `ethers::types::Address` which is a byte array with necessary type conversion logics. I think it'd be a better canonical type than string or lower case string.

Before we do, I'm ok to leave these as is since they won't be used anyways. wdty?

37ng: To answer your question, yes the test will fail because we have `to_lowercase()` all over the code base now.

xmtp_id/src/associations/signature.rs
richardhuaaa: Ahhh that explains a lot, thank you!

richardhuaaa (approved):
Looks good! The changes to signing and verification are likely breaking changes - do we need to reset the dev network and/or give a heads up?

PR #776, opened 2024-05-23
Fail `Client` on mismatch of given address & stored identity

### LINKED ISSUE #763: Client instantiation should fail when the expected account address does not match the stored identity
When creating an identity with a strategy of `CreateIfNotFound`, the SDK should fail to instantiate if the stored identity does not match the address provided.

insipx:
closes #763


insipx:
Close in favor of #792

PR #777, opened 2024-05-24
All String lower case

37ng:
Make all string lower case to avoid mismatch addresses like `0xABC` and `0xabc`


richardhuaaa (approved):
Looks good! This problem has bitten us a few times now, some potential ideas if we ever have the bandwidth:
- We could define an 'Address' type and use strong-typing to make all addresses of that type. The type can internally lower-case the address when doing comparisons and when casting to string.
- We could add unit tests with the raw bytes of signatures that are signed by addresses with strange casing, and verify that they recover correctly

37ng:
@richardhuaaa Love the idea. Thinking about the implementation - we can have a `LowerCaseString` or stricter `LowerCaseAddress` that has private String field and public `new()` that converts everything lower case, and `getter`. Then we just need to replace addresses with it.

PR #778, opened 2024-05-24
Msg History: adds tests for preparing groups, uploading and downloading

tuddman:
## Summary

- adds groups prep fn to upload
- adds `upload_history_bundle()` and tests.
- adds `download_history_bundle()` and tests.
- other small lints


xmtp_mls/Cargo.toml
```diff
@@ -38,6 +38,8 @@ openmls_rust_crypto = { workspace = true }
 openmls_traits = { workspace = true }
 prost = { workspace = true, features = ["prost-derive"] }
 rand = { workspace = true }
+reqwest = { version = "0.12.4", features = ["stream"] }
```
insipx, line 41: maybe not something to do in the future, but since the message history stuff is only used by the message-history crate (or anybody else impl message history) what do you think of feature-gating `message-history` in order to reduce the resulting WASM binary for stuff that doesn't need message history

xmtp_mls/src/groups/message_history.rs
```diff
@@ -42,19 +43,23 @@ use crate::{
 
 const ENC_KEY_SIZE: usize = 32; // 256-bit key
 const NONCE_SIZE: usize = 12; // 96-bit nonce
+const HISTORY_SERVER_HOST: &str = "0.0.0.0";
```
insipx, line 46: should host/server be only in `mod test`?

xmtp_mls/src/groups/message_history.rs
```diff
@@ -199,7 +212,17 @@ where
     }
 }
 
-#[allow(dead_code)]
+fn write_groups_file(groups: Vec<StoredGroup>) -> Result<PathBuf, MessageHistoryError> {
+    let tmp_dir = std::env::temp_dir();
+    let file_path = tmp_dir.join("groups.jsonl");
```
insipx, line 217: Might be good to make "groups.jsonl" a const maybe

tuddman: I have a follow up PR that I'll put up soon™ that refactors this to make it generic.  i.e. this is going away.

xmtp_mls/src/groups/message_history.rs
tuddman: Yep. I've moved them to `mod tests`  .  I haven't finalized on how/where integrators should declare the `host:port` for their own message-history-server.  Likely an `std::env::var` that defaults back to ours if not provided?
Another thing to discuss and roll into a follow up PR.

Issue #773, opened 2024-05-23
Validate groups when created from welcome message


neekolas:
We aren't doing any validation on the group membership (beyond what comes for free from OpenMLS) when creating a group from a welcome.

We really need to get your local association state in sync with the `GroupMembership` and make sure that only installations specified in the `GroupMembership` are actually in the group.

Issue #769, opened 2024-05-22
Identity Release Bug - members() func returning zero for non group creator


cameronvoell:
### Describe the bug

See PR for test case reproduction: https://github.com/xmtp/libxmtp/pull/768/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3R926

### Steps to reproduce the bug

```rust
    // Test members function from non group creator
    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
    async fn test_members_func_from_non_creator() {
        let amal = ClientBuilder::new_test_client(&generate_local_wallet()).await;
        let bola = ClientBuilder::new_test_client(&generate_local_wallet()).await;

        let amal_group = amal.create_group(None).unwrap();
        amal_group
            .add_members_by_inbox_id(&amal, vec![bola.inbox_id()])
            .await
            .unwrap();

        // Get bola's version of the same group
        let bola_groups = bola.sync_welcomes().await.unwrap();
        let bola_group = bola_groups.first().unwrap();

        // Call sync for both
        amal_group.sync(&amal).await.unwrap();
        bola_group.sync(&bola).await.unwrap();

        // Verify bola can see the group name
        let bola_group_name = bola_group.group_name().unwrap();
        assert_eq!(bola_group_name, "New Group");

        // Check if both clients can see the members correctly
        let amal_members = amal_group.members().unwrap();
        let bola_members = bola_group.members().unwrap();

        assert_eq!(amal_members.len(), 2);
        assert_eq!(bola_members.len(), 2); // failing here, see len == 0
    }
```

neekolas:
Odd. I can take a look into this.

neekolas:
OK. I at least know what's going on. There are two TODOs in the integration work that both play into this issue

1. [When you join a group from a welcome, we don't currently go and download all the identity updates from the network for all the group members and make sure that the association state is correct.
](https://github.com/xmtp/libxmtp/issues/773)
2. [When calling `group.members()`, we only check our local cache for the association state to fill in each group member. Once we solve (1) that cache will be populated. Currently it only gets populated the next time you create or read a commit.
](https://github.com/xmtp/libxmtp/issues/750)

Once the first one is fixed this issue should resolve itself in most real cases. And once the second one is fixed this situation will result in an error.

cameronvoell:
@neekolas Verified that this was fixed by #781  🎉 

[...]

PR #781, opened 2024-05-27
Add validation of group membership

neekolas:
## tl;dr

- Validates that the `GroupMembership` struct matches the actual membership when you join a group

## Issues
- https://github.com/xmtp/libxmtp/issues/773
- https://github.com/xmtp/libxmtp/issues/769


xmtp_mls/src/groups/mod.rs
```diff
@@ -774,6 +778,39 @@ fn build_group_config(
...
+
+        expected_installation_ids.extend(association_state.installation_ids());
+    }
+
+    let actual_installation_ids: HashSet<Vec<u8>> = mls_group
+        .members()
+        .map(|member| member.signature_key)
+        .collect();
+
+    if expected_installation_ids != actual_installation_ids {
```
neekolas, line 807: I've left this as a strict validation where the expected and actual memberships need to match exactly.

There is a looser form where we just validate that every actual installation is allowed to be there, which would allow for cases where key packages were broken or invalid at the time the user was added to the group. I think this is the right call: we should think about the "installations who should be in the group but can't" problem more holistically.

xmtp_mls/src/groups/mod.rs
```diff
@@ -825,6 +877,50 @@ mod tests {
         messages.pop().unwrap()
     }
 
+    // Adds a member to the group without the usual validations on group membership
+    // Used for testing adversarial scenarios
+    async fn force_add_member<ApiClient: XmtpApi>(
```
neekolas, line 882: FYI, this helper is going to be useful for a lot more tests. It adds a member to the group without properly validating the commit, so you can create `GroupMembership` structs that are malicious and test if the receiving side properly rejects them.

insipx (approved):

PR #782, opened 2024-05-28
Adds the ability to add/remove admins and super admins

cameronvoell:
Part of #664 

Depends on proto update:
- https://github.com/xmtp/proto/pull/184

This PR adds the following:
- add / remove admin
- add / remove super admin
- admin_list super_admin_list
- is_admin, is_super_admin
- updated bindings
- New rule => can't remove super admin from a group (super_admin can remove super_admin status from an inbox_id, including themselves, but only if there is at least 1 super admin still remaining in the group).

Before merging:

- [x] Set `dev/gen_protos.sh` back to `main`


xmtp_mls/src/groups/validated_commit.rs
```diff
@@ -628,6 +628,7 @@ fn extract_metadata_changes(
             immutable_metadata,
             old_mutable_metadata,
         ),
+        num_super_admins: old_mutable_metadata.super_admin_list.len() as u32,
```
neekolas, line 631: Shouldn't this be the `new_mutable_metadata`?

bindings_ffi/src/mls.rs
```diff
@@ -627,6 +628,97 @@ impl FfiGroup {
...
+        );
+
+        let super_admin_list = group.super_admin_list()?;
+
+        Ok(super_admin_list)
+    }
+
+    pub fn is_admin(&self, account_address: &String) -> Result<bool, GenericError> {
+        let admin_list = self.admin_list()?;
+        Ok(admin_list.contains(account_address))
```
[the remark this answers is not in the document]
cameronvoell: 👍 updated a7cbdced26a6694f6aca5f3832737696f9057ae6

xmtp_mls/src/groups/validated_commit.rs
cameronvoell: Yea I had a check in a world where you can only remove 1 super admin at a time, that there must be more than 1 before you remove. But checking that the new mutable metadata has at least one super admin is the better option so that removing multiple superadmins would evaluate correctly as well. fixed here - d5d06f48829d0a7637ee67ffc16ea94deb220df1

nplasterer (approved):

PR #791, opened 2024-05-29
Message History - more

tuddman:
## Summary

- more concise wording for some functions and variables
- makes writing `group`s and `messages` to a file generic over `serde::Serialize`
- encapsulates generating the history reply into one callable function.
- adds a few more tests


xmtp_mls/src/groups/message_history.rs
```diff
@@ -797,6 +884,22 @@ mod tests {
...
+    async fn test_prepare_history_reply() {
+        let wallet = generate_local_wallet();
+        let amal_a = ClientBuilder::new_test_client(&wallet).await;
+        let amal_b = ClientBuilder::new_test_client(&wallet).await;
+        assert_ok!(amal_b.allow_history_sync().await);
+
+        amal_a.sync_welcomes().await.expect("sync_welcomes");
+
+        let request_id = new_request_id();
+        let url = "http://0.0.0.0:5558/upload";
```
tuddman, line 900: ignore this for now.  I was testing with a locally running `message-history-server`; I'll fix this test in a minute..

tuddman: the upload server is now mocked in this test.

xmtp_mls/src/groups/message_history.rs
```diff
@@ -182,7 +184,45 @@ where
...
 
-    pub(crate) async fn prepare_groups_to_sync(&self) -> Result<Vec<StoredGroup>, StorageError> {
+    async fn prepare_history_reply(
+        &self,
+        request_id: &str,
+        url: &str,
+    ) -> Result<HistoryReply, MessageHistoryError> {
+        let (history_file, enc_key) = self.write_history_bundle().await?;
+
+        let signing_key = HistoryKeyType::new_chacha20_poly1305_key();
```
neekolas, line 194: This is an encryption key, not a signing key, right?

neekolas, line 194: We should use a different function name for this if it's not actually returning a chacha20 key used for encryption

tuddman: fixed to actually use the `chacha20` cipher in [9896f89](https://github.com/xmtp/libxmtp/pull/791/commits/9896f892ae2d30776d9b170635c5bbafe32c12cb)

neekolas: It doesn't really matter since both are just a bunch of random bytes, but what I was getting at is that you aren't using it as a chacha20 key. You're using it as a MAC key.

neekolas (approved):

PR #792, opened 2024-05-29
Fail client instantiation if address does not match stored identity

### LINKED ISSUE #763: Client instantiation should fail when the expected account address does not match the stored identity

insipx:
This is just #776 but on `main`

closes #763


37ng (approved):
love the tests

neekolas (approved):
Agree these tests are great

PR #768, opened 2024-05-22
Adds test for group members function for non group creator client

cameronvoell:

PR #794, opened 2024-05-29
Adds test for group members function for non group creator client

cameronvoell:
Adds a test for #769 

Cherry-picked commits from https://github.com/xmtp/libxmtp/pull/768, which was targeting a branch that got squashed into `main`. Also removed some unwraps. 

#769 looks like it was indeed fixed by https://github.com/xmtp/libxmtp/pull/781. Seems like a good idea to add the test so any breaks will be caught in the future.

### DIFF
--- a/xmtp_mls/src/groups/members.rs
+++ b/xmtp_mls/src/groups/members.rs
@@ -44,14 +44,15 @@ impl MlsGroup {
 
         let conn = provider.conn_ref();
         let association_state_map = StoredAssociationState::batch_read_from_cache(conn, &requests)?;
+        let mutable_metadata = self.mutable_metadata()?;
         // TODO: Figure out what to do with missing members from the local DB. Do we go to the network? Load from identity updates?
         // Right now I am just omitting them
         let members = association_state_map
             .into_iter()
             .map(|association_state| {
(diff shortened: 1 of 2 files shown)


xmtp_mls/src/groups/members.rs
```diff
@@ -49,28 +49,25 @@ impl MlsGroup {
...
             .into_iter()
             .map(|association_state| {
-                let is_admin = self
-                    .is_admin(association_state.inbox_id().to_string())
-                    .unwrap();
-                let is_super_admin = self
-                    .is_super_admin(association_state.inbox_id().to_string())
-                    .unwrap();
+                let inbox_id_str = association_state.inbox_id().to_string();
+                let is_admin = self.is_admin(inbox_id_str.clone())?;
```
neekolas, line 53: This is going to load the MLS group from the DB on every call, isn't it?

Think we can just load the mutable metadata at the top and use a single instance of it for all the checks.

cameronvoell: good catch, fixed that here: f3c97a0b19a601536456ae53850406c5e18bc401

neekolas (approved):

PR #795, opened 2024-05-30
Timestamp and message order bug

nplasterer:
Reported here: https://github.com/xmtp/xmtp-react-native/issues/394

Reproduced here: https://github.com/xmtp/xmtp-android/pull/250


nplasterer:
This was actually a bug in the SDK.

It is to be expected that the optimistic order of messages may be incorrect and that message order is only correct after calling sync. It is also expected that the timestamp will get updated after the publish is successful reflecting the actual time that the message was published.

PR #796, opened 2024-05-30
Fix `Identity::new()`

37ng:
- Identity/Client creation now requires `inbox_id` & `nonce` on top of `account_address`. (`nonce` is only required when creating new inbox)
- Added/Modified tests to catch all cases.
- Removed `LegacyIdentitySource`


bindings_ffi/src/mls.rs
```diff
@@ -52,6 +52,27 @@ pub enum LegacyIdentitySource {
...
+/// if !inbox_id {
+///     if !legacy_key { nonce = random_u64() }
+///     inbox_id = generate_inbox_id(account_address, nonce)
+/// }
+/// db_path = $inbox_id-$env
+///
+/// xmtp.create_client(account_address, nonce, inbox_id, Option<legacy_signed_private_key_proto>)
+/// ```
+///
+/// Note that in the case where the `account_address` is already associated with the `inbox_id`, `nonce` will be ignored.
```
37ng, line 75: requesting extra scrutiny for the doc here

xmtp_mls/src/api/identity.rs
```diff
@@ -251,135 +241,4 @@ mod tests {
...
-        let identity = IdentityStrategy::CreateIfNotFound(network_address.clone(), None);
-        let err = identity
-            .initialize_identity(&wrapper, &store)
-            .await
-            .unwrap_err();
-
-        assert!(
-            matches!(err, IdentityError::InboxIdMismatch{ id, address, stored} if id == "other" && address == network_address && stored == inbox_id)
-        );
-    }
```
37ng, line 384: deleted these because they don't work for new Identity, added a few more tests in the builder.rs.

bindings_ffi/src/mls.rs
```diff
@@ -92,19 +115,20 @@ pub async fn create_client(
...
     log::info!("Creating XMTP client");
-    let legacy_key_result =
-        legacy_signed_private_key_proto.ok_or("No legacy key provided".to_string());
     // TODO: uncomment
     // let legacy_identity = match legacy_identity_source {
     //     LegacyIdentitySource::None => LegacyIdentity::None,
     //     LegacyIdentitySource::Static => LegacyIdentity::Static(legacy_key_result?),
     //     LegacyIdentitySource::Network => LegacyIdentity::Network(legacy_key_result?),
     //     LegacyIdentitySource::KeyGenerator => LegacyIdentity::KeyGenerator(legacy_key_result?),
     // };
```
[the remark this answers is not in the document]
neekolas: Yes, we can delete. We don't treat these any differently.

bindings_ffi/src/mls.rs
```diff
@@ -1047,15 +1057,18 @@ mod tests {
 
     async fn new_test_client() -> Arc<FfiXmtpClient> {
         let ffi_inbox_owner = LocalWalletInboxOwner::new();
+        let nonce = 1;
```
neekolas, line 1060: Why a nonce of 1 here?

37ng: It doesn't provide a legacy key so it cannot use nonce 0.

neekolas: Hmm. @richardhuaaa what do you think of that change to the protocol? In the XIP we don’t make any restrictions on using nonce 0 without a legacy key. 

But it does preserve nonce 0 for later use with a legacy key if we did.

37ng: Why is nonce 0 preserved for legacy keys?

Currently, if a wallet has registered with an inbox, next time it won't be able to register another inbox with nonce 0 & legacy key unless it revoked the first inbox.(I know @neekolas mentioned that it doesn't need to revoke the first inbox but only needs to register with the new inbox but I'm not sure if that works. Because the installation keys added by the wallet in the first inbox will not be removed if the wallet is not specifically revoked in the first inbox)

Would love the clarity on this - want to make sure the identity creation logic is solid

37ng: Correct me if I'm wrong - this `only be used once` has yet to be implemented in libxmtp
```
XIP-46 

// This signature type can ONLY be used for CreateInbox and AddAssociation
// payloads, and can only be used once in xmtpv3.
message LegacyDelegatedSignature {
  xmtp.message_contents.SignedPublicKey delegated_key = 1;
  RecoverableEcdsaSignature signature = 2;
}

```

richardhuaaa: > Why is nonce 0 preserved for legacy keys?

For context, this is what the XIP states:
> The ‘Create Identity’ signature of the delegated_key is implicitly treated as a CreateInbox action, with a nonce of 0.

So we use nonce 0 here because we don't have any way to retroactively specify a nonce for a `Create Identity` signature from v2.

> Currently, if a wallet has registered with an inbox, next time it won't be able to register another inbox with nonce 0 & legacy key unless it revoked the first inbox.

Is this purely an issue for unit tests, where we might want to create multiple Inbox ID's for test purposes? In a production environment, we want a wallet (and all of its installations) to always resolve to the same inbox, so there's no conflict if a legacy key registers the inbox ID with nonce 0, and a later installation uses the same nonce (or vice versa).

neekolas: > `this only be used once` has yet to be implemented in libxmtp

It's enforced in the identity update processing. Legacy signatures have to be used on nonce 0, and it will trigger a replay error if it is used more than once in a sequence of identity updates.

neekolas: > the installation keys added by the wallet in the first inbox will not be removed if the wallet is not explicitly revoked in the first inbox

For all cases where you are moving inbox_ids we can assume that you'll be working with a fresh DB and new installation keys

neekolas (approved):

neekolas:
Approving to unblock, but I think there's still more discussion required around nonce 0

37ng:
I'm working on another one that makes `signature_request()` return `None`, will merge this after that one.

PR #797, opened 2024-05-31
Message history:  insert records

tuddman:
## Summary

This patch 
- adds the ability to insert downloaded (and decrypted) history-bundle files into the local DB.
- ensures the requester inbox_id is added to all of the responder's groups.
- responder generates and sends the message bundle "reply" upon processing the `MessageHistoryRequest` message.
- after insertion, `sync_welcomes` then `sync` all the groups.
- other refactorings


xmtp_mls/src/groups/message_history.rs
```diff
@@ -62,6 +63,13 @@ pub enum MessageHistoryError {
     Storage(#[from] StorageError),
 }
 
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+enum SyncableTables {
+    StoredGroup(StoredGroup),
```
neekolas, line 69: If we are going to be inviting the new device into all known groups, do we actually need to send the group list over as part of the backup? Without any of the MLS group data stored, it'll likely throw an error trying to interact with it on the receiving side.

I guess it would make a difference for groups that the sender has been kicked out of, but maybe that is a secondary concern.

richardhuaaa: Tentative flow we discussed -

The backup responder performs the following in sequence:
1. Add the backup requester to all groups
2. Send the backup requester a list of all the groups
3. Send the backup requester all the messages in those groups

The backup requester performs the following in sequence:
1. Add all of the groups to DB
2. Add all of the group messages to DB
3. Call sync() on all the groups (and start receiving messages subsequently)
4. App is now ready for use

neekolas (approved):

PR #798, opened 2024-05-31
Break up large requests to GetIdentityUpdates and do in parallel

neekolas:
## tl;dr

When querying `GetIdentityUpdates` for a large group, the list of inboxes requested might be really long. This breaks the list of requests into groups of 50 and does them all in parallel.


37ng (approved):
Parallel processing is 🔥

insipx (approved):

Issue #788, opened 2024-05-28
Bug: creator ID not readable from Immutable Group Metadata


cameronvoell:
### Describe the bug

See link to branch with test for repro, and code snippet below

### Expected behavior

I think we should still keep creator inbox id in Immutable data and it should be readable. It is useful to know who created a group. 

### Steps to reproduce the bug

https://github.com/xmtp/libxmtp/compare/cv/group-creator-id-bug?expand=1

```rust
#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
    async fn test_can_read_group_creator_inbox_id() {
        let amal = ClientBuilder::new_test_client(&generate_local_wallet()).await;
        let policies = Some(PreconfiguredPolicies::AllMembers);
        let amal_group = amal.create_group(policies).unwrap();
        amal_group.sync(&amal).await.unwrap();
    
        let mutable_metadata = amal_group.mutable_metadata().unwrap();
        assert_eq!(mutable_metadata.admin_list.len(), 1);
        assert_eq!(mutable_metadata.admin_list[0], amal.inbox_id());
        
        let protected_metadata: GroupMetadata = amal_group.metadata().unwrap();
        assert_eq!(protected_metadata.conversation_type, ConversationType::Group);
        assert_eq!(protected_metadata.creator_inbox_id, amal.inbox_id());  // Fails here!
        /*
         * thread 'groups::tests::test_can_read_group_creator_inbox_id' 
         * panicked at xmtp_mls/src/groups/mod.rs:1742:9:assertion `left == right` failed
         * left: ""
         * right: "8c623aceafbd8a142e5327e5eb4fbee1d519c641730eec7eba84aa82e95abf47"
         */
    }
```

neekolas:
I see the bug. Good news is that it's a one-line fix.

cameronvoell:
closed in https://github.com/xmtp/libxmtp/pull/802

PR #800, opened 2024-05-31
Add node bindings

rygine:
# Summary

This PR adds node bindings for the MLS client via [napi-rs](https://napi.rs/).

### Notes

This PR will not include a CI workflow for publishing updates. That will be a fast-follow. Once this merges, I'll manually publish a package to the NPM registry.


bindings_node/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+xmtp_cryptography = { path = "../xmtp_cryptography" }
+xmtp_mls = { path = "../xmtp_mls", features = ["grpc", "native"] }
+xmtp_proto = { path = "../xmtp_proto", features = ["proto_full"] }
+xmtp_id = { path = "../xmtp_id" }
+rand = "0.8.5"
+
+[build-dependencies]
+napi-build = "2.0.1"
+
+[profile.release]
```
rygine, line 32: this could likely be improved to cut down on output file size, but that's not super important right now

bindings_node/src/encoded_content.rs
```diff
@@ -0,0 +1,81 @@
...
+    NapiContentTypeId {
+      authority_id: content_type_id.authority_id,
+      type_id: content_type_id.type_id,
+      version_major: content_type_id.version_major,
+      version_minor: content_type_id.version_minor,
+    }
+  }
+}
+
+impl Into<ContentTypeId> for NapiContentTypeId {
```
rygine, line 26: do i need an `Into` trait when there's a `From` trait?

bindings_node/src/encoded_content.rs
```diff
@@ -0,0 +1,81 @@
...
+      r#type,
+      parameters: content.parameters,
+      fallback: content.fallback,
+      compression: content.compression,
+      content: content.content.into(),
+    }
+  }
+}
+
+impl Into<EncodedContent> for NapiEncodedContent {
```
rygine, line 64: same question as above, but perhaps this is different since i'm converting a into a `Vec<u8>`?

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,209 @@
...
+use xmtp_mls::builder::ClientBuilder;
+use xmtp_mls::identity::IdentityStrategy;
+use xmtp_mls::storage::{EncryptedMessageStore, EncryptionKey, StorageOption};
+use xmtp_mls::Client as MlsClient;
+
+pub type NapiXmtpClient = MlsClient<TonicApiClient>;
+
+#[napi]
+pub struct NapiClient {
+  inner_client: Arc<NapiXmtpClient>,
```
rygine, line 20: is this `Arc` wrapper necessary?

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,209 @@
...
+
+  #[napi]
+  pub async fn db_reconnect(&self) -> Result<()> {
+    Ok(
+      self
+        .inner_client
+        .reconnect_db()
+        .map_err(|e| Error::from_reason(format!("{}", e)))?,
+    )
+  }
```
rygine, line 203: i saw these DB functions in the FFI bindings. not sure if they're necessary for Node.

bindings_node/src/streams.rs
```diff
@@ -0,0 +1,38 @@
...
+  Arc, Mutex,
+};
+use tokio::sync::oneshot::Sender;
+
+use napi_derive::napi;
+
+#[napi]
+pub struct NapiStreamCloser {
+  close_fn: Arc<Mutex<Option<Sender<()>>>>,
+  is_closed_atomic: Arc<AtomicBool>,
```
rygine, line 12: these types were taken from the FFI bindings. not sure if they should remain the same for Node.

bindings_node/src/groups.rs
```diff
@@ -0,0 +1,403 @@
...
+
+#[napi]
+pub struct NapiGroupMetadata {
+  inner: GroupMetadata,
+}
+
+#[napi]
+impl NapiGroupMetadata {
+  #[napi]
+  pub fn creator_inbox_id(&self) -> String {
```
rygine, line 59: this function doesn't return an Inbox ID, just an empty string. thoughts?

cameronvoell: saw the same, here is an issue with test for reproducing - https://github.com/xmtp/libxmtp/issues/788

bindings_node/src/encoded_content.rs
neekolas: That automatically gets generated for you based on the `From`

bindings_node/src/mls_client.rs
richardhuaaa: We added the `Arc` in the FFI bindings because Uniffi requires objects passed across the boundary to be an `Arc`. It seems `Napi` has a different way of handling this (maybe they're already wrapping it in Arc internally within their macro?)

My suggestion would be to just remove it and see if it compiles - if it compiles, it should be okay.

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,209 @@
...
+use napi_derive::napi;
+use xmtp_api_grpc::grpc_api_helper::Client as TonicApiClient;
+use xmtp_cryptography::signature::ed25519_public_key_to_address;
+use xmtp_id::associations::{Erc1271Signature, RecoverableEcdsaSignature};
+use xmtp_mls::builder::ClientBuilder;
+use xmtp_mls::identity::IdentityStrategy;
+use xmtp_mls::storage::{EncryptedMessageStore, EncryptionKey, StorageOption};
+use xmtp_mls::Client as MlsClient;
+
+pub type NapiXmtpClient = MlsClient<TonicApiClient>;
```
richardhuaaa, line 16: My suggestion would be to prefix objects passed directly across the boundary to be prefixed with `Napi`, and objects that live only in Rust with `Rust`. That's similar to the naming scheme we used in the FFI bindings. In this case, this is a rust-only object.

bindings_node/src/mls_client.rs
richardhuaaa: I used `git blame` to find the PR that added this method in the FFI bindings: https://github.com/xmtp/libxmtp/pull/759

It seems that this won't be relevant for Node, seeing as you don't have the app backgrounding issue that iOS has

bindings_node/src/streams.rs
richardhuaaa: The only constraint is that these methods must be exposed to the platform SDK (Node), so that they know if the stream is closed and have the ability to close it: https://github.com/xmtp/libxmtp/blob/main/bindings_ffi/src/mls.rs#L875-L888

I suspect we need to keep the `Mutex` and `AtomicBool` for Rust concurrency purposes, but I'm unsure if the `Arc` can be removed or not

bindings_node/src/mls_client.rs
rygine: when i remove the `Arc` wrapper, i can no longer `.clone()` the `inner_client`. how do i get around this without deriving the `Clone` trait on `xmtp_mls::Client`? passing by reference seems to bring up lifetime concerns.

richardhuaaa: Ah, so that answers the question then - we need to share a reference to the `inner_client` in multiple places in Rust, but we don't want to handle lifetimes so we use an `Arc'. It's a Rust issue rather than a bindings issue.

Technically, after Andrew's refactor, I think we could refactor all of the other structs (like `FfiGroup`) to not need to hold that client reference, hence removing the need to clone it. But honestly I don't think it matters that much if it works today.

richardhuaaa: Let me know if I didn't describe that clearly - happy to give more context on Rust stuff as I think it'll be useful going forward

rygine: yep, great explanation! my limited understanding of `Arc` fits that. definitely curious about the refactor and how a reference would no longer be needed.

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,189 @@
...
+      .inner_client
+      .can_message(account_addresses)
+      .await
+      .map_err(|e| Error::from_reason(format!("{}", e)))?;
+
+    Ok(results)
+  }
+
+  #[napi]
+  pub async fn register_ecdsa_identity(&self, signature_bytes: Uint8Array) -> Result<()> {
```
37ng, line 96: This is fine now, but if you register with legacy key, you won't need to provide `siganture_bytes`

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    let mut signature_request = match self.inner_client.identity().signature_request() {
+      Some(signature_req) => signature_req,
+      None => return Err(Error::from_reason("No signature request found")),
+    };
+    let signature_text = match self.signature_text() {
+      Some(text) => text,
+      None => return Err(Error::from_reason("No signature text found")),
+    };
```
37ng, line 110: comment for myself: We probably need to make this signature_text() simpler by having only one function that returns Option<String> so that it's easy to tell whether we need signatures from users.

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    let signature = Box::new(
+      Erc1271Signature::new_with_rpc(
+        signature_text,
+        signature_bytes.deref().to_vec(),
+        self.account_address.clone(),
```
37ng, line 156: we should take `account_address` from arguments instead of client. Client should represent inbox instead of an account address.

bindings_node/src/conversations.rs
```diff
@@ -0,0 +1,172 @@
...
+          group.group_id,
+          group.created_at_ns,
+        )
+      })
+      .collect();
+
+    Ok(convo_list)
+  }
+
+  #[napi(ts_args_type = "callback: (err: null | Error, result: NapiGroup) => void")]
```
neekolas, line 126: Wow. This typed callback support in napi is pretty cool

bindings_node/src/encoded_content.rs
neekolas: All From's automatically generated the inverse `Into`. One of those neat magic Rust features.

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,189 @@
...
+    return Ok(());
+  }
+
+  #[napi]
+  pub fn signature_text(&self) -> Option<String> {
+    match self.inner_client.identity().signature_request() {
+      Some(signature_req) => Some(signature_req.signature_text()),
+      None => None,
+    }
+  }
```
37ng, line 183: There might be multiple signature required from that `signature_request` hence you can't have `register_erc1271_identity` and `register_ecdsa_identity `. 
Here is a flow that makes sense to me 
```
let signature_request = client.signature_request()
let signature_text = signature_request.signature_text()
signature_request.add_signature(Box::new(Erc1271 or ECDSA signature))
client.register_identity(signature_request)
```
So you need to have a napi client that exposes `register_identity()`, `signature_request()`, and a napi siganture_request that exposes `signature_text()`, `add_signature()`m which is identical to ffi bindings.

37ng (changes requested):

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,189 @@
...
+        .map_err(|_| Error::from_reason("Malformed 32 byte encryption key".to_string()))?;
+      EncryptedMessageStore::new(storage_option, key)
+        .map_err(|_| Error::from_reason("Error creating encrypted message store"))?
+    }
+    None => EncryptedMessageStore::new_unencrypted(storage_option)
+      .map_err(|_| Error::from_reason("Error creating unencrypted message store"))?,
+  };
+
+  let identity_strategy =
+    IdentityStrategy::CreateIfNotFound(account_address.clone().to_lowercase(), None);
```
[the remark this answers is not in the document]
rygine: i originally had that as an option, but this client won't support legacy identity to keep things simple.

bindings_node/src/mls_client.rs
rygine: i originally went down this path, but passing `SignatureRequest` between node and rust creates a lot of complexity. my understanding is that only 1 signature is required per client.

happy to sync on this to better understand how this should work and how best to pass these more complex data types between node and rust.

37ng: > my understanding is that only 1 signature is required per client

There are cases where more than 1 signature is required e.g. it requires two signatures to revoke a wallet, one from recovery address & one from revoked wallet. 
This is fine for creating new inboxes, but I'm afraid we will need to figure out complex data types very soon.

37ng (approved):
Nice work!

We will need to figure out how to exchange complex data types(e.g. SignatureRequest) between rust and node in the follow ups.

bindings_node/src/mls_client.rs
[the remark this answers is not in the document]
37ng: I'm not sure but it should handle legacy identity properly after this [PR](https://github.com/xmtp/libxmtp/pull/796) is merged.

bindings_node/src/mls_client.rs
```diff
@@ -114,56 +118,75 @@ impl NapiClient {
...
-    signature_request
-      .add_signature(signature)
-      .await
-      .map_err(|e| Error::from_reason(format!("{}", e)))?;
+    let signature = Box::new(Erc1271Signature::new(
+      signature_text,
+      signature_bytes.deref().to_vec(),
+      account_id,
+      chain_rpc_url,
+      block_number.get_u64().1,
```
37ng, line 155: If `BigInt` isn't doing much, would it make sense to just use `u64`?

bindings_node/src/mls_client.rs
```diff
@@ -3,21 +3,28 @@ use std::ops::Deref;
...
+#[derive(Eq, Hash, PartialEq)]
+enum SignatureType {
+  RecoverableEcdsaSignature,
+  Erc1271Signature,
+}
+
 #[napi]
 pub struct NapiClient {
   inner_client: Arc<RustXmtpClient>,
+  signatures: HashMap<SignatureType, Box<dyn Signature>>,
```
37ng, line 27: Like we discussed, the key of type should be `MemberIdentifier` instead of `SignatureType`.

bindings_node/src/mls_client.rs
rygine: `BigInt` is required for larger numbers.

see: https://napi.rs/docs/concepts/values#bigint

bindings_node/src/mls_client.rs
rygine: updated: [Update signatures key type](https://github.com/xmtp/libxmtp/pull/800/commits/a9e841a2b238c5cef40d3e3a292966c186fe1570)

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    if self.signatures.is_empty() {
+      return Err(Error::from_reason(
+        "No client signatures found, add at least 1 before registering",
+      ));
+    }
+
+    let mut signature_request = match self.inner_client.identity().signature_request() {
+      Some(signature_req) => signature_req,
+      None => return Err(Error::from_reason("No signature request found")),
```
37ng, line 179: If `None` signature request is returned, you can pass `None` to register_identity and then it will register the identity without additional signatures from users.

register_identity doesn't accept None for now, but it will shortly.

bindings_node/src/mls_client.rs
```diff
@@ -0,0 +1,210 @@
...
+    };
+
+    for signature in self.signatures.values() {
+      signature_request
+        .add_signature(signature.clone())
```
37ng, line 184: This could error because not all in the signatures HashMap is needed. Try using `signature_request.missing_signatures() -> Vec<MemberIdentifier>`?

neekolas (approved):

bindings_node/src/mls_client.rs
rygine: updated the logic here: [Update register_identity](https://github.com/xmtp/libxmtp/pull/800/commits/be09af822c82c667b895a4c6dc9607f9624ba1e0)

37ng (approved):

bindings_node/src/mls_client.rs
richardhuaaa: I looked at it again, and I don't think a refactor would be helpful here! You don't need to spend a lot of time on this, but Andrew's refactor is here: https://github.com/xmtp/libxmtp/pull/727/files

Basically, an option could be to make the method signatures of `NapiGroup` match the method signatures of `MlsGroup`, where only the `context()` is passed in on construction, and the client is passed in by the caller on every method that needs it. But that requires the Node caller to manually specify the client every time it is needed, which might be inconvenient.

This sounds more complex than it is, and isn't super helpful, so don't worry too much about it!

bindings_node/src/streams.rs
richardhuaaa: (For clarity, it's fine to leave as-is, if it's currently working)

PR #802, opened 2024-05-31
Fix creator_inbox_id

neekolas:
## tl;dr

There was a missing cleanup item from our migration to Identity. This fixes that and solves https://github.com/xmtp/libxmtp/issues/788.

Thanks to @cameronvoell for both diagnosing the issue and writing a test.


cameronvoell (approved):
thanks!

PR #803, opened 2024-05-31
Run clippy on all targets (i.e tests/benchmarks)

insipx:
Quick PR to add clippy to run on all targets, which includes tests/benchmarks

### DIFF
--- a/xmtp_mls/src/groups/intents.rs
+++ b/xmtp_mls/src/groups/intents.rs
@@ -501,7 +501,7 @@ mod tests {
         let intent =
             UpdateGroupMembershipIntentData::new(membership_updates, vec!["bar".to_string()]);
 
-        let as_bytes: Vec<u8> = intent.clone().try_into().unwrap();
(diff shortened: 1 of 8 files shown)


xmtp_mls/src/groups/intents.rs
```diff
@@ -501,7 +501,7 @@ mod tests {
         let intent =
             UpdateGroupMembershipIntentData::new(membership_updates, vec!["bar".to_string()]);
 
-        let as_bytes: Vec<u8> = intent.clone().try_into().unwrap();
+        let as_bytes: Vec<u8> = intent.clone().into();
```
richardhuaaa, line 504: Will this end up being the same code?

richardhuaaa (approved):

xmtp_mls/src/groups/intents.rs
insipx: Yeah, `TryFrom` is auto-impl for anything that impls `From`

PR #808, opened 2024-06-03
Simplify missing signature logic in node bindings

rygine:
# Summary

* Used lowercase addresses in signature storage
* Simplify missing signature logic

### DIFF
--- a/bindings_node/src/mls_client.rs
+++ b/bindings_node/src/mls_client.rs
@@ -115,7 +115,7 @@ impl NapiClient {
     ));
 
     self.signatures.insert(
-      MemberIdentifier::Address(self.account_address.clone()),
+      MemberIdentifier::Address(self.account_address.clone().to_lowercase()),
       signature,
     );
 
@@ -153,7 +153,7 @@ impl NapiClient {
     ));
 
     self.signatures.insert(
-      MemberIdentifier::Address(account_address.clone()),
+      MemberIdentifier::Address(account_address.clone().to_lowercase()),
       signature,
     );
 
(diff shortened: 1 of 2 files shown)


bindings_node/src/mls_client.rs
```diff
@@ -182,14 +182,14 @@ impl NapiClient {
 
     // check for missing signatures
     let missing_signatures = signature_request.missing_signatures();
-    let missing = missing_signatures
+    let signatures_to_add = missing_signatures
       .iter()
-      .filter(|id| !self.signatures.contains_key(id))
+      .filter(|id| self.signatures.contains_key(id))
```
37ng, line 187: You will need to add all missing_signatures, not only the ones in the `self.signatures`

rygine: this function will error out if a signature is missing, which is intentional.

am i missing something?

37ng: `filter` won't return error, it's just filtering out missing signatures that are not in the `self.signatures` hash map

rygine: yep, the error will come from `self.inner_client.register_identity(signature_request)`

37ng: https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#883-886

37ng: ok then I guess you don't need to filter, just add_signature() from self.signatures and it will error if the signature is not needed

37ng (approved):

bindings_node/src/mls_client.rs
rygine: oh, that's even better! i'll update the logic again.

rygine: does this update look good? [Simplify signature logic](https://github.com/xmtp/libxmtp/pull/808/commits/746fcc8fa125d41a6dcdd09f780a32c843a50df0)

37ng: yep

PR #811, opened 2024-06-04
Enable legacy key when `account_address` is already associated with an `inbox_id`

37ng:
Now, no signature request when creating clients if legacy key is provided.


nplasterer (approved):

xmtp_mls/src/builder.rs
```diff
@@ -295,6 +295,42 @@ mod tests {
...
+            .await
+            .build()
+            .await
+            .unwrap();
+        let signature_request = client1.context.signature_request().unwrap();
+        assert!(signature_request.is_ready());
+        client1.register_identity(signature_request).await.unwrap();
+
+        let client2: Client<GrpcClient> = ClientBuilder::new(identity_strategy)
+            .temp_store()
```
neekolas, line 322: In the case @nplasterer is running into, both instances would share the same DB. So we should have the same dbPath here

37ng: Great catch on the db_path! I was wondering why second is client is not using db and here it is

PR #813, opened 2024-06-04
Register legacy keys on first use

neekolas:
## tl;dr

When creating a client with legacy keys there are no signature requests, so we need to call `identity.register()` after the identity updates are published as part of `Identity::new()`. Otherwise it will leave clients in a half-registered state with no way for callers to fix.

## Next up

We need tests that generate new legacy keys (not just static fixtures) to properly test this case. We should add those next.


37ng (approved):
This fixes the problem but I think we should find a way to call `register_identity` instead of copying the content.

neekolas:
That's fair. Can do as a refactor later.

PR #814, opened 2024-06-04
Add tests for 2nd time client creation

37ng:
1. Added `generate_random_legacy_key` for testing
2. Added several test cases for client creation.

### DIFF
--- a/xmtp_mls/src/builder.rs
+++ b/xmtp_mls/src/builder.rs
@@ -128,6 +128,7 @@ mod tests {
     use xmtp_api_grpc::grpc_api_helper::Client as GrpcClient;
     use xmtp_cryptography::signature::h160addr_to_string;
     use xmtp_cryptography::utils::{generate_local_wallet, rng};
+    use xmtp_id::associations::ValidatedLegacySignedPublicKey;
     use xmtp_id::associations::{
         generate_inbox_id, test_utils::rand_u64, RecoverableEcdsaSignature,
     };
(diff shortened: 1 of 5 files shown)


xmtp_mls/src/builder.rs
```diff
@@ -295,6 +295,76 @@ mod tests {
         }
     }
 
+    // First, create a client1 using legacy key and then test following cases:
+    // - create client2 from same db with [IdentityStrategy::CachedOnly]
+    // - create client3 from same db with [IdentityStrategy::CreateIfNotFound]
+    // - create client4 with different db.
+    #[tokio::test]
+    async fn test_2nd_time_client_creation() {
```
neekolas, line 303: Will this test fail if you run it 2X against the same backend without clearing the backend's DB?

I would expect it to, since the first time it will register using the legacy key and publish an `IdentityUpdate`. The second time it will attempt to register the same inbox_id using the legacy key, which will cause `client_1` to behave like `client_4`, which _does_ require a signature request.

neekolas: Ideally we would have a helper function that could generate new legacy keys for tests

37ng: Ya, and other tests seem failing because using the same `legacy_account_address`, this can be solved by generating random legacy_key & legacy_account_address.

37ng: Added `generate_random_legacy_key` for this

37ng:
oops FFI error, fixing it

xmtp_mls/src/builder.rs
```diff
@@ -192,6 +203,49 @@ mod tests {
         }
     }
 
+    /// Generate a random legacy key proto bytes and corresponding account address.
+    async fn generate_random_legacy_key() -> (Vec<u8>, String) {
```
neekolas, line 207: This is great

neekolas (approved):

Issue #678, opened 2024-04-19
Block resumption PSK from being used


richardhuaaa:
Here is the relevant section of the MLS spec:
https://www.rfc-editor.org/rfc/rfc9420.html#name-reinitialization

We don't have the resources to support this feature properly, therefore we should block it from happening

franziskuskiefer:
The reinit PSKs need to be stored because OpenMLS needs them as part of the group state right now.
Two checks are needed to prevent reinit PSKs from being used
* Block any ReInit proposal
* Block any Welcome with a PreSharedKeyID of type resumption with usage reinit

PR #817, opened 2024-06-06
Block resumption psks

nplasterer:
Closes https://github.com/xmtp/libxmtp/issues/678

https://www.rfc-editor.org/rfc/rfc9420.html#section-12.1
https://openmls.tech/openmls/doc/openmls/schedule/psk/enum.Psk.html

- [x] Block any ReInit proposal
- [x] Block any Welcome with a PreSharedKeyID of type resumption with usage reinit


nplasterer:
@franziskuskiefer do you mind taking a look at this? I'm not sure if this is correct way to block psks or if there are different functions I should be using.

xmtp_mls/src/groups/mod.rs
```diff
@@ -326,6 +328,9 @@ impl MlsGroup {
         let welcome = deserialize_welcome(&welcome_bytes)?;
 
         let join_config = build_group_join_config();
+        if join_config.number_of_resumption_psks > 0 {
```
nplasterer, line 331: Just noting I had to make this public so I could use it https://github.com/xmtp/openmls/pull/30

franziskuskiefer: This isn't doing what you think it's doing I'm afraid. This only prevents you from joining a group that allow resumption PSKs. It's a property you probably want to set when creating groups to avoid generating unused resumption PSKs.

But to achieve what you want, you need to look into the group secrets. That's actually not possible yet. But it is with openmls/openmls#1585. I just added a function there to [get the psks](https://github.com/openmls/openmls/pull/1586/commits/703fb763a2eef40d401c9ecfcbc9bab188663b6a). On the result from that function you can get the `processed_welcome.psks().psk()`, which returns a [`PSK`](https://openmls.tech/openmls/doc/openmls/schedule/psk/enum.Psk.html) enum. There you can error out when it contains a resumption psk (or any, depending on what you want to achieve).

franziskuskiefer (commented):
The proposal check looks good. The other one needs to be different I think (see comment).

xmtp_mls/src/groups/mod.rs
```diff
@@ -326,8 +331,12 @@ impl MlsGroup {
         let welcome = deserialize_welcome(&welcome_bytes)?;
 
         let join_config = build_group_join_config();
-        let staged_welcome =
-            StagedWelcome::new_from_welcome(provider, &join_config, welcome.clone(), None)?;
+        let processed_welcome =
+            ProcessedWelcome::new_from_welcome(provider, &join_config, welcome.clone())?;
+        if processed_welcome.psks().is_empty() {
+            return Err(GroupError::NoPSKSupport);
+        }
```
nplasterer, line 338: We may eventually want to allow external psks depending on how we want to do conversation history. But in the meantime I'm okay with blocking both resumption and external psks.

xmtp_mls/src/groups/mod.rs
nplasterer: I pulled in your PR to my fork and I seem to be able to get the psks now. 🙏

xmtp_mls/src/groups/validated_commit.rs
```diff
@@ -226,6 +228,11 @@ impl ValidatedCommit {
             &mutable_metadata,
         )?;
 
+        // Block any ReInit proposals
+        if staged_commit.psk_proposals().any(|_| true) {
```
franziskuskiefer, line 232: Should this be added to the comment on `ValidatedCommit` to list this check as well?

franziskuskiefer (commented):
lgtm!

xmtp_mls/src/groups/mod.rs
```diff
@@ -326,8 +331,12 @@ impl MlsGroup {
         let welcome = deserialize_welcome(&welcome_bytes)?;
 
         let join_config = build_group_join_config();
-        let staged_welcome =
-            StagedWelcome::new_from_welcome(provider, &join_config, welcome.clone(), None)?;
+        let processed_welcome =
+            ProcessedWelcome::new_from_welcome(provider, &join_config, welcome.clone())?;
+        if processed_welcome.psks().is_empty() {
```
nplasterer, line 336: @franziskuskiefer I've got some test failures because this `is_empty()` call runs forever and never completes. Any ideas of why that would happen?

keks: @nplasterer we had a look at this and noticed that the functions returns the `NoPSKSupport` error if the Welcome does not contain PSKs. After negating the condition of the `if` on like 336 all tests pass and we don't see any hangs. Does this work for you as well?

nplasterer: Of course. This call is supposed to be !empty not isEmpty. Blame that on Friday brain. 🤦‍♀️ Appreciate y'all taking the time. 🙏

richardhuaaa (approved):
I've looked at the streaming test failure, and the error message seems unusual (Storage(NotFound)). Given other tests pass, I don't think it's related to your changes. Feel free to ignore/land, I'll look at it separately

PR #820, opened 2024-06-06
Clear pending commit state after commit validation fails

### LINKED ISSUE #805: Bug: After a metadata update fails as expected for insufficient permissions, subsequent valid updates also fail for that client
### Describe the bug

Error is `Sync([CreateGroupContextExtProposalError(MlsGroupStateError(PendingCommit))])`

See reproducing test below

https://github.com/xmtp/libxmtp/compare/cv/metadata-update-fails-after-failed-commit?expand=1

```rust
#[tokio::test]
    async fn test_can_update_gce_after_failed_commit() {
        // Step 1: Amal creates a group
        let amal = ClientBuilder::new_test_client(&generate_local_wallet()).await;
        let policies = Some(PreconfiguredPolicies::AllMembers);
        let amal_group = amal.create_group(policies).unwrap();
        amal_group.sync(&amal).await.unwrap();

        // Step 2:  Amal adds Bola to the group
        let bola = ClientBuilder::new_test_client(&generate_local_wallet()).await;
        amal_group.add_members_by_inbox_id(&amal, vec![bola.inbox_id()]).await.unwrap();

        // Step 3: Verify that Bola can update the group name, and amal sees the update
        bola.sync_welcomes().await.unwrap();
        let bola_groups = bola.find_groups(None, None, None, None).unwrap();
        let bola_group: &MlsGroup = bola_groups.first().unwrap();
        bola_group.sync(&bola).await.unwrap();
        bola_group.update_group_name(&bola, "Name Update 1".to_string()).await.unwrap();
        amal_group.sync(&amal).await.unwrap();
        let name = amal_group.group_name().unwrap();
        assert_eq!(name, "Name Update 1");

        // Step 4:  Bola attempts an action that they do not have permissions for like add admin, fails as expected
        let result = bola_group.update_admin_list(&bola, UpdateAdminListType::Add, bola.inbox_id()).await;
        if let Err(e) = &result {
            eprintln!("Error updating admin list: {:?}", e);
        }
        // Step 5: Now have Bola attempt to update the group name again => It is failing for some reason
        let result = bola_group.update_group_name(&bola, "Name Update 2".to_string()).await;
        if let Err(e) = &result {
            eprintln!("FAILING WHEN UPDATING GROUP NAME AFTER FAILED UPDATE COMMIT: {:?}", e);
        }

        // Step 6: Verify that the group name has been updated
        amal_group.sync(&amal).await.unwrap();
        let binding = amal_group.mutable_metadata().expect("msg");
        let amal_group_name: &String = binding
            .attributes
            .get(&MetadataField::GroupName.to_string())
            .unwrap();
        assert_eq!(amal_group_name, "Name Update 2"); // <= Currently failing because error above on second name update is unexpectedly showing up

        // If you comment out step 4, the test passes
    }
```

### Expected behavior

_No response_

_No response_

cameronvoell:
Fixes #805 

Changes for preventing groups being stuck in a pending-commit state:
1. when a staged commit fails validation, clear pending commit state and set intent state to failed =>
https://github.com/xmtp/libxmtp/pull/820/files#diff-202bbf85eac4d977447dc52cf6f01c12b56e9b8e72e579844bfa5bc43862acecR233-R243
3. do not return an error for functions passed into `process_for_id` if they contain db transactions you do not wish to be rolled back
https://github.com/xmtp/libxmtp/pull/820/files#diff-202bbf85eac4d977447dc52cf6f01c12b56e9b8e72e579844bfa5bc43862acecR244-R246
4. do not retry intents with `IntentState::Error`, instead return an error
https://github.com/xmtp/libxmtp/pull/820/files#diff-202bbf85eac4d977447dc52cf6f01c12b56e9b8e72e579844bfa5bc43862acecR154-R161

### DIFF
--- a/xmtp_mls/src/groups/sync.rs
+++ b/xmtp_mls/src/groups/sync.rs
@@ -151,6 +151,13 @@ impl MlsGroup {
                     return Ok(());
                 }
                 Ok(Some(intent)) => {
+                    if intent.state == IntentState::Error {
+                        log::warn!(
+                            "not retrying intent ID {}. since it is in state Error",
+                            intent.id,
+                        );
+                        return Ok(());
+                    }
                     log::warn!(
                         "retrying intent ID {}. intent currently in state {:?}",
                         intent.id,
@@ -222,7 +229,7 @@ impl MlsGroup {
                 .await;
 
                 if maybe_validated_commit.is_err() {
-                    log::error!("error validating commit: {:?}", maybe_validated_commit);
+                    debug!("error validating commit: {:?}", maybe_validated_commit);
                     match openmls_group.clear_pending_commit(provider.storage()) {
                         Ok(_) => (),
                         Err(_) => {
@@ -231,6 +238,9 @@ impl MlsGroup {
                             ))
                         }
                     }
+                    conn.set_group_intent_error(intent.id)?;
+                    // Return before merging commit since it does not pass validation
+                    // An error will roll back clearing pending commit, so we return Ok here
                     return Ok(());
                 }
 
--- a/xmtp_mls/src/groups/sync.rs
+++ b/xmtp_mls/src/groups/sync.rs
@@ -14,10 +14,13 @@ use super::{
     GroupError, MlsGroup,
 };
 use crate::{
-    client::MessageProcessingError,
+    client::{self, MessageProcessingError},
     codecs::{group_updated::GroupUpdatedCodec, ContentCodec},
     configuration::{DELIMITER, MAX_INTENT_PUBLISH_ATTEMPTS, UPDATE_INSTALLATIONS_INTERVAL_NS},
-    groups::{intents::UpdateMetadataIntentData, validated_commit::ValidatedCommit},
+    groups::{
+        intents::UpdateMetadataIntentData,
+        validated_commit::{CommitValidationError, ValidatedCommit},
+    },
     hpke::{encrypt_welcome, HpkeError},
     identity::parse_credential,
     identity_updates::load_identity_updates,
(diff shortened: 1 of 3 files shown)


xmtp_mls/src/groups/sync.rs
```diff
@@ -151,6 +151,15 @@ impl MlsGroup {
                     return Ok(());
                 }
                 Ok(Some(intent)) => {
+                    if intent.state == IntentState::Error {
+                        log::warn!(
+                            "not retrying intent ID {}. since it is in state Error",
+                            intent.id,
+                        );
+                        return Err(last_err.unwrap_or(GroupError::Generic(
```
insipx, line 159: I personally would prefer errors like this to be made into concrete enu mvariants like `GroupError::CouldNotBeCommitted` with the message as an `#[error("Group Intent could not be permitted")]` rather than throwing everything into the `Generic` bucket. Should make these errors easy to match on should we need to, + can use the `Retryable` trait on them (in this case no but for other errors)

insipx (approved):
looks good to me other than error nit!

neekolas (approved):

xmtp_mls/src/groups/sync.rs
cameronvoell: @insipx I agree, I added a comment on the follow up issue here https://github.com/xmtp/libxmtp/issues/822#issuecomment-2153361265.

Think it's better to update this error message while looking at that issue so that this category of errors are more cohesive/ better organized. Thanks for the idea though, I'll make sure this gets updated 🙏

PR #826, opened 2024-06-06
Fix NPM publishing

rygine:
# Summary

* Update publish permission to include `id-token: write`
* Added use of an action to publish the package

### DIFF
--- a/.github/workflows/release-node-bindings.yml
+++ b/.github/workflows/release-node-bindings.yml
@@ -251,3 +251,5 @@ jobs:
         with:
           token: ${{ secrets.NPM_TOKEN }}
           package: bindings_node
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release-node-bindings.yml
+++ b/.github/workflows/release-node-bindings.yml
@@ -223,6 +223,8 @@ jobs:
           retention-days: 1
 
   publish:
+    permissions:
+      id-token: write
     runs-on: warp-ubuntu-latest-x64-8x
     needs: [build-linux, build-macos, build-windows]
     steps:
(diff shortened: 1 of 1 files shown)


neekolas (approved):

.github/workflows/release-node-bindings.yml
```diff
@@ -245,9 +247,7 @@ jobs:
...
       - name: Publish to NPM
-        run: |
-          cd bindings_node
-          npm publish
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        uses: JS-DevTools/npm-publish@v3
+        with:
+          token: ${{ secrets.NPM_TOKEN }}
```
neekolas, line 252: Don't you also need to pass in the GITHUB_TOKEN secret as well?

rygine: shouldn't need that token as this action doesn't seem to interact with the GitHub API. at least according to the action docs, they don't mention needing it.

neekolas: Then I'm not sure what adding the id-token permission does above.

Although now that I've looked at the docs for the action, you're right that it probably won't matter until you set the `provenance` flag to true.

rygine: hmm, in that case, i'll add the token back. the publish config does enable provenance.

neekolas: Based on their docs I have no idea how you're supposed to configure use both the Github token and a NPM token, since there's only one field for either. But putting it into the env (like you just did) feels like a logical guess. Otherwise you might have to dig through the code.

PR #829, opened 2024-06-07
Apply lint suggestions, add lint workflow

rygine:
# Summary

* Applied clippy suggestions
* Added GitHub action to run clippy and check format when the node bindings change

### DIFF
--- a/.github/workflows/lint-node-bindings.yaml
+++ b/.github/workflows/lint-node-bindings.yaml
@@ -1,0 +1,26 @@
+name: Lint Node Bindings
+
+on:
+  pull_request:
+    paths:
+      - "bindings_node/**"
+      - ".github/workflows/lint-node-bindings.yaml"
+
+jobs:
+  lint:
+    runs-on: warp-ubuntu-latest-x64-8x
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


.github/workflows/lint-node-bindings.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Run clippy and fail on warnings
+        run: cargo clippy --manifest-path bindings_node/Cargo.toml --all-features --all-targets --no-deps -- -Dwarnings
```
insipx, line 24: should we add this to the general `lint.yaml`? It's already running clippy for bindings_ffi and the general lib/bin targets

rygine: wanted to create a lint action that only runs when the node bindings change. perhaps that's unnecessary?

insipx: no that makes sense!

insipx (approved):

PR #831, opened 2024-06-07
replace json with tls_codec serialization

tuddman:
## Summary

This patch is in `DRAFT` status until tls_codec serde (sans `serde_json`) works for our needs.

### Approach

Right now [sql_key_store.rs](https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/src/storage/sql_key_store.rs) , like [memorary_storage/lib.rs](https://github.com/openmls/openmls/blob/main/memory_storage/src/lib.rs), rely heavily on using `serde_json::to_vec` and `serde_json::from_slice` et al, to de/serialize data into/out of the format that is needed for local storage and retrieval.  Mostly the `Key` and `Entity`  are  conversions from/to `Vec<u8>` which should make this fairly straightforward.

The aim here is to modify or replace what we need to (without expanding scope) locally in this repo and/or in our fork of `openmls` in order to have a usable storage interface that uses the `tls_codec` but does not rely on `serde_json` nor its conversion to a json `Value`.

read `TODO`s in code comments for more thoughts.

Need to implement the serialization for the following public types

- [x] `TreeSync`
- [ ] `openmls::prelude_test::hash_ref::HashReference`
- [x] `QueuedProposal`
 - [x] `GroupContext`
 - [x] `InterimTranscriptHash`
 - [x] `ConfirmationTag`
 - [x] `SignatureKeyPair`
 - [x] `KeyPackage`
 - [x] `HpkeKeyPair`
 - [x] `GroupState`
 - [x] `MessageSecrets`
 - [x] `ResumptionPskStore`
 - [x] `LeafNodeIndex`
 - [x] `GroupEpochSecrets`
 - [x] `MlsGroupJoinConfig`

---

### Other approach

Finish implementing any and all of the remaining serializers (and deserializers?) (spike needed - determine what's left)  in https://github.com/RustCrypto/formats/tree/master/tls_codec following the pattern laid out in e.g. https://github.com/openmls/openmls/blob/12c73315a257f8377b6b7bbe8cdb694808b6e1c2/openmls/src/group/mls_group/ser.rs

### DIFF
--- a/xmtp_mls/src/storage/sql_key_store.rs
+++ b/xmtp_mls/src/storage/sql_key_store.rs
@@ -6,8 +6,9 @@ use diesel::{
 };
 use log::error;
 use openmls_traits::storage::*;
-use serde::Serialize;
-use serde_json::{from_slice, from_value, Value};
+use openmls::prelude::tls_codec::{Serialize, Deserialize};
+// use serde::Serialize;
+// use serde_json::{from_slice, from_value, Value};
 
 const SELECT_QUERY: &str =
     "SELECT value_bytes FROM openmls_key_value WHERE key_bytes = ? AND version = ?";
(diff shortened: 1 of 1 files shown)


neekolas:
Given that this will break any local DBs, I'd suggest we keep this PR open until we're confident the serialization work is done. Then there's just a single breaking update for us to manage instead of multiple.

tuddman:
Closing, in favor of https://github.com/xmtp/libxmtp/pull/845 as an alternative approach.

PR #833, opened 2024-06-10
Add tests for bindings_node

rygine:
# Summary

* Added initial tests for node bindings
* Added test workflow in CI for node bindings

### Notes

These tests are not meant to re-test existing code, but rather to ensure that functions run as expected with the given parameters. The intent is to catch any underlying API changes so that the bindings can be updated accordingly.

### DIFF
--- a/.github/workflows/test-node-bindings.yml
+++ b/.github/workflows/test-node-bindings.yml
@@ -1,0 +1,71 @@
+name: Test Node Bindings
+
+on:
+  push:
+    branches:
+      - main
+
+  pull_request:
+    # only run tests when related changes are made
+    paths:
+      - ".github/workflows/test-node-bindings.yml"
+      - "bindings_node/**"
+      - "dev/**"
+      - "mls_validation_service/**"
+      - "xmtp_api_grpc/**"
+      - "xmtp_cryptography/**"
+      - "xmtp_id/**"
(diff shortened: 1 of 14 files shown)


.github/workflows/test-node-bindings.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Start Docker containers
+        run: dev/up
+
+      - name: Run cargo test on bindings_node
+        run: |
+          cargo test --manifest-path bindings_node/Cargo.toml -- --test-threads=2
```
37ng, line 48: Would it make sense to reuse `test.yml` so that we just need to add the `bindings_node` to cache and then run the tests.

37ng (approved):

.github/workflows/test-node-bindings.yml
rygine: this is another instance where running these tests is optimized by path changes. see `paths-ignore` above.

37ng: Yeah CI does't run the test when the changes are only in `bindings_node` but I feel creating a new file would make the `workflows/` folder a bit messy while not saving a lot. 
Anyways, it's very nit feel free to disregard.

PR #835, opened 2024-06-11
Refactor of stream_all_messages, fix flaky stream tests

### LINKED ISSUE #772: Fix Flaky Test

https://github.com/xmtp/libxmtp/blob/75ee4aeb869d66340226893ff98d8daf188166ff/bindings_ffi/src/mls.rs#L1044

Test is flaking b/c the streamed messages are arriving too slow to one end of the `stream`. Calls to `tokio::sleep` are supposed to give time for messages to arrive but depending on architecture/system speed this varies. 

It would be great to use `async-barrier` for these tests in order to guarantee sync points and avoid flakiness

_No response_

### Additional details

_No response_

insipx:
- refactors stream_all_messages to return a `impl Stream`
    - Limits missed messages by draining old stream and using std::mem::replace
        - missed messages are still possible, but should occur much less frequently since the surface area went from the future of creating a new stream, to the time it takes to replace the stream with `mem::replace`.
    - the callback now just spawns a task which loops over the stream
- use `tokio::sync::Notify` to create formal sync points across streaming tests
- remove `StreamCloser` duplication and instead use `tokio::task::JoinHandle` as StreamCloser in ffi code
    - `StreamCloser` now has another `end` function, `end_and_wait`. `end_and_wait` waits for the stream to end before exiting the function. the `end` function is still fire-and-forget (allowing it to remain synchronous). 
- stream functions are all uniformly synchronous which fixes the disparity between ffi & node. Stream functions return a `StreamHandle` that has async methods:
    -  `wait_for_ready`: async waits until the stream is ready to accept messages
    - `end_and_wait`: ends the stream and waits for it to shutdown
- `end` is still sync and just tells the stream to stop without blocking
- new test for missed messages

fixes #772 

Fixes most of the flaky tests, except for `test_can_stream_group_messages_for_updates` and `test_can_stream_and_update_name_without_forking_group` which are unrelated to the stream changes, and instead one of them suffers from the `dblock` issue (although only in CI), and the other is problematic I think b/c of `publish_intents` conflicts (also only in CI).


xmtp_mls/src/subscriptions.rs
```diff
@@ -464,81 +443,82 @@ mod tests {
...
             .create_group(None, GroupMetadataOptions::default())
             .unwrap();
         alix_group_2
             .add_members_by_inbox_id(&alix, vec![caro.inbox_id()])
             .await
             .unwrap();
-        tokio::time::sleep(std::time::Duration::from_millis(300)).await;
+        
+        // TODO:
+        // theres missed messages here IF:
```
insipx, line 492: Got stuck here debugging for a bit with lots of improvised logs before figuring out what the issue is:

If a message is sent right after a new group is created, we can run into an issue where we actually miss messages on the stream. The message will be sent/published while the old stream is active, and then once the new stream is created it mysteriously disappears. At least, that's what i believe the issue to be. 

I think this is because of the round-trip to the gRPC and back. I reproduced this issue with the code on `main` before refactor, too, so it's not a new issue but was masked because of the `std::time::sleeps`.

Not sure if there is a good way to avoid this, or we just have to live with it. Maybe some kind of update function into gRPC instead of re-initalizing the stream? Would love your input here @richardhuaaa or @neekolas 

Maybe not a huge deal though. I'm assuming sdks have functions for filling in missing messages?

insipx: While typing this out I realized a possible solution which i did [here](https://github.com/xmtp/libxmtp/pull/835/commits/2bae690cf1625297ebd91d5c258c182681fe6f52)

One of the problems is that the condition happens when the message is sent _just as_ the new stream is initializing. The new stream initializes and dropped while the message is still present.

Also created a test to better represent this case

nplasterer (approved):

bindings_ffi/src/mls.rs
```diff
@@ -1023,22 +1008,16 @@ impl FfiGroup {
...
             self.created_at_ns,
             move |message| message_callback.on_message(message.into()),
-        )?;
+        );
 
-        Ok(Arc::new(FfiStreamCloser {
-            close_fn: stream_closer.close_fn,
-            is_closed_atomic: stream_closer.is_closed_atomic,
-        }))
+        FfiStreamCloser::new(handle)
```
richardhuaaa, line 1020: Why do we hold the error inside of the `FfiStreamCloser` instead of returning it to the caller of this method?

richardhuaaa (approved):
Well done, this is a number of very helpful improvements! TIL about `tokio::sync::notify`, that helps a lot with the flaky tests. And handling the changeover between streams more gracefully is very helpful for reducing missed messages.

I think we need to propagate some changes up to the mobile SDK's, now that the stream methods are synchronous?

bindings_ffi/src/mls.rs
insipx: The stream is started inside of the `tokio::spawn`, so the handle is returning the error once the stream is `Joined`/awaited. 

I think a good followup is [this](https://github.com/xmtp/libxmtp/issues/899) to allow the sdks to handle the errors as necessary

insipx:
> I think we need to propagate some changes up to the mobile SDK's, now that the stream methods are synchronous?

Yeah there do need to be some minor sdk changes -- mostly from async to sync and one method has an error return removed

PR #839, opened 2024-06-12
modify max group size according to benchmarks/docs, 250 -> 400

insipx:


neekolas (approved):
If you're confident it'll work reliably, let's do it

PR #840, opened 2024-06-12
Create a group with a name

nplasterer:
This adds the ability to create a group with a name.

Refactors into a opts field so that we can add more metadata later.


nplasterer:
@rygine when I make changes like this to the bindingsFFI would it be helpful if I also made them in Node?

rygine:
> @rygine when I make changes like this to the bindingsFFI would it be helpful if I also made them in Node?

totally up to you! i'm happy to make the changes as well.

alexrisch (approved):

bindings_ffi/src/mls.rs
```diff
@@ -1377,7 +1432,13 @@ mod tests {
 
         let alix_group = alix
             .conversations()
-            .create_group(vec![caro.account_address.clone()], None)
+            .create_group(
+                vec![caro.account_address.clone()],
+                FfiCreateGroupOptions {
```
insipx, line 1437: Could derive 'Default' to make all these 'FfiCreateGroupOptions::default()'.

insipx (approved):

bindings_ffi/src/mls.rs
nplasterer: Learn a new rust trick every day!

bindings_ffi/src/mls.rs
```diff
@@ -330,6 +330,15 @@ impl FfiConversations {
                 .add_members(&self.inner_client, account_addresses)
                 .await?;
         }
+
+        if let Some(ref name) = opts.group_name {
+            if !name.is_empty() {
+                convo
+                    .update_group_name(&self.inner_client, name.to_string())
```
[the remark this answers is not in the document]
nplasterer: Ya I was curious about that. Felt like it would have multiple intents getting committed just on create especially as the metadata on create grows. I'll dig deeper.

cameronvoell (approved):
perfect, thanks for adding!

PR #844, opened 2024-06-14
Update create group, add getter/setter for group image URL

rygine:
# Summary

* Added options to set group name and image URL during creation
* Added getter/setter for group image URL

### DIFF
--- a/bindings_node/src/groups.rs
+++ b/bindings_node/src/groups.rs
@@ -469,6 +469,37 @@ impl NapiGroup {
     Ok(group_name)
   }
 
+  #[napi]
+  pub async fn update_group_image_url_square(&self, group_image_url_square: String) -> Result<()> {
+    let group = MlsGroup::new(
+      self.inner_client.context().clone(),
+      self.group_id.clone(),
+      self.created_at_ns,
+    );
+
+    group
+      .update_group_image_url_square(&self.inner_client, group_image_url_square)
+      .await
+      .map_err(|e| Error::from_reason(format!("{}", e)))?;
+
+    Ok(())
+  }
+
+  #[napi]
+  pub fn group_image_url_square(&self) -> Result<String> {
+    let group = MlsGroup::new(
+      self.inner_client.context().clone(),
+      self.group_id.clone(),
+      self.created_at_ns,
+    );
+
+    let group_image_url_square = group
+      .group_image_url_square()
+      .map_err(|e| Error::from_reason(format!("{}", e)))?;
+
+    Ok(group_image_url_square)
+  }
+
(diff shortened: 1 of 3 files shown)


bindings_node/src/groups.rs
```diff
@@ -469,6 +469,37 @@ impl NapiGroup {
     Ok(group_name)
   }
 
+  #[napi]
+  pub async fn update_group_image_url_square(&self, group_image_url_square: String) -> Result<()> {
```
tuddman, line 473:
```suggestion
  pub async fn update_group_image_url_square(&self, group_image_url_square: &str) -> Result<()> {
```

tuddman: If you can get away with it; it's typically better/more efficient to pass `str`-slices rather than allocated `String`s

bindings_node/src/groups.rs
```diff
@@ -469,6 +469,37 @@ impl NapiGroup {
...
+    let group = MlsGroup::new(
+      self.inner_client.context().clone(),
+      self.group_id.clone(),
+      self.created_at_ns,
+    );
```
tuddman, line 478: Would it be better to: rather than pass the whole client just to get a group with `group_id` instantiated, to instead put this function (as a method) on the `MLSGroup` struct directly?

tuddman (approved):
Love to see this!
I left some minor rusty suggestions.   Do what works for now.

bindings_node/src/groups.rs
rygine: would love to understand this more. i was just matching the signature of `MlsGroup::update_group_image_url_square`.

bindings_node/src/groups.rs
rygine: i'm not sure i understand this question

bindings_node/src/groups.rs
tuddman: ahh. okay. you can probably get away with leaving that argument on your end to be passed by reference, a `&str` ; but you'll have to call `group_image_url_square.to_owned()` on the inner call (which defeats the benefit of not allocating) but you're reducing the places where we're passing `String` around if/when it is refactored to only allocate a `String` where its strictly necessary.   Most of the time, it's not.  And is better/more memory efficient to pass references.

understanding the differences between `&str` and `String`, really...

bindings_node/src/groups.rs
tuddman: nvm.  Looking at this PR more closely, I see you're converting over from a `NapiGroup` to an `MLSGroup` to call the method; passing through.  I mistakenly thought this is being implemented on the client.

Please proceed with merging 😆

PR #846, opened 2024-06-14
Don't process same seq id welcome message twice

cameronvoell:
Fixes https://github.com/xmtp/xmtp-react-native/issues/416

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -279,7 +279,7 @@ impl FfiXmtpClient {
     }
 }
 
-#[derive(uniffi::Record)]
+#[derive(uniffi::Record, Default)]
 pub struct FfiListConversationsOptions {
     pub created_after_ns: Option<i64>,
     pub created_before_ns: Option<i64>,
@@ -455,7 +455,7 @@ pub enum FfiPermissionLevel {
     SuperAdmin,
 }
 
-#[derive(uniffi::Record)]
+#[derive(uniffi::Record, Default)]
 pub struct FfiListMessagesOptions {
     pub sent_before_ns: Option<i64>,
     pub sent_after_ns: Option<i64>,
@@ -998,8 +998,7 @@ impl FfiGroupPermissions {
 #[cfg(test)]
 mod tests {
     use crate::{
-        get_inbox_id_for_address, inbox_owner::SigningError, logger::FfiLogger,
-        FfiConversationCallback, FfiCreateGroupOptions, FfiInboxOwner, GroupPermissions,
+        get_inbox_id_for_address, inbox_owner::SigningError, logger::FfiLogger, FfiConversationCallback, FfiCreateGroupOptions, FfiInboxOwner, FfiListConversationsOptions, FfiListMessagesOptions, GroupPermissions
     };
     use std::{
         env,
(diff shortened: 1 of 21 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -71,6 +71,7 @@ pub async fn create_client(
     nonce: u64,
     legacy_signed_private_key_proto: Option<Vec<u8>>,
 ) -> Result<Arc<FfiXmtpClient>, GenericError> {
+    init_logger(logger);
```
nplasterer, line 74: If you add this here. Can you remove it from line 133 https://github.com/xmtp/libxmtp/pull/846/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8fR133

The double init causes errors in android for some reason and error logs will no longer show up in the android.

xmtp_mls/src/groups/sync.rs
```diff
@@ -515,7 +532,20 @@ impl MlsGroup {
                     openmls_group,
                 )
                 .await?;
-                openmls_group.merge_staged_commit(provider, sc)?;
```
neekolas, line 518: This change will no longer return early on error

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -184,15 +185,21 @@ impl EncryptedMessageStore {
...
             }
-            Err(err) => match conn.raw_query(|conn| {
-                PoolTransactionManager::<AnsiTransactionManager>::rollback_transaction(&mut *conn)
-            }) {
-                Ok(()) => Err(err),
-                Err(Error::BrokenTransactionManager) => Err(err),
-                Err(rollback) => Err(rollback.into()),
-            },
+            Err(err) => {
+                log::info!("Transaction being rolled back");
```
neekolas, line 192: We can probably switch some of these to debug logs when we're ready

bindings_ffi/src/mls.rs
```diff
@@ -1046,11 +1048,13 @@ mod tests {
         }
     }
 
-    pub struct MockLogger {}
+    pub struct MockLogger {
+        inbox_id: String,
```
neekolas, line 1052: Nice change. Makes life much easier.

bindings_ffi/src/mls.rs
cameronvoell: removed here 0c68a692a8fd82bc2b5f8f69df2c62d4b833ea88

xmtp_mls/src/groups/sync.rs
cameronvoell: good catch, im checking all the other logs I added. This is fixed here c8db82024e12b1f43f85d3ec61c960eb8654c04a

xmtp_mls/src/storage/encrypted_store/mod.rs
cameronvoell: updated some here, will look a bit for other candidates a40d4459b670fa90175e886b564e811117e32633

nplasterer (approved):
Are this jniLibs the latest?

bindings_ffi/src/mls.rs
richardhuaaa: It actually doesn't work currently :( There is one global MockLogger used, and the first one wins - so the second inbox ID in tests just logs with the first inbox's logger. We can probably cook up some clever way to log inbox ID's by default in the future, but it can't be done by storing it on the logger.

xmtp_mls/src/identity_updates.rs
```diff
@@ -117,7 +116,6 @@ where
             last_sequence_id,
             association_state.clone(),
         )?;
-        log::debug!("Wrote association state to cache");
```
neekolas, line 120: There's a test that relies on this log line. Even if it's not the most useful log line, probably fine to just put it back.

cameronvoell: added back here https://github.com/xmtp/libxmtp/pull/846/commits/f5aa568f938f0868195d0b5023e803f61c63c5bb

neekolas (approved):

neekolas:
Merging this as-is, and we can figure out the streaming/database locking issue as a separate PR

PR #849, opened 2024-06-20
Message History - FFI & Node Bindings

tuddman:


bindings_ffi/src/mls.rs
```diff
@@ -430,6 +444,11 @@ impl FfiConversations {
             is_closed_atomic: stream_closer.is_closed_atomic,
         }))
     }
+
+    pub async fn request_history_sync(&self) -> Result<(), GenericError> {
+        self.inner_client.send_history_request().await?;
+        Ok(())
+    }
```
nplasterer, line 451: Is there a reason this is on conversations and not on client?

bindings_ffi/src/mls.rs
```diff
@@ -70,6 +70,7 @@ pub async fn create_client(
     account_address: String,
     nonce: u64,
     legacy_signed_private_key_proto: Option<Vec<u8>>,
+    history_sync_url: Option<String>,
```
nplasterer, line 73: Where does this URL come from?

nplasterer (commented):
Was just looking at the code here and trying to think through this again. Is the flow

I open an app and see a dialog to recover messageHistory before I create a client?

Be nice to have a diagram of which client is calling which (sorry if this already exists) like:

NewClient - calls `request_history_sync`
OldClient - confirms request
NewClient - gets recreated with a `history_sync_url`?

bindings_ffi/src/mls.rs
tuddman: Integrators will supply it when they create a client-app - if they want to run their own Msg History Server; else, when we stand up our own, we can point to ours as a fallback/default.

nplasterer (approved):
This looks incrementally good but I think we're missing the new topic work for actually subscribing to the message request response. Or am I just missing that here?

PR #860, opened 2024-06-22
[WIP] DB Retries

neekolas:
## tl;dr

Add more retries to things that might fail due to locked databases

### DIFF
--- a/xmtp_mls/src/storage/encrypted_store/mod.rs
+++ b/xmtp_mls/src/storage/encrypted_store/mod.rs
@@ -24,6 +24,7 @@ pub mod schema;
 use std::{
     borrow::Cow,
     sync::{Arc, RwLock},
+    time::Duration,
 };
 
 use diesel::{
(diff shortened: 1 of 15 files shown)


nplasterer (approved):

xmtp_mls/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"]
```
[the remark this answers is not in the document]
insipx: The recursion limit being high is a result first showed up after adding `tracing::instrument` attribute macros everywhere as a result of benchmarking. These macros are nested within each other depending on the codepath resulting in us reaching the default limit. The easiest is to just remove the tracing::instrument macros. I haven't tried it but we can also specify a [static verbosity level](https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters) to skip compiling these macros unless we specify a trace level at compile time. This would probably get rid of the attribute for the `xmtp_mls` crate, but would require the attribute for benchmarks or anything making use of the instrumentation.

insipx (approved):

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -146,6 +146,8 @@ impl EncryptedMessageStore {
             conn.batch_execute(&format!("PRAGMA key = \"x'{}'\";", hex::encode(key)))?;
         }
 
+        conn.batch_execute("PRAGMA busy_timeout = 5000;")?;
```
richardhuaaa, line 149: Are we seeing busy errors that take more than 5s?

richardhuaaa (approved):
Very nice reliability improvement

xmtp_mls/src/storage/encrypted_store/mod.rs
neekolas: Tbh, this change did nothing. We aren't seeing busy errors at all, since the DB locks are a different error code altogether.

PR #861, opened 2024-06-24
Fetch Group & Message by id & InboxId from address

nplasterer:
Part of https://github.com/xmtp/xmtp-react-native/issues/422 & https://github.com/xmtp/xmtp-react-native/issues/426

Fetch message and group by ID from local database.
Also add a method to fetch the inboxId from an address.

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -253,7 +253,7 @@ where
         &self,
         address: String,
     ) -> Result<Option<String>, ClientError> {
-        let results = self.api_client.get_inbox_ids(vec![address.clone()]).await?;
+        let mut results = self.api_client.get_inbox_ids(vec![address.clone()]).await?;
         Ok(results.get(&address).cloned())
     }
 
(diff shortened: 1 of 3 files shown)


alexrisch (approved):

xmtp_mls/src/client.rs
```diff
@@ -249,6 +249,14 @@ where
         self.context.inbox_id()
     }
 
+    pub async fn find_inbox_id_from_address(
+        &self,
+        address: String,
+    ) -> Result<Option<String>, ClientError> {
+        let results = self.api_client.get_inbox_ids(vec![address.clone()]).await?;
```
insipx, line 256:
```suggestion
        let mut results = self.api_client.get_inbox_ids(vec![address.clone()]).await?;
```

xmtp_mls/src/client.rs
```diff
@@ -249,6 +249,14 @@ where
...
+        Ok(results.get(&address).cloned())
```
insipx, line 257:
```suggestion
        Ok(results.remove(&address))
```

insipx: can avoid the clone by making results mutable and using [`HashMap::remove`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.remove) instead of `get`

insipx (approved):

PR #862, opened 2024-06-24
Move message lookup to client

rygine:
# Summary

This PR moves the find message by ID method to the client in `xmtp_mls` and `bindings_ffi`.


nplasterer (approved):
Thanks!

Issue #857, opened 2024-06-21
Feature request: remove members should be admin only even in AllMembers PolicySet


cameronvoell:

cameronvoell:
fixed in https://github.com/xmtp/libxmtp/pull/867

PR #867, opened 2024-06-24
Fix invalid policy error

### LINKED ISSUE #851: Invalid Policy error

cameronvoell:
Fixes #851 

We were throwing an error when reading group policies whenever the policy did not exactly match the libxmtp defined policy set for our preconfigured policy set of All Members, or Admin Only. When we added a new metadata option, those preconfigured policies would add a policy for the new metadata option, and the existing group policy sets would not have that, so we would not match any of the preconfigured policy sets.

Part 1 of the fix is to make it so that if all the metadata policies match the all members / admin only policy set default, that they can still match as being All Members / Admin Only, even if they are missing or have an new metadata policy.

Part 2 of the fix is to not return an error if the group policy set does not perfectly match a preconfigured policy set, and instead return `GroupPermissions::CustomPolicy`. This makes the SDK more robust against updates. For example this PR updates the remove member policy to default to "admin only" for "All Members". That means existing groups who already have that policy set to "allow all" will show up in new versions as a Custom Policy instead of just erroring. 

Related PRs:
- Android: https://github.com/xmtp/xmtp-android/pull/263
- Libxmtp-swift: 
- iOS:
- React Native:

### DIFF
--- a/xmtp_mls/src/groups/group_permissions.rs
+++ b/xmtp_mls/src/groups/group_permissions.rs
@@ -233,11 +233,11 @@ impl TryFrom<MetadataPolicyProto> for MetadataPolicies {
                 2 => Ok(MetadataPolicies::deny()),
                 3 => Ok(MetadataPolicies::allow_if_actor_admin()),
                 4 => Ok(MetadataPolicies::allow_if_actor_super_admin()),
-                _ => Err(PolicyError::InvalidPolicy),
+                _ => Err(PolicyError::InvalidMetadataPolicy),
             },
             Some(MetadataPolicyKindProto::AndCondition(inner)) => {
                 if inner.policies.is_empty() {
-                    return Err(PolicyError::InvalidPolicy);
+                    return Err(PolicyError::InvalidMetadataPolicy);
                 }
                 let policies = inner
                     .policies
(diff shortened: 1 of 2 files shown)


xmtp_mls/src/groups/group_permissions.rs
```diff
@@ -994,7 +1054,7 @@ pub(crate) fn policy_all_members() -> PolicySet {
     }
     PolicySet::new(
         MembershipPolicies::allow(),
-        MembershipPolicies::allow(),
+        MembershipPolicies::allow_if_actor_admin(),
```
cameronvoell, line 1057: Remove members should require admin /super admin to be aligned with Permissions XIP. Fixes https://github.com/xmtp/libxmtp/issues/857

xmtp_mls/src/groups/group_permissions.rs
```diff
@@ -1034,12 +1094,12 @@ impl PreconfiguredPolicies {
     }
 
     pub fn from_policy_set(policy_set: &PolicySet) -> Result<Self, PolicyError> {
```
cameronvoell, line 1096: This function was causing the `Invalid Policy` Errors on existing groups after adding group image url. The reason was that the existing policies no longer matched the "default policies" for "Add Members" or "Admin only" because those policy sets now had one new property in them.

neekolas (approved):

PR #868, opened 2024-06-25
make v2 streaming like v3

### LINKED ISSUE #859: Update V2 Rust Streaming to Callbacks
The V2 rust subscribe methods https://github.com/xmtp/libxmtp/blob/03b1a89e70fe5b4b016f758963116d2ac2f41fa1/bindings_ffi/src/v2.rs#L366-L372

Do not implement the callback structure like the V3 methods do making it nearly impossible to use in Android.
Can we get these streaming methods updated to perform more like the V3 stuff.

Here is how the V2 stuff is implemented in iOS. https://github.com/xmtp/xmtp-ios/blob/main/Sources/XMTPiOS/ApiClient.swift#L156-L178 Unfortunately that is very difficult/nearly impossible to do in Android.

insipx:
fixes #859

### DIFF
--- a/bindings_ffi/src/v2.rs
+++ b/bindings_ffi/src/v2.rs
@@ -293,7 +293,7 @@ pub trait FfiV2SubscriptionCallback: Send + Sync {
 pub struct FfiV2Subscription {
     tx: mpsc::Sender<FfiV2SubscribeRequest>,
     abort: AbortHandle,
-    // we require Arc<Mutex<>> here because uniffi doesn't like &mut or self in impl
+    // we require Arc<Mutex<>> here because uniffi doesn't like &mut self, or the owned version of self on exported methods 
     #[allow(clippy::type_complexity)]
     handle: Arc<Mutex<Option<JoinHandle<()>>>>,
 }
@@ -334,7 +334,7 @@ impl FfiV2Subscription {
 }
 
 impl FfiV2Subscription {
-    pub async fn subscribe(
+    async fn subscribe(
         mut subscription: GrpcMutableSubscription,
         callback: Box<dyn FfiV2SubscriptionCallback>,
     ) -> Self {
--- a/bindings_ffi/src/v2.rs
+++ b/bindings_ffi/src/v2.rs
@@ -8,7 +8,8 @@ use xmtp_proto::api_client::{
 use xmtp_proto::xmtp::message_api::v1::IndexCursor;
 use xmtp_v2::{hashes, k256_helper};
 
-use xmtp_api_grpc::grpc_api_helper::Client as GrpcClient;
+use tokio::{sync::mpsc, task::JoinHandle};
+use xmtp_api_grpc::grpc_api_helper::{Client as GrpcClient, GrpcMutableSubscription};
 use xmtp_proto::xmtp::message_api::v1::{
     cursor::Cursor as InnerCursor, BatchQueryRequest, Cursor, Envelope, PublishRequest,
     QueryRequest, SortDirection,
@@ -67,7 +68,7 @@ pub struct FfiV2QueryRequest {
     paging_info: Option<FfiPagingInfo>,
 }
 
-#[derive(uniffi::Record)]
+#[derive(uniffi::Record, Debug)]
 pub struct FfiEnvelope {
     content_topic: String,
     timestamp_ns: u64,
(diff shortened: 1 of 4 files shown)


bindings_ffi/src/v2.rs
```diff
@@ -279,46 +283,85 @@ impl From<FfiV2SubscribeRequest> for SubscribeRequest {
...
+    ) -> Result<Self, GenericError> {
+        let (tx, mut rx): (_, mpsc::Receiver<FfiV2SubscribeRequest>) = mpsc::channel(10);
+
+        let handle = tokio::spawn(async move {
+            loop {
+                tokio::select! {
+                    item = subscription.next() => {
+                        match item {
+                            Some(Ok(envelope)) => callback.on_message(envelope.into()),
+                            Some(Err(e)) => log::error!("Stream error {}", e),
```
insipx, line 349: Unlike the previous v2 implement, we don't return errors here. it's possible to pass them into the callback but Uniffi is finicky with passing `Result` types into the callback. V3 streaming also just logs the errors and moves on, so I did the same here

bindings_ffi/src/v2.rs
```diff
@@ -279,46 +283,90 @@ impl From<FfiV2SubscribeRequest> for SubscribeRequest {
...
+    /// Check if the subscription is closed
+    pub fn is_closed(&self) -> bool {
+        self.abort.is_finished()
+    }
+
+    /// Update subscription with new topics
     pub async fn update(&self, req: FfiV2SubscribeRequest) -> Result<(), GenericError> {
-        let mut sub = self.inner_subscription.lock().await;
-        sub.update(req.into()).await?;
+        self.tx.send(req).await.map_err(|_| GenericError::Generic {
```
richardhuaaa, line 329: What is the pro/con of sending it through via mpsc, compared to calling it on the subscription directly like in the previous implementation?

richardhuaaa (approved):
LGTM!

bindings_ffi/src/v2.rs
```diff
@@ -279,46 +283,90 @@ impl From<FfiV2SubscribeRequest> for SubscribeRequest {
...
-        Self {
-            inner_subscription: Arc::new(futures::lock::Mutex::new(subscription)),
-        }
-    }
+/// Subscription to a stream of V2 Messages
+#[derive(uniffi::Object)]
+pub struct FfiV2Subscription {
+    tx: mpsc::Sender<FfiV2SubscribeRequest>,
+    abort: AbortHandle,
+    // we require Arc<Mutex<>> here because uniffi doesn't like &mut or self in impl
```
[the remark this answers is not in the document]
insipx: It didn't like just consuming `self` either. Though maybe I misinterpreted it the error message. It was talking about self not being copy, and originated from the macro. Using `self` would be odd anyway b/c would need to re-assign the subscription _after_ closing to check if its actually closed, which would be confusing

bindings_ffi/src/v2.rs
insipx: That's a good question, other than the usual edge-casey footguns with mpsc,

I think the biggest downside is the extra memory overhead for the channel itself, which in this case is bounded to 10 updates. Since it is bounded, if `update` is called continually like in a loop we might lose some updates after reaching the limit on the channel. This shouldn't really be an issue, though, since the caller should be putting the topics in a `Vec<>` to the channel `FfiV2SubscribeRequest` type anyway. The limit is needed to avoid a infinitely growing channel.

richardhuaaa: That makes sense -  I guess more specifically, what would be the drawback of storing the subscription directly on FFIV2Subscription, and directly updating it here as in the old implementation, versus the new approach of storing the tx and sending a message over it?

insipx: By storing the subscription locally in the task, we can avoid using a `Mutex` to access the subscription. This has a few benefits, and it more generally/lightly follows the `actor` model of concurrency rather than locking:

- It's easier to reason about what's happening to the subscription, because all the events are handled in one event loop the task is spawning
- we avoid any footguns we might run into with `Mutex's`, which lets us avoid a class of errors that occurs with `Mutexes`, which makes `update` more difficult to misuse outside of libxmtp. We still use `Mutexes` a bunch which is fine, esp considering this is with uniffi, but more mutexes means more potential for deadlocks and the such. Mutexes are more performant, though, but I doubt using a channel vs mutex will make a huge diff in that direction here
- if we want to modify the subscription in other ways, can extend it by just adding new events to the loop

I guess its a bit arbitrary, when writing streams I do default to this strategy rather than `Mutex`'s, unless absolutely necessary for performance or to satisfy the borrow checker w.r.t uniffi or the like

PR #870, opened 2024-06-25
Android binding release improvements

nplasterer:
Make better tooling for releasing android bindings
Releases now triggered here https://github.com/xmtp/libxmtp/actions/runs/9674675295

And the jniBindings can be found in the releases. https://github.com/xmtp/libxmtp/releases/tag/kotlin-bindings-ddb96dd

### DIFF
--- a/.github/workflows/release-kotlin-bindings.yml
+++ b/.github/workflows/release-kotlin-bindings.yml
@@ -69,20 +69,10 @@ jobs:
           retention-days: 1
       
       - name: Prepare JNI libs
+        working-directory: bindings_ffi
         run: |
-          LIBRARY_NAME="libxmtpv3"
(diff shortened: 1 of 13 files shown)


.github/workflows/release-kotlin-bindings.yml
```diff
@@ -0,0 +1,127 @@
+name: Release Kotlin Bindings
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - np/update-android-bindings
```
neekolas, line 7: We can do this on main, right

nplasterer: Totally. Just added this to trigger it for testing. Will remove. Thought I had it working but just made a few more tweaks.

neekolas (approved):

PR #872, opened 2024-06-26
Even more locking fixes

neekolas:
## tl;dr

- Disables the `backtrace` default feature on OpenMLS which was causing serious performance impacts
- Implements `RetyableError` in more places to cover OpenMLS errors that may wrap DB errors
- More advanced backoff strategy for retries. Now starts at 50ms and multiplies by 3 on each attempt


nplasterer (approved):

nplasterer:
I'm about to build some bindings for the V2 streaming stuff. But will wait till this lands.

alexrisch (approved):

PR #873, opened 2024-06-27
Adds update permissions intent and validation

cameronvoell:
Finishes #664 

protos pr: https://github.com/xmtp/proto/pull/192


bindings_ffi/src/mls.rs
```diff
@@ -1892,4 +2068,161 @@ mod tests {
...
+            remove_admin_policy: FfiPermissionPolicy::SuperAdmin,
+            update_group_name_policy: FfiPermissionPolicy::Allow,
+            update_group_description_policy: FfiPermissionPolicy::Allow,
+            update_group_image_url_square_policy: FfiPermissionPolicy::Allow,
+        };
+        assert_eq!(alix_permission_policy_set, expected_permission_policy_set);
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 5)]
+    async fn test_permissions_updates() {
```
cameronvoell, line 2134: See new test in bindings on how we can now read all updatable permissions and update each of them.

xmtp_mls/src/groups/group_mutable_metadata.rs
```diff
@@ -45,7 +45,7 @@ pub enum MetadataField {
 }
 
 impl MetadataField {
-    fn as_str(self) -> &'static str {
+    pub fn as_str(&self) -> &'static str {
```
insipx, line 48:
```suggestion
    pub const fn as_str(&self) -> &'static str {
```

might as well make it `const` as well

insipx (approved):

xmtp_mls/src/groups/group_mutable_metadata.rs
cameronvoell: updated 519e0c30e0a7a4df1fdeea91c04dcba0422e3032

PR #888, opened 2024-07-02
Adding pinned frame metadata field

cameronvoell:
Adds new metadata field `GroupPinnedFrameUrl`

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -902,10 +902,7 @@ impl FfiGroup {
         Ok(group_description)
     }
 
-    pub async fn update_pinned_frame(
-        &self,
-        pinned_frame: String,
-    ) -> Result<(), GenericError> {
+    pub async fn update_pinned_frame(&self, pinned_frame: String) -> Result<(), GenericError> {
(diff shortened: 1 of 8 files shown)


alexrisch (approved):

bindings_ffi/src/mls.rs
```diff
@@ -647,6 +651,7 @@ impl FfiCreateGroupOptions {
             name: self.group_name,
             image_url_square: self.group_image_url_square,
             description: self.group_description,
+            pinned_frame: self.pinned_frame,
```
nplasterer, line 654: Should this be pinned frame url to be more explicit it's a url?

nplasterer (approved):

bindings_ffi/src/mls.rs
cameronvoell: probably right I'll update that and fix these node bindings tests before merging 👌

cameronvoell: updated https://github.com/xmtp/libxmtp/pull/888/commits/656c2c69ae27b20a7f07f13e0e69d9c13563bd8a

Issue #693, opened 2024-04-30
Tracking: 1 to 1 DMs MLS (Mobile)


nplasterer:
### MVP
- [ ] Permissions - Ability to create a group chat that only has two members and cannot have additional members added (Medium)
	- [ ] New permission type to identify a 1to1 conversations
	- [ ] Way to handle adding just one person to the group after create (possibly and immutable list called allowed_members)
	- [ ] Validation
		- [ ] 1to1 Group should not be able to be named
		- [ ] You cannot leave a 1to1 Group
		- [ ] Should not see a group membership change codec on first send
- [ ] Duel Sending - Can still receive messages from non V3 clients (Medium)
  - [ ] Welcome messages for 1to1 should be hidden in libxmtp for clients that don't support 1to1 yet
  - [ ] Resolver logic for sending messages
		- [ ] Resolve the `wallet_address` to an `inbox_id` 
	    - [ ] If it can’t be resolved, abort the process. Dual-send isn’t possible here.
		- [ ] Check if there is an existing 1:1 convo with that `inbox_id`
	    - [ ] If not, create one and invite the other member
		- [ ] Send the message
  - [ ] Validation
	  - [ ] A client that supports V3 can still recieve a V2 only message
	  - [ ] A client that supports only V2 can still get messages sent from a V3 client
	  - [ ] A client that supports V3 can send a V2 and V3 message
	  - [ ] A client that supports V3 does not show double messages
- [ ] Conversation History - New installations only see one 1to1 conversation (NEEDS SCOPING - Large)

### Follow ups
- [ ] V2/V1 Removal
	- [ ] Period of read only
	- [ ] Clear datadog metric tracking
	- [ ]	Import of all V2 messages into V3 database
		- [ ] Dedupe on message unique identifier

bwcDvorak:
Note to self: hold off on converting any of these into GH issues/importing into ZH.

richardhuaaa:
I think we also need a replacement for personal preferences, otherwise there's no point in starting the 1:1 migration earlier than that? And I think that's a fairly substantial work item. What do you think @neekolas?

PR #901, opened 2024-07-14
Adds functions for creating a DM group

cameronvoell:
Part of #693 

DM Permissions setup:
1. We create a group with new immutable DM Members fields (two unordered inboxes)
2. For DM groups, you can always add a member with one of the two inbox ids
3. adding anyone else or removing is set to deny
4. there are no admins or super admins
5. both members can update all mutable metadata

Before Merging:
- [x] set `dev/gen_protos.sh` back to `main` branch


nplasterer (approved):
Looks good to start the long running dm branch. 👍

PR #902, opened 2024-07-15
Message Failed State

### LINKED ISSUE #671: Message failed state
In the follow up PR we should do the following

1. Mark the optimistic message as FAILED whenever its intent is marked as failed (set_group_intent_error)
2. Make sure that every possible [error case in this method](https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/src/groups/sync.rs#L156) results in the corresponding intent either being marked as TO_PUBLISH or FAILED
3. At some point in the future revisit whether there is a possibility of an infinite loop of intent retries. Maybe we should call increment_intent_publish_attempt_count() wherever we reset an intent to TO_PUBLISH, and just set a really high retry limit, like 10?

For (2), maybe the most fool-proof way is to mark all MessageProcessingErrors as either retryable or non-retryable, and have the caller of the method catch all errors and mark the intent as TO_PUBLISH for retryable errors and FAILED for non-retryable errors

_Originally posted by @nplasterer in https://github.com/xmtp/libxmtp/issues/669#issuecomment-2062578221_

insipx:
fixes #671 

> Mark the optimistic message as FAILED whenever its intent is marked as failed (set_group_intent_error)

added `message_id` function to `StoredGroupIntent` that calculates a `message_id` if the intent is of type `SendMessage`. On intent fail we can access the message and set it to failed in `publish_intents`.

> Make sure that every possible [error case in this method](https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/src/groups/sync.rs#L156) results in the corresponding intent either being marked as TO_PUBLISH or FAILED

Verified that the intent is only returned as failed from `publish_intents`

> At some point in the future revisit whether there is a possibility of an infinite loop of intent retries. Maybe we should call increment_intent_publish_attempt_count() wherever we reset an intent to TO_PUBLISH, and just set a really high retry limit, like 10?

Intent retries are limited by 3 in `sync_until_intent_resolved` and also to 3 in `publish_intents`.

### DIFF
--- a/xmtp_mls/src/storage/encrypted_store/group_intent.rs
+++ b/xmtp_mls/src/storage/encrypted_store/group_intent.rs
@@ -7,14 +7,15 @@ use diesel::{
     sql_types::Integer,
     sqlite::Sqlite,
 };
+use prost::Message;
 
 use super::{
     db_connection::DbConnection,
     group,
     schema::{group_intents, group_intents::dsl},
 };
-use crate::{impl_fetch, impl_store, storage::StorageError, Delete};
-
+use crate::{impl_fetch, impl_store, storage::StorageError, Delete, groups::intents::SendMessageIntentData, utils::id::calculate_message_id};
+use xmtp_proto::xmtp::mls::message_contents::{plaintext_envelope::{Content, V1}, PlaintextEnvelope};
 pub type ID = i32;
 
 #[repr(i32)]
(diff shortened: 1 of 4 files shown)


xmtp_mls/src/storage/encrypted_store/group_intent.rs
```diff
@@ -67,6 +77,37 @@ pub struct StoredGroupIntent {
...
 }
 
+impl StoredGroupIntent {
+    /// Calculate the message id for this intent.
+    ///
+    ///
+    /// # Returns
+    /// Returns [`Option::None`] if [`StoredGroupIntent`] is not [`IntentKind::SendMessage`] or if
+    /// an error occurs during decoding of intent data for [`IntentKind::SendMessage`].
+    pub fn message_id(&self) -> Result<Option<Vec<u8>>, IntentError> {
```
neekolas, line 87: My only worry about this refactor is that people are going to use it _and_ decode the proto somewhere else separately to get at the other fields, so we're going to end up doing the same work twice.

`message_id` sounds like it would be a cheap little operation, but it's actually pretty expensive.

neekolas (approved):

xmtp_mls/src/storage/encrypted_store/group_intent.rs
insipx: Added a comment explaining how that is a bad idea, we should watch out for that but I also am not sure when we would access the contents of a PlainText envelope outside of publish success/failure. It could be worth storing the message_id separately on `StoredGroupIntent` as an `Option` which would avoid the problem altogether

PR #916, opened 2024-07-20
Sqlcipher bundling

nplasterer:
WIP


nplasterer:
It appears the sqlcipher is correctly being linked but react native is conflicting and removing it. Fixes this upstream in react native instead

PR #917, opened 2024-07-22
Ensure message send succeeds even when out of sync

nplasterer:
We have configured the `max_past_epochs` value to 3, which means that we keep around message encryption keys for 3 epochs before deleting them. This means that if we are 3 commits behind when we send a message, nobody else will be able to decrypt it, because they process everything sequentially, and they'll have already deleted their encryption keys by the time they see it.

The fix is as follows:
- When pulling down messages from a group, if we see a message we previously published, we check that the message is no more than 3 epochs behind. If the check passes, the message send intent is updated to COMMITTED, otherwise it's reset to TO_PUBLISH so that the message can be sent again.
- After sending a message, we should go ahead and pull down the messages afterwards, to make sure the message send succeeded (and retry via intents otherwise).

This has the following implications:
1. It's not required to sync the group before sending a message
2. Confirming that a message sent successfully (i.e. waiting for `send_message()` to complete) is slower - there is an extra round trip to pull down the messages afterwards (+more if the message needs to be retried)

My justification for the slower message send is that we've already set up optimistic message sends, with separate prepare and publish steps. In the event that multiple optimistic message sends happen back-to-back, you can call a single publish at the end. Perhaps we can recommend using optimistic message sends, with debounced publishes, in the docs somewhere.

\- Rich

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -615,7 +615,7 @@ impl FfiConversations {
         Ok(convo_list)
     }
 
-    pub fn stream(&self, callback: Box<dyn FfiConversationCallback>) -> FfiStreamCloser {
+    pub async fn stream(&self, callback: Box<dyn FfiConversationCallback>) -> FfiStreamCloser {
         let client = self.inner_client.clone();
         let handle =
             RustXmtpClient::stream_conversations_with_callback(client.clone(), move |convo| {
@@ -629,7 +629,7 @@ impl FfiConversations {
         FfiStreamCloser::new(handle)
     }
 
-    pub fn stream_all_messages(
+    pub async fn stream_all_messages(
         &self,
         message_callback: Box<dyn FfiMessageCallback>,
     ) -> FfiStreamCloser {
@@ -1117,7 +1117,7 @@ impl FfiGroup {
         Ok(())
     }
 
-    pub fn stream(&self, message_callback: Box<dyn FfiMessageCallback>) -> FfiStreamCloser {
+    pub async fn stream(&self, message_callback: Box<dyn FfiMessageCallback>) -> FfiStreamCloser {
         let inner_client = Arc::clone(&self.inner_client);
         let handle = MlsGroup::stream_with_callback(
             inner_client,
@@ -1266,7 +1266,7 @@ impl FfiStreamCloser {
     }
 }
 
-#[uniffi::export]
+#[uniffi::export(async_runtime = "tokio")]
 impl FfiStreamCloser {
     /// Signal the stream to end
     /// Does not wait for the stream to end.
@@ -2010,7 +2010,8 @@ mod tests {
         let message_callbacks = RustStreamCallback::default();
         let stream_messages = bo
             .conversations()
-            .stream_all_messages(Box::new(message_callbacks.clone()));
+            .stream_all_messages(Box::new(message_callbacks.clone()))
+            .await;
         stream_messages.wait_for_ready().await;
 
         // Create group and send first message
@@ -2067,7 +2068,8 @@ mod tests {
         let message_callbacks = RustStreamCallback::default();
         let stream_messages = bo
             .conversations()
-            .stream_all_messages(Box::new(message_callbacks.clone()));
+            .stream_all_messages(Box::new(message_callbacks.clone()))
+            .await;
         stream_messages.wait_for_ready().await;
 
         let first_msg_check = 2;
@@ -2140,7 +2142,8 @@ mod tests {
 
         let stream = bola
             .conversations()
-            .stream(Box::new(stream_callback.clone()));
+            .stream(Box::new(stream_callback.clone()))
+            .await;
 
         amal.conversations()
             .create_group(
@@ -2188,7 +2191,8 @@ mod tests {
 
         let stream = caro
             .conversations()
-            .stream_all_messages(Box::new(stream_callback.clone()));
+            .stream_all_messages(Box::new(stream_callback.clone()))
+            .await;
         stream.wait_for_ready().await;
 
         alix_group.send("first".as_bytes().to_vec()).await.unwrap();
@@ -2234,7 +2238,7 @@ mod tests {
         let bola_group = bola.group(amal_group.group_id.clone()).unwrap();
 
         let stream_callback = RustStreamCallback::default();
-        let stream_closer = bola_group.stream(Box::new(stream_callback.clone()));
+        let stream_closer = bola_group.stream(Box::new(stream_callback.clone())).await;
 
         stream_closer.wait_for_ready().await;
 
@@ -2273,7 +2277,8 @@ mod tests {
         let stream_callback = RustStreamCallback::default();
         let stream_closer = bola
             .conversations()
-            .stream_all_messages(Box::new(stream_callback.clone()));
+            .stream_all_messages(Box::new(stream_callback.clone()))
+            .await;
         stream_closer.wait_for_ready().await;
 
         amal_group.send(b"hello1".to_vec()).await.unwrap();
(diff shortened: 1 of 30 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -2056,6 +2056,73 @@ mod tests {
...
+        ]).await.unwrap();
+        alix_group.add_members(vec![
+            eri.account_address.clone(),
+            frankie.account_address.clone(),
+        ]).await.unwrap();
+
+        bo_group.send("bo3".as_bytes().to_vec()).await.unwrap();
+        alix_group.send("alix3".as_bytes().to_vec()).await.unwrap();
+        bo_group.send("bo4".as_bytes().to_vec()).await.unwrap();
+        bo_group.send("bo5".as_bytes().to_vec()).await.unwrap();
```
nplasterer, line 2110: You can just keep adding more and more bo messages and alix will still be 8.

nplasterer:
I can't review since I opened the PR but the logic makes sense to me. 👍

xmtp_mls/src/groups/mod.rs
```diff
@@ -459,7 +461,8 @@ impl MlsGroup {
         let update_interval = Some(5_000_000);
         self.maybe_update_installations(conn.clone(), update_interval, client)
             .await?;
-        self.publish_intents(conn, client).await?;
+        self.publish_intents(conn.clone(), client).await?;
+        self.sync_until_all_intents_resolved(conn, client).await?;
```
neekolas, line 465: This is a lot of overhead to handle a pretty fringe-y edge case. Will have impacts on latency (now we have to wait for at least 2 sequential API requests to complete before returning on every message send) and on rate-limiting.

Given that most clients are going to receive the message from a stream anyways I'm not sure it's necessary.

xmtp_mls/src/groups/mod.rs
```diff
@@ -440,10 +440,12 @@ impl MlsGroup {
         let message_id = self.prepare_message(message, &conn);
 
         // Skipping a full sync here and instead just firing and forgetting
-        if let Err(err) = self.publish_intents(conn, client).await {
+        if let Err(err) = self.publish_intents(conn.clone(), client).await {
```
neekolas, line 443: Why the new clone here? I don't see anything changed about publish_intents

xmtp_mls/src/groups/sync.rs
```diff
@@ -133,6 +134,28 @@ impl MlsGroup {
         Ok(())
     }
 
+    pub(super) async fn sync_until_all_intents_resolved<ApiClient>(
```
neekolas, line 137: Might be more accurate to name this `sync_until_last_jntent_resolved`. Earlier intents can still fail and this will succeed.

xmtp_mls/src/groups/mod.rs
richardhuaaa: The last paragraph of the PR description has a justification for this - WDYT?

> Will have impacts on latency 
Not to perceived latency if optimistic message sends are used. If non-optimistic sends are used, i.e. confirmation of message send is desired, I'm not sure if it's honest to return success until we have synced.

> on rate-limiting
I think it'll be negligible, as we have significantly smaller publish volume compared to reads

> Given that most clients are going to receive the message from a stream anyways I'm not sure it's necessary.
Receiving it from a stream won't cause the client to retry

xmtp_mls/src/groups/mod.rs
richardhuaaa: `publish_intents()` consumes the connection, and we are using it again below in `sync_until_all_intents_resolved()`. This seems to be a common pattern, see `maybe_update_installations()` above, where the conn is cloned

xmtp_mls/src/groups/mod.rs
neekolas: > Not to perceived latency if optimistic message sends are used. If non-optimistic sends are used, i.e. confirmation of message send is desired, I'm not sure if it's honest to return success until we have synced.

For an inbox app, separating prepare and send makes sense. But this is also going to effect bulk senders/bots who are almost certainly not using optimistic sends.

> Receiving it from a stream won't cause the client to retry

That's fair. Updated my comment.


Maybe the better question is: if we are going to pay the price for adding a read to every write, couldn't we just sync first and prevent this problem in almost all cases? It would take a hell of a race condition to get 3 commits in between the sync and the send.

richardhuaaa: Also keep in mind that perceived latency on the receiver side is not affected - the publish is going to take the same amount of time regardless

xmtp_mls/src/groups/sync.rs
```diff
@@ -281,6 +337,16 @@ impl MlsGroup {
...
             }
             IntentKind::SendMessage => {
+                if !Self::is_valid_epoch(
+                    self.context.inbox_id(),
+                    intent.id,
+                    group_epoch,
+                    message_epoch,
+                    3, // max_past_epochs, TODO: expose from OpenMLS MlsGroup
+                ) {
+                    conn.set_group_intent_to_publish(intent.id)?;
```
neekolas, line 347: I just realized there is another case we should think about:

What if the recipient is streaming messages (or receiving push notifs) and is also out of date with the current epoch. In that case the original message will decrypt successfully AND they will eventually receive a duplicate when they next sync. I _think_ the message ID will be the same between these retries, but it's probably worth writing a test for that.

xmtp_mls/src/groups/mod.rs
neekolas: Ahh, I see it now

xmtp_mls/src/groups/mod.rs
richardhuaaa: With the next version of the protocol, there's an opportunity to have this validation done server-side (reject the message if a more recent commit has been seen), perhaps we could remove this extra round-trip when that happens

richardhuaaa: Sorry didn't see your last comment until just now! Funny given we're talking about stale local state.

>  this is also going to effect bulk senders/bots who are almost certainly not using optimistic sends

Are bulk senders sending multiple messages back-to-back within the same conversation, or single messages to many conversations? Is it the case that they need to be up-to-date with their groups regardless?

>  if we are going to pay the price for adding a read to every write, couldn't we just sync first and prevent this problem in almost all cases? It would take a hell of a race condition to get 3 commits in between the sync and the send

We could. I'm not sure if it's that much of a race condition in large groups, and it seems we've already run into it in our own testing. Also, integrating apps are not guaranteed to have a subscription running, and we're not guaranteed that a running subscription isn't broken or has a network delay etc.

The problem is even if we accept that it's rare, we have no recovery mechanism otherwise. Those messages are permanently lost. We could talk about a different way to recover, but it might involve a large refactor. I'm inclined to wait until our work on the new decentralized backend ships, when we will have some degree of server-side validation.

neekolas: > it's not clear that getting client apps to sync every time they open a thread, or on some interval, will create a lower overall volume than syncing every time a send happens.

It might be the same volume, but I think that has to happen anyways to catch missing messages. So this would be purely additional

neekolas (approved):

richardhuaaa:
Summary of offline discussion: there's a tradeoff between reliability and performance here, for now we choose reliability. It's a reversible decision - removing the extra round-trip is a one line change. I think the next iteration of the server will also help:
- Verifying epochs within the same round trip as the publish
- Reliable subscriptions so that further syncs are unnecessary

I'd also like to add a test case for the scenario Nick mentioned, but am running into some problems - going to land this first to unblock release, and put that test case up as a separate PR.

richardhuaaa:
The build is broken, but it looks like there is a fix (updating the time crate) in https://github.com/xmtp/libxmtp/pull/923 - will land regardless

PR #921, opened 2024-07-24
fix: drop the pin to openssl-sys which was building against 1.1.x.

tuddman:
## Summary

Experiment to find out if dynamically linking to OpenSSL `1.1.1` was itself the issue.

As rust-openssl now builds against OpenSSL `3.x.x`  https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/Cargo.toml#L28

### DIFF
--- a/bindings_ffi/Cargo.lock
+++ b/bindings_ffi/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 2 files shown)


tuddman:
Closing.  As I suspected, without an explicit commit to reference that clearly solves the underlying linking issue, then pinning the `openssl-sys` version (or not) has no effect.

PR #927, opened 2024-07-29
Make `NoChanges` NoOp

insipx:
This converts the `NoChanges` err to NoOp in two places. It leaves the error on `get_membership_update_intent`.


neekolas (approved):
Seems like a sensible way of handling these cases without erroring

Issue #900, opened 2024-07-11
Removed members may still be able to send messages


richardhuaaa:
I suspect removed members can still send messages on a group, because all other members accept messages from previous epochs. The way we can test this is to write a unit test that removes a member, send a message to the group from the same member, and verify that the remover does not receive the message. We also need to verify that if the unit test passes, it's because there is logic on the receiver that is rejecting the message, not logic on the sender (which can be circumvented).

nplasterer:
Looks like this test has been written.

PR #929, opened 2024-07-30
removed members should not be able to send messages

tuddman:
## Summary

From https://github.com/xmtp/libxmtp/issues/900, a member who has been from from a group should NOT be able to send messages to that group - after they have been removed from that group.   The test in this patch confirms they (still) can.

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -1774,7 +1774,7 @@ mod tests {
             .any(|m| m.inbox_id == charlie.inbox_id()));
 
         group.sync(&amal).await.expect("sync failed");
-        
+
         group
             .send_message(b"hello", &bola)
             .await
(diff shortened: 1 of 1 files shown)


xmtp_mls/src/groups/mod.rs
```diff
@@ -1737,6 +1737,62 @@ mod tests {
...
+        group.sync(&amal).await.expect("sync failed");
+
+        let amal_messages = group
+            .find_messages(Some(GroupMessageKind::Application), None, None, None, None)
+            .unwrap()
+            .into_iter()
+            .collect::<Vec<StoredGroupMessage>>();
+
+        // FIXME:st this is passing ONLY because the message IS being sent to the group
+        assert_eq!(amal_messages.len(), 1);
```
insipx, line 1793: It would be nice to confirm that the message is the message we expect, `hello`. I think we also send a message on group invite/join?

tuddman: agree. I was logging the decrypted message bytes, and it is indeed the case that `hello` is coming through.  I can explicitly test for this, though 👍

insipx (approved):

PR #930, opened 2024-07-31
Refactor message history sync, add to CLI

### LINKED ISSUE #877: End to End App test of Message History

tuddman:
### Refactored message history sync

* Exposed `message_history` module
* Exported official message history sync URLs
* Changed storage of history message content from delimited strings to serialized structs
* Removed use of `bundle_hash` and `sign_key` when managing history bundles
* Removed automatic reply and bundle processing during sync
* Always use the client's `history_sync_url` for syncing
* Added more message history errors
* Updated all message history functions to use `MessageHistoryError` when appropriate
* Added `get_sync_group` function to get sync group and its ID
* Refactored `allow_history_sync` to use existing sync group or create a new one
* Refactored `send_history_request`
  * Added guard to only send a history request if there isn't one already pending
  * Updated return value to include both `request_id` and `pin_code`
* Refactored `send_history_reply`
  * Added guard to only send a history reply when there's a pending request
* Added `get_pending_history_request` function that returns the `request_id` and `pin_code` of a pending history request
* Added `reply_to_history_request` function that replies to a pending history request
* Added `get_latest_history_reply` function that returns a history reply if it's the last message in a sync group
* Added `process_history_reply` function that downloads a history bundle using the latest history reply message

#### Notes

* The guards added in `send_history_request` and `send_history_reply` are currently very basic and don't account for spamming (no time based checks). Since all installations use the same sync group, the sync group will always contain a request followed by a reply. There cannot be multiple requests or replies in a row.
* The `verify_pin` function is not currently being used. I'm not sure what the use case is.

#### Future enhancements

* Allow for history bundles with constraints (e.g. `startTime`) to limit size and processing time

#### TODO

- [x] Add tests for `get_pending_history_request`
- [x] Add tests for `get_latest_history_reply`
- [x] Add tests for `reply_to_history_request`

### Enable message history sync-ing from the `examples/cli` mini-app.

**Note**: This will only work if you provide the **same** `--seed-phrase` when creating __both__ client instances:

```sh
## tab 1
./xli.sh --db user1_a.db3 register --seed-phrase "celery home axis version spike fiction vehicle actual hair absurd diesel body"

## tab 2
./xli.sh --db user1_b.db3 register --seed-phrase "celery home axis version spike fiction vehicle actual hair absurd diesel body"
```

#### Request history sync
```sh
## in tab 1, for consistency
./xli.sh --db user1_a.db3 request-history-sync
```

#### Send history sync reply

```sh
## in tab 2
./xli.sh --db user1_b.db3 reply-to-history-sync-request
```

#### Process history sync reply

This will download and process a history bundle.

```sh
## in tab 1, for consistency
./xli.sh --db user1_a.db3 process-history-sync-reply
```

#### List history sync messages

This will display all messages in the history sync group.

```sh
./xli.sh --db user1_a.db3 list-history-sync-messages
```

#### Inspect

```sh
(after request-sync-history has been sent...)

## load sqlite3
sqlite3 user1_b.db3
> select hex(id) from groups;
B940E26CBD7FB35BD9041D06B46D1F0E

## in another tab
sqlite3 user1_a.db3
> select hex(id) from groups;
B940E26CBD7FB35BD9041D06B46D1F0E  <- user1_a should be joined into the sync group
```

Closes #877


tuddman:
@neekolas @richardhuaaa 

Tying up lingering loose ends here...

There's a db lock issue when the responder is attempting to gather the groups + messages (from the local db) to send to the requester: 
```
xmtp_mls::groups::subscriptions retrying function that failed with error=receive error: generic:could not send history reply: storage error: Diesel result error: database is locked
```

I'm confident one or both of you can solve it in short order.   

There's also the matter of standardizing on the message-history "api contract" i.e. make the endpoints strictly REST-ful to send to e.g. `POST /bundles` and for the other side to fetch the bundle from e.g. `GET /bundles/<uuid>`

Fix those and this whole feature-set should just work™

nplasterer (approved):

nplasterer:
Looks like just a lint error on the usage of an `unwrap`

xmtp_mls/src/groups/message_history.rs
```diff
@@ -78,9 +105,31 @@ impl<ApiClient> Client<ApiClient>
...
+    pub fn get_sync_group(&self) -> Result<(Vec<u8>, MlsGroup), GroupError> {
+        let conn = self.store().conn()?;
+        let sync_group_id = conn
+            .find_sync_groups()?
+            .pop()
+            .ok_or(GroupError::GroupNotFound)?
+            .id;
+        let sync_group = self.group(sync_group_id.clone())?;
+
+        Ok((sync_group_id, sync_group))
```
neekolas, line 117: Do we actually need two return values here? Can't we just get the ID off of the sync group?

xmtp_mls/src/groups/message_history.rs
```diff
@@ -78,9 +105,31 @@ impl<ApiClient> Client<ApiClient>
...
+    }
+
     pub async fn allow_history_sync(&self) -> Result<(), GroupError> {
```
neekolas, line 120: It would be helpful to have a comment explaining what this function actually is allowing. Or maybe renaming it.

All the function does is get or create a sync group, and syncs it. But I'm assuming this is opting me in to respond to message history requests, right?

xmtp_mls/src/groups/message_history.rs
```diff
@@ -178,37 +272,158 @@ where
...
     }
 
-    pub(crate) fn provide_pin(&self, pin_challenge: &str) -> Result<(), GroupError> {
-        let conn = self.store().conn()?;
-        let sync_group_id = conn
-            .find_sync_groups()?
-            .pop()
-            .ok_or(GroupError::GroupNotFound)?
-            .id;
+    pub async fn get_pending_history_request(
```
neekolas, line 275: I'm OK with having this not support there being more than one pending history request for now, but we should probably think through how we can support that down the line.

xmtp_mls/src/groups/message_history.rs
rygine: good point, i think `enable_history_sync` is a better name. we could also just get rid of this function altogether and incorporate the creation of a sync group in the `get_sync_group` fn.

xmtp_mls/src/groups/message_history.rs
rygine: very true. i'll make it just return the `MlsGroup`.

xmtp_mls/src/groups/message_history.rs
```diff
@@ -129,37 +198,62 @@ where
...
-            contents
-                .signing_key
-                .as_ref()
-                .ok_or(MessageProcessingError::InvalidPayload)?,
-            contents.bundle_hash.as_slice()
-        )
-        .into_bytes();
+        let content = MessageHistoryContent::Reply(contents.clone());
+        let content_bytes = serde_json::to_vec(&content)
+            .map_err(|e| MessageHistoryError::Generic(format!("{e}")))?;
```
insipx, line 256: instead of `map_err(/* ... */) and a Generic history message, you can add a variant to MessageHistoryError:


```
#[derive(Debug, Error)]
enum MessageHistoryError {
    #[error("Serialization/Deserialization Error {0}")]
    Serde(#[from] serde_json::Error)
}
```

then you can take advantage of an implicit `From` implementation on the `?` operator to clean up this line a bit:

```rust
let content_bytes = serde_json::to_vec(&content)?;
```


If an error occurs, this will auto-convert the error to `MessageHistory::Serde` variant

xmtp_mls/src/groups/message_history.rs
```diff
@@ -178,37 +270,158 @@ where
...
+                let message_history_content =
+                    serde_json::from_slice::<MessageHistoryContent>(&msg.decrypted_message_bytes);
+                match message_history_content {
+                    // if the last message is a request, return its request ID and pin code
+                    Ok(MessageHistoryContent::Request(request)) => {
+                        Some((request.request_id, request.pin_code))
+                    }
+                    _ => None,
+                }
+            }
```
neekolas, line 300: We should probably log something in the case that deserialization fails here

xmtp_mls/src/groups/message_history.rs
```diff
@@ -178,37 +270,158 @@ where
...
+
+        Ok(history_request)
+    }
+
+    pub async fn reply_to_history_request(
+        &self,
+    ) -> Result<MessageHistoryReply, MessageHistoryError> {
+        let pending_request = self.get_pending_history_request().await?;
+
+        if let Some((request_id, _)) = pending_request {
```
neekolas, line 312: How do i know which request I'm replying to? What happens if a recent request just snuck in

xmtp_mls/src/groups/message_history.rs
```diff
@@ -178,37 +270,158 @@ where
...
             None,
+        )?;
+        let last_message = messages.last();
+        let history_request: Option<(String, String)> = match last_message {
+            Some(msg) => {
+                let message_history_content =
+                    serde_json::from_slice::<MessageHistoryContent>(&msg.decrypted_message_bytes);
+                match message_history_content {
+                    // if the last message is a request, return its request ID and pin code
+                    Ok(MessageHistoryContent::Request(request)) => {
```
insipx, line 295: if you add the serde error, you can use the `?` here as well

xmtp_mls/src/groups/message_history.rs
```diff
@@ -178,37 +270,158 @@ where
...
+            None,
+            None,
             None,
             None,
+        )?;
+        let last_message = messages.last();
+        let history_request: Option<(String, String)> = match last_message {
+            Some(msg) => {
+                let message_history_content =
+                    serde_json::from_slice::<MessageHistoryContent>(&msg.decrypted_message_bytes);
```
insipx, line 292: if you add the serde error, you can use the ? here as well

xmtp_mls/src/groups/message_history.rs
```diff
@@ -178,37 +270,158 @@ where
...
-            sync_group_id,
+        let messages = sync_group.find_messages(
+            Some(GroupMessageKind::Application),
+            None,
+            None,
             None,
             None,
+        )?;
+        let last_message = messages.last();
+        let history_request: Option<(String, String)> = match last_message {
```
insipx, line 289: if we don't need the `None` branch, can use `if let Some`:

```rust
let history_request: Option<(String, String)> = if let Some(msg) = last_message {
   let message_history_content = /* ... */
}

```

xmtp_mls/Cargo.toml
```diff
@@ -28,67 +37,68 @@ diesel = { version = "2.2.2", features = [
...
-tracing.workspace = true
 openmls = { workspace = true, features = ["test-utils"] }
 openmls_basic_credential = { workspace = true }
 openmls_rust_crypto = { workspace = true }
 openmls_traits = { workspace = true }
+parking_lot = "0.12.3"
 prost = { workspace = true, features = ["prost-derive"] }
 rand = { workspace = true }
 reqwest = { version = "0.12.4", features = ["stream"] }
-ring = "0.17.8"
```
insipx, line 46: 🙏

xmtp_mls/src/groups/message_history.rs
rygine: this will reply the latest request message in the sync group, which is sync'd prior to checking. does that make sense?

xmtp_mls/src/groups/message_history.rs
```diff
@@ -129,37 +197,61 @@ where
...
+
+        let messages = sync_group.find_messages(
+            Some(GroupMessageKind::Application),
+            None,
+            None,
+            None,
+            None,
+        )?;
+
+        let last_message = match messages.last() {
```
insipx, line 222: Nesting match statements can get messy & make it difficult to intuit what's happening


We can compose the `Result<>` and `Option<>` in order to retain one level of indentation

But, we also need the result of the deserialization `from_slice` operation



So, before matching we can map the option:

```rust
let last_message = messages.last().map(|m| serde_json::from_slice(msg.decrypted_message_bytes)).ok().flatten();
match last_message {
    Some(Ok(MessageHistoryContent::Request(request))) => {
        /* ... */
    }
    Some(Ok(MessageHistoryContent::Reply(_) => {
        /* */
    },
    None => return Err(MessageHistoryError::NoPendingRequest)
}
```

xmtp_mls/src/groups/message_history.rs
neekolas: I think that's fine for now. Just want to walk through a scenario.

1. AppB receives a message history request from AppA and shows it to the user. They look at it and click OK.
2. App receives a second message history request from App C
3. libxmtp replies to the request from App C because it's the newest request at the time of reply

We can solve that issue later, but it's something we should write a ticket for

xmtp_mls/src/groups/message_history.rs
```diff
@@ -178,37 +269,162 @@ where
...
+                {
+                    None
+                } else {
+                    let message_history_content = serde_json::from_slice::<MessageHistoryContent>(
+                        &msg.decrypted_message_bytes,
+                    )?;
+                    match message_history_content {
+                        // if the last message is a reply, return it
+                        MessageHistoryContent::Reply(reply) => Some(reply),
+                        _ => None,
```
neekolas, line 353: Should log the error from here

xmtp_mls/src/groups/message_history.rs
```diff
@@ -510,9 +698,10 @@ impl From<HistoryReply> for MessageHistoryReply {
         MessageHistoryReply {
             request_id: reply.request_id,
             url: reply.url,
-            bundle_hash: reply.bundle_hash,
-            signing_key: Some(reply.signing_key.into()),
             encryption_key: Some(reply.encryption_key.into()),
```
neekolas, line 701: Feel free to update the protos whenever

xmtp_mls/src/groups/message_history.rs
```diff
@@ -65,6 +76,22 @@ pub enum MessageHistoryError {
...
+    NoReplyToProcess,
+    #[error("generic: {0}")]
+    Generic(String),
+    #[error("missing history sync url")]
+    MissingHistorySyncUrl,
+    #[error("invalid history message payload")]
+    InvalidPayload,
+    #[error("invalid history bundle url")]
+    InvalidBundleUrl,
 }
```
neekolas, line 95: Can be done in a follow up, but we're going to need to define which of these error variants are retryable or not. Probably just the `Storage` directly, but `Client` and `Group` should defer to those error type's `RetryableError` implementation to see if the underlying error is retryable too.

examples/cli/cli-client.rs
```diff
@@ -350,16 +419,20 @@ async fn create_client(cli: &Cli, account: IdentityStrategy) -> Result<Client, C
...
         );
+        builder = builder.history_sync_url(MessageHistoryUrls::LOCAL_ADDRESS);
     } else {
         info!("Using dev network");
         builder = builder.api_client(
             ApiClient::create("https://grpc.dev.xmtp.network:443".into(), true)
                 .await
                 .unwrap(),
         );
+        builder = builder.history_sync_url(MessageHistoryUrls::DEV_ADDRESS);
```
neekolas, line 430: You should be able to chain these calls. The compiler won't let you?

xmtp_mls/src/groups/message_history.rs
rygine: won't the error surface with the `?` on line 359? i re-did the `serde_json` errors in a recent commit.

examples/cli/cli-client.rs
rygine: oh, i was just following what was already there. i'll adjust to use chaining!

neekolas (approved):

xmtp_mls/src/groups/message_history.rs
rygine: after implementing your suggestion, i realized that i need access to the value of `msg` inside of the `match`. any tips on how the accomplish that without a nested `match`?

xmtp_mls/src/groups/message_history.rs
```diff
@@ -384,58 +592,50 @@ pub(crate) fn decrypt_history_file(
...
+    if response.status().is_success() {
+        Ok(response.text().await?)
+    } else {
+        eprintln!(
+            "Failed to upload file. Status code: {} Response: {:?}",
+            response.status(),
+            response
+        );
+        Err(MessageHistoryError::Reqwest(
+            response.error_for_status().unwrap_err(),
```
rygine, line 615: @insipx `clippy` is complaining about this line in CI, but doesn't complain when i run it locally. any suggestions on a solution? how can i get `clippy` to recognize this issue locally?

rygine: nvm, i see that it's been fixed in `main`

insipx: ah weird, it should who locally. The only difference for CI is that we add -Dwarnings to fail the build. Maybe you're on an old clippy build, possibly `rustup update stable`?

running just `cargo clippy` will show a warning but not fail the build

xmtp_mls/src/groups/message_history.rs
```diff
@@ -384,58 +592,52 @@ pub(crate) fn decrypt_history_file(
...
-        .body(content)
-        .send()
-        .await?;
+    let response = client.post(url).body(content).send().await?;
 
-    Ok(())
+    if response.status().is_success() {
+        Ok(response.text().await?)
+    } else {
+        eprintln!(
```
insipx, line 609: sorry last nit: prefer `tracing::error!()` here to be consistent with the rest of our logging

feel free to do that in one of the followups

insipx (approved):

PR #933, opened 2024-08-01
Removes unnecessary group_id parameter

tuddman:
## Summary

Small patch to remove the unnecessary `group_id` param for the `stream_with_callback` impl method on Groups.   It already has that information from `self.group_id`


tuddman:
Closing.  I see why we're calling this fn with the param.  We're calling an impl method on the base struct, not an instance, in `bindings_ffi`.

neekolas (approved):

PR #935, opened 2024-08-03
The first message sent gets lost when Epochs behind 

nplasterer:
When the epoch of a group is changed a users messages won't successfully send until a sync is performed.

Possible related to our `max_past_epochs` setting not being honored by openmls. That would explain why, after the epoch is pushed forward by 1, the first message isn't decryptable. But we sync after sending, so all messages afterwards are decryptable


cameronvoell (approved):

nplasterer:
🎉

Issue #931, opened 2024-07-31
Membership lists inconsistent across users in same group


nplasterer:
Members in same group are seeing different membership lists.
People in the same group across devices are seeing different membership lists.

**Naomi iOS**

![Image](https://github.com/user-attachments/assets/b689d678-1624-4cf2-98bf-4629933926ed)

**Naomi Android**

![Image](https://github.com/user-attachments/assets/ee6b1900-c1b9-45e1-9692-2c76ff997f6d)

**Rich iOS**

![Image](https://github.com/user-attachments/assets/c022cef5-e21e-4855-b4aa-581dc47211d7)

**Cameron Android**

![Image](https://github.com/user-attachments/assets/718fecbc-7b4d-4027-94bf-a75e62574593)

**Andrew iOS**

![Image](https://github.com/user-attachments/assets/2033422a-d573-42db-818f-8940cef14cc6)

nplasterer:
It appears that the membership list is somehow always consistently correct on Android. But that may be because we have less usage on Android at the moment. TBD.

nplasterer:
Another odd behavior that may be related or not. Is that certain members cannot be added to the group because it says that they are still apart of the group even tho they are not present in anyones membership list. For example in this specific group chat cvoell.converse can not be added because the group thinks for some reason he is already a part of the group.

PR #937, opened 2024-08-05
When a new installation is added the members list becomes incorrect

nplasterer:
Fixes https://github.com/xmtp/libxmtp/issues/931


richardhuaaa (approved):
Great find

PR #944, opened 2024-08-06
Fix missing members due to cache inconsistency

richardhuaaa:
Depends on https://github.com/xmtp/libxmtp/pull/937

We currently write the association state cache whenever we compute an association state from scratch. Also, whenever we read an association state, we check the cache, and if it's not there, we go ahead and re-compute it. However, there are two codepaths that behave a little differently, causing us to underreport the members in a group:
1. One codepath computes incremental updates to an existing state, and does not update the cache.
2. One codepath batch reads multiple association states from the cache, and silently omits anything it doesn't find rather than re-computing it.

I've done a few things in this PR:

1. When computing incremental updates, make sure we also write to the cache
2. When batch reading from the cache, bake in a hard assumption that the cache will be up-to-date, and throw errors if we don't find what we expect
3. Some small improvements to our logging

### DIFF
--- a/bindings_ffi/src/logger.rs
+++ b/bindings_ffi/src/logger.rs
@@ -20,7 +20,7 @@ impl log::Log for RustLogger {
             self.logger.lock().expect("Logger mutex is poisoned!").log(
                 record.level() as u32,
                 record.level().to_string(),
-                format!("[libxmtp] {}", record.args()),
+                format!("[libxmtp][t:{}] {}", thread_id::get(), record.args()),
             );
         }
     }
--- a/xmtp_mls/src/groups/members.rs
+++ b/xmtp_mls/src/groups/members.rs
@@ -40,19 +40,25 @@ impl MlsGroup {
             .members
             .into_iter()
             .map(|(inbox_id, sequence_id)| (inbox_id, sequence_id as i64))
+            .filter(|(_, sequence_id)| *sequence_id != 0) // Skip the initial state
             .collect::<Vec<_>>();
 
         let conn = provider.conn_ref();
-        let association_state_map =
+        let association_states =
             StoredAssociationState::batch_read_from_cache(conn, requests.clone())?;
         let mutable_metadata = self.mutable_metadata()?;
-        // TODO: Figure out what to do with missing members from the local DB. Do we go to the network? Load from identity updates?
-        // Right now I am just omitting them
-        if association_state_map.len() != requests.len() {
-            // Cache miss - should either error if not expected, or should fetch from network if it is expected
-            log::error!("Failed to load all members for group: {:?}", requests);
+        if association_states.len() != requests.len() {
+            // Cache miss - not expected to happen because:
+            // 1. We don't allow updates to the group metadata unless we have already validated the association state
+            // 2. When validating the association state, we must have written it to the cache
+            log::error!(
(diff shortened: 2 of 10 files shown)


bindings_ffi/src/logger.rs
```diff
@@ -20,7 +20,7 @@ impl log::Log for RustLogger {
             self.logger.lock().expect("Logger mutex is poisoned!").log(
                 record.level() as u32,
                 record.level().to_string(),
-                format!("[libxmtp] {}", record.args()),
```
richardhuaaa, line 23: Adds thread ID to production logs (previously only included on the MockLogger)

xmtp_mls/src/groups/members.rs
```diff
@@ -40,14 +40,25 @@ impl MlsGroup {
             .members
             .into_iter()
             .map(|(inbox_id, sequence_id)| (inbox_id, sequence_id as i64))
+            .filter(|(_, sequence_id)| *sequence_id != 0) // Skip the initial state
```
richardhuaaa, line 43: The original behavior is as follows:

1. When a group is created, the metadata will read `(creator_inbox_id, 0)`. So a sequence_id of 0
2. The old logic would return no association state, and therefore an empty member list

I'm not sure if this is by design, or a side effect. Either way, adding this filter preserves that original behavior, while ensuring that the new error handling logic on L50 doesn't trigger in this case.

xmtp_mls/src/identity_updates.rs
```diff
@@ -148,8 +152,26 @@ where
...
+        if ending_sequence_id.is_some()
+            && last_sequence_id.is_some()
+            && last_sequence_id != ending_sequence_id
+        {
+            log::error!(
+                "Did not find the expected last sequence id. Expected: {:?}, Found: {:?}",
+                ending_sequence_id,
+                last_sequence_id
+            );
+            return Err(AssociationError::MissingIdentityUpdate.into());
```
richardhuaaa, line 171: This is a new error scenario I think is possible in theory, maybe more so once replication is implemented. When we get a group metadata update for a specific sequence ID, it's possible that we fail to fetch the update at that specific sequence ID, for example if the server has a stale version of the identity log. I looked at the rest of the code and I don't think we currently attempt to detect if this happens, which is fine given the current server implementation, but we might need to address this eventually.

Heads up @neekolas

xmtp_mls/src/identity_updates.rs
```diff
@@ -159,6 +181,16 @@ where
...
         }
 
+        log::debug!("Final state at {:?}: {:?}", last_sequence_id, final_state);
+        if last_sequence_id.is_some() {
+            StoredAssociationState::write_to_cache(
+                conn,
+                inbox_id.as_ref().to_string(),
+                last_sequence_id.unwrap(),
+                final_state.clone(),
+            )?;
```
richardhuaaa, line 191: This is what fixes the test

nplasterer (approved):

xmtp_mls/src/identity_updates.rs
neekolas: That's a good call out. We'll have to come up with some way of handling those errors and maybe retrying when we don't have full consistency on messages.

PR #945, opened 2024-08-07
Abort if retry fails

neekolas:
## tl;dr

This fixes a bug where transient errors can turn into a forked group state.

Imagine the following situation:

There are 2 unprocessed messages on the network, the first is a valid commit and the second is an application message.
- Client attempts to process message 1, but the database is locked and all retries are exhausted
- Client moves on to message 2 and is able to successfully process it. This updates the cursor in the database, so message 1 will never be tried again

There are many similar variations to this story (API downtime, cases where conflicting commits are inconsistently applied).

This fix changes the logic so that if an error is categorized as retryable, we will abort processing all further payloads. This does put even more pressure on us to correctly identify all retryable errors. If we categorize an error that will never resolve on its own as "retryable" the group will never be able to sync past this point. We should monitor these cases and fix them, rather than ever let a client move on and potentially fork group state.

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -1199,6 +1199,7 @@ fn build_group_join_config() -> MlsGroupJoinConfig {
 
 #[cfg(test)]
 mod tests {
+    use diesel::connection::SimpleConnection;
     use openmls::prelude::{tls_codec::Serialize, Member, MlsGroup as OpenMlsGroup};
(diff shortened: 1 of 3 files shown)


insipx (approved):

nplasterer (approved):

neekolas:
@richardhuaaa agree that network disconnects are the most likely way to cause this error right now, since we made the DB locks resolve themselves fast enough that things usually work on the second try

Issue #801, opened 2024-05-31
Personal Preferences


nplasterer:
Move this to V3

nplasterer:
Personal Preferences are stored in our V3 database

1. Same as conversation history ie you need to sync from another device that has the personal preferences.
You also need to sync all changes from that point forward. You need to relay that change from that point forward. A persistant identity group chat. Label it as group metadata for those that sync later.

Potentially breaking: Mutable metadata on the sync group? The way we do sync groups currently is not compatible with this. Which might require us to support external joins. State snapshot as part of the message history sync. 

2. ~~ The same as V2 ~~
Anytime a user signs in on a new device we ask them for a wallet signature. We give them the same payload to sign every time. Derive an encryption key from those bytes. A secret key that everyone of your devices can get. As long as they have access to your wallet. Wallet apis dont give us anything else but the ability to sign. Smart wallets don't have signatures over time and you might not have access to all wallets on each device.
UX burden to request another signature from a user
Deriving a encryption key from signature bytes is not standard 

3. Make the user transport a secret between devices.
They set a passcode or we give them a passcode. They have to remember this passcode on each login. When the sign in on a new device it pulls that file. A lot of UX burden. Authenticating double inside a app.

4. Threshhold cryptography or Shamir secret sharing
You can have multiple backends each backend holds a piece of the secret. But not enough to get the full secret. Requires the backends to collide together. We would need to re implement the algorithm. We would need to store the keys.

nplasterer:
Edge case
If you loose your device create a new account entirely.

This is only a PPPP issue since conversations will still get messages and would show up eventually.
We could let a split brain happen and they would merge eventually. Once you sign in and sync your consent the broadcast that to the new group. Whenever you log into another device with an inconstant state it can broadcast. Consent state is written in group metadate in one group just for consent.

We store the consent state and do a join table for them.
We have a group that is only your self and it consists of all of your installation. There needs to be mechanism to catch up a device that is behind. You can tell the device is behind because it sent a message history request that nobody responded to. Still requires a device online to catch up the new device.
We don't wipe we let you continue without the existing consent state. 
If you don't sync everything will be in pending. When you come online from an old device then it will fix everything. Conflict resolution.


In order to get your history you need to have your other device.
Let the converse app do backups for you to the iCloud/GoogleCloud.


Work to be done
- A database of personal preference records stored in V3 with a join table on groups
[...]

PR #946, opened 2024-08-07
Private Preferences DB

nplasterer:
Part of https://github.com/xmtp/libxmtp/issues/801

Adds a new table for personal preference records in the local db


xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE "private_preferences"(
+    -- The inbox_id the private preferences are owned by
+    "inbox_id" text NOT NULL,
```
nplasterer, line 3: We might not need an inbox_id since every item in this list will be for the person who is logged in they shouldn't be getting other peoples preferences.

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE "private_preferences"(
```
nplasterer, line 1: Should this be `personal_preferences`? or `private_preferences` ?

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
```diff
@@ -0,0 +1,11 @@
...
+    -- The inbox_id the private preferences are owned by
+    "inbox_id" text NOT NULL,
+    -- Enum of the PRIVATE_PREFERENCE_TYPE (GROUP, INBOX, etc..)
+    "valueType" int NOT NULL,
+    -- Enum of PREFERENCE_STATE (ALLOWED, DENIED, etc..)
+    "state" int NOT NULL,
+    -- The value of what has a preference
+    "value" text NOT NULL,
+    PRIMARY KEY (valueType, value)
+);
```
nplasterer, line 11: Want to make sure I'm reasoning about this correctly we will want a way to get the current consent state for a group or inbox id performantly so the primary key should be the value and the valueType so that you can fetch for type group and value group_id and get what the state is.

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE "private_preferences"(
+    -- The inbox_id the private preferences are owned by
+    "inbox_id" text NOT NULL,
+    -- Enum of the PRIVATE_PREFERENCE_TYPE (GROUP, INBOX, etc..)
+    "valueType" int NOT NULL,
```
nplasterer, line 5: I think we will want to have some logic in the code that says if type is ADDRESS find the inbox_id for that address and actually only store and fetch inbox ids. Probably a migration needed to convert all the V2 items to inbox ids.

nplasterer (commented):
@xmtp/engineering would love some preliminary heads thinking about this data structure for personal preferences. 🙏

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
neekolas: I thought we settled on private

neekolas, line 1: I wonder if we really want to call this "consent state".

Private preferences sounds more generic than what is really here. The schema is very specific to consent preferences (which is also the only kind we have right now).

nplasterer: From the topic I actually think we call it `user_preferences`

https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/messages/Topic.kt#L33

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
```diff
@@ -0,0 +1,11 @@
...
+    -- Enum of PREFERENCE_STATE (ALLOWED, DENIED, etc..)
+    "state" int NOT NULL,
+    -- The value of what has a preference
+    "value" text NOT NULL,
```
neekolas, line 9: What do we think about calling it the "entity" just for clarity. I could read value to mean either "the thing that has a preference" or "the preference itself".

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
neekolas: That does feel like what we'll want at the end of the process. Very cheap to look up, has all the information a

One thing that was a struggle in the past is just making sure that we processed all the events in the correct order, so that if you allowed and then blocked the same group you'd get a consistent value in the end. What happens if you received a preferences event that you couldn't read before, but now you can? How do we rewind things so that you can rebuild the state? Maybe the right answer is to truncate this table and start from scratch in that scenario?

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
neekolas: Oh, you're right. Sorry, we changed this name so many times.

In any case, I'm not sure that should be the name for this table. User preferences are going to be handling more than just consent state (for example, HMAC keys). This table is all about consent state.

nplasterer: @neekolas in that case would we make a separate table just for the hmac keys?

neekolas: I think so. I don't think there are a million different cases of private preferences. There's probably 3-5, ever. Might as well store them in a nice way vs one-size-fits-all

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
insipx: does it make sense to make an index here, like we did for group id?

```
CREATE INDEX idx_value_type_value ON private_preferences(valuType, value ASC);
```

xmtp_mls/src/storage/encrypted_store/consent_record.rs
```diff
@@ -0,0 +1,217 @@
...
+        record: StoredConsentRecord,
+    ) -> Result<(), StorageError> {
+        self.raw_query(|conn| {
+            let _ = diesel::insert_into(dsl::consent_records)
+                .values(&record)
+                .on_conflict((dsl::entity_type, dsl::entity))
+                .do_update()
+                .set(dsl::state.eq(excluded(dsl::state)))
+                .execute(conn)
+                .map_err(StorageError::from);
```
insipx, line 75: should throw error here `?` rather than using `let _ =`

insipx:
sans nits/lints + tests, it looks good to me

xmtp_mls/migrations/2024-08-07-213816_create-private-preference-store/up.sql
nplasterer: My understanding is that the primary key made an index by default so we shouldn't need to add another index.

xmtp_mls/src/groups/group_metadata.rs
```diff
@@ -150,8 +150,12 @@ impl TryFrom<DmMembersProto> for DmMembers {
 
     fn try_from(value: DmMembersProto) -> Result<Self, Self::Error> {
         Ok(Self {
-            member_one_inbox_id: value.dm_member_one.unwrap().inbox_id.clone(),
-            member_two_inbox_id: value.dm_member_two.unwrap().inbox_id.clone(),
```
nplasterer, line 154: These unwraps were just on the mls_dms feature branch already. Just cleaning up the linter in this PR. cc @cameronvoell

insipx (approved):

PR #949, opened 2024-08-08
Add HTTP gRPC crate

insipx:
- This adds `xmtp_api_http` for using the http API with gRPC. It adds some utility traits to `xmtp_mls` for switching out the Client backend depending on feature flags. 
- gets rid of `tokio::spawn` in stream_all_messages
- add HTTP api workflow for tests
- replaces all mutexes with `parking_lot::Mutex`
- removes `block_on` from `InboxOwner::Sign` for `LocalWallet`

bumping tonic to the latest version (0.13) will break TLS -- needs further investigation. ~Further, unifying `pbjson` version between crate in the workspace breaks the compile. Currently we have two different `pbjson` version. Will open separate issue.~


Cargo.toml
```diff
@@ -50,11 +49,16 @@ serde_json = "1.0"
...
-tracing = { version = "0.1", features = ["release_max_level_debug"] }
+tokio = { version = "1.35.1", default-features = false }
+async-stream = "0.3"
+tracing = { version = "0.1" }
 tracing-subscriber = "0.3"
 url = "2.5.0"
+tonic = "^0.11" 
+prost = "^0.12"
+prost-types = "^0.12"
+pbjson = "0.5.1"
```
neekolas, line 60: Have we run into any cases where `pbjson` doesn't serialize the same way that GRPC gateway expects? Or is it guaranteed to be 1:1?

insipx: Looked at this again with a fresh head, this was an issue trying to update `pbjson` to `0.7`, but making everything point to `0.6` works. Updating `pbjson` to 0.7 forces an update for `prost` & `tonic` too, because of trait implementations. But updating those dependencies results in test failures because of TLS problems which can be tackled separately.

neekolas (approved):

PR #950, opened 2024-08-08
skip empty key packages

insipx:
if an installation id has a keypackage of length `0`, don't include it

### DIFF
--- a/xmtp_mls/src/api/mls.rs
+++ b/xmtp_mls/src/api/mls.rs
@@ -236,13 +236,20 @@ where
             .key_packages
             .into_iter()
             .enumerate()
-            .map(|(idx, key_package)| {
-                (
+            .filter_map(|(idx, key_package)| {
+                if key_package.key_package_tls_serialized.is_empty() {
+                    log::warn!(
+                        "installation key 0x{} has empty key package",
+                        hex::encode(&installation_keys[idx])
+                    );
+                    return None;
+                }
+
+                Some((
                     installation_keys[idx].to_vec(),
                     key_package.key_package_tls_serialized,
-                )
+                ))
             })
-            .filter(|(_, key_package)| key_package.len() > 0)
             .collect();
 
         Ok(mapping)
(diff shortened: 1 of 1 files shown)


nplasterer (approved):
Thanks for the fix!

xmtp_mls/src/api/mls.rs
```diff
@@ -236,11 +236,19 @@ where
             .key_packages
             .into_iter()
             .enumerate()
-            .map(|(idx, key_package)| {
-                (
+            .filter_map(|(idx, key_package)| {
```
neekolas, line 239: Thinking about this a little more, we probably want to test this in an actual group somehow.

My worry is that looking at the identity update logs everyone is going to expect N new installations but the commit will have < N new installations. That'll throw an error when someone else tries to read a commit, since it doesn't match the expectations which looks fishy.

I wonder if the better thing to do here is to either
A) Enforce that a key package must be uploaded and valid on the server at the time of adding an installation to the identity history. That way this situation is impossible.
B) Go back to throwing an error when trying to add the member to the group initially, rather than kicking the can down the road to when the commit is read and validated.

Also, if the key package is empty I have no idea why our backend is even accepting it. That should have never been saved in the DB

nplasterer: So @cameronvoell and I are testing this in the example app now with his identity that has an empty key package. I can successfully create a group with his identity now. However his identity can't see any messages from me or the other members of the group and vice versa. Is that what we would expect? Going to hold off on pushing this all the way through the SDKs till next week just to make sure.

neekolas: You'd see some sort of `CommitValidationError` in the logs if that were happening.

I took another look at the code, and I can't see where we are checking for installations that _should_ be added but weren't. We do check that all installations that were added or removed are in the diff between the old and new inbox state. 

I'd have to look closer to see if that is a feature to protect against this case or a bug. The risk of not doing that check is that you can selectively omit a new installation from someone else and block those devices from ever joining the group.

Issue #948, opened 2024-08-08
Bug: Investigate failing test with `Notify` instead of `sleep`


insipx:
### Describe the bug

changing the test `test_subscribe_membership_changes` from using `sleep` to using `tokio::sync::Notify`/`Delivery` results in test failure. The stream does not receive the group membership change message.

```rust
    #[tokio::test(flavor = "current_thread")]
    async fn test_subscribe_membership_changes() {
        let amal = Arc::new(ClientBuilder::new_test_client(&generate_local_wallet()).await);
        let bola = ClientBuilder::new_test_client(&generate_local_wallet()).await;

        let amal_group = amal
            .create_group(None, GroupMetadataOptions::default())
            .unwrap();

        let amal_ptr = amal.clone();
        let amal_group_ptr = amal_group.clone();
        let notify = Delivery::new(Some(Duration::from_secs(20)));
        let notify_ptr = notify.clone();

        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
        let mut stream = UnboundedReceiverStream::new(rx);
        tokio::spawn(async move {
            let mut stream = amal_group_ptr.stream(amal_ptr).await.unwrap();
            while let Some(item) = stream.next().await {
                let _ = tx.send(item);
                notify_ptr.notify_one();
            }
        });

        amal_group
            .add_members_by_inbox_id(&amal, vec![bola.inbox_id()])
            .await
            .unwrap();
        notify
            .wait_for_delivery()
            .await
            .expect("Never received group membership change from stream");
        let first_val = stream.next().await.unwrap();
        assert_eq!(first_val.kind, GroupMessageKind::MembershipChange);

        amal_group
            .send_message("hello".as_bytes(), &amal)
            .await
            .unwrap();
        notify
            .wait_for_delivery()
            .await
            .expect("Never received second message from stream");
        let second_val = stream.next().await.unwrap();
        assert_eq!(second_val.decrypted_message_bytes, "hello".as_bytes());
    }
```
The test fails with 
```
2024-08-12T15:05:59.909954Z ERROR xmtp_mls::subscriptions: Error processing stream entry: Generic("Group intent could not be committed")
```
before the stream times out

<details>
    <summary><h2><b>Log output</b></h2></summary>

```

running 1 test
2024-08-12T15:05:59.713631Z  INFO xmtp_mls::storage::encrypted_store: Setting up DB connection pool
[...]

PR #958, opened 2024-08-14
Reproduce perf test in rust

insipx:
Copied naomis perf test from [here](https://github.com/xmtp/xmtp-react-native/pull/457#top) into rust. Was able to reproduce UNIQUE constraint errors, but no group intent errors or IntentPending errors

Group intent not committed is repro'd consistently with #948. This seems to track since both issues include adding members to a group, which is where this intent error seems to occur

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -229,7 +229,7 @@ impl XmtpMlsLocalContext {
         self.identity.installation_keys.to_public_vec()
     }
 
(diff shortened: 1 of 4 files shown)


nplasterer:
Do you think this unique constraint error is causing any down stream problems or is it just noise?

insipx:
Seems like its ok to ignore the UNIQUE constraint errors for now. looks like something where we're trying to publish the same intent twice and one of them goes through while the other tries to store but it already exists b/c the first one finished first

PR #962, opened 2024-08-14
Upload key package before publishing identity updates

neekolas:
### TL;DR
Implements parts of https://github.com/xmtp/xmtp-node-go/issues/399

- Uses `UploadKeyPackage` instead of `RegisterInstallation` so we can deprecate the register installation endpoint
- Uploads key package _before_ publishing the identity updates for new installations. This ensures that any installation seen in an identity update must have a matching key package.
- No longer checks the lifetime of key packages when verifying

### What changed?

- Removed expiration checks and set expiration to 0 in MLS validation service
- Updated the client registration process to register identity before applying signature request
- Renamed `register_installation` to `upload_key_package` in API client calls
- Removed lifetime validity checks from key package verification

### DIFF
--- a/mls_validation_service/src/handlers.rs
+++ b/mls_validation_service/src/handlers.rs
@@ -198,7 +198,8 @@ async fn validate_inbox_id_key_package(
         error_message: "".into(),
         credential: Some(kp.credential),
         installation_public_key: kp.installation_public_key,
-        expiration: kp.inner.life_time().not_after(),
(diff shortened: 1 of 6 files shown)


nplasterer (approved):

neekolas:
### Merge activity

* **Aug 14, 5:32 PM PDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/962).

PR #964, opened 2024-08-14
Stop setting new lifetimes

neekolas:
## tl;dr

- Stops setting new lifetimes on key packages, so that we don't run in to cases of users with invalid key packages in 6 months
- Anyone with a key package with no lifetime will appear as being invalid to an older client version, which will prevent anyone from adding them to a group. Reading existing commits that add that member to a group will also fail on client SDK versions that still check for lifetimes.
- Ideally we roll this out concurrently to wiping all existing installations, and we get everyone upgraded to a newer SDK version.

### DIFF
--- a/xmtp_mls/src/identity.rs
+++ b/xmtp_mls/src/identity.rs
@@ -24,7 +24,6 @@ use openmls::{
     extensions::{
         ApplicationIdExtension, Extension, ExtensionType, Extensions, LastResortExtension,
     },
-    key_packages::Lifetime,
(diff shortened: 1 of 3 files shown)


cameronvoell (approved):

neekolas:
### Merge activity

* **Aug 27, 10:34 AM PDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/964).

PR #969, opened 2024-08-16
Clean up empty intents

neekolas:
## tl;dr

- We [recently made a change](https://github.com/xmtp/libxmtp/pull/927/files) where `get_publish_intent_data` returns an `Option`, but we were not properly handling the None value of the return.
- In the case where there is no publish intent data, we should delete the intent since there is nothing left to do.


nplasterer (approved):
Nice find!

PR #973, opened 2024-08-19
Use latest MLS

neekolas:
## tl;dr

- Upgrades to the latest MLS which includes some bug fixes
- Moves `max_past_epochs` values to the configuration.rs file
- Checks `is_valid_epoch` in line with the `MAX_PAST_EPOCHS` configuration (now possible with the fixes in OpenMLS)
- Simplifies testing for old epoch messages

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -72,7 +72,7 @@ use crate::{
     client::{deserialize_welcome, ClientError, MessageProcessingError, XmtpMlsLocalContext},
     configuration::{
         CIPHERSUITE, GROUP_MEMBERSHIP_EXTENSION_ID, GROUP_PERMISSIONS_EXTENSION_ID, MAX_GROUP_SIZE,
-        MUTABLE_METADATA_EXTENSION_ID,
+        MAX_PAST_EPOCHS, MUTABLE_METADATA_EXTENSION_ID,
     },
     hpke::{decrypt_welcome, HpkeError},
     identity::{parse_credential, Identity, IdentityError},
@@ -1164,7 +1164,7 @@ fn build_group_config(
         .capabilities(capabilities)
         .ciphersuite(CIPHERSUITE)
         .wire_format_policy(WireFormatPolicy::default())
-        .max_past_epochs(3) // Trying with 3 max past epochs for now
+        .max_past_epochs(MAX_PAST_EPOCHS) // Trying with 3 max past epochs for now
         .use_ratchet_tree_extension(true)
         .build())
 }
@@ -1211,7 +1211,7 @@ async fn validate_initial_group_membership<ApiClient: XmtpApi>(
 fn build_group_join_config() -> MlsGroupJoinConfig {
     MlsGroupJoinConfig::builder()
         .wire_format_policy(WireFormatPolicy::default())
-        .max_past_epochs(3) // Trying with 3 max past epochs for now
+        .max_past_epochs(MAX_PAST_EPOCHS) // Trying with 3 max past epochs for now
         .use_ratchet_tree_extension(true)
         .build()
 }
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -1164,7 +1164,7 @@ fn build_group_config(
         .capabilities(capabilities)
         .ciphersuite(CIPHERSUITE)
         .wire_format_policy(WireFormatPolicy::default())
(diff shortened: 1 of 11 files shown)


xmtp_mls/src/storage/sql_key_store.rs
```diff
@@ -1130,31 +1072,31 @@ mod tests {
...
-        let store = EncryptedMessageStore::new(
-            StorageOption::Persistent(db_path),
-            EncryptedMessageStore::generate_enc_key(),
-        )
-        .unwrap();
-        let conn = store.conn().unwrap();
-        let key_store = SqlKeyStore::new(conn.clone());
-        let provider = XmtpOpenMlsProvider::new(conn);
-        let group_id = GroupId::random(provider.rand());
+    // #[test]
```
neekolas, line 1075: Will have to rewrite this test, since it uses aad to test list removes.

neekolas:
I did some testing with the CLI, where I created a group using `main` and then switched over to this branch and added more members and sent messages. Everything seems to work as expected

nplasterer (approved):

xmtp_mls/src/groups/mod.rs
```diff
@@ -1164,7 +1164,7 @@ fn build_group_config(
         .capabilities(capabilities)
         .ciphersuite(CIPHERSUITE)
         .wire_format_policy(WireFormatPolicy::default())
-        .max_past_epochs(3) // Trying with 3 max past epochs for now
+        .max_past_epochs(MAX_PAST_EPOCHS) // Trying with 3 max past epochs for now
```
richardhuaaa, line 1167: Can remove the comment, here and below

xmtp_mls/src/groups/sync.rs
```diff
@@ -344,8 +345,9 @@ impl MlsGroup {
                     intent.id,
                     group_epoch,
                     message_epoch,
-                    1, // max_past_epochs, TODO: expose from OpenMLS MlsGroup
+                    MAX_PAST_EPOCHS,
                 ) {
+                    log::warn!("[{}] Message was sent in an old epoch. Attempting to re-send. Group epoch: {group_epoch}. Message epoch: {message_epoch}", self.context.inbox_id());
```
richardhuaaa, line 350: There is already very similar logging inside of `is_valid_epoch`

xmtp_mls/src/storage/errors.rs
```diff
@@ -89,7 +89,16 @@ impl RetryableError for openmls::group::CreateCommitError<sql_key_store::SqlKeyS
...
+            Self::LeafNodeUpdateError(leaf_node_update) => retryable!(leaf_node_update),
+            _ => false,
+        }
+    }
+}
+
+impl RetryableError for openmls::treesync::LeafNodeUpdateError<sql_key_store::SqlKeyStoreError> {
+    fn is_retryable(&self) -> bool {
+        match self {
+            Self::Storage(storage) => retryable!(storage),
```
richardhuaaa, line 101: Are all storage errors retryable here?

richardhuaaa (approved):

xmtp_mls/src/storage/errors.rs
neekolas: Not all storage errors are retryable, but most are. Connection pool is a perfect example of a retryable error.

Answering this question I noticed some newer DB error types that aren't marked as retryable, so I'm going to add them in here. 😬

xmtp_mls/src/storage/errors.rs
```diff
@@ -63,11 +63,12 @@ impl RetryableError for diesel::result::Error {
...
     fn is_retryable(&self) -> bool {
         match self {
-            Self::DieselConnect(connection) => {
-                matches!(connection, diesel::ConnectionError::BadConnection(_))
-            }
+            Self::DieselConnect(_) => true,
             Self::DieselResult(result) => retryable!(result),
             Self::Pool(_) => true,
+            Self::Lock(_) => true,
+            Self::PoolNeedsConnection => true,
```
insipx, line 70: PoolNeedsConn/SqlCipherNot loaded can't really be retryable here I don't think, since a SqlCipher error indicates a compile-time linking issue and PoolNeedsConnection would be an error on SDK side that can't be resolved from libxmtp

Also, my bad, I should be better with remembering to add these to the `retryable` 😓

insipx (approved):

xmtp_mls/src/storage/errors.rs
neekolas: Thanks for clarifying. I misunderstood the sqlcipher error as something that could happen if the pragma statement didn’t apply. Will remove.

PR #975, opened 2024-08-20
re-use `prepare_message` for history

insipx:
- re-uses the `prepare_message` function for message history
- introduces a closure for `prepare_message` to return a context-specific `PlaintextEnvelope`

message history functions were publishing and storing an intent, but never stored the actual history request/reply locally as a `StoredGroupMessage`. Clients who externally received the message would have the message stored because of `process_external_message` correctly forms & inserts the `StoredGroupMessage` into the database. Intent creation + message storage for history is the same as for a normal `SendMessage`, except for the proto `PlaintextEnvelope`, which differs in its version + contents.

### DIFF
--- a/xmtp_mls/src/groups/message_history.rs
+++ b/xmtp_mls/src/groups/message_history.rs
@@ -7,7 +7,6 @@ use aes_gcm::{
     aead::{Aead, KeyInit},
     Aes256Gcm,
 };
-use prost::Message;
 use rand::{
     distributions::{Alphanumeric, DistString},
     Rng, RngCore,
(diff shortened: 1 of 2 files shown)


xmtp_mls/src/groups/message_history.rs
```diff
@@ -112,22 +108,21 @@ where
...
 
-        // build the intent
-        let mut encoded_envelope = vec![];
-        envelope
-            .encode(&mut encoded_envelope)
-            .map_err(GroupError::EncodeError)?;
-        let intent_data: Vec<u8> = SendMessageIntentData::new(encoded_envelope).into();
-        let intent = NewGroupIntent::new(IntentKind::SendMessage, sync_group_id, intent_data);
-        intent.store(&conn)?;
+        let content_bytes = format!(
```
insipx, line 112: @rygine will have to change this to use the bincode from your PR

rygine (approved):

PR #976, opened 2024-08-20
Removed user from group should no longer update

nplasterer:
After a user is removed from a group the group should no longer update for them.


nplasterer:
Working as expected but good to add a test. See documentation here: https://openmls.tech/book/user_manual/remove_members.html

neekolas (approved):

PR #978, opened 2024-08-21
Expose API to revoke all other installations

neekolas:
## tl;dr

- Adds a helper FFI method to "Revoke all other installations"
- Allows for callers to get the full state of the current inbox (all addresses, all installation IDs, recovery address). The caller can choose whether to refresh from the network first

## Notes

The `revoke_all_other_installations` method may go away later, since it's unclear how generally useful it is. But it is helpful for resolving a situation where a bad client release allowed for installations missing key packages. We want to revoke them in bulk.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -333,12 +333,9 @@ impl FfiXmtpClient {
         existing_wallet_address: &str,
         new_wallet_address: &str,
     ) -> Result<Arc<FfiSignatureRequest>, GenericError> {
-        let inbox_id = self.inner_client.inbox_id();
-        let signature_request = self.inner_client.associate_wallet(
-            inbox_id,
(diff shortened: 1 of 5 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -295,6 +296,31 @@ impl FfiXmtpClient {
...
+    /**
+     * Get the client's inbox state.
+     *
+     * If `refresh_from_network` is true, the client will go to the network first to refresh the state.
+     * Otherwise, the state will be read from the local database.
+     */
+    pub async fn inbox_state(
+        &self,
+        refresh_from_network: bool,
+    ) -> Result<FfiInboxState, GenericError> {
```
nplasterer, line 309: Awesome this will be helpful for the `addWallet` stuff.

nplasterer (approved):

PR #981, opened 2024-08-21
Sync all groups at once

nplasterer:
Creates a sync all method for groups. 
https://github.com/xmtp/xmtp-react-native/issues/474

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -752,11 +752,11 @@ impl FfiConversations {
         Ok(())
     }
 
-    pub async fn sync_all(&self) -> Result<(), GenericError> {
+    pub async fn sync_all_groups(&self) -> Result<(), GenericError> {
         let inner = self.inner_client.as_ref();
         let groups = inner.find_groups(None,None,None,None)?;
 
-        inner.sync_all(groups).await?;
+        inner.sync_all_groups(groups).await?;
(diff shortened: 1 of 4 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -752,6 +752,14 @@ impl FfiConversations {
         Ok(())
     }
 
+    pub async fn sync_all(&self) -> Result<(), GenericError> {
+        let inner = self.inner_client.as_ref();
+        let groups = inner.find_groups(None,None,None,None)?;
```
nplasterer, line 757: Should this be from local or should this be inner.sync?

xmtp_mls/src/groups/mod.rs
```diff
@@ -253,7 +253,10 @@ impl MlsGroup {
 
     // Load the stored MLS group from the OpenMLS provider's keystore
     #[tracing::instrument(level = "trace", skip_all)]
-    fn load_mls_group(&self, provider: impl OpenMlsProvider) -> Result<OpenMlsGroup, GroupError> {
+    pub fn load_mls_group(
```
nplasterer, line 256: Doesn't seem like a big deal to make these all public but makes me wonder if this function should live inside `xmtp_mls/src/groups` instead?

insipx (approved):
nice!

PR #982, opened 2024-08-21
Force Provider to take ownership of `DbConnection`

insipx:
- removes Clone derive for `DbConnection`. DbConnections should not be cloned.
- forces MLS Provider to take ownership of a connection
   - this forces any function that uses a provider, to make use of a reference from provider rather than a cloned reference beforehand
- remove owned `.conn()` functions
    - `.conn()` 'implicitly' clones the underlying DbConnection
- add logging in async transactions to check how many references we have
- add `From` for conn/provider
- make `mls_provider` functions pull a db conn and instantiate provider

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4491,6 +4491,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "scoped-futures"
+version = "0.1.3"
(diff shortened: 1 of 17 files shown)


nplasterer (approved):

nplasterer:
Looks like I might have given you some merge conflicts... hopefully simple.

PR #983, opened 2024-08-21
Remove lots of unwraps

insipx:
- adds directive to disallow unwraps in code that is not test/benchmarks

This PR removes a bunch of unwraps found with clippy. An unwrap would panic libxmtp and crash the client, so we should never use it unless we are 100% sure it will not panic. Unwrapping in this context may also cause errors to go completely unreported in logs (if a task or something crashes or the logs are not picked up by a logger)

Added a warn lint for `unwrap` that clippy will pick up if ran without `test`,`test-utils`, or `bench` feature, i.e `cargo clippy --no-deps -- -Dwarnings`

there are a few unwraps here that could definitely been causing issues/causing crashes in probably less-common circumstances (like the one in insert_or_replace_group and permissions)

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -180,6 +180,8 @@ pub enum GroupError {
     PublishPanicked,
     #[error("Missing metadata field {name}")]
     MissingMetadataField { name: String },
+    #[error("Message was processed but is missing")]
+    MissingMessage,
 }
 
 impl RetryableError for GroupError {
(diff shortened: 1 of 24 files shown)


xmtp_mls/src/groups/mod.rs
```diff
@@ -1074,7 +1076,11 @@ pub fn build_extensions_for_permissions_update(
         PermissionUpdateType::UpdateMetadata => {
             let mut metadata_policy = existing_policy_set.update_metadata_policy.clone();
             metadata_policy.insert(
-                update_permissions_intent.metadata_field_name.unwrap(),
```
insipx, line 1077: this one seems like we could make a default, like `unwrap_or("DEFAULT METADATA FIELD NAME")` instead of an error

nplasterer (approved):
🔥 🔥 🔥

PR #984, opened 2024-08-22
Move staged commit to intents

neekolas:
## tl;dr

- Moves `staged_commits` from being handled by OpenMLS to being stored on each intent
- Adds a mutex for each group's sync operations. Only one sync may happen in parallel at any time per group
- Creates database columns for storing the staged commit and the epoch the commit was published
- Updates a test for message conflicts to better check for forked group states
- I found some cases where we were calling `publish_intents` and then immediately calling `sync`. That should not be necessary, since we call `publish_intents` from inside the sync method
- Adds tests for parallel  and reentrant syncs

## More Info
[https://github.com/xmtp/libxmtp/issues/979](https://github.com/xmtp/libxmtp/issues/979#issuecomment-2303541848)

### DIFF
--- a/dev/docker/compose
+++ b/dev/docker/compose
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -eou pipefail
 
-docker-compose -f dev/docker/docker-compose.yml -p "libxmtp" "$@"
+docker compose -f dev/docker/docker-compose.yml -p "libxmtp" "$@"
--- a/xmtp_mls/migrations/2024-08-22-044745_add_staged_commit/up.sql
+++ b/xmtp_mls/migrations/2024-08-22-044745_add_staged_commit/up.sql
@@ -1,0 +1,7 @@
+-- Your SQL goes here
+ALTER TABLE group_intents
+    ADD COLUMN staged_commit BLOB;
+
+ALTER TABLE group_intents
+    ADD COLUMN published_in_epoch BIGINT;
+
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -31,6 +31,7 @@ use openmls::{
 use openmls_traits::OpenMlsProvider;
 use prost::Message;
 use thiserror::Error;
+use tokio::sync::Mutex;
 
 pub use self::group_permissions::PreconfiguredPolicies;
 pub use self::intents::{AddressesOrInstallationIds, IntentError};
@@ -182,6 +183,10 @@ pub enum GroupError {
     MissingMetadataField { name: String },
     #[error("Message was processed but is missing")]
     MissingMessage,
+    #[error("sql key store error: {0}")]
+    SqlKeyStore(#[from] sql_key_store::SqlKeyStoreError),
+    #[error("No pending commit found")]
+    MissingPendingCommit,
 }
 
 impl RetryableError for GroupError {
@@ -210,6 +215,7 @@ pub struct MlsGroup {
     pub group_id: Vec<u8>,
     pub created_at_ns: i64,
     context: Arc<XmtpMlsLocalContext>,
+    mutex: Arc<Mutex<()>>,
 }
 
 #[derive(Default)]
@@ -226,6 +232,7 @@ impl Clone for MlsGroup {
             context: self.context.clone(),
             group_id: self.group_id.clone(),
             created_at_ns: self.created_at_ns,
+            mutex: self.mutex.clone(),
         }
     }
 }
@@ -241,10 +248,12 @@ pub enum UpdateAdminListType {
 impl MlsGroup {
     // Creates a new group instance. Does not validate that the group exists in the DB
     pub fn new(context: Arc<XmtpMlsLocalContext>, group_id: Vec<u8>, created_at_ns: i64) -> Self {
+        let mut mutexes = context.mutexes.clone();
         Self {
             context,
-            group_id,
+            group_id: group_id.clone(),
             created_at_ns,
+            mutex: mutexes.get_mutex(group_id),
         }
     }
 
@@ -455,11 +464,6 @@ impl MlsGroup {
             Self::into_envelope(message, now)
         });
 
-        // Skipping a full sync here and instead just firing and forgetting
-        if let Err(err) = self.publish_intents(&provider, client).await {
-            log::error!("Send: error publishing intents: {:?}", err);
-        }
-
         self.sync_until_last_intent_resolved(&provider, client)
             .await?;
 
@@ -479,7 +483,6 @@ impl MlsGroup {
         let update_interval = Some(5_000_000);
         self.maybe_update_installations(&provider, update_interval, client)
             .await?;
-        self.publish_intents(&provider, client).await?;
         self.sync_until_last_intent_resolved(&provider, client)
             .await?;
         Ok(())
@@ -927,10 +930,14 @@ impl MlsGroup {
         ApiClient: XmtpApi,
     {
         let conn = self.context.store.conn()?;
-        let intent = NewGroupIntent::new(IntentKind::KeyUpdate, self.group_id.clone(), vec![]);
-        intent.store(&conn)?;
+        let intent = conn.insert_group_intent(NewGroupIntent::new(
+            IntentKind::KeyUpdate,
+            self.group_id.clone(),
+            vec![],
+        ))?;
 
-        self.sync_with_conn(&conn.into(), client).await
+        self.sync_until_intent_resolved(&conn.into(), intent.id, client)
+            .await
     }
 
     pub fn is_active(&self, provider: impl OpenMlsProvider) -> Result<bool, GroupError> {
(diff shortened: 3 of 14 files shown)


neekolas:
* **#984** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

dev/docker/compose
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -eou pipefail
 
-docker-compose -f dev/docker/docker-compose.yml -p "libxmtp" "$@"
+docker compose -f dev/docker/docker-compose.yml -p "libxmtp" "$@"
```
neekolas, line 4: The GH action runner has been upgraded and no longer has a separate `docker-compose` command. It's now a subcommand of `docker` only.

xmtp_mls/src/groups/mod.rs
```diff
@@ -455,11 +464,6 @@ impl MlsGroup {
             Self::into_envelope(message, now)
         });
 
-        // Skipping a full sync here and instead just firing and forgetting
-        if let Err(err) = self.publish_intents(&provider, client).await {
```
neekolas, line 459: This gets handled inside `sync_until_intent_resolved`. No need to call it twice

xmtp_mls/src/groups/mod.rs
```diff
@@ -479,7 +483,6 @@ impl MlsGroup {
         let update_interval = Some(5_000_000);
         self.maybe_update_installations(&provider, update_interval, client)
             .await?;
-        self.publish_intents(&provider, client).await?;
```
neekolas, line 482: Same as above

xmtp_mls/src/groups/subscriptions.rs
```diff
@@ -66,7 +66,8 @@ impl MlsGroup {
         );
 
         if let Some(GroupError::ReceiveError(_)) = process_result.as_ref().err() {
-            self.sync(&client).await?;
+            self.sync_with_conn(&client.mls_provider()?, &client)
```
neekolas, line 69: This change alone is going to resolve a lot of thrash and noise. Previously we were calling sync when we receive a commit from a stream. That would attempt to add missing members first.

Every member of the group who was online would try and do it at the same time, leading to a bunch of commits that would get thrown out.

We don't need to do this as part of stream processing.

cameronvoell (approved):

xmtp_mls/src/groups/subscriptions.rs
cameronvoell: Nice catch. I'm thinking that another optimization would be to update `sync` so that if we sync during the `maybe_update_installations` process (`add_missing_installations` => `sync_until_intent_resolved`), we don't need to do a second `sync_with_conn` right afterwards.

xmtp_mls/src/groups/sync.rs
```diff
@@ -283,41 +293,46 @@ impl MlsGroup {
...
                     client,
                     conn,
-                    maybe_pending_commit.expect("already checked"),
+                    &pending_commit,
                     openmls_group,
                 )
                 .await;
 
                 if maybe_validated_commit.is_err() {
-                    log::warn!("error validating commit: {:?}", maybe_validated_commit);
```
richardhuaaa, line 313: Should we keep this log line? Think it might be important. It's unclear if this scenario is expected or not, my assumptions:
1. We validate the commit at the time we created the intent
2. If we are here, it means no new commits have happened since we last validated it

Therefore it seems unexpected, and should be logged as an error?

Separately to this PR, another case I'm unsure if we handle or not - we want to add [A, B] to the group, but somebody else adds [B, C] and their commit comes in before ours. Ideally we still add [A], even though [B] is no longer valid to be added.

xmtp_mls/src/groups/sync.rs
```diff
@@ -1309,6 +1350,22 @@ fn get_removed_leaf_nodes(
         .collect()
 }
 
+fn get_and_clear_pending_commit(
+    openmls_group: &mut OpenMlsGroup,
+    provider: &XmtpOpenMlsProvider,
+) -> Result<Option<Vec<u8>>, GroupError> {
+    // TODO: remove clone
+    if let Some(commit) = openmls_group.clone().pending_commit() {
```
richardhuaaa, line 1358: Wondering if we can use the commit as a reference rather than cloning the group?

xmtp_mls/src/groups/sync.rs
```diff
@@ -77,6 +78,13 @@ use xmtp_proto::xmtp::mls::{
     },
 };
 
+#[derive(Debug)]
+struct PublishIntentData {
+    staged_commit: Option<Vec<u8>>,
+    post_commit_action: Option<Vec<u8>>,
+    message: Vec<u8>,
```
richardhuaaa, line 85: I wonder if `payload` or `payload_to_publish` would be a clearer name here, as `message` could be confused with application message, and it also doesn't convey what stage of encryption/marshaling it is in (from looking at the rest of the code, I think it's fully encrypted and marshaled here)

xmtp_mls/src/groups/sync.rs
```diff
@@ -825,29 +835,41 @@ impl MlsGroup {
...
+                    log::debug!(
+                        "client [{}] set stored intent [{}] to state `published`",
+                        client.inbox_id(),
+                        intent.id
+                    );
 
                     client
                         .api_client
                         .send_group_messages(vec![payload_slice])
                         .await?;
```
richardhuaaa, line 861: Now that we're marking the intent as published before sending it, if the publish fails, either:
- Network error returned from `send_group_messages`
- App dies before the request goes through

Do we have any way of making sure the intent gets re-sent?

xmtp_mls/src/groups/sync.rs
```diff
@@ -825,29 +835,41 @@ impl MlsGroup {
...
-                    )?;
-                    log::debug!(
-                        "client [{}] set stored intent [{}] to state `published`",
-                        client.inbox_id(),
-                        intent.id
-                    );
+                    if has_staged_commit {
+                        log::info!("Canceling all further publishes, since a commit was found");
+                        return Err(GroupError::PublishCancelled);
+                    }
```
richardhuaaa, line 872: This works if this method (publish_intents) is only called once, but won't further publishes still happen if `publish_intents` is called again? Then we would skip past this published intent because it is no longer in `TO_PUBLISH` state?

richardhuaaa: Also, are we sure we want to return an error here, rather than returning success? Won't an error here stop us from calling `receive()` within `sync_with_conn()`, and finishing off the intent lifecycle?

neekolas:
### Merge activity

* **Aug 23, 3:37 PM PDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/984).

richardhuaaa (commented):
These are critical reliability improvements, well done. Left a few questions about things I'm unsure about, but I could be missing details

xmtp_mls/migrations/2024-08-22-044745_add_staged_commit/up.sql
```diff
@@ -0,0 +1,7 @@
+-- Your SQL goes here
```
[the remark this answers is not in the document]
neekolas: Good call

xmtp_mls/src/groups/sync.rs
neekolas: That's a good call. Easy enough to change.

xmtp_mls/src/groups/sync.rs
neekolas: I don't think we have a test for that, but I think it _should_ work. Worth investigating to find out.

I'll add the log line back in. Got a little aggressive pruning old code.

xmtp_mls/src/groups/sync.rs
neekolas: In the case of a network error we bubble the error back up to the caller and they can choose to retry.

In the case of the app dying, that's the big trade-off of this approach. It's just permanently in limbo. The difference from some of the alternative solutions is that it won't brick the group.

xmtp_mls/src/groups/sync.rs
neekolas: We continue through publish_intents errors in `sync_with_conn`. In the old method we would get them frequently if you had two intents lined up that both created commits, because they would trigger `PendingCommit` errors.

I did actually make that change in one of the down-stack PRs just to clean up error handling.

xmtp_mls/src/groups/sync.rs
neekolas: Compiler didn't like that because `openmls_group` is a mutable reference. I'm sure there's some hack to make it work.

xmtp_mls/src/groups/sync.rs
richardhuaaa: Just FYI, we are not bubbling it up in the case of `send_message<ApiClient>()`. And in the case of `send_message_optimistic`, the message would have already been rendered to the UI, and the error from the publish may or may not be bubbled up to the user

neekolas: > , we are not bubbling it up in the case of send_message<ApiClient>(). 

We would be because `sync_until_last_intent_resolved` will never finish

PR #985, opened 2024-08-22
Refactor process_own_message to not mutate intent state directly

neekolas:
## tl;dr

- Simplifies `process_own_message` to return the desired intent state, instead of having a bunch of DB operations happen in the method

### DIFF
--- a/xmtp_mls/src/groups/sync.rs
+++ b/xmtp_mls/src/groups/sync.rs
@@ -267,9 +267,9 @@ impl MlsGroup {
         message: ProtocolMessage,
         envelope_timestamp_ns: u64,
         allow_epoch_increment: bool,
-    ) -> Result<(), MessageProcessingError> {
+    ) -> Result<IntentState, MessageProcessingError> {
         if intent.state == IntentState::Committed {
-            return Ok(());
+            return Ok(IntentState::Committed);
         }
         let message_epoch = message.epoch();
         let group_epoch = openmls_group.epoch();
@@ -298,13 +298,12 @@ impl MlsGroup {
                     let group_epoch_u64 = group_epoch.as_u64();
 
                     if published_in_epoch_u64 != group_epoch_u64 {
-                        conn.set_group_intent_to_publish(intent.id)?;
                         log::warn!(
                             "Intent was published in epoch {} but group is currently in epoch {}",
                             published_in_epoch_u64,
                             group_epoch_u64
                         );
-                        return Ok(());
+                        return Ok(IntentState::ToPublish);
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


neekolas:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/985?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#985** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/985?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#984** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/984?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

cameronvoell (approved):

neekolas:
### Merge activity

* **Aug 23, 3:37 PM PDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/985).

xmtp_mls/src/groups/sync.rs
```diff
@@ -345,7 +343,7 @@ impl MlsGroup {
                 );
                 if let Err(err) = openmls_group.merge_staged_commit(&provider, pending_commit) {
                     log::error!("error merging commit: {}", err);
-                    conn.set_group_intent_to_publish(intent.id)?;
```
richardhuaaa, line 348: Interesting, I wonder if this was a bug before? In the previous code it would have continued to L371 and set it to committed

xmtp_mls/src/groups/sync.rs
```diff
@@ -612,26 +607,39 @@ impl MlsGroup {
...
+                    .await?
+                {
+                    IntentState::ToPublish => {
+                        Ok(provider.conn_ref().set_group_intent_to_publish(intent_id)?)
+                    }
+                    IntentState::Committed => {
+                        Ok(provider.conn_ref().set_group_intent_committed(intent_id)?)
+                    }
+                    IntentState::Published => {
+                        log::warn!("Unexpected behaviour: returned intent state published from process_own_message");
```
richardhuaaa, line 636: For unexpected behavior, I think we should log errors, so that we are more likely to see them. Especially if we choose not to update the state here

richardhuaaa (approved):

xmtp_mls/src/groups/sync.rs
neekolas: The problem is that we aren't returning the error in a lot of cases. We only return the error if we want to hard fail without changing the intent status (like a DB connection error). That'll get returned to the calling function, which does log it.

So, we have to log inside `process_own_message` before returning OK

richardhuaaa: What I meant is, as a general rule, we should do `log::error` instead of `log::warn` for cases we expect never to happen - not that we need to return an error to the caller

neekolas: Ahh, I see. I can update that

PR #990, opened 2024-08-23
Remove scoped futures

insipx:
I was not able to reproduce #987, but the only thing that really changed was the use of `ScopedFutures`, so I removed that in this PR and created a test. I also changed `join_all` to the more performant `FuturesUnordered` in the case that large amounts (30+) of groups were being synced at once.


neekolas (approved):

xmtp_mls/src/client.rs
```diff
@@ -881,6 +876,30 @@ mod tests {
...
+        let peers = test::create_bulk_clients(20).await;
+        let bo = peers[0].clone();
+
+        // create 50 groups with 5 messages each
+        test::create_groups(&alix, &peers, 50, 5).await.unwrap();
+        let groups = bo.sync_welcomes().await.unwrap();
+        assert_eq!(groups.len(), 50);
+        bo.sync_all_groups(groups).await.unwrap();
+
+        let groups = bo
```
neekolas, line 892: What is this really testing? The groups were already in the DB before `sync_all_groups` was called, so all it is really verifying is that the groups didn't get deleted along the way.

neekolas: If you added a second installation for Bo after creating the group, `sync_all_groups` would automatically add them to the group. Then you could verify after the sync that the installation was actually added.

insipx: ah you're right, I think I misunderstood how that worked :/ -- as far as this test goes, I was also curious how fast the syncing would be, but didn't want to spend time adding it to benchmarks yet. 

Since there's another test for syncing all groups, I'll remove this test and maybe revisit it for benchmarks in the future

PR #991, opened 2024-08-23
address review comments

neekolas:
## tl;dr

- Add some missing logging
- Rename a field on `PublishIntentData` to be more clear what the purpose is

### DIFF
--- a/xmtp_mls/src/groups/sync.rs
+++ b/xmtp_mls/src/groups/sync.rs
@@ -82,7 +82,7 @@ use xmtp_proto::xmtp::mls::{
 struct PublishIntentData {
     staged_commit: Option<Vec<u8>>,
     post_commit_action: Option<Vec<u8>>,
-    message: Vec<u8>,
+    payload_to_publish: Vec<u8>,
 }
 
 impl MlsGroup {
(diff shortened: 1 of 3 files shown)


neekolas:
* **#991** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/991?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

xmtp_mls/src/groups/sync.rs
```diff
@@ -324,7 +324,12 @@ impl MlsGroup {
                 )
                 .await;
 
-                if maybe_validated_commit.is_err() {
+                if let Err(err) = maybe_validated_commit {
+                    log::warn!(
```
richardhuaaa, line 328: I think it would be better to do `log::error` here, given we expect this to never happen, and the consequence is that we lose a commit

richardhuaaa (approved):

xmtp_mls/src/groups/sync.rs
neekolas: Done

PR #992, opened 2024-08-23
Sync all groups only syncs groups if they are active

cameronvoell:


bindings_ffi/src/mls.rs
```diff
@@ -2294,6 +2294,76 @@ mod tests {
...
+            .await
+            .unwrap()
+        {
+            group.send("hello".as_bytes().to_vec()).await.unwrap();
+        }
+
+        // Verify that sync all groups is faster after bo is removed
+        let start_time_2 = std::time::Instant::now();
+        bo.conversations().sync_all_groups().await.unwrap();
+        let duration_2 = start_time_2.elapsed();
```
neekolas, line 2357: I guess you have to do it this way because sync_all_groups returns (). 

What if we just made it return the number of groups synced? Feels like that would be useful beyond just test cases.

xmtp_mls/src/client.rs
```diff
@@ -607,18 +607,21 @@ where
...
                     async move {
+                        let mls_group = group.load_mls_group(provider_ref)?;
                         log::info!("[{}] syncing group", self.inbox_id());
                         log::info!(
                             "current epoch for [{}] in sync_all_groups() is Epoch: [{}]",
                             self.inbox_id(),
-                            group.load_mls_group(provider_ref)?.epoch()
+                            mls_group.epoch()
                         );
+                        if mls_group.is_active() {
```
neekolas, line 617: This is going to have merge conflicts with https://github.com/xmtp/libxmtp/pull/990

FYI @insipx

neekolas (approved):

bindings_ffi/src/mls.rs
cameronvoell: good call - added here https://github.com/xmtp/libxmtp/pull/992/commits/856ac68775c9087f21dbb7f9bf407e723840da5c

xmtp_mls/src/client.rs
```diff
@@ -593,32 +600,41 @@ where
...
+        let active_group_count = Arc::new(AtomicUsize::new(0));
+
         let sync_futures: Vec<_> = groups
             .into_iter()
             .map(|group| {
                 async {
                     // create new provider ref that gets moved, leaving original
                     // provider alone.
                     let provider_ref = &provider;
+                    let active_group_count = Arc::clone(&active_group_count);
```
nplasterer, line 620: Ah thanks for this fix. I had a thought this might be an issue and asked in slack. But should have validated. 🙏

PR #1008, opened 2024-08-28
Fix send update interval ns calculation, move to configuration

cameronvoell:


bindings_ffi/src/mls.rs
```diff
@@ -2464,6 +2464,8 @@ mod tests {
         // Recreate client2 (new installation)
         let client2 = new_test_client_with_wallet(wallet2).await;
 
+        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
```
insipx, line 2467: what happens if we don't sleep here?

As long as we're not spawning anything above, we shouldn't need to wait -- and if we do spawn something, it'd be better to devise a way to use `Notify` or `Barrier` + timeouts so that CI doesn't get tripped up

cameronvoell: Updated to make this more obvious 7f67702ee52a3ff59b80fa399bea8df5b910ad3a

This PR fixes an incorrect calculation that was causing us to sync installations on every message send, when it should only occur if 5 seconds has passed (or now whatever is stored in SEND_MESSAGE_UPDATE_INSTALLATIONS_INTERVAL_NS)

Let me know if that makes sense, or if you still think this should be updated.

richardhuaaa: Would it help to have some kind of test method that triggers installation syncing? Could maybe save the extra 10s this/other tests would need on each run

richardhuaaa: I suppose it depends on what we're testing - if we're testing the installation syncing logic, then we shouldn't do that

cameronvoell: > some kind of test method that triggers installation syncing?

this seems like a good call. especially if we end up increasing the update interval later on. Ill go ahead and add that.

insipx: > This PR fixes an incorrect calculation that was causing us to sync installations on every message send, when it should only occur if 5 seconds has passed (or now whatever is stored in SEND_MESSAGE_UPDATE_INSTALLATIONS_INTERVAL_NS)

That makes sense, good improvements!

richardhuaaa (approved):

PR #1009, opened 2024-08-28
Generate latest protos, update affected types

rygine:
# Summary

* Generated latest proto defs
* Fixed issues related to updated proto defs


xmtp_mls/src/groups/group_metadata.rs
```diff
@@ -46,6 +46,7 @@ impl GroupMetadata {
             conversation_type: conversation_type as i32,
             creator_inbox_id: self.creator_inbox_id.clone(),
             creator_account_address: "".to_string(), // TODO: remove from proto
+            dm_members: None,
```
rygine, line 49: @neekolas is this right?

xmtp_proto/Cargo.toml
```diff
@@ -24,7 +24,7 @@ grpc = ["tonic"]
 # @@protoc_deletion_point(features)
 # This section is automatically generated by protoc-gen-prost-crate.
 # Changes in this area may be lost on regeneration.
-proto_full = ["xmtp-identity","xmtp-identity-api-v1","xmtp-identity-associations","xmtp-keystore_api-v1","xmtp-message_api-v1","xmtp-message_contents","xmtp-mls-api-v1","xmtp-mls-database","xmtp-mls-message_contents","xmtp-mls_validation-v1"]
+proto_full = ["xmtp-identity","xmtp-identity-api-v1","xmtp-identity-associations","xmtp-keystore_api-v1","xmtp-message_api-v1","xmtp-message_contents","xmtp-mls-api-v1","xmtp-mls-database","xmtp-mls-message_contents","xmtp-mls_validation-v1","xmtp-xmtpv4"]
```
rygine, line 27: unsure on whether or not the v4 stuff should be included. and if so, how to exclude it.

xmtp_mls/src/groups/group_metadata.rs
neekolas: @cameronvoell is probably the best to ask

cameronvoell: Hey @rygine that looks correct. 

For context the dm_members libxmtp update [PR](https://github.com/xmtp/libxmtp/pull/901/files) was decided to go on a feature branch, but I forgot to leave a note for future proto updates that we could just set that to None in the mean time. sorry about the confusion.

xmtp_proto/Cargo.toml
richardhuaaa: We will need these for client changes eventually. It might be churny to include before we start on the client changes, but not a big deal either way

cameronvoell (approved):

Issue #996, opened 2024-08-27
Retry Pool Disconnection Errors


nplasterer:
a retry mechanism, if libxmtp hits a pool disconnected error, it could retry 3 times with a delay but never reconnect by itself

This is incase the app is trying to reconnect and it will reconnect eventually gives it a bit of a time delay to get that reconnection before fully erroring.

PR #1010, opened 2024-08-28
Retry PoolNeedsConnection Errors

nplasterer:
Makes `PoolNeedsConnection` Errors retryable for actions that require backgrounding the app while doing an action like receiving a signature on adding and revoking installations and applying a signature request.

Also adds a new error type that can be caught for the group create issue.

Fixes https://github.com/xmtp/libxmtp/issues/996


xmtp_mls/src/storage/errors.rs
```diff
@@ -68,6 +68,7 @@ impl RetryableError for StorageError {
             Self::Pool(_) => true,
             Self::Lock(_) => true,
             Self::SqlCipherNotLoaded => true,
+            Self::PoolNeedsConnection => true,
```
nplasterer, line 71: This means the error will retry 5 times before erroring. That might be okay? I also took a pass at just doing it in the code and retrying 3 times here https://github.com/xmtp/libxmtp/pull/1010/commits/45adbfe9ee1641694db84a3778a11700e16c5e2d open to others thoughts on how we should handle this.

neekolas: This definitely feels like it should be retryable. Thanks for fixing

alexrisch (approved):

insipx:
Just reposting here for visibility :


Was just thinking about this a little

During these sessions, is the XMTP client still in memory/we're not re-instantiating the struct?

Instead of retrying, which may still fail depending on the threading situation -- sortof feels like we're just waiting for a race which reconnects the db in another thread -- we could instead add a flag on the struct and a boolean passed to release_connection like `reconnect_on_reentry`, that would auto-re-establish a database connection instead of throwing an error. This seems to check if you're already calling to release the db connection. I don't fully understand what the app is back active callback is, though. would be happy to look through the sdk code for that if someone can link it.


also, this retry would only work if the reconnect is already in a codepath that is in a `retry!()` block. We could surround this in another `retry` block, but since this is a function used in lots of other places, may inadvertently cause double-retry times

nmalzieu:
> Just reposting here for visibility :
> 
> Was just thinking about this a little
> 
> During these sessions, is the XMTP client still in memory/we're not re-instantiating the struct?
> 
> Instead of retrying, which may still fail depending on the threading situation -- sortof feels like we're just waiting for a race which reconnects the db in another thread -- we could instead add a flag on the struct and a boolean passed to `release_connection` like `reconnect_on_reentry`, that would auto-re-establish a database connection instead of throwing an error. This seems to check if you're already calling to release the db connection. I don't fully understand what the `app is back active callback` is, though. would be happy to look through the sdk code for that if someone can link it.
> 
> also, this retry would only work if the reconnect is already in a codepath that is in a `retry!()` block. We could surround this in another `retry` block, but since this is a function used in lots of other places, may inadvertently cause double-retry times

Hey @insipx , I'm not 100% sure I get what you mean, but one thing we DON'T want is for libxmtp tu auto-re-establish a database connection. Because libxmtp has no idea if the app is active or backgrounded, and on iOS opening a connection to the libxmtp db when the app gets suspended will lead to a crash

xmtp_mls/src/groups/mod.rs
```diff
@@ -187,6 +187,8 @@ pub enum GroupError {
     SqlKeyStore(#[from] sql_key_store::SqlKeyStoreError),
     #[error("No pending commit found")]
     MissingPendingCommit,
+    #[error("Sync failed to wait for intent: {0}")]
+    SyncFailedToWait(String),
```
nplasterer, line 191: @neekolas adding this so it's easier to catch the error that is causing group creates to fail.

xmtp_mls/src/identity_updates.rs
```diff
@@ -259,9 +267,14 @@ where
...
-            .get_association_state(&self.store().conn()?, &inbox_id, None)
-            .await?;
+
+        let current_state = retry_async!(
+            Retry::default(),
+            (async {
+                self.get_association_state(&self.store().conn()?, &inbox_id, None)
+                    .await
+            })
+        )?;
```
nplasterer, line 277: This should cover the case converse is hitting when the database is trying to reconnect after getting a signature from a different app for revoking installations.

xmtp_mls/src/groups/mod.rs
```diff
@@ -188,6 +188,8 @@ pub enum GroupError {
     SqlKeyStore(#[from] sql_key_store::SqlKeyStoreError),
     #[error("No pending commit found")]
     MissingPendingCommit,
+    #[error("Sync failed to wait for intent: {0}")]
```
insipx, line 191:
```suggestion
    #[error("Sync failed to wait for intent")]
```

xmtp_mls/src/groups/mod.rs
```diff
@@ -188,6 +188,8 @@ pub enum GroupError {
...
+    SyncFailedToWait(String),
```
insipx, line 192:
```suggestion
    SyncFailedToWait,
```

xmtp_mls/src/groups/sync.rs
```diff
@@ -221,7 +221,9 @@ impl MlsGroup {
             num_attempts += 1;
         }
 
-        Err(last_err.unwrap_or(GroupError::Generic("failed to wait for intent".to_string())))
+        Err(last_err.unwrap_or(GroupError::SyncFailedToWait(
```
insipx, line 224:
```suggestion
        Err(last_err.unwrap_or(GroupError::SyncFailedToWait)
```

xmtp_mls/src/groups/sync.rs
```diff
@@ -221,7 +221,9 @@ impl MlsGroup {
...
+            "failed to wait for intent".to_string(),
```
insipx, line 225:
```suggestion
```

xmtp_mls/src/groups/sync.rs
```diff
@@ -221,7 +221,9 @@ impl MlsGroup {
...
+        )))
```
insipx, line 226:
```suggestion
        ))
```

insipx (approved):
Looks good!

When writing errors, the `#[error("display impl")]` is the `Display` impl for `to_string`, so an extra string is redundant

neekolas (approved):

Issue #941, opened 2024-08-05
Test Diesel ORM/WebAssembly libxmtp E2E


insipx:
- get `xmtp_mls` tests working in the browser with the wasm sqlite backend

PR #1026, opened 2024-09-03
compile `xmtp_id` and `xmtp_mls` to wasm32

### LINKED ISSUE #939: Fix `xmtp_mls` compile to `wasm32-unknown-unknown`
we can't have XMTP Mls library for web if the `xmtp_mls` crate does not compile to wasm. `xmtp_mls` currently has hard-dependences on some things which depend on C system libraries, namely:
    - libsqlite3
    - openssl
   
These need to be put under feature flags to allow for xmtp_mls to be included in projects with a `wasm32-unknown-unknown` target.

Some early work exists for this [here](https://github.com/xmtp/libxmtp/blob/5c198d0dbc8c7535ffbd9d6f7493576058ae90fc/xmtp_mls/Cargo.toml) which may be used as a reference/starting point

### LINKED ISSUE #940: Include `diesel-wasm-sqlite` crate in `xmtp_mls`
Requires #939 

Include the diesel WASM sqlite backend in `xmtp_mls` under a `web` feature flag. This may require some changes to diesel `Connection` traits for `AsyncConnection` from [diesel-async](https://github.com/weiznich/diesel_async). 

This is required for web, which cannot have any blocking calls.

### LINKED ISSUE #951: Fix `tokio` for `wasm32-unknown-unknown` target

_No response_

_No response_

_No response_

### Additional details

_No response_

insipx:
- feature flag/`Send` bounds

- slightly changes streams. All streams except for `_with_callback` are stack-based/no more `Box<Stream>`. This makes it easier to modify the `Send` bound for wasm32. 
- general dependency pruning where possible. 
    - Remove the `smart-default` crate since it was used in only one place.
    - replace `flume` with `tokio::sync::mpsc`
- ~blocked on some missing trait implementations for `WasmSqlite` backend which require a patched `diesel`.~
- remove `InboxOwner` in `xmtp_mls` and use the one in `xmtp_id`
    - this allows us to remove `xmtp_mls` dependency on `ethers`. `xmtp_id` depends on `ethers` anyway
- remove `ethers_core` and use the `ethers::core` re-export

90% of this PR is adding `cfg`/`cfg_attr` to tests/items + hoisting individual types like `Box<dyn Signature>` into type alias to make it possible to vary `Send/Sync` bounds based on wasm/not wasm

closes #951 closes #939 closes #940 

this PR makes tests compile in WebAssembly, but not necessarily pass. Rel: #941


xmtp_id/src/utils/mod.rs
```diff
@@ -1,3 +1,11 @@
-pub mod time;
+use wasm_timer::{SystemTime, UNIX_EPOCH};
```
insipx, line 1: wasm_timer uses normal `std` system_time libraries if on native, and the JS web_sys/js_sys shim if on web

xmtp_mls/src/owner/evm_owner.rs
```diff
@@ -1,16 +0,0 @@
-pub use ethers::signers::{LocalWallet, Signer};
-
-use xmtp_cryptography::signature::{h160addr_to_string, RecoverableSignature, SignatureError};
-
-use crate::InboxOwner;
-
-impl InboxOwner for LocalWallet {
```
insipx, line 7: I preferred the implementation in `xmtp_id`, enabling us to remove the `ethers` dependency in `xmtp_mls`

xmtp_mls/src/subscriptions.rs
```diff
@@ -236,8 +235,9 @@ where
     ) -> StreamHandle<Result<(), ClientError>> {
         let (tx, rx) = oneshot::channel();
 
-        let handle = tokio::spawn(async move {
-            let mut stream = client.stream_conversations().await?;
+        let handle = crate::spawn(async move {
+            let stream = client.stream_conversations().await?;
+            futures::pin_mut!(stream);
```
insipx, line 240: this should be a minor performance improvement, streams are entirely stack-based, even in the `with_callback` methods.

Issue #1025, opened 2024-09-03
failed to create group from welcome: storage error: Diesel result error: welcome id already exists


nplasterer:
failed to create group from welcome: storage error: Diesel result error: welcome id already exists

Getting this error randomly on iOS and Android when trying to stream messages to a group. 

Steps:
Create Group
Start Stream All Messages
Send message to group
Cancel Stream All Message
Send message to group
Start Stream All Messages
Send message to group 💥  --> blows up with `welcome id already exists`

This feels odd since no welcome should be getting created at this point?

nplasterer:
Seems suspicious that the group ids wouldn't match for a group that clearly exists. 🤔 

https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/src/storage/encrypted_store/group.rs#L224-L229

nplasterer:
Test still pass so probably not a huge deal.

neekolas:
Could happen if we are trying to process the same welcome more than one time.

nplasterer:
Maybe this should be a warning instead of a error.

PR #1031, opened 2024-09-04
Warn instead of error when syncing existing welcome id

cameronvoell:
Closes https://github.com/xmtp/libxmtp/issues/1025


xmtp_mls/src/client.rs
```diff
@@ -607,7 +607,20 @@ where
                                 match result {
                                     Ok(mls_group) => Ok(Some(mls_group)),
                                     Err(err) => {
-                                        log::error!("failed to create group from welcome: {}", err);
+                                        if let GroupError::Storage(StorageError::DieselResult(diesel::result::Error::DatabaseError(
+                                            diesel::result::DatabaseErrorKind::UniqueViolation,
+                                            ref error_info,
+                                        ))) = err
+                                        {
```
insipx, line 614: I feel like we should create a standalone function that accepts a `Result<T, Error>`, printing any kind of log messages for the error and returning `T` if no error. 

This way we don't need to burden our functions with these checks, and can re-use the function for other `UniqueViolations` or errors when needed

Also, we should have a comment on the function explaining that if diesel changes any of these error messages, then this message turns into a hard-error

nplasterer (approved):

PR #1041, opened 2024-09-06
Stream/General improvements

insipx:
Non-wasm specific improvements pulled from work from wasm:

- improve streams. Since [Rust 1.75](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html) `async fn` in trait is supported, along with a much lighter replacement to [`async_trait`](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html). 
    - This allows us to make our _native_ streams completely stack-based, removing `Pin<Box<dyn Stream>>`
    - removes `async-trait` from everywhere except a trait in `xmtp_id` which requires dynamic dispatch
    - removing `async-trait` should also generally improve compile times
    - more flexible stream traits with associated types
    - using stack-based streams, means we can get rid of lots of `clones()` and use references instead
    - sets us up nicely for these new upcoming rust [2024 edition](https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules.html) trait rules
- removes some unnecessary dependencies
    - `flume`, used in one place
    - `smart-default` used in once place

- removes an expensive clone we were doing in `stream_all_messages` preferring an `Arc<>` clone instead
- replaces `ethers_core` with `ethers::core`
- removes a random `unsafe` code block in `xmtp_id`
- unifies `trait InboxOwner` (uses one in xmtp_id, deletes elsewhere)


xmtp_mls/src/subscriptions.rs
```diff
@@ -167,48 +167,50 @@ where
...
+
         let messages_subscription = self.api_client.subscribe_group_messages(filters).await?;
 
         let stream = messages_subscription
             .map(move |res| {
-                let context = self.context.clone();
-                let client = self.clone();
-
-                let group_id_to_info = group_id_to_info.clone();
+                let group_info = group_id_to_info.clone();
```
insipx, line 190: Previously, this would clone the entire `HashMap<>` for every single group message, where the key is a `Vec<u8>`

xmtp_mls/src/groups/subscriptions.rs
```diff
@@ -121,21 +118,21 @@ impl MlsGroup {
...
     }
 
-    pub async fn stream<ApiClient>(
-        &self,
-        client: Arc<Client<ApiClient>>,
-    ) -> Result<Pin<Box<dyn Stream<Item = StoredGroupMessage> + Send + '_>>, GroupError>
+    pub async fn stream<'a, ApiClient>(
+        &'a self,
+        client: &'a Client<ApiClient>,
+    ) -> Result<impl Stream<Item = StoredGroupMessage> + '_, GroupError>
```
insipx, line 124: This lifetime capture (`+ '_`) becomes the default behavior in rust 2024 edition + allows us to attach more flexible lifetime captures wherever we use `impl Trait`, which will allow eliding the `'a`lifetime here

nplasterer (approved):
Nice!

neekolas (approved):
This is a huge set of improvements

PR #1046, opened 2024-09-10
DM updates - default to not displaying dm groups

cameronvoell:
See https://github.com/xmtp/libxmtp/issues/1074

As a pre-requisite for dual sending 1 to 1 messages, we will be creating DM groups that we do not want to show up for users until we switch to V3 only clients. This PR filters DM groups from list and stream conversations functions that are used in bindings and downstream SDKs


nplasterer (approved):
Maybe you covered this somewhere but is there a way to override the filter out functionality? So for these pure V3 clients that they can still get conversations?

bindings_ffi/src/mls.rs
```diff
@@ -790,7 +805,16 @@ impl FfiConversations {
 
     pub async fn sync_all_groups(&self) -> Result<u32, GenericError> {
         let inner = self.inner_client.as_ref();
-        let groups = inner.find_groups(None, None, None, None)?;
+        let groups = inner.find_groups(FindGroupParams {
+            include_dm_groups: true,
+            ..FindGroupParams::default()
+        })?;
+
+        println!(
```
insipx, line 813:
```suggestion
        log::info!(
```

xmtp_mls/src/subscriptions.rs
```diff
@@ -242,7 +243,12 @@ where
             futures::pin_mut!(stream);
             let _ = tx.send(());
             while let Some(convo) = stream.next().await {
-                convo_callback(convo)
+                let provider = client.context.mls_provider()?;
+                // Don't execute callback for dms unless include_dm is true
+                if include_dm || convo.metadata(provider)?.conversation_type != ConversationType::Dm
```
insipx, line 248: it would be better to do this filter in `stream_conversations` itself rather than the callback method -- the callback methods should retain the same behavior as the Non-callback variants, but just accept a function instead. this makes the logic for both easier to follow

Here is a place you could insert a check/filter for DM Group (i.e another filter/filter map call just above that, or even just save the iterator and if the dm_group flag is on, filter the iterator again): https://github.com/xmtp/libxmtp/blob/998623d7d3a9fd1a1ecdc08779a144f6b9c53105/xmtp_mls/src/subscriptions.rs#L160

cameronvoell: good call 👌 , fixed here fc02806f7e7f47b402bb019c56379112d4fda97e

cameronvoell:
> Maybe you covered this somewhere but is there a way to override the filter out functionality? So for these pure V3 clients that they can still get conversations?

@nplasterer For streams to include Dm groups, we just need to pass in true to `stream_conversations_with_callback` here:

https://github.com/xmtp/libxmtp/blob/5d58c4c731b5e52ba14f48ac7be0eec8c19b6f1b/bindings_ffi/src/mls.rs#L864

For `list` to include Dm groups, we need to pass in true for `include_dm_groups` in  `FindGroupParams` struct here:
https://github.com/xmtp/libxmtp/blob/eb506f4e982cf3d07c42c51b049511e9911869cf/bindings_ffi/src/mls.rs#L839

PR #1055, opened 2024-09-11
Consent Database

nplasterer:
Pulled this over from here https://github.com/xmtp/libxmtp/pull/946 to the main branch since this enables group chat consent and should probably be separate from the mls dms work.


cameronvoell (approved):
looks good, agreed changes from https://github.com/xmtp/libxmtp/pull/946 are prob better to separate from the `mls-dm-release` feature branch 👍

PR #1059, opened 2024-09-12
Allocate sql queries in wasm memory space

codabrink:
Skip an extra copy per query in JS space by allocating directly in wasm memory space.

Handles #1027


insipx:
First commit!

insipx (approved):

PR #1024, opened 2024-08-31
added `serialize` and `deserialize` functions to `diesel-wasm-sqlite` backend

vking45:
Resolves #1003 

This PR has been made after doing thorough research on the serialize and deserialize sqlite3 functions, all the necessary changes have been made. 

- [x] added new foreign function interfaces in `ffi.rs`
- [x] implemented the logic for the function in both `package.js` and `connection/raw.rs`

Please let me know if there are any changes/fixes needed to this PR.

Issue #1003, opened 2024-08-28
add `serialize` and `deserialize` functions to `diesel-wasm-sqlite` backend


insipx:
Serialize docs: https://sqlite.org/c3ref/serialize.html
Deserialize docs: https://sqlite.org/c3ref/deserialize.html
SQLite WASM: https://www.sqlite.org/wasm
sqlite3 WebAssembly C-Style API documentation: https://www.sqlite.org/wasm/doc/trunk/api-c-style.md
`wasm-bindgen` (rust -> js ffi) docs: https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/index.html 

the `serialize` and `deserialize` methods on `RawConnection` are commented out:
https://github.com/xmtp/libxmtp/blob/bb18b61df2e310b7c23ad222cb2c167fa9b28ae9/diesel-wasm-sqlite/src/connection/raw.rs#L89

They still use the raw C-bindings from `libsqlite3-sys` rust crate. We want to convert these to use the WASM bindings.

- Add the required methods to `ffi.rs` and `package.json` for serialize/deserialize
- add the methods to `raw.rs`

vking45:
Hey, I would like to work on this issue. Can you please assign this issue to me ?

codabrink:
Picking up this issue where @vking45 left off.

PR #1060, opened 2024-09-12
WASM Diesel Serialize/Deserialize

codabrink:
Issue #1003 continuation of PR #1024


diesel-wasm-sqlite/src/connection/raw.rs
```diff
@@ -84,45 +90,60 @@ impl RawConnection {
...
         unsafe {
-            let mut size: ffi::sqlite3_int64 = 0;
-            let data_ptr = ffi::sqlite3_serialize(
-                self.internal_connection.as_ptr(),
-                std::ptr::null(),
-                &mut size as *mut _,
-                0,
-            );
-            SerializedDatabase::new(data_ptr, size as usize)
+            ffi::raw_copy_from_sqlite(data_ptr, len as u32, data.as_mut_slice());
```
insipx, line 108: since we're copying this data from sqlite, we can probably call the `sqlite3_free` on the `data_ptr` returned from `sqlite3_serialize` after copying, so that sqlite doesn't carry this data around anymore

diesel-wasm-sqlite/src/ffi.rs
```diff
@@ -82,11 +82,13 @@ pub fn raw_copy_to_sqlite<B: Into<Uint8Array>>(bytes: B, dst: *mut u8) {
 pub unsafe fn raw_copy_from_sqlite(src: *mut u8, len: u32, buf: &mut [u8]) {
     let wasm = crate::get_sqlite_unchecked().inner().wasm();
     let mem = wasm.heap8u();
-    let offset = (src as u32) / std::mem::size_of::<u8>() as u32;
-    // this is safe because we view the slice and immediately copy it into
-    // our memory.
-    let view = Uint8Array::new_with_byte_offset_and_length(&mem, offset, len);
-    view.raw_copy_to_ptr(buf.as_mut_ptr())
+    mem.slice(src as u32, src as u32 + len)
```
insipx, line 85: this is interesting! I wonder why the other method didn't work. It seems to be doing the same/similar thing

insipx (approved):

diesel-wasm-sqlite/src/connection/raw.rs
codabrink: I was freeing it, but I was running into out of bounds issues. I believe it's running into a race condition. I'll play around with it to see if I can get that fixed.

codabrink: Oh, I realized I was freeing from the wrong memory space. It's fixed now.

PR #1062, opened 2024-09-13
Introduce gRPC replication client api

mkysel:


xmtp_api_grpc/src/grpc_api_helper.rs
```diff
@@ -335,12 +351,22 @@ impl MutableApiSubscription for GrpcMutableSubscription {
 impl XmtpMlsClient for Client {
     #[tracing::instrument(level = "trace", skip_all)]
     async fn upload_key_package(&self, req: UploadKeyPackageRequest) -> Result<(), Error> {
-        let client = &mut self.mls_client.clone();
+        if self.use_replication_v4 {
```
insipx, line 354: This is OK, we could also do something like this:

1.) define a feature in api_grpc, `replication`
2.) Create a new `impl XmtpMlsClient` definition exactly like the non-v4 one but for v4
3.) use `#[cfg(feature = "replication")]` and `#[cfg(not(feature = "replication"))]`


This would feature-gate replication, and result in easy access to the client through a flag during testing/compiling, i.e `cargo test --feature replication`. Mixing replication and v3 backend also isn't terribly useful in itself, so I would argue for using compile-time features here (i'm also just a fan of compile time features)

mkysel: I was considering that option as well. I decided against it, because I want the ability to point the client (or any SDK implementation) against any backend using run-time configuration. Otherwise you have to compile different binaries or recompile every time you switch the backend. This makes shipping the binaries a bit annoying.

Maybe a good idea would be to hide the new implementation behind a feature gate. Basically hide the new experimental code behind a feature thats off by default. And allow developers to compile it in, if they so desire.

We might have to mix v3 and v4 in some cases unfortunately.

richardhuaaa (commented):
Looks good so far, just left a bunch of nits!

Depending on how far you want to go, we could also consider modifying the setup for existing unit tests to test both API clients, as a shortcut for writing your own tests. Not super critical at this stage either way though

mkysel:
Superseded by a combo of newer PRs

Issue #1063, opened 2024-09-13
Mutate identity signature request after registering a client


rygine:
After registering a new client, calling `client.identity().is_ready()` returns `false` because the underlying `signature_request` is not updated during registration.

PR #1066, opened 2024-09-16
Mark client as ready after registering

codabrink:
Solves #1063

### DIFF
--- a/xmtp_mls/src/builder.rs
+++ b/xmtp_mls/src/builder.rs
@@ -148,6 +148,8 @@ where
 
 #[cfg(test)]
 mod tests {
+    use std::sync::atomic::AtomicBool;
+
     use crate::api::ApiClientWrapper;
     use crate::builder::ClientBuilderError;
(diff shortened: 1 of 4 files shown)


xmtp_mls/src/builder.rs
```diff
@@ -529,7 +530,7 @@ mod tests {
             inbox_id: inbox_id.clone(),
             installation_keys: SignatureKeyPair::new(SignatureScheme::ED25519).unwrap(),
             credential: Credential::new(CredentialType::Basic, rand_vec()),
-            signature_request: None,
+            signature_request: Mutex::default(),
```
insipx, line 533: Is there a way we can pull this out of `xmtp_mls`, and use a `Mutex` directly in `bindings_ffi` and `bindings_node`?

codabrink: Switching to an atomic to avoid potential deadlocks in the future.

insipx (approved):
nice job!

PR #1067, opened 2024-09-16
Bulk set consent records

nplasterer:
In this interim period it will be easier to bulk set consent.

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -313,15 +313,31 @@ where
         &self,
         address: String,
     ) -> Result<Option<String>, ClientError> {
-        if let Some(sanitized_address) = sanitize_evm_addresses(vec![address])?.pop() {
-            let mut results = self
-                .api_client
-                .get_inbox_ids(vec![sanitized_address.clone()])
-                .await?;
-            Ok(results.remove(&sanitized_address))
+        let results = self
+            .find_inbox_ids_from_addresses(vec![address.clone()])
+            .await?;
+        if let Some(first_result) = results.into_iter().next() {
+            Ok(first_result)
         } else {
             Ok(None)
         }
+    }
+
+    pub async fn find_inbox_ids_from_addresses(
+        &self,
+        addresses: Vec<String>,
+    ) -> Result<Vec<Option<String>>, ClientError> {
+        let sanitized_addresses = sanitize_evm_addresses(addresses.clone())?;
+        let mut results = self
+            .api_client
+            .get_inbox_ids(sanitized_addresses.clone())
+            .await?;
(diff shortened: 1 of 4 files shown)


xmtp_mls/src/storage/encrypted_store/consent_record.rs
```diff
@@ -57,14 +57,14 @@ impl DbConnection {
...
+    /// Insert consent_records, and replace existing entries
+    pub fn insert_or_replace_consent_records(
         &self,
-        record: StoredConsentRecord,
+        records: Vec<StoredConsentRecord>,
     ) -> Result<(), StorageError> {
         self.raw_query(|conn| {
             diesel::insert_into(dsl::consent_records)
-                .values(&record)
+                .values(&records) // Use a vector of records instead of a single record
```
nplasterer, line 67: @insipx sqlite doesnt allow this. Would just wrapping this in a transaction and keeping it as before doing it one by one the most performant way?

nplasterer:
I don't know how big of a deal this is but we iterate through a whole list of consent and set them in the database every time you load V2 consent in android. It felt like it might be faster to just do this in a transaction all at once. https://github.com/xmtp/xmtp-android/pull/297/files#r1759609079

Open to others thoughts if they feel this isn't necessary.

xmtp_mls/src/storage/encrypted_store/consent_record.rs
insipx: Not sure if below about consent in android is related but wanted to put some context:


this kind of insertion for SQLite database is standard, and the same way diesel works for normal bulk inserts: https://github.com/diesel-rs/diesel/blob/f7e9819303877b5a255143c4b32fa1e5745aaad0/diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs#L241

With WAL logging, this will still result in one overall write to the DB, so would be the best way to get this done I think

xmtp_mls/src/client.rs
```diff
@@ -344,28 +344,38 @@ where
...
-                ))?;
+        // Collect new records to add after the loop
+        let mut new_records = Vec::new();
+
+        // Iterate over the records and check if the entity is an address
+        for record in &records {
+            if record.entity_type == ConsentType::Address {
+                // Find the inbox ID associated with the address asynchronously
+                if let Some(inbox_id) = self
+                    .find_inbox_id_from_address(record.entity.clone())
```
neekolas, line 361: Would be nice to batch this call, since it goes to the network

nplasterer: Good callout! I just added the batched API call.

insipx (approved):
looks good!

PR #1072, opened 2024-09-17
Add message-history feature to xmtp_mls

rygine:
# Summary

* Added `message-history` to `xmtp_mls` crate
* Gated all message history code behind `message-history` feature
* Enabled `message-history` feature for CLI and all bindings

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -223,6 +223,7 @@ impl From<&str> for ClientError {
 pub struct Client<ApiClient> {
     pub(crate) api_client: ApiClientWrapper<ApiClient>,
     pub(crate) context: Arc<XmtpMlsLocalContext>,
+    #[cfg(feature = "message-history")]
     pub(crate) history_sync_url: Option<String>,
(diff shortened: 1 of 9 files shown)


xmtp_mls/src/client.rs
```diff
@@ -465,6 +467,7 @@ where
         Ok(group)
     }
 
+    #[cfg(feature = "message-history")]
```
rygine, line 470: not sure about this one. AFAIK, sync groups can/will be used for other purposes.

xmtp_mls/src/groups/mod.rs
```diff
@@ -405,6 +408,7 @@ impl MlsGroup {
         Self::create_from_welcome(client, provider, welcome, inbox_id, welcome_id).await
     }
 
+    #[cfg(feature = "message-history")]
```
rygine, line 411: same as other comment

xmtp_mls/src/client.rs
neekolas: Fine to gate for now, and if it breaks something we can always un-gate.

neekolas (approved):

Issue #1044, opened 2024-09-09
Expose installation created at date


nplasterer:
It would be helpful to understand when an installation was created.

nplasterer:
Exposed out through the sdks! 🎉

PR #1073, opened 2024-09-18
Provide timestamps in identity action updates

codabrink:
Addresses #1044


neekolas (approved):
This looks good to me

In case you're wondering why we have both a `client_timestamp_ns` and a `server_timestamp_ns`, maybe this is a good time to explain. When you create the `IdentityUpdate` we include a timestamp included the text that gets signed, so that you can do the same action multiple times and get a unique signature. That lets you associate a wallet, revoke it, and then re-associate the same wallet. If two actions have identical signatures, we will reject the second one.

The server_timestamp_ns only comes once you've uploaded the signed request to our nodes.

xmtp_mls/src/identity_updates.rs
```diff
@@ -117,10 +117,18 @@ where
...
-        if let Some(association_state) =
-            StoredAssociationState::read_from_cache(conn, inbox_id.to_string(), last_sequence_id)?
+        match StoredAssociationState::read_from_cache(conn, inbox_id.to_string(), last_sequence_id)
         {
-            return Ok(association_state);
+            Ok(Some(state)) => {
+                return Ok(state);
+            }
+            // If the cache fails to deserialize, it's because the structure of the cache
+            // has changed, and we need to reload.
```
cameronvoell, line 126: I'm not 100% sure that deserialization will fail if the `AssociationStateProto` (via the `MemberProto`) has been [updated](https://github.com/xmtp/proto/pull/208/files) to include a new optional field. So I think we still need a way to skip the cache after this update.

codabrink: The bytes from the sql blob won't align with the proto. It should fail.

neekolas: That sounds right. Protobuf is very forgiving

codabrink: Created a migration to clear out the table in question.

xmtp_mls/src/identity_updates.rs
```diff
@@ -117,10 +117,18 @@ where
             return Err(AssociationError::MissingIdentityUpdate.into());
         }
 
-        if let Some(association_state) =
-            StoredAssociationState::read_from_cache(conn, inbox_id.to_string(), last_sequence_id)?
+        match StoredAssociationState::read_from_cache(conn, inbox_id.to_string(), last_sequence_id)
```
neekolas, line 120: We can undo this change as well

codabrink: Done

xmtp_mls/src/groups/members.rs
```diff
@@ -48,19 +53,28 @@ impl MlsGroup {
...
-            log::error!(
-                "Failed to load all members for group - metadata: {:?}, computed members: {:?}",
-                requests,
-                association_states
-            );
-            return Err(GroupError::InvalidGroupMembership);
+            // Attempt to rebuild the cache before erroring out due to a cache miss.
+            let requests: Vec<_> = requests
+                .into_iter()
+                .map(|(id, sequence)| (id, Some(sequence)))
```
neekolas, line 63: In the narrow case of your PR, this is fine. But really we should only be doing these lookups for items that are missing from the cache.

codabrink: Got it. Filtering those out now.

xmtp_id/src/associations/state.rs
```diff
@@ -163,13 +163,9 @@ impl AssociationState {
...
         let identifier = MemberIdentifier::Address(account_address.clone());
-        let new_member = Member::new(identifier.clone(), None);
+        let new_member = Member::new(identifier.clone(), None, None);
         Self {
-            members: {
-                let mut members = HashMap::new();
-                members.insert(identifier, new_member);
-                members
-            },
+            members: HashMap::from_iter([(identifier, new_member)]),
```
neekolas, line 168: Nice cleanup

neekolas (approved):
Heads up @nplasterer this is going to change the `group.members()` function to be async, which will need to get updated in iOS and Android.

cameronvoell (approved):
great work here!

cameronvoell:
#1044 wont be finished until we expose the timestamps for installations in the bindings inbox_state function, but I advise we do that in a follow up PR since changes in here are quite large

PR #1091, opened 2024-09-25
Expose Installations with timestamps on the FFI

codabrink:


nplasterer (approved):
This looks right! Thanks for fixing it up.

PR #1096, opened 2024-09-30
consistent logging to enable better debugging

insipx:
Makes logging more consistent by only using the `tracing` crate wherever possible. the `tracing` crate provides contextual tracing useful for debugging async code, but can also fallback to emitting `log` events if no subscriber is set.

- removes `tracing-test` crate preferring a small custom closure to capture logs from tests instead
    - this closure only works with `tokio`s `current_thread` runtime. It can be made to work with a multi-thread environment, but not needed since its only used in a couple places so far. could make use of `tracing-subscriber` [reload](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/reload/index.html) handle with rusts `LazyLock` global initializer to temporarily switch out a custom global subscriber that logs to a internal memory buffer only in the closure.
    - removing this allows us to avoid setting the `traced-test` global subscriber in `ctor`, giving us the ability to filter logs during debugging
- Default log level for tests is still INFO
- Logs in tests can be filtered on module again, which is useful for debugging
   - EX: if i'm debugging something related to sync, I can do `RUST_LOG=xmtp_mls::groups=DEBUG` to only get logs from `xmtp_mls/src/groups/mod.rs`
- `tracing` also has useful adapters for debugging WebAssembly code
- removes `env_logger` from `mls_validation` service preferring `tracing-subscriber`

`bindings_ffi` still uses the `log` crate. The `FfiLogger` trait doesn't fit well into the `tracing::Subscriber` trait -- enabling tracing there could be a different issue/PR but for now the benefit would be small to switch it out

### DIFF
--- a/.github/workflows/lint-ffi-bindings.yaml
+++ b/.github/workflows/lint-ffi-bindings.yaml
@@ -1,16 +1,15 @@
 name: Lint FFI Bindings
-
 on:
   push:
     branches:
       - main
-
   pull_request:
     paths:
       - "bindings_ffi/**"
       - ".github/workflows/lint-ffi-bindings.yaml"
       - "rustfmt.toml"
-
+env:
+  CARGO_TERM_COLOR: always
 jobs:
   lint:
     name: Lint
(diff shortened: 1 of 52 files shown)


codabrink:
Thank you!

codabrink (approved):

PR #1098, opened 2024-09-30
MLS Validation Service for SCW

codabrink:


bindings_ffi/src/mls.rs
```diff
@@ -203,17 +203,22 @@ impl FfiSignatureRequest {
     pub async fn add_scw_signature(
         &self,
         signature_bytes: Vec<u8>,
-        address: String,
+        account_address: String,
         chain_id: u64,
         block_number: u64,
+        hash: Vec<u8>,
```
neekolas, line 209: The hash should not be passed in here. It can come from hashing the inner signature text for security.

xmtp_mls/src/credential/grant_messaging_access_association.rs
```diff
@@ -0,0 +1,290 @@
+use chrono::DateTime;
```
neekolas, line 1: We can still remove this. We shouldn't need it anywhere in the code anymore, right?

We want to keep the old APIs up for _old_ SDK versions, but the current SDK can leave them stubbed out

xmtp_mls/src/credential/legacy_create_identity_association.rs
```diff
@@ -0,0 +1,211 @@
+use prost::Message;
```
neekolas, line 1: Same as above. We can leave this code out

xmtp_mls/src/credential/mod.rs
```diff
@@ -0,0 +1,185 @@
+mod grant_messaging_access_association;
```
neekolas, line 1: Same as above. Can leave deleted

xmtp_mls/src/verified_key_package.rs
```diff
@@ -0,0 +1,116 @@
+use openmls::{
```
neekolas, line 1: Can remove

bindings_wasm/src/mls_client.rs
```diff
@@ -164,9 +164,8 @@ impl WasmClient {
     signature_bytes: Uint8Array,
     chain_id: u64,
     account_address: String,
-    // TODO:nm Remove this
-    _chain_rpc_url: String,
     block_number: u64,
+    hash: Uint8Array,
```
neekolas, line 168: Same as below, we can get the hash from the inner signature request

bindings_ffi/src/mls.rs
codabrink: Got it

mls_validation_service/src/handlers.rs
```diff
@@ -158,6 +187,30 @@ async fn validate_inbox_id_key_package(
...
 }
 
+async fn verify_smart_contract_wallet_signatures(
+    signatures: Vec<SmartContractWalletSignature>,
+    scw_verifier: &dyn SmartContractSignatureVerifier,
+) -> Result<Response<VerifySmartContractWalletSignaturesResponse>, Status> {
+    let mut futures = vec![];
+
+    for sig in signatures {
+        let hash = sha256_bytes(&sig.signature)
```
neekolas, line 197: The hash should be of the signature text from the inner signature request

xmtp_id/src/associations/unverified.rs
```diff
@@ -344,14 +348,24 @@ pub struct UnverifiedSmartContractWalletSignature {
...
 }
 
 impl UnverifiedSmartContractWalletSignature {
-    pub fn new(signature_bytes: Vec<u8>, account_id: AccountId, block_number: u64) -> Self {
+    pub fn new(
+        signature_bytes: Vec<u8>,
+        account_id: AccountId,
+        block_number: u64,
+        chain_id: u64,
+        hash: [u8; 32],
```
neekolas, line 361: None of the other unverified signature types store the hash, because the hash gets provided as part of signature verification (check out the `to_verified`) method of `IdentityUpdate`. 

We should really keep that consistent or it's going to cause confusion for devs.

insipx (commented):
Looks good!

Should we wait until https://github.com/xmtp/proto/pull/212 is merged to merge this?

doing `./dev/gen_protos` on this branch gives a different history then the generated protos, is that something to worry about?

mls_validation_service/Cargo.toml
```diff
@@ -10,6 +10,8 @@ path = "src/main.rs"
 [dependencies]
 clap = { version = "4.4.6", features = ["derive"] }
 ed25519-dalek = { workspace = true, features = ["digest"] }
+env_logger = "0.11"
```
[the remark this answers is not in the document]
codabrink: Oops, bad merge. Good call out.

codabrink:
@insipx Correct, I'd like to merge the proto branch (https://github.com/xmtp/proto/pull/211) before this.

mls_validation_service/src/handlers.rs
```diff
@@ -158,6 +198,43 @@ async fn validate_inbox_id_key_package(
     })
 }
 
+async fn verify_smart_contract_wallet_signatures(
+    signatures: Vec<UnverifiedSmartContractWalletSignature>,
+    scw_verifier: &dyn SmartContractSignatureVerifier,
+) -> Result<Response<VerifySmartContractWalletSignaturesResponse>, Status> {
+    let mut responses = vec![];
+    for request in signatures {
+        let Some(signature) = request.scw_signature else {
```
neekolas, line 207: If the client is expecting to zip together the requests and the responses by index then we should insert a `false` here rather than skipping the entry. Otherwise the request at index 1 might get matched with the response at index 2.

codabrink: Do you know why this field is optional at all? It's not declared as optional in the protos, but comes in as optional in the generated spec. https://github.com/xmtp/proto/pull/211/files#diff-a8714baf14c9f3a47e527cf66f630eb43d2e063fc68413d73e8ee5bf561fc44eR62

codabrink: Got it, that's an interesting limitation. I was aware of the mismatched index issue, and was perplexed about that being generated as optional when the spec wasn't marked as such, and meant to bring it up. I'll switch this logic out to put in a false where there's a `None`.

codabrink: Updated

neekolas (approved):

PR #1101, opened 2024-10-01
Update node bindings

rygine:
# Summary

- Added `getLatestInboxState` to client
- Added installation timestamps to `inboxState`
- Updated `send_optimistic` to return the message ID as a hex string
- Added consent state methods to groups and client
- Updated CHANGELOG and version

### DIFF
--- a/bindings_node/src/groups.rs
+++ b/bindings_node/src/groups.rs
@@ -107,7 +107,7 @@ impl NapiGroup {
 
   /// send a message without immediately publishing to the delivery service.
   #[napi]
-  pub fn send_optimistic(&self, encoded_content: NapiEncodedContent) -> Result<Vec<u8>> {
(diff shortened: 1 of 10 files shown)


bindings_node/src/groups.rs
```diff
@@ -119,7 +121,7 @@ impl NapiGroup {
       .send_message_optimistic(encoded_content.encode_to_vec().as_slice())
       .map_err(ErrorWrapper::from)?;
 
-    Ok(id)
+    Ok(hex::encode(id.clone()))
```
insipx, line 124: maybe not important, but keep in mind that `hex::encode` will return a string without `0x`, it could trip some apps that expect that. `format!("0x{}", hex::encode(id.clone())` could add the `0x`

insipx (approved):

nplasterer (approved):

insipx (approved):

PR #1112, opened 2024-10-03
Make scw block number optional for front-end

codabrink:
We want to make the block number optional for the front-end since it can often be difficult for them and we can easily retrieve the latest number.

To keep the block number optionality close to the surface, I expose the current block number functionality from the scw verifier in the `SmartContractSignatureVerifier` trait, to allow for immediate retrieval near the surface of the request when `block_number` is `None`.


mls_validation_service/src/handlers.rs
```diff
@@ -220,7 +219,7 @@ async fn verify_smart_contract_wallet_signatures(
                         GrpcServerError::Deserialization(DeserializationError::InvalidHash)
                     })?,
                     signature.signature.into(),
-                    Some(BlockNumber::Number(U64::from(signature.block_number))),
+                    None,
```
codabrink, line 222: This might need more thought.

xmtp_id/src/associations/serialization.rs
```diff
@@ -174,7 +174,7 @@ impl TryFrom<SignatureWrapperProto> for UnverifiedSignature {
                 UnverifiedSmartContractWalletSignature::new(
                     sig.signature,
                     sig.account_id.try_into()?,
-                    sig.block_number,
+                    None,
```
codabrink, line 177: Same here.

xmtp_id/src/scw_verifier/chain_rpc_verifier.rs
```diff
@@ -87,14 +87,20 @@ impl SmartContractSignatureVerifier for RpcSmartContractWalletVerifier {
         ];
         let data = constructor.encode_input(code, tokens)?;
         let tx: TypedTransaction = TransactionRequest::new().data(data).into();
-        let block_number = if let Some(number) = block_number {
-            number
-        } else {
-            BlockNumber::Number(self.provider.get_block_number().await?)
+        let block_number = match block_number {
+            Some(bn) => bn,
+            None => BlockNumber::Number(self.current_block_number(&signer).await?),
```
insipx, line 92: why do we need to pass in the signer here?

xmtp_id/src/scw_verifier/chain_rpc_verifier.rs
```diff
@@ -87,14 +87,20 @@ impl SmartContractSignatureVerifier for RpcSmartContractWalletVerifier {
...
         };
         let res = self.provider.call(&tx, Some(block_number.into())).await?;
         Ok(res == Bytes::from_hex("0x01").expect("Hardcoded hex will not fail."))
     }
+
+    async fn current_block_number(&self, _account_id: &AccountId) -> Result<U64, VerifierError> {
```
insipx, line 98: I like this approach, but we shouldn't need the account_id to get a block_number


The block number is independent of any account, although it will differ based on `chain_rpc_url` so I think the rest is correct but we can remove this part

insipx (changes requested):

xmtp_id/src/scw_verifier/chain_rpc_verifier.rs
codabrink: That's how I had it initially, but saw that the account_id is being used to get the appropriate `verifier` [here](https://github.com/xmtp/libxmtp/pull/1112/files#diff-696d43c0087a23db3e463f659837d15a5fedf997f09aaedb9aaa612b9ac5b697R139)

xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -116,16 +121,24 @@ impl SmartContractSignatureVerifier for MultiSmartContractSignatureVerifier {
...
 
         todo!()
     }
+
+    async fn current_block_number(&self, chain_id: &str) -> Result<U64, VerifierError> {
+        let id: u64 = chain_id.parse().unwrap();
+        if let Some(verifier) = self.verifiers.get(&id) {
+            return verifier.current_block_number(chain_id).await;
+        }
+        todo!()
```
insipx, line 142:
```suggestion
```

xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -116,16 +121,27 @@ impl SmartContractSignatureVerifier for MultiSmartContractSignatureVerifier {
...
+                .is_valid_signature(account_id, hash, signature, block_number)
                 .await
                 .unwrap());
         }
 
         todo!()
     }
+
+    async fn current_block_number(&self, chain_id: &str) -> Result<U64, VerifierError> {
+        let id: u64 = chain_id.parse().unwrap();
```
insipx, line 138: Should turn `unwrap` into error here too, any unwrap would crash converse

xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -116,16 +121,27 @@ impl SmartContractSignatureVerifier for MultiSmartContractSignatureVerifier {
         account_id: AccountId,
         hash: [u8; 32],
         signature: Bytes,
-        _block_number: Option<BlockNumber>,
+        block_number: Option<BlockNumber>,
     ) -> Result<bool, VerifierError> {
         let id: u64 = account_id.chain_id.parse().unwrap();
```
insipx, line 126: replace with an error

xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -116,16 +121,27 @@ impl SmartContractSignatureVerifier for MultiSmartContractSignatureVerifier {
...
         if let Some(verifier) = self.verifiers.get(&id) {
             return Ok(verifier
-                .is_valid_signature(account_id, hash, signature, None)
+                .is_valid_signature(account_id, hash, signature, block_number)
                 .await
                 .unwrap());
```
insipx, line 131: Error here too

insipx (approved):

PR #1117, opened 2024-10-04
Repro of max past epochs issue

richardhuaaa:


richardhuaaa:
Merged in https://github.com/xmtp/libxmtp/pull/1120

PR #1119, opened 2024-10-04
Remote signature verifier

codabrink:
Sibling: https://github.com/xmtp/proto/pull/216


xmtp_mls/src/utils/test.rs
```diff
@@ -104,10 +107,11 @@ impl ClientBuilder<TestClient> {
...
     }
 
-    pub async fn local_client(mut self) -> Self {
-        let local_client = <TestClient as XmtpTestClient>::create_local().await;
-        self = self.api_client(local_client);
-        self
+    pub async fn local_client(self) -> Self {
+        self.api_client(<TestClient as XmtpTestClient>::create_local().await)
+            .scw_signature_verifier(MultiSmartContractSignatureVerifier::new_from_file(
+                "chain_urls.json",
```
insipx, line 113: is `chain_urls` going to be in CI or should we use the global constant here

insipx (approved):

insipx:
nice!

xmtp_mls/src/utils/test.rs
codabrink: Either should be good. If it doesn't find the file, it'll just use the default. We can also swap this out for a mock.

PR #1122, opened 2024-10-07
Update WASM bindings

rygine:
# Summary

This PR adds more WASM bindings to match our other bindings. Tests and CI workflows to follow.

### DIFF
--- a/bindings_wasm/src/mls_client.rs
+++ b/bindings_wasm/src/mls_client.rs
@@ -1,41 +1,54 @@
 use js_sys::Uint8Array;
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use wasm_bindgen::prelude::{wasm_bindgen, JsError};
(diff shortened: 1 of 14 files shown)


bindings_wasm/src/mls_client.rs
```diff
@@ -1,41 +1,54 @@
...
+#[wasm_bindgen(js_name = createClient)]
 pub async fn create_client(
   host: String,
   inbox_id: String,
   account_address: String,
+  db_path: String,
   encryption_key: Option<Uint8Array>,
   history_sync_url: Option<String>,
 ) -> Result<WasmClient, JsError> {
+  xmtp_mls::utils::wasm::init().await;
```
insipx, line 47: we shouldn't include the tracing setup in prod, instead we should come up with some system similar to the `bindings_ffi` to pass a logger from javascript/let the client sdk setup logging by calling the init fn themselves. Could be a good followup issue since it's not necessary yet

PR #1128, opened 2024-10-09
Upgrade default urls to paid/private if appropriate env vars are present

codabrink:
Update MLS validation service to consume env vars from https://github.com/ephemeraHQ/infrastructure/pull/416/files

### DIFF
--- a/xmtp_id/src/scw_verifier/mod.rs
+++ b/xmtp_id/src/scw_verifier/mod.rs
@@ -8,7 +8,13 @@ use ethers::{
     providers::{Http, Provider, ProviderError},
     types::{BlockNumber, Bytes},
 };
-use std::{collections::HashMap, fs, path::Path, str::FromStr};
+use std::{
+    collections::HashMap,
+    env,
+    fs::{self},
+    io,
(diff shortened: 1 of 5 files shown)


xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -87,32 +101,26 @@ impl MultiSmartContractSignatureVerifier {
...
-            })
-            .collect();
+        Ok(Self::new(urls))
+    }
 
-        Self::new(urls)
+    /// Upgrade the default urls to paid/private urls if the env vars are present.
+    pub fn upgrade(mut self) -> Self {
+        self.verifiers.iter_mut().for_each(|(id, verif)| {
+            if let Ok(url) = env::var(format!("CHAIN_RPC_{id}")) {
```
neekolas, line 119: It would be helpful to log if any are missing, so we can't forget to set an environment variable and silently fall back to the defaults.

codabrink: Good point

neekolas (approved):

PR #1133, opened 2024-10-10
get addresses by inbox id

insipx:


bindings_ffi/src/mls.rs
```diff
@@ -329,6 +329,24 @@ impl FfiXmtpClient {
         Ok(state.into())
     }
 
+    /**
+     * Get the inbox state for each `inbox_id`.
+     *
+     * If `refresh_from_network` is true, the client will go to the network first to refresh the state.
+     * Otherwise, the state will be read from the local database.
+     */
+    pub async fn addresses_from_inbox_id(
```
rygine, line 338: i think this comment was copied from `inbox_state`

rygine (approved):
left a comment, but LGTM

bindings_ffi/src/mls.rs
insipx: Yeah i modified the comment a little bit to refer to all addresses but otherwise its the same functions

bindings_ffi/src/mls.rs
```diff
@@ -329,6 +329,24 @@ impl FfiXmtpClient {
...
+        &self,
+        refresh_from_network: bool,
+        inbox_ids: Vec<String>,
+    ) -> Result<Vec<FfiInboxState>, GenericError> {
```
nplasterer, line 342: @insipx what do you think about changing this so it just returns the list of addresses? Is it weird to show other people all the installations a user has?

nplasterer: Or maybe it fine to show what other peoples recovery address and installations are? Maybe @richardhuaaa has thoughts.

richardhuaaa: It's all stored publicly on the network anyway, so I don't think it matters much from a privacy perspective. It has to be public in order for your installation to know which installations to include in the conversation. Not sure what is important here from a usability perspective though!

nplasterer: Okay seems no harm in exposing it all in the sdk. Is the recovery address helpful in anyway from a integrator perspective? Is it like the main address if you had to choose just one to display?

richardhuaaa: The recovery address is the only address that can perform revocations on your account. It's a separate class of permissions because otherwise any compromised key could remove all other keys and take over the account. It's useful for a user to know their own recovery address, can't think of a use-case today for you to query someone else's recovery address though

nplasterer: In the inverse it almost seems bad to make it easy to access others recovery addresses as a attacker could target that address in particular to try to gain master access?

nplasterer: Tho I guess it's public so people who really want to access that information can access it either way so I guess not a huge deal. The amount of money you have in your wallet is also public 😄

Issue #1129, opened 2024-10-09
Find DM group by peer address or inbox id


nplasterer:
Add the ability to find a dm group by inbox id so that we can dual send messages to V2 conversations and V3.

In V2 world conversations are found exclusively by the peer address since there can only be 1 one to one conversation. It would be nice to be able to closely emulate that until we get conversation threading fully implemented maybe just return the first group dm for the 2 people.

Issue #696, opened 2024-05-01
Conversation History - New installations only see one 1to1 conversation (NEEDS SCOPING - Large)


nplasterer:

richardhuaaa:
I think there's two questions here - whether v2 and v3 1:1 conversations can exist simultaneously, and whether there can be multiple duplicate v3 1:1 conversations.

I think the answer to the first question should be yes, as we need to support dual send/receive as a migration path. And ideally the answer to the second question would be no.

I think in order to answer the second question, we need to iron out exactly how we are going to support personal preferences (which is needed for consent, for example, even if we don't support message history). It's possible we could piggy-back on however that works.

richardhuaaa:
Ideas from discussion:

1. Stitch conversations
2. Use consent lists + external join
3. Use deterministic topic ID (with per-user DH keys in personal prefs) + external join
4. Make message history sync mandatory for at least the conversation list

nplasterer:
1. Stitch conversations

Complexity on our side. Tracking all these conversations and tracking them together. We all agree this is not ideal. 

2. We already track consent

We may be able to derive the whole conversation from that consent list. Could check the consent list to see if you already have a conversation. Knowing you have a conversation doesn't allow you to send message. Add an installation to that conversation. MLS feature called external joins. We would need to implement.

3. Use deterministic topic ID (with per-user DH keys in personal prefs) + external join

2 doesnt handle race conditions. This is similar to 2. There could be a large window for this race condition. Lets make sure for any two participants they will always end up on the same topic id. We generate a diff helmen key pair and hash the results for the topic. (kind of how we do it in V2 for deterministic topics) This is a little complicated because we have to generate new keys and sync it. 
All devices have the same key in V2 but all devices have different keys in V3. We're trying to get away from this for security reasons. Similar to the HMAC key we would use the same mechanism 

4. Make message history sync mandatory for at least the conversation list

Message history sync is optional. But make the conversation list of it as required. Simple. But if lost in a river the list is gone.  
V2 does not require a live device but it has a security trade off. And a UX trade off because of the wallet signature.

nplasterer:
Default on for message history with the option to turn off. But you can't turn off personal preference syncing.

New device is still going to make duplicate conversations if it doesn't do the message history because the device isn't active. 
Stitching will query both conversations but only write to the latest. Latest will be come the same for both users after a message is sent there. 1to1 does not have special permission policies so shouldn't be an issue for two group chats to get out of sync. 

[...]

PR #1137, opened 2024-10-11
Find DM conversation from target's inbox_id

codabrink:
https://github.com/xmtp/libxmtp/issues/1129

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -363,21 +363,6 @@ impl FfiXmtpClient {
         Ok(state.into_iter().map(Into::into).collect())
     }
 
-    /**
-     * Get the inbox_id for each address.
-     */
-    pub async fn inbox_ids_from_addresses(
-        &self,
-        addresses: Vec<String>,
-    ) -> Result<Vec<Option<String>>, GenericError> {
-        let inbox_ids = self
-            .inner_client
-            .find_inbox_ids_from_addresses(addresses)
-            .await?;
-
-        Ok(inbox_ids)
-    }
-
     pub async fn get_latest_inbox_state(
         &self,
         inbox_id: String,
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -347,10 +347,27 @@ impl FfiXmtpClient {
         Ok(state.into_iter().map(Into::into).collect())
     }
 
-    pub async fn dm_from_address(&self, address: String) -> Result<FfiGroup, GenericError> {
-        Err(GenericError::Generic {
-            err: "Not found changeme".to_string(),
-        })
+    /**
+     * Get the inbox_id for each address.
+     */
+    pub async fn inbox_ids_from_addresses(
+        &self,
+        addresses: Vec<String>,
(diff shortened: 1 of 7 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -347,6 +363,21 @@ impl FfiXmtpClient {
         Ok(state.into_iter().map(Into::into).collect())
     }
 
+    /**
+     * Get the inbox_id for each address.
+     */
+    pub async fn inbox_ids_from_addresses(
```
nplasterer, line 369: I think you can delete this because we already have it here: https://github.com/xmtp/libxmtp/blob/main/bindings_ffi/src/mls.rs#L163-L180

bindings_node/src/conversations.rs
```diff
@@ -121,6 +121,20 @@ impl NapiConversations {
     ))
   }
 
+  #[napi]
+  pub fn find_dm_group_by_target_inbox_id(&self, target_inbox_id: String) -> Result<NapiGroup> {
```
nplasterer, line 125: I try to avoid the word group in these bindings when it's referring to a "conversation" But this is probably fine since the node bindings need a major refactor to support dms. cc @rygine 

```suggestion
  pub fn find_dm_by_target_inbox_id(&self, target_inbox_id: String) -> Result<NapiGroup> {
```

xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -185,13 +185,34 @@ impl DbConnection {
...
+            .order(dsl::created_at_ns.asc())
+            .filter(dsl::dm_inbox_id.eq(Some(&dm_target_inbox_id)));
+
+        let groups: Vec<StoredGroup> = self.raw_query(|conn| query.load(conn))?;
+        if groups.len() > 1 {
+            tracing::error!(
+                "More than one group found for dm_inbox_id {}",
+                dm_target_inbox_id
+            );
+        }
```
nplasterer, line 214: So I think this is likely going to happen frequently. With the way installations work you can over time get lots of dms with the same person. We are going to implement conversation threading eventually to make all these dm groups look like just one conversation from the user perspective. So I would just select the first or last? Group and not error. cc @insipx You can read more about this issue here https://github.com/xmtp/libxmtp/issues/696

xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -185,13 +185,34 @@ impl DbConnection {
...
         let groups: Vec<StoredGroup> = self.raw_query(|conn| query.load(conn))?;
         if groups.len() > 1 {
             tracing::error!("More than one group found for welcome_id {}", welcome_id);
         }
-        // Manually extract the first element
+
+        Ok(groups.into_iter().next())
+    }
+
+    pub fn find_dm_group(
```
nplasterer, line 200: Be worth while to add a test for this. 🙏

xmtp_mls/src/storage/encrypted_store/group.rs
codabrink: Got it, I'll downgrade this to info!.

nplasterer (approved):

Issue #1130, opened 2024-10-10
Make clear distinctions between DMs & Groups with common Conversation


nplasterer:
With DMs being a group under the hood I'm afraid the lines between the two are going to get increasingly more confusing and potentially cause bugs. 
To make this easier I think we should make very clear distinctions between DMs and Groups with a common interface of Conversation when doing both in the bindings.

Here are some functions that should exist
- [x] listGroups - (filters groups on NOT type dm)
- [x] listDms - (filters groups on type dm)
- [x] listConversations - (no filter aka dms and groups)
- [x] streamGroups
- [x] streamDms
- [x] streamConversations
- [x] findConversation(id) - (probably no longer need the findGroup method)
- [x] consentType - (GROUP_ID, INBOX_ID, ADDRESS) should be -> (CONVERSATION_ID, INBOX_ID, ADDRESS)
- [x] streamAllGroupMessages -> we need this in the dualsending interim when we don't want to stream dm groups
- [x] streamAllDMMessages -> we need this in the dualsending interim when we don't want to stream dm groups
- [x] streamAllMessages -> this will be used to stream all messages in a V3 only world

richardhuaaa:
This sounds great to me. Do any of these involve breaking changes to the meaning of existing methods?

nplasterer:
I don't think there is enough meaningful data in the local consent database to warrent a major issue with renaming `GROUP_ID` to `CONVERSATION_ID` but that would be the only breaking change.

PR #1138, opened 2024-10-13
Make general conversation to encompass group dms and groups.

nplasterer:
Closes https://github.com/xmtp/libxmtp/issues/1130

Brings the language in the bindings more inline with conversation than group to make it easier to understand with groupDMs.

Adds a bunch of specific dm and group functionality to allow backwards compatibility for V2 migrations.

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -41,8 +41,7 @@ use xmtp_proto::xmtp::mls::api::v1::{
 use crate::{
     api::ApiClientWrapper,
     groups::{
-        group_permissions::PolicySet, validated_commit::CommitValidationError, GroupError,
-        GroupMetadataOptions, IntentError, MlsGroup,
+        group_metadata::ConversationType, group_permissions::PolicySet, validated_commit::CommitValidationError, GroupError, GroupMetadataOptions, IntentError, MlsGroup
     },
     identity::{parse_credential, Identity, IdentityError},
     identity_updates::{load_identity_updates, IdentityUpdateError},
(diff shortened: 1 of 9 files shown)


xmtp_mls/src/client.rs
```diff
@@ -224,7 +225,7 @@ pub struct FindGroupParams {
     pub created_after_ns: Option<i64>,
     pub created_before_ns: Option<i64>,
     pub limit: Option<i64>,
-    pub include_dm_groups: bool,
+    pub conversation_type: Option<ConversationType>,
```
nplasterer, line 228: I changed all these `include_dm` booleans to `conversation_type` so that we could have all 3 options
just dms, just groups, and both.

xmtp_mls/src/storage/encrypted_store/consent_record.rs
```diff
@@ -85,8 +85,8 @@ impl DbConnection {
 #[diesel(sql_type = Integer)]
 /// Type of consent record stored
 pub enum ConsentType {
-    /// Consent is for a group
-    GroupId = 1,
+    /// Consent is for a conversation
+    ConversationId = 1,
```
nplasterer, line 89: Since it is just a number under the hood this rename should have no effect on existing records.

xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -122,7 +124,7 @@ impl DbConnection {
         created_after_ns: Option<i64>,
         created_before_ns: Option<i64>,
         limit: Option<i64>,
-        include_dm_groups: bool,
+        conversation_type: Option<ConversationType>,
```
codabrink, line 127: What does it mean when the conversation type is `None`?

codabrink (approved):

insipx:
nice! Much prefer enums to booleans

PR #1141, opened 2024-10-15
Full stack SCW tests with external Anvil instance

codabrink:
Create an instance of anvil in our dev suite and expose it to tests for full-stack testing.


xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -128,6 +128,15 @@ impl MultiSmartContractSignatureVerifier {
                 info!("No upgraded chain url for chain {id}, using default.");
             };
         });
+
+        if let Ok(url) = env::var("FOUNDRY_URL") {
```
insipx, line 132: Since this is specific to testing + the validation service, we shuld

xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -128,6 +129,15 @@ impl MultiSmartContractSignatureVerifier {
...
```
insipx, line 133: Since this is specific to tests + the validation service, WDYT about putting this in `mls_validation_service` under `cfg(test)`? It could go right where we're building the verifier afaiu


PR looks good otherwise!

insipx (changes requested):

xmtp_id/src/scw_verifier/mod.rs
codabrink: The problem is it's a dev feature, not a test feature. We could put this under a `#[cfg(debug_assertions)]` flag.

mls_validation_service/Cargo.toml
```diff
@@ -25,7 +25,7 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 tracing.workspace = true
 warp = "0.3.6"
 xmtp_cryptography = { path = "../xmtp_cryptography" }
-xmtp_id.workspace = true
+xmtp_id = { workspace = true, features = ["test-utils"] }
```
neekolas, line 28: I don't think we want this in the prod build. I added it just for debugging our issues with local dev.

codabrink: Oh, woops. I thought this was dev-dependencies. Let me remove it.

insipx (approved):

PR #1150, opened 2024-10-17
V2 Conversation Database

nplasterer:
We need to store v2 conversations in the V3 database so that we can easily look up to see if a group dm should be shown or not.


xmtp_mls/migrations/2024-10-17-173334_create_v2_conversations/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE v2_conversations (
+    topic TEXT NOT NULL PRIMARY KEY,
+    created_at_ns BIGINT NOT NULL,
+    peer_address TEXT NOT NULL,
+    envelope_bytes BLOB NOT NULL
```
neekolas, line 5: The most expensive thing about listing v2 conversations is decrypting this envelope.

What if we actually stored the plaintext here so it only has to happen once?

xmtp_mls/src/storage/encrypted_store/v2_conversation.rs
```diff
@@ -0,0 +1,155 @@
...
+        Ok(self.raw_query(|conn| -> diesel::QueryResult<_> {
+            dsl::v2_conversations.load(conn)
+        })?)
+    }
+
+    /// Insert v2_conversations
+    pub fn insert_or_replace_v2_conversation(&self, v2_conversation: StoredV2Conversation) -> Result<StoredV2Conversation, StorageError> {
+        tracing::info!("Trying to insert v2 conversation");
+        let stored_v2_conversation = self.raw_query(|conn| {
+            let maybe_inserted_conversation: Option<StoredV2Conversation> = diesel::insert_into(dsl::v2_conversations)
```
neekolas, line 72: If this is using `on_conflict_do_nothing` then it's actually an "insert or ignore" not an "insert or replace"

neekolas: There is also a helpful Diesel helper function `insert_or_ignore_into`

xmtp_mls/src/storage/encrypted_store/v2_conversation.rs
```diff
@@ -0,0 +1,155 @@
...
+                .values(&v2_conversation)
+                .on_conflict_do_nothing()
+                .get_result(conn)
+                .optional()?;
+
+            if maybe_inserted_conversation.is_none() {
+                let existing_conversation: StoredV2Conversation = dsl::v2_conversations.find(v2_conversation.topic).first(conn)?;
+                if existing_conversation.topic == v2_conversation.topic {
```
neekolas, line 80: If you are looking this up by topic, I don't see why you would need this check

Issue #1113, opened 2024-10-03
Sync Preferences


nplasterer:
Add ability to sync preferences just like we have `sync` on Conversations and Group.

Right now everything with consent V3 happens with the local DB. But we also want to send and sync these consent messages from the network as well.

You can see the local work here: https://github.com/xmtp/libxmtp/pull/1057

- [ ] Can send consent records to the network
- [ ] Can sync consent records from the network

codabrink:
## Conflict reconciliation
This issue focuses on the initial consent sync that will be ideally run when a new installation is created, so that there will be no local state to reconcile with the incoming payloads. However, there are a few edge cases we should still consider.

# What if the installation receives multiple sync responses.
The user will have to choose the device to sync from, putting in a pin on an existing device that matches the pin in the sync request to ensure that there is a user in the loop.

# What if the user consents to a conversation they had denied on their previous installation before the payload is received?
We should assume two things here.
  1. The consent to the conversation on the new installation was intentional, and a desired outcome.
  2. The old installation made a best effort to stream all consent updates from the sync group before sending the payload.
With these considerations, if there is a conflict between the local consent, and the consent in the incoming payload, the resolution should be the more open of the two options.
This does mean that if the user **denies** consent on a new installation that they had previously **approved** on the old installation while the old installation is uploading the payload, they will have to deny the consent again after the new payload arrives.

richardhuaaa:
This is a really good balance between handling edge cases and keeping things simple.

> What if the user consents to a conversation they had denied on their previous installation before the payload is received?

There are some edge cases with the described solution, but I don't think they matter much - we're talking about a single user, so concurrency should be limited in practice.

> The user will have to choose the device to sync from, putting in a pin on an existing device that matches the pin in the sync request to ensure that there is a user in the loop.

Adding a user interaction on the existing device puts UX requirements on all implementing apps that they may or may not want to service. One thing that would be desirable is if all apps *by default* can serve sync requests, so that we minimize the possibility that a user logs in on a different app and has no means of getting their backup from their existing app (as there is an interoperability risk here). Is there a world in which libxmtp will serve backup requests silently by default, but with strong encouragement in our docs for them to implement a pin UX?

codabrink:
[...]

PR #1152, opened 2024-10-18
Consent Sync

codabrink:
Issue: https://github.com/xmtp/libxmtp/issues/1113
Proto PR: https://github.com/xmtp/proto/pull/222

## Summary by CodeRabbit

- **New Features**
	- Introduced device synchronization functionalities, including new methods for handling sync requests and replies.
	- Added support for consent synchronization with dedicated methods.
	- Enhanced group management with new intent structures for updating admin lists and permissions.

- **Bug Fixes**
	- Improved error handling in various methods to ensure robustness during synchronization processes.

- **Refactor**
	- Simplified parameter handling across multiple methods by passing references instead of owned values, improving memory efficiency.

- **Documentation**
	- Updated README and documentation to reflect changes in command usage and feature flags.

- **Chores**
	- Cleaned up and reorganized dependency management in several Cargo.toml files.


xmtp_mls/benches/group_limit.rs
```diff
@@ -172,7 +177,7 @@ fn add_to_100_member_group_by_inbox_id(c: &mut Criterion) {
                             .add_members_by_inbox_id(
                                 // it is OK to take from the back for now because we aren't getting
                                 // near MAX_IDENTITIES
-                                inbox_ids.iter().rev().take(100).cloned().collect(),
+                                &inbox_ids,
```
neekolas, line 180: Do we know why we were originally doing this 100 at a time? If not, we should probably leave it as-is

codabrink: We're still taking 100 at a time. https://github.com/xmtp/libxmtp/pull/1152/files/829289369ec1172c170929014df33eade5e8ab2b#diff-cd11ded27b36e44d16af4f4f8426b315cba5e17487db308d9d6a92483ff9fb9bR156

xmtp_mls/migrations/2024-10-18-200341_create_key_value_store/up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE IF NOT EXISTS key_value_store (
```
neekolas, line 1: I'd love to understand what data we expect to put into this table over time? Is there no way we can have a better typed table for specific problems? Gets very hard to migrate a generic key/value store if the shape of the data changes.

xmtp_mls/benches/group_limit.rs
codabrink: Or are you saying we would rather leave it in the loop so that the benches aren't distorted?

xmtp_mls/src/groups/device_sync/consent_sync.rs
```diff
@@ -0,0 +1,174 @@
...
+            KVStore::get(&conn, &Key::ConsentSyncRequestId).map_err(DeviceSyncError::Storage)?;
+        let Some(request_id) = request_id else {
+            return Err(DeviceSyncError::NoReplyToProcess);
+        };
+
+        // process the reply
+        self.process_sync_reply(&request_id).await
+    }
+
+    fn syncable_consent_records(&self) -> Result<Vec<Syncable>, DeviceSyncError> {
```
neekolas, line 49: Is there some document I can look at to see how we are expecting to merge these consent records?

I'm thinking about situations like:

Client A blocks Address A
Client B allows Address A
Client C gets created

(and a bunch of variations on that idea)

How do we make sure that all clients get a consistent view of who is allowed and blocked?

xmtp_proto/src/gen/mod.rs
```diff
@@ -72,24 +72,10 @@ pub mod xmtp {
...
-            include!("xmtp.xmtpv4.message_api.rs");
-            // @@protoc_insertion_point(xmtp.xmtpv4.message_api)
-        }
-        #[cfg(feature = "xmtp-xmtpv4-payer_api")]
-        // @@protoc_insertion_point(attribute:xmtp.xmtpv4.payer_api)
-        pub mod payer_api {
-            include!("xmtp.xmtpv4.payer_api.rs");
-            // @@protoc_insertion_point(xmtp.xmtpv4.payer_api)
-        }
+        include!("xmtp.xmtpv4.rs");
```
neekolas, line 78: I think this is out of date. The old code is more current than these changes

codabrink: Just updated :)

neekolas:
This PR is so massive and cross-cutting I'm worried we aren't going to be able to give it the quality of review it deserves.

I'm also not really sure what requirements or design this is implementing. Especially how local updates and remote updates get merged and reconciled. Is there some document that outlines the design somewhere?

xmtp_mls/benches/group_limit.rs
neekolas: Nevermind. This looks fine

insipx: This is just setup for the benchmark, its taking 100 from the back of a very large generated list of identities either on local or dev. The benchmark itself is progressing by the predetermined steps from the front of the list. Moving the iteration out of here should be fine, since its not the part being benchmarked

codabrink:
The reconciliation logic is a good call-out. I had followed the pattern of the message sync reconciliation implementation, which was more or less just overwriting. But there are some good edge cases that we should consider and document before moving forward. I'll write some updates for the issue.

xmtp_mls/src/groups/device_sync/consent_sync.rs
```diff
@@ -0,0 +1,159 @@
...
+        let reply = self
+            .send_syncables(
+                &request.request_id,
+                &[consent_records],
+                DeviceSyncKind::Consent,
+            )
+            .await?;
+
+        Ok(reply)
+    }
```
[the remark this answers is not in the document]
codabrink: It's not missing. This is intentional.

xmtp_mls/src/storage/encrypted_store/consent_record.rs
```diff
@@ -78,6 +78,30 @@ impl DbConnection {
...
+            if maybe_inserted_consent_record.is_none() {
+                return Ok(dsl::consent_records
+                    .find((&record.entity_type, &record.entity))
+                    .first(conn)
+                    .optional()?);
+            }
+
+            Ok(None)
+        })
+    }
```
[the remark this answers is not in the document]
codabrink: I have a test that proves otherwise.

xmtp_mls/src/storage/encrypted_store/schema.rs
```diff
@@ -84,6 +84,13 @@ diesel::table! {
     }
 }
 
+diesel::table! {
```
neekolas, line 87: Is this still being used?

richardhuaaa:
Would it make sense to do a code walkthrough? I always like to suggest that for large or complex PR's, if it's too late to split into smaller pieces

xmtp_mls/src/storage/encrypted_store/schema.rs
codabrink: Ah no, I missed it when I removed kvstore. I'll purge this.

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -0,0 +1,596 @@
...
+        self.sync_welcomes().await?;
+
+        let groups = conn.find_groups(None, None, None, None, Some(ConversationType::Group))?;
+        for crate::storage::group::StoredGroup { id, .. } in groups.into_iter() {
+            let group = self.group(id)?;
+            Box::pin(group.sync()).await?;
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
codabrink: This is intentional to allow a time slip of 1h in either direction.

xmtp_mls/src/groups/node_sync.rs
```diff
@@ -995,15 +992,15 @@ where
     pub(super) async fn get_membership_update_intent(
         &self,
         provider: &XmtpOpenMlsProvider,
-        inbox_ids_to_add: Vec<InboxId>,
-        inbox_ids_to_remove: Vec<InboxId>,
+        inbox_ids_to_add: &[InboxId],
```
insipx, line 995: love all the cleanup converting `Vec<>` to `&`

xmtp_mls/src/groups/device_sync/message_sync.rs
```diff
@@ -0,0 +1,291 @@
...
+            for msg in messages {
+                all_messages.push(Syncable::GroupMessage(msg));
+            }
+        }
+
+        Ok(all_messages)
+    }
+}
+
+#[cfg(all(not(target_arch = "wasm32"), test))]
```
insipx, line 71: sucks that we don't have any history sync tests for wasm, I created a ticket to eventually create some integration tests, when the message-history/consent sync server is ever added to docker-compose: https://github.com/xmtp/libxmtp/issues/1201

I skimmed some http-mock libraries, but none have wasm support b/c of the nature of http, so the docker route is probably the only way to go

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -0,0 +1,596 @@
...
+    /// Look for sync reply by kind, returns NoReplyToProcess error if not found.
+    async fn sync_reply(
+        &self,
+        kind: DeviceSyncKind,
+    ) -> Result<DeviceSyncReplyProto, DeviceSyncError> {
+        let sync_group = self.get_sync_group()?;
+
+        sync_group.sync().await?;
+        let messages = sync_group
+            .find_messages(&MsgQueryArgs::default().kind(GroupMessageKind::Application))?;
```
insipx, line 244: `MsgQueryArgs` 🔥

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -0,0 +1,596 @@
...
+            let content = DeviceSyncContent::Reply(contents);
+            let content_bytes = serde_json::to_vec(&content)?;
+            let DeviceSyncContent::Reply(contents) = content else {
+                unreachable!("This is a reply.");
+            };
+
+            (content_bytes, contents)
+        };
+
+        let _message_id = sync_group.prepare_message(&content_bytes, &conn, move |_time_ns| {
```
insipx, line 191: noob Q: How come we need both `content_bytes` and `content` here? aren't they the same thing?

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -0,0 +1,596 @@
...
+        kind: DeviceSyncKind,
+    ) -> Result<(StoredGroupMessage, DeviceSyncRequestProto), DeviceSyncError> {
+        let sync_group = self.get_sync_group()?;
+
+        sync_group.sync().await?;
+
+        let messages = sync_group
+            .find_messages(&MsgQueryArgs::default().kind(GroupMessageKind::Application))?;
+
+        for msg in messages.into_iter().rev() {
```
insipx, line 218: would it be worth adding a field to `group_messages` here, or maybe something else to make this lookup faster? @neekolas

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -0,0 +1,596 @@
...
+
+        let enc_payload = download_history_payload(&reply.url).await?;
+        insert_encrypted_syncables(&conn, enc_payload, &enc_key.try_into()?)?;
+
+        self.sync_welcomes().await?;
+
+        let groups = conn.find_groups(None, None, None, None, Some(ConversationType::Group))?;
+        for crate::storage::group::StoredGroup { id, .. } in groups.into_iter() {
+            let group = self.group(id)?;
+            Box::pin(group.sync()).await?;
```
insipx, line 286: related to connection use, wherever we do a `sync`, we should probably actually be doing a `sync_with_conn` in order to re-use the connection. That requires the `XmtpOpenMlsProvider` which can be created from an owned value of `DbConnection`.

Anything not explicitly requiring the provider (like normal queries) can still access the connection with `conn_ref`

insipx (commented):
good work!

Mostly I would like to see the connection use optimized, since that's been an issue in the SDKs in the past. Otherwise looks really good

xmtp_mls/src/groups/mod.rs
```diff
@@ -510,14 +507,13 @@ impl<ScopedClient: ScopedGroupClient> MlsGroup<ScopedClient> {
...
 
-    #[cfg(feature = "message-history")]
     pub(crate) fn create_and_insert_sync_group(
         client: Arc<ScopedClient>,
     ) -> Result<MlsGroup<ScopedClient>, GroupError> {
         let context = client.context();
         let conn = context.store().conn()?;
         // let my_sequence_id = context.inbox_sequence_id(&conn)?;
-        let creator_inbox_id = context.inbox_id().to_string();
+        let creator_inbox_id = context.inbox_id() as String;
```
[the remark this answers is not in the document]
codabrink: Likely! Some of this is old code that was merely shifted around.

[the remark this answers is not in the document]
codabrink: It returns type `InboxId` which is an alias for `String`.

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -0,0 +1,596 @@
...
+
+        Ok(())
+    }
+
+    async fn send_sync_request(
+        &self,
+        request: DeviceSyncRequest,
+    ) -> Result<(String, String), DeviceSyncError> {
+        // find the sync group
+        let conn = self.store().conn()?;
```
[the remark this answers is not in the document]
codabrink: Nice, good suggestion!

xmtp_mls/src/groups/device_sync.rs
codabrink: Not sure. This is legacy code that I moved around, but didn't really check. Let me do some poking around to see what I can find.

codabrink: Looking at this a bit more, we're not using both. The `content_bytes` is serialized, and we're not actually using the `content` value here, that's just a field. :) Maybe you looked over it too quickly?

xmtp_mls/src/groups/device_sync.rs
codabrink: I'd definitely be down for this optimization later if it becomes an issue. Right now I don't think it'll be much of an issue since the design right now is to create a new sync group every time a device is added, which becomes the default sync group for all installations.

xmtp_mls/src/groups/device_sync/consent_sync.rs
```diff
@@ -0,0 +1,178 @@
+use super::*;
+use crate::{Client, XmtpApi};
+use xmtp_id::scw_verifier::SmartContractSignatureVerifier;
+
+impl<ApiClient, V> Client<ApiClient, V>
+where
+    ApiClient: XmtpApi + Clone,
+    V: SmartContractSignatureVerifier + Clone,
+{
+    pub async fn send_consent_sync_request(&self) -> Result<(String, String), DeviceSyncError> {
```
insipx, line 10: Can accept a `XmtpOpenMlsProvider` here too

xmtp_mls/src/groups/device_sync/message_sync.rs
```diff
@@ -0,0 +1,318 @@
...
+use crate::{storage::group::StoredGroup, Client};
+use xmtp_id::scw_verifier::SmartContractSignatureVerifier;
+
+impl<ApiClient, V> Client<ApiClient, V>
+where
+    ApiClient: XmtpApi + Clone,
+    V: SmartContractSignatureVerifier + Clone,
+{
+    // returns (request_id, pin_code)
+    pub async fn send_history_sync_request(&self) -> Result<(String, String), DeviceSyncError> {
```
insipx, line 13: Can accept a provider here too

insipx (approved):
nice!

PR #1153, opened 2024-10-18
Fix rustls transport error / use nix releases

insipx:
A rustls dependency upgrade that broke android + swift tests. Originally thought it was an openssl issue, but openssl was completely unrelated so i removed all the openssl pins (the patch we had was never being applied). The only dep using openssl is diesel (and it is vendored).

I pinned rustls and tonic to the versions present in main before the wasm PR and it works again. Probably, android is very fragile about the webpki version we use. We should consider  moving over to [rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier) at some point

CI builds should be alot faster using native dependencies + use the same rust cache (`/target` folder), and so much easier to test locally. It was invaluable being able to quickly build the jniLibs and copy them over/add logging/repeat. I'm leaving the old cross builds just in case, but pushing new workflows with a -nix ending. These have the benefit of being able to run `./dev/release-kotlin/` & `./dev/release-swift` in libxmtp and create the jni/swift artifacts locally & quickly for testing, but do require installing nix which is included in the script (the nix version curated by [determinate systems](https://determinate.systems/)). 

If you'd rather not use nix, you can also `cargo install cargo-ndk` and run the command locally, but there's no guarantee of the same environment as CI. In my experience this hasn't mattered much, unless there are non-standard/unexpected sqlite/openssl versions on the system.

`swift` artifacts were already possible to build without any magic & `cargo`, as long as a mac machine is used. A nix build is still used there just for consistency and assurance of env

## Summary by CodeRabbit

- **New Features**
	- Introduced automated release workflows for Kotlin and Swift bindings, enhancing the release process for multiple architectures.
	- Added a new Flake Shell configuration for building release artifacts for Swift and Kotlin applications.

- **Bug Fixes**
	- Improved error handling and logging in various methods to enhance clarity and efficiency.

- **Documentation**
	- Enhanced comments and documentation throughout the codebase to clarify functionality and error handling.

- **Chores**
	- Updated dependencies and versions in `Cargo.toml` files to improve compatibility and functionality.


nplasterer (approved):
Getting better each time! 🤞

PR #1156, opened 2024-10-21
MLS Validation Service Version

mkysel:
For node-partners we require publicly versioned docker images.

I picked 0.1.0 since `mls-validation-service` has been around for a bit and it does not feel like `0.0.1`

Also added some convenience methods to print `--version`

The format is as follows:
```
$ docker run --rm af4389d1bb47 --version
Version: 0.1.0-a7bd4b5
```

Once this is merged, I will update CI to react to tags.


mls_validation_service/src/main.rs
```diff
@@ -24,6 +26,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .init();
 
     let args = Args::parse();
+
+    if args.version {
+        println!("Version: {0}", get_version());
```
insipx, line 31: Should be an info! log

insipx (approved):
Looks good other than small nit !

mls_validation_service/src/main.rs
mkysel: with `info!` I get this.
```
 $ cargo run --bin mls-validation-service -- -v
   Compiling mls_validation_service v0.1.0 (/Users/martinkysel/work/libxmtp/mls_validation_service)
    Finished `dev` profile [unoptimized] target(s) in 1.69s
     Running `target/debug/mls-validation-service -v`
```

Any ideas?

mkysel: I would have to run it like this. Which is suboptimal.
```
$ RUST_LOG=info cargo run --bin mls-validation-service -- --version
   Compiling mls_validation_service v0.1.0 (/Users/martinkysel/work/libxmtp/mls_validation_service)
    Finished `dev` profile [unoptimized] target(s) in 1.29s
     Running `target/debug/mls-validation-service --version`
2024-10-21T13:29:30.722812Z  INFO mls_validation_service: Version: 0.1.0-a314006
```

mkysel: done!

PR #1160, opened 2024-10-21
MLS Release from tag

mkysel:


.github/workflows/release-mls-validation-service.yml
```diff
@@ -0,0 +1,45 @@
...
+            type=ref,event=tag
+            type=semver,pattern={{version}}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v6
+        id: push
+        with:
+          context: .
+          file: ./dev/validation_service/Dockerfile
+          push: true
```
richardhuaaa, line 43: The one in xmtpd has `push: ${{ github.event_name != 'pull_request' }}`, is there a reason for the difference?

richardhuaaa (approved):

.github/workflows/release-mls-validation-service.yml
mkysel: this is copied from one place, and the other one from another place :)

In either case the event is not a pull request, so both will be true.

Issue #899, opened 2024-07-11
Feature request: Pass stream errors through to callbacks


insipx:
This should make handling stream errors possible from the SDK's

Need a good way to pass a `Result<>` through the FFI boundary for uniffi & napi. This could either be by using the traditional `Result` type or crafting some custom type for XMTP Stream errors.

the fn signatures in question: 
`pub async fn stream_conversations`
`pub(crate) async fn stream_messages`
`pub async fn stream_all_messages`
`pub fn stream_conversations_with_callback`
`pub(crate) fn stream_messages_with_callback`
`pub fn stream_all_messages_with_callback`

the `with_callback` variant is just the non-`with-callback` variant, except they take a callback and iterate on the stream in a spawned `tokio` task. This makes it more ergonomic for native client sdks to use the stream, since these native SDK's have no notion of a rust `Stream` type.

Notice, that in the `with_callback` variants, instead of passing the error into the callback, we instead match on the result and log an error. This makes it difficult/impossible for external SDKs to handle stream errors. we need a way to pass the error into the callback that is compatible with [`uniffi`](https://mozilla.github.io/uniffi-rs/latest/) and [`napi`](https://napi.rs/)

these bindings are in `bindings_node/` and `bindings_ffi`/ folders respectively.

This will allow external native SDK's to handle stream errors rather than silently failing if an error occurs in a stream.

Issue #1149, opened 2024-10-17
Bug: Accounts die silently and streams don't resume on restart


humanagent:
### Describe the bug

- After the latest Node SDK bindings release, logs were added for the Node SDK bots.
- Upon restarting, the message stream does not resume, preventing the bot from receiving or responding to messages.
- The stream only resumes correctly after removing and re-adding the bot.
- Another workaround is wiping the DB and restarting the bot.

[log (1).txt](https://github.com/user-attachments/files/17421178/log.1.txt)

https://github.com/user-attachments/assets/d9ba3222-025a-4a9c-a420-0f63b663d8d9

```
2024-10-17T17:06:26.184081Z  INFO xmtp_mls::groups::sync: Processing envelope with hash "9d9a4b28377e782279e59b964a2ea21e2dd48e85cb1351483150c2276b4d8ed9"

2024-10-17T17:06:26.184086Z  INFO xmtp_mls::groups::sync: client [b92c13db581e91ff4b5c1d5af598a4f4a815f3d9b960d5fc79febb1ae3c71a05] is about to process external envelope [39658]

2024-10-17T17:06:26.184088Z ERROR openmls::group::public_group::validation: Wrong Epoch: message.epoch() 68 > 65 self.group_context().epoch()    

2024-10-17T17:06:26.184090Z DEBUG xmtp_mls::storage::encrypted_store: Transaction async being rolled back

2024-10-17T17:06:26.184095Z  INFO xmtp_mls::groups::sync: error is not retryable. OpenMlsProcessMessage(ValidationError(WrongEpoch))

2024-10-17T17:06:26.184097Z DEBUG xmtp_mls::storage::encrypted_store: Transaction async beginning

2024-10-17T17:06:26.184099Z DEBUG xmtp_mls::storage::encrypted_store: Pulling connection from pool, idle_connections=23, total_connections=25

2024-10-17T17:06:26.184120Z  INFO xmtp_mls::groups::sync: Processing envelope with hash "71b9d064d93754f61f49097fe996b953857ace3ee242f2090f493eeda3b48313"

2024-10-17T17:06:26.184125Z  INFO xmtp_mls::groups::sync: client [b92c13db581e91ff4b5c1d5af598a4f4a815f3d9b960d5fc79febb1ae3c71a05] is about to process external envelope [39659]

[...]

PR #1163, opened 2024-10-21
Restart streams if service unavailable detected

### LINKED ISSUE #1154: Retry strategy for service unavailable
By adding better context to our gprc/network subscriptions, we can create a better retry strategy for reconnecting dead grpc streams, specifically by using the tonic `https://docs.rs/tonic/latest/tonic/enum.Code.html` `Unvailable` error code, which would be returned if a stream is suddenly disconnected because of a network upgrade for instance.

rel #

insipx:
fix #1154 
rel #1149
fix https://github.com/xmtp/libxmtp/issues/899

### DIFF
--- a/.github/workflows/test-workspace-against-dev.yml
+++ b/.github/workflows/test-workspace-against-dev.yml
@@ -1,0 +1,44 @@
+name: Test Workspace
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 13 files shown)


insipx:
#1601 accomplishes what this set out to do

PR #1164, opened 2024-10-22
Add ability to order messages

nplasterer:
Messages always come in ascending order which is not useful with messaging as you actually always want them in descending order. This allows us to specify a direction so that limit works correctly still select the correct N items based on the sort direction.

## Release Notes

- **New Features**
  - Enhanced messaging framework with improved consent management and message retrieval options.
  - Added sorting capabilities for message retrieval, allowing users to specify ascending or descending order.
  - Updated command-line interface (CLI) for better message querying flexibility.

- **Bug Fixes**
  - Improved error handling in message synchronization and retrieval processes.

- **Tests**
  - Expanded test coverage for new functionalities, ensuring robust performance and reliability.

### DIFF
--- a/xmtp_mls/src/storage/encrypted_store/group.rs
+++ b/xmtp_mls/src/storage/encrypted_store/group.rs
@@ -1,8 +1,10 @@
 //! The Group database table. Stored information surrounding group membership and ID's.
 
+use diesel::query_dsl::QueryDsl;
 use diesel::{
     backend::Backend,
     deserialize::{self, FromSql, FromSqlRow},
+    dsl::sql,
     expression::AsExpression,
     prelude::*,
     serialize::{self, IsNull, Output, ToSql},
@@ -12,8 +14,9 @@ use diesel::{
 use serde::{Deserialize, Serialize};
 
 use super::{
+    consent_record::ConsentState,
     db_connection::DbConnection,
-    schema::{groups, groups::dsl},
+    schema::groups::{self, dsl},
     Sqlite,
 };
 use crate::{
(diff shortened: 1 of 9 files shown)


xmtp_mls/src/storage/encrypted_store/group_message.rs
```diff
@@ -140,6 +141,14 @@ impl DbConnection {
             query = query.filter(dsl::delivery_status.eq(status));
         }
 
+        if let Some(dir) = direction {
+            query = match dir {
+                SortDirection::Ascending => query.order(dsl::sent_at_ns.asc()),
+                SortDirection::Descending => query.order(dsl::sent_at_ns.desc()),
+                SortDirection::Unspecified => query,
```
neekolas, line 148: Do we need Unspecified if the direction argument is optional?

neekolas: Ahhh. Because you're taking the proto enum, which always comes with Unspecified.

Could make a specific enum for the DB that only has two options and convert Unspecified -> None

codabrink: It is like putting a hat on a hat. What would you think of making direction non-optional in the args?

insipx: Since its defined in the proto it is required here, or using `_ =>` but doesn't matter much in this case

xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -126,40 +129,56 @@ impl DbConnection {
...
-        query = query.filter(dsl::purpose.eq(Purpose::Conversation));
+        if let Some(state) = consent_state {
+            query =
+                query
+                    .inner_join(consent_dsl::consent_records.on(
+                        sql::<diesel::sql_types::Text>("hex(groups.id)").eq(consent_dsl::entity),
+                    ))
+                    .filter(consent_dsl::state.eq(state))
+                    .select(groups_dsl::groups::all_columns());
+        }
```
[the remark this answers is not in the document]
insipx: This one is tricky, I'll have to come back to it. Left a message in the diesel gitter chat in case anyone in the community knows. 

one way I know how to make this work now, would be to create another fn that is able to filter on consent state, and re-use the current filters by splitting it into an inner fn that accepts a box query. Then we can do the join at the top-level. The reason this doesn't work is b/c the inner table type is now `Join<Left, Right>` instead of just `table`

xmtp_mls/src/storage/encrypted_store/group_message.rs
```diff
@@ -140,6 +147,13 @@ impl DbConnection {
             query = query.filter(dsl::delivery_status.eq(status));
         }
 
+        if let Some(dir) = direction {
```
neekolas, line 150: This is going to change the default behavior (if None is specified it won't be sorted at all). 

Would be a safer change if we default to ASC so we don't surprise any devs who might leave the sort direction out (or us)

neekolas (approved):

xmtp_mls/src/storage/encrypted_store/group.rs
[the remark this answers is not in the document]
insipx: Screenshot 2024-10-23 at 8 34 48 AM


I can create that common codepath real quick. I'll just separate out the filters from the table formation, WDYT?

PR #1173, opened 2024-10-22
Filter conversations by consent

nplasterer:
Adds the ability to filter a conversation list by the consent state so that we can show just a list of allowed conversations and in the future just sync those lists as well.


xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -126,40 +129,56 @@ impl DbConnection {
...
-        query = query.filter(dsl::purpose.eq(Purpose::Conversation));
+        if let Some(state) = consent_state {
+            query =
+                query
+                    .inner_join(consent_dsl::consent_records.on(
+                        sql::<diesel::sql_types::Text>("hex(groups.id)").eq(consent_dsl::entity),
+                    ))
+                    .filter(consent_dsl::state.eq(state))
+                    .select(groups_dsl::groups::all_columns());
+        }
```
nplasterer, line 173: @insipx pulled this into a separate PR so it wasn't blocking the message order stuff. Just FYI.

insipx (approved):

PR #1182, opened 2024-10-23
Pass errors into Stream Callbacks

### LINKED ISSUE #899: Feature request: Pass stream errors through to callbacks

insipx:
closes #899 

rel: #1149 

## Summary by CodeRabbit

- **New Features**
	- Introduced new error handling mechanisms with the `SubscribeError` enum for improved subscription processes.
	- Enhanced logging and error reporting in synchronization and message processing methods.

- **Bug Fixes**
	- Streamlined error handling in various methods, ensuring clearer feedback and robustness in message processing.

- **Documentation**
	- Updated scripts and commands for generating bindings to reflect new structures and error handling.

- **Chores**
	- Modified dependency versions and configurations for better compatibility and performance.


bindings_ffi/src/mls.rs
```diff
@@ -1799,14 +1575,16 @@ impl FfiStreamCloser {
     }
 }
 
-#[uniffi::export(callback_interface)]
+#[uniffi::export(with_foreign)]
```
insipx, line 1578: `callback_interface` was deprecated

nplasterer (approved):

Issue #1187, opened 2024-10-25
Enforce consistent length of the `account_address` input for inbox IDs


neekolas:
### Impact
The method for creating an XIP-46 Inbox ID is not collision-resistant, which may result in two distinct addresses being associated with the same Inbox ID.

### Description
The XIP-46 specification uses an Inbox ID as the primary identifier for a user within XMTP, and provides a framework for associating multiple addresses with said Inbox ID.

> In this model, users of the protocol will be identified by Inbox IDs. An Inbox ID can be treated as an opaque string by applications but is constrained to the hash of the first associated address and a nonce.

More specifically, the Inbox ID is computed via a simple hash of the concatenation of the first address and a nonce:

> The inbox_id is derived via SHA256(CONCAT($account_address, $nonce))

The nonce appears to be a mechanism by which a single address may be associated with more than one XMTP inbox. The above is implemented in a straightforward manner:

```rust
fn sha256_string(input: String) -> String {
    let mut hasher = Sha256::new();
    hasher.update(input.as_bytes());
    let result = hasher.finalize();
    format!("{:x}", result)
}

pub fn generate_inbox_id(account_address: &str, nonce: &u64) -> String {
    sha256_string(format!("{}{}", account_address.to_lowercase(), nonce))
}

It was observed that no length information is incorporated into this hash, which means it is trivial to produce multiple (address, nonce) pairs that produce the same Inbox ID. For example,

	•	(“abc123”, 0)
	•	(“abc12”, 30)
	•	(“abc1”, 230)
	•	(“abc”, 1230)

all produce the same Inbox ID according to the specified and implemented approach.

When concatenating variable-length data for hashing, it is generally recommended to incorporate the length of each field as part of the hash input to avoid the above scenario. For example,

	•	(“6”, “abc123”, “1”, “0”)
	•	(“5”, “abc12”, “2”, “30”)
	•	(“4”, “abc1”, “3”, “230”)
	•	(“3”, “abc”, “4”, “1230”)

would each produce a different Inbox ID when concatenated and hashed. Similarly, encoding data using a standardized type-length-value (TLV) format would implicitly provide this approach. Alternatively, a different construction, such as that used in HMAC, would also avoid the issue, where the Inbox ID is calculated using SHA256(SHA256(address) || SHA256(nonce)).

[...]

PR #1202, opened 2024-10-30
Enforce 42 characters for address on inboxId create

nplasterer:
Fixes https://github.com/xmtp/libxmtp/issues/1187


bindings_ffi/src/mls.rs
```diff
@@ -167,7 +167,7 @@ pub async fn get_inbox_id_for_address(
 #[allow(unused)]
 #[uniffi::export]
 pub fn generate_inbox_id(account_address: String, nonce: u64) -> String {
-    xmtp_id_generate_inbox_id(&account_address, &nonce)
+    xmtp_id_generate_inbox_id(&account_address, &nonce)?
```
neekolas, line 170: To use the ? this function needs to return a Result

xmtp_id/src/associations/hashes.rs
```diff
@@ -1,12 +1,27 @@
...
     hasher.update(input.as_bytes());
     let result = hasher.finalize();
     format!("{:x}", result)
 }
 
-pub fn generate_inbox_id(account_address: &str, nonce: &u64) -> String {
-    sha256_string(format!("{}{}", account_address.to_lowercase(), nonce))
+/// Validates that the account address is exactly 42 characters, starts with "0x", 
+/// and contains only valid hex digits.
+fn is_valid_address(account_address: &str) -> bool {
```
neekolas, line 15: Looks right to me

bindings_ffi/src/mls.rs
codabrink: Also works with Optionals

insipx (approved):

PR #1207, opened 2024-10-31
Automatic sync - sync worker

codabrink:


xmtp_mls/src/groups/mod.rs
```diff
@@ -228,6 +228,10 @@ pub struct MlsGroup<C> {
     mutex: Arc<Mutex<()>>,
 }
 
+// C is wrapped in an Arc, and so is the Mutex, so this is fine.
+unsafe impl<C> Send for MlsGroup<C> {}
+unsafe impl<C> Sync for MlsGroup<C> {}
```
codabrink, line 233: @insipx This is so we can use this feature on wasm since the Client is neither Send nor Sync there. Everything in MlsGroup that isn't Send/Sync already is wrapped in an Arc, so this should be fine.

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,145 @@ pub enum DeviceSyncError {
...
+
+        crate::spawn(None, {
+            let client = self.clone();
+
+            let receiver = client.local_events.subscribe();
+            let mut sync_stream = Box::pin(receiver.stream_sync_messages());
+
+            async move {
+                while let Err(err) = client.sync_worker(&mut sync_stream).await {
+                    tracing::error!("Sync worker error: {err}");
```
mchenani, line 133: Question: What would happen if something were to happen here?

codabrink: It logs it and marches on to the next sync request. We don't want the sync worker to outright crash.

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,145 @@ pub enum DeviceSyncError {
...
+            while tries < 3 {
+                if let Some(msg) = conn.get_group_message(message_id)? {
+                    return Ok(msg);
+                }
+                thread::sleep(Duration::from_millis(20));
+                tries += 1;
             }
-            DeviceSyncKind::Unspecified => return Err(DeviceSyncError::UnspecifiedDeviceSyncKind),
+
+            Err(DeviceSyncError::NoPendingRequest)
```
mchenani, line 165: question: is this case a true NoPendingRequest?

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,145 @@ pub enum DeviceSyncError {
...
-                vec![self.syncable_groups(conn)?, self.syncable_messages(conn)?]
+        let provider = self.mls_provider()?;
+
+        let get_message_debounce = |conn: &DbConnection, message_id: &[u8]| {
+            let mut tries = 0;
+            while tries < 3 {
+                if let Some(msg) = conn.get_group_message(message_id)? {
+                    return Ok(msg);
+                }
+                thread::sleep(Duration::from_millis(20));
```
insipx, line 161: we need to avoid sleep in async contexts, should use `crate::sleep` shim which is tokios equivalent/js timeout for wasm

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,145 @@ pub enum DeviceSyncError {
...
-        let (_msg, request) = self.pending_sync_request(provider, kind).await?;
-        let records = match kind {
-            DeviceSyncKind::Consent => vec![self.syncable_consent_records(conn)?],
-            DeviceSyncKind::MessageHistory => {
-                vec![self.syncable_groups(conn)?, self.syncable_messages(conn)?]
+        let provider = self.mls_provider()?;
+
+        let get_message_debounce = |conn: &DbConnection, message_id: &[u8]| {
+            let mut tries = 0;
+            while tries < 3 {
```
insipx, line 157: Could also use `crate::retry::retry_async!` macro here: [macro](https://xmtp.github.io/libxmtp/xmtp_mls/macro.retry_async.html) and optional [builder](https://xmtp.github.io/libxmtp/xmtp_mls/retry/index.html)

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,145 @@ pub enum DeviceSyncError {
...
+            let msg = msg?;
+            match msg {
+                SyncMessage::Reply { message_id } => {
+                    let conn = provider.conn_ref();
+                    let msg = get_message_debounce(conn, &message_id)?;
+
+                    let msg_content: DeviceSyncContent =
+                        serde_json::from_slice(&msg.decrypted_message_bytes)?;
+                    let DeviceSyncContent::Reply(reply) = msg_content else {
+                        // This should never happen
```
mchenani, line 178: question: what if it happened? what we can do?

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,145 @@ pub enum DeviceSyncError {
...
+                    }
+                }
+                SyncMessage::Request { message_id } => {
+                    let conn = provider.conn_ref();
+                    let msg = get_message_debounce(conn, &message_id)?;
+
+                    let msg_content: DeviceSyncContent =
+                        serde_json::from_slice(&msg.decrypted_message_bytes)?;
+                    let DeviceSyncContent::Request(request) = msg_content else {
+                        // This should never happen
```
mchenani, line 194: same as above.

xmtp_mls/src/groups/device_sync.rs
codabrink: Then the code is wrong. I was tempted to put an `unreachable!();` here. :wink: Thinking about it a bit more, it would probably make more sense to do that. Let me update.

xmtp_mls/src/groups/device_sync.rs
codabrink: Hah, I missed this. Thanks for the callout. Let me update this to a StorageError

xmtp_mls/src/groups/mod.rs
insipx: If this is only for wasm, we should have it under `#[cfg(target_arch = "wasm32")]`

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,141 @@ pub enum DeviceSyncError {
...
+        &self,
+        provider: &XmtpOpenMlsProvider,
+    ) -> Result<(), DeviceSyncError> {
+        self.sync_init(provider).await?;
+
+        crate::spawn(None, {
+            let client = self.clone();
+
+            let receiver = client.local_events.subscribe();
+            let mut sync_stream = Box::pin(receiver.stream_sync_messages());
```
insipx, line 129: we can avoid boxing by instead using `pin_mut!` right after moving it into the future/first thing in `async_move`

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,141 @@ pub enum DeviceSyncError {
...
     V: SmartContractSignatureVerifier,
 {
-    pub async fn reply_to_sync_request(
+    pub(crate) async fn sync_worker(
         &self,
-        provider: &XmtpOpenMlsProvider,
-        kind: DeviceSyncKind,
-    ) -> Result<DeviceSyncReplyProto, DeviceSyncError> {
-        let conn = provider.conn_ref();
+        sync_stream: &mut Pin<Box<impl Stream<Item = Result<SyncMessage, SubscribeError>>>>,
```
insipx, line 149: if we avoid the box, this becomes `&mut (impl Stream<Item = Result<SyncMessage, SubscribeError>> + Unpin)`

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -101,57 +107,141 @@ pub enum DeviceSyncError {
     UnspecifiedDeviceSyncKind,
     #[error("sync reply is too old")]
     SyncReplyTimestamp,
+    #[error(transparent)]
+    Subscribe(#[from] SubscribeError),
+}
+
+impl<ApiClient, V> Client<ApiClient, V>
+where
+    ApiClient: XmtpApi + 'static,
```
insipx, line 116: we can remove unsafe `impl Send` by requiring `Send + Sync` here, and it will work in wasm -- its the same that we do for our stream functions (look at `stream_all_messages` for as a reference). It works, because `MlsGroup` _is_ actually `Send + Sync` in wasm, its the futures that aren't

since this is a separate `impl` we avoid the headache of making `Send + Sync` in the larger `impl` block

insipx (changes requested):

nplasterer:
I tested this in the android SDK and works like expected 🎉

xmtp_mls/src/groups/device_sync.rs
codabrink: A silly oversight on my part, thanks for catching this!

xmtp_mls/src/groups/mod.rs
```diff
@@ -230,6 +230,10 @@ pub struct MlsGroup<C> {
     mutex: Arc<Mutex<()>>,
 }
 
+// C is wrapped in an Arc, and so is the Mutex, so this is fine.
+// unsafe impl<C> Send for MlsGroup<C> {}
```
insipx, line 234: can get rid of the comment now too

insipx (approved):

PR #1214, opened 2024-11-04
Teach example/CLI how to talk to V4/Replication Node

mkysel:
This is the first (large) step in the right direction. By no means is the migration to decentralization complete.

This teaches the CLI to do basic operations via xtmpd such as:
- registering
- creating groups
- adding members to groups
- messaging

One has to simply provide the `--testnet` flag to the CLI and everything will work against the new API.

The easiest way to spin up the new environment is to wait for https://github.com/xmtp/libxmtp/pull/1281.

This refactors the Error class into its own file and adds an Internal error.

Also adds a new command to the CLI for debugging.


xmtp_proto/src/convert.rs
```diff
@@ -16,3 +44,146 @@ mod inbox_id {
         }
     }
 }
+
+impl From<UploadKeyPackageRequest> for PublishClientEnvelopesRequest {
+    fn from(req: UploadKeyPackageRequest) -> Self {
+        PublishClientEnvelopesRequest {
+            envelopes: vec![ClientEnvelope {
+                aad: Some(AuthenticatedData::with_topic(get_key_package_topic(
+                    req.key_package.as_ref().unwrap(),
```
neekolas, line 53: Instead of using unwraps, can we make this a TryFrom and return an error?

xmtp_proto/src/convert.rs
```diff
@@ -16,3 +44,146 @@ mod inbox_id {
...
+                ))),
+                payload: Some(Payload::UploadKeyPackage(req)),
+            }],
+        }
+    }
+}
+
+impl From<PublishIdentityUpdateRequest> for PublishClientEnvelopesRequest {
+    fn from(req: PublishIdentityUpdateRequest) -> Self {
+        let identity_update = req.identity_update.unwrap();
```
neekolas, line 63: Same as above. Easier to start with no unwraps than to remove them later

Cargo.toml
```diff
@@ -96,7 +96,7 @@ xmtp_proto = { path = "xmtp_proto" }
 [profile.dev]
 # Disabling debug info speeds up builds a bunch,
 # and we don't rely on it for debugging that much.
-debug = 0
+#debug = 0
```
mkysel, line 99:
```suggestion
```

This needs to be undone before merge

xmtp_api_grpc/src/replication_client.rs
```diff
@@ -170,207 +160,166 @@ impl XmtpApiClient for ClientV4 {
...
         req: QueryGroupMessagesRequest,
     ) -> Result<QueryGroupMessagesResponse, Error> {
-        unimplemented!();
+        let client = &mut self.client.clone();
+        let res = client
+            .query_envelopes(QueryEnvelopesRequest {
+                query: Some(EnvelopesQuery {
+                    topics: vec![build_group_message_topic(req.group_id.as_slice())],
+                    originator_node_ids: vec![],
+                    last_seen: None,
```
richardhuaaa, line 237: As a note - the `PagingInfo` of the `QueryGroupMessagesRequest` has an `id_cursor` that we could potentially convert into vector clock format here as a short-term hack. Unsure if we get subtle bugs without this, as the client expects to query from where it left off in the queue. In the long-term we need to do a bigger refactor to properly support vector clocks though, just making a note in case you run into bugs in the meantime

xmtp_api_grpc/src/replication_client.rs
```diff
@@ -170,207 +160,166 @@ impl XmtpApiClient for ClientV4 {
...
+                }),
+                limit: 100,
```
richardhuaaa, line 239: There is a `limit` on the `PagingInfo` we should honor here

xmtp_api_grpc/src/replication_client.rs
```diff
@@ -170,207 +160,166 @@ impl XmtpApiClient for ClientV4 {
...
+        let response = QueryGroupMessagesResponse {
+            messages: envelopes
+                .iter()
+                .map(|envelope| {
+                    let unsigned_originator_envelope =
+                        extract_unsigned_originator_envelope(envelope).unwrap();
+                    let client_envelope = extract_client_envelope(envelope).unwrap();
+                    let Payload::GroupMessage(group_message) = client_envelope.payload.unwrap()
+                    else {
+                        panic!("Payload is not a group message");
```
richardhuaaa, line 254: Could we return an error rather than panic? In general we want to avoid giving malicious clients or node operators the ability to arbitrarily crash our client

xmtp_api_grpc/src/replication_client.rs
```diff
@@ -170,207 +160,166 @@ impl XmtpApiClient for ClientV4 {
...
+                    };
+
+                    let group_message_input::Version::V1(v1_group_message) =
+                        group_message.version.unwrap();
+
+                    GroupMessage {
+                        version: Some(group_message::Version::V1(group_message::V1 {
+                            id: unsigned_originator_envelope.originator_sequence_id,
```
richardhuaaa, line 262: Just making a note, eventually we'll need to refactor the client to store both the originator node id and sequence ID in order to set the vector clock properly, but this is great for now

xmtp_api_grpc/src/replication_client.rs
```diff
@@ -462,3 +443,83 @@ impl XmtpTestClient for ClientV4 {
...
+        let requests = topics.iter().map(|topic| async {
+            let client = &mut self.client.clone();
+            let v4_envelopes = client
+                .query_envelopes(QueryEnvelopesRequest {
+                    query: Some(EnvelopesQuery {
+                        topics: vec![topic.clone()],
+                        originator_node_ids: vec![],
+                        last_seen: None,
+                    }),
+                    limit: 100,
```
richardhuaaa, line 461: To avoid bugs, I think we should take in the `last_seen` and `limit` from the caller of this method, and honor whatever semantics were previously used

xmtp_mls/src/hpke.rs
```diff
@@ -75,6 +75,7 @@ pub fn decrypt_welcome(
         // With the hash reference we can read the key package.
         let key_package: Option<KeyPackageBundle> = provider.storage().key_package(&hash_ref)?;
 
+        // MKYSEL Marker - storage key_package is empty!
```
richardhuaaa, line 78: Can remove this now probably

xmtp_proto/src/error.rs
```diff
@@ -0,0 +1,138 @@
+use openmls::prelude::tls_codec::Error as TlsCodecError;
```
richardhuaaa, line 1: Is this `error.rs` file generated? Maybe I missed something, why do we need a new error type different from before?

richardhuaaa (approved):
Well done! The structure of the helper methods is very clean.

We'll also need to add validation of the payloads coming from the server eventually (e.g. signed correctly and from the expected originator), but we can document this properly in the XIP first.

xmtp_mls/src/hpke.rs
mkysel: thanks!

xmtp_proto/src/error.rs
mkysel: I just moved it to its own file as a drive-by refactor.

I added the InternalError and all of its subtypes to allow for error conversions.

xmtp_api_grpc/src/replication_client.rs
mkysel: where do I find it?

PR #1215, opened 2024-11-04
Mc/add mit license to all crates

mchenani:
**Description**
Set the licensing for all crates and the workspace to the MIT license.

**Issue:**
We need to specify the license for Cargo-Deny.


insipx (approved):
nice, first commit!

Issue #1192, opened 2024-10-25
Update dependencies with known RustSec advisories


neekolas:
### Impact
Vulnerabilities in third-party dependencies may be inherited by the application. Even if known vulnerabilities do not apply, the presence of vulnerable dependencies may still affect the perceived security posture of the application and its maintainers.

### Description
The Rust ecosystem provides several tools for managing dependencies, such as `cargo audit` for identifying known security issues, `cargo outdated` for identifying stale dependencies, and `cargo deny` for blocking or allowing various vulnerable or outdated dependencies.

The `cargo audit` tool identifies two vulnerable dependencies:

- **diesel 2.2.2** - "Binary Protocol Misinterpretation caused by Truncating or Overflowing Casts"
- **tonic 0.12.2** - "Remotely exploitable Denial of Service in Tonic"

along with 3 dependencies with warnings:

- **ansi_term 0.12.1** - Unmaintained
- **dirs 5.0.1** - Unmaintained
- **futures-util 0.3.30** - Yanked

In general, it is recommended to actively address or update any applicable RustSec advisories that affect the application. The `cargo deny` tool can be used to automatically fail builds if a vulnerable crate is detected, and also supports a list of exceptions so that the inclusion of such a package can be explicitly reviewed and annotated. Additionally, GitHub’s [Dependabot Service](https://github.blog/2021-04-22-introducing-dependabot-security-updates/) can be configured to scan for and open issues or PRs when updated dependencies are found.

It is emphasized that this review is a point-in-time evaluation of an evolving project, and the presence of outdated dependencies is expected. The affecting advisories for vulnerable crates are recent and do not appear indicative of any deeper issue within the project.

### Recommendation
Consider one or more of the following:

- Adoption of `cargo deny` to automatically detect new RustSec advisories.
- Adoption of Dependabot to automatically update dependencies.
- Ensure that dependencies are reviewed, updated, and tested as part of any documented release ceremonies.

### Location
`Cargo.toml`

PR #1218, opened 2024-11-04
Remove dependency on openmls test-utils

insipx:
@neekolas or @richardhuaaa would appreciate your review on this one if possible. This touches some sensitive code around our installation keys. We were using `ed25519_dalek::SigningKey` in a sortof roundabout way, so I formalized into `xmtp_cryptography::basic_credential`. Then removed some clones to remove a dependency on `test-utils`

wrote some tests to triple-check `basic_credential` is OK, but think this should simplify & centralize installation key operations a bit

This allows us to pass all RUSTSEC advisories. Removing test utils feature also removes a bunch of unused crates in the crate graph for the production binary/ WASM

Part of #1192


xmtp_cryptography/src/basic_credential.rs
```diff
@@ -0,0 +1,336 @@
...
+use openmls_basic_credential::SignatureKeyPair;
+use openmls_traits::signatures::Signer;
+use openmls_traits::{signatures, types::SignatureScheme};
+use serde::de::Error;
+use sha2::Sha512;
+use tls_codec::SecretTlsVecU8;
+use zeroize::Zeroizing;
+
+/// DO NOT CHANGE. SIGNATURES WILL BREAK
+const INSTALLATION_KEY_SIGNATURE_CONTEXT: &[u8] = b"IDENTITY UPDATE SIGNATURE";
```
neekolas, line 14: Feels like this should get passed in, given it’s in a whole other crate from where this context comes from.

neekolas (approved):
This is great work. The test coverage that is there gives me good confidence this isn't going to break something.

If we want to be really sure, an easy smoke test would be to create a new identity using the CLI on `main` and upgrading to this version (since we use store the installation keys in the DB). Then we can send a message to another CLI client that is back on main.

PR #1219, opened 2024-11-04
chore(dep): add cargo-deny config

mchenani:
**Description:**
Add cargo-deny config file to the project.

To check cargo-deny results use `cargo deny check`

Want more? [👀📖](https://github.com/EmbarkStudios/cargo-deny)

**Notes**:
[RUSTSEC-2021-0139 ansi_term is Unmaintained](https://rustsec.org/advisories/RUSTSEC-2021-0139.html)
I intentionally set cargo-deny to ignore it for now, but we need to review this dependency and choose an alternative.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,33 @@
+[advisories]
+ignore = [
+    {id="RUSTSEC-2021-0139", reason="The maintainer has advised that this crate is deprecated and will not receive any maintenance. https://rustsec.org/advisories/RUSTSEC-2021-0139.html"},
+]
+
+[licenses]
+allow = [
+    "Apache-2.0",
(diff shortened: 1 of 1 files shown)


insipx (approved):

deny.toml
```diff
@@ -0,0 +1,33 @@
+[advisories]
+ignore = [
+    {id="RUSTSEC-2021-0139", reason="The maintainer has advised that this crate is deprecated and will not receive any maintenance. https://rustsec.org/advisories/RUSTSEC-2021-0139.html"},
```
mchenani, line 3: @insipx do we need to create an issue for this to decide about it? we need to user other alternatives.

PR #1224, opened 2024-11-06
Clean up node bindings

rygine:
# Summary

- Removed all `Napi` prefixes
- Fixed stream callback argument types
- Renamed group to `Conversation`
- Removed `AsyncStream` from testing, use callback instead
- Added `lint` command
- Updated `build` command

### Notes

`Napi` prefixes were removed to create more consistent named exports for SDKs. The WASM bindings will follow suit.


insipx (approved):
nice!

PR #1226, opened 2024-11-06
feat(inbox): cache wallet address lookup

### LINKED ISSUE #1143: Make wallet address look up cached
Right now anytime we go from inbox id to address or vice versa we hit the network. Since these will change in frequently and we call them often it should just be mostly a local lookup.

Lets add a refresh param similiar to how `inboxState` works to the existing method to see if the user wants to fetch from the network or local.

mchenani:
**Description:**
Implemented a caching mechanism to reduce unnecessary API calls when fetching wallet addresses associated with inboxIds.

**Problem:**
Issue #1143 highlights that every lookup from inbox_id to wallet_address (and vice versa) currently triggers a network request. Given that these mappings change infrequently and are accessed often, a local cache is more efficient.

**Changes:**
* Added a new database table, wallet_addresses, with columns for inbox_id and wallet_address.
* Introduced logic to first check the database for the requested wallet addresses. If all requested addresses are found, they are returned directly from the cache.
* If any requested wallet address is missing from the cache, an API call is made to retrieve the missing addresses. These are then stored in the database for future lookups.

This update minimizes network calls and improves performance by leveraging cached data for wallet address lookups.

Closes #1143


xmtp_mls/migrations/2024-11-05-191238_cache_wallet_addresses/up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE wallet_addresses(
+    inbox_id TEXT NOT NULL,
```
codabrink, line 2: We'll probably want to create an index for this column alone, or just make it the primary key instead of using a compound primary key.

codabrink: Oh, I read the PR title and thought we're looking up addresses, but apparently it's the other way around? :smile: If so, then I mean the other column.

xmtp_mls/src/client.rs
```diff
@@ -370,13 +356,48 @@ where
...
-            .await?;
-        let inbox_ids: Vec<Option<String>> = sanitized_addresses
+        let conn = self.store().conn()?;
+
+        let local_results: Vec<WalletEntry> = conn.fetch_wallets_list_with_key(&sanitized_addresses)?;
+
+        let mut results: HashMap<String, String> = local_results
             .into_iter()
-            .map(|address| results.remove(&address))
+            .map(|entry| (entry.wallet_address.clone(), entry.inbox_id.clone()))
```
codabrink, line 365: Probably no need to clone here, since we're `.into_iter()`

xmtp_mls/src/client.rs
```diff
@@ -370,13 +356,48 @@ where
...
+        let missing_addresses: Vec<String> = sanitized_addresses
+            .iter()
+            .filter(|address| !results.contains_key(*address))
+            .cloned()
+            .collect();
+
+        if missing_addresses.is_empty() {
+            let inbox_ids: Vec<Option<String>> = sanitized_addresses
+                .iter()
+                .map(|address| results.get(address).cloned())
```
codabrink, line 377: You can also avoid a `.clone()` here by using `.remove(...)` instead.

xmtp_mls/src/client.rs
```diff
@@ -370,13 +356,48 @@ where
...
+                .collect();
+            return Ok(inbox_ids);
+        }
+
+        let web_results = self.api_client.get_inbox_ids(missing_addresses.clone()).await?;
```
codabrink, line 382: No need for `.clone()`

xmtp_mls/src/client.rs
```diff
@@ -370,13 +356,48 @@ where
...
+
+        for (address, inbox_id) in &web_results {
+            results.insert(address.clone(), inbox_id.clone()).unwrap_or_default();
+        }
+
+        for (address, inbox_id) in &web_results {
```
codabrink, line 388: We can combine these, and remove the borrow in the `web_results` we're iterating on.

xmtp_mls/src/client.rs
```diff
@@ -370,13 +356,48 @@ where
...
+            if let id = inbox_id.clone() {
```
codabrink, line 389: This always matches. No need for `if let`.

xmtp_mls/src/client.rs
```diff
@@ -370,13 +356,48 @@ where
...
+                    inbox_id: InboxId::from(id),
+                    wallet_address: address.clone(),
+                };
+                new_entry.store(&conn).ok();
+            }
+        }
+
+        let inbox_ids: Vec<Option<String>> = sanitized_addresses
+            .iter()
+            .map(|address| results.get(address).cloned())
```
codabrink, line 400:
```suggestion
            .iter()
            .map(|address| results.remove(address))
```

xmtp_mls/migrations/2024-11-05-191238_cache_wallet_addresses/up.sql
```diff
@@ -0,0 +1,7 @@
...
+    wallet_address TEXT NOT NULL,
+    PRIMARY KEY (inbox_id, wallet_address)
```
codabrink, line 4: It'd probably be better to make `wallet_address` the primary key.

codabrink: And then add a unique index on `inbox_id`

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -362,6 +363,39 @@ macro_rules! impl_fetch {
...
+macro_rules! impl_fetch_list_with_key {
+    ($model:ty, $table:ident, $key:ty, $column:ident) => {
+        impl $crate::FetchListWithKey<$model>
+            for $crate::storage::encrypted_store::db_connection::DbConnection
+        {
+            type Key = $key;
+            fn fetch_list_with_key(
+                &self,
+                keys: &[Self::Key],
+            ) -> Result<Vec<$model>, $crate::StorageError> {
```
codabrink, line 390: What would you think of having this return `Vec<Option<$model>>` instead so that way we can zip the output with the input of this function?

xmtp_mls/migrations/2024-11-05-191238_cache_wallet_addresses/up.sql
mchenani: I'm not sure if we have a `wallet_address` connected to multiple `inbox_id`s, or it's always 1 `wallet_address` connected to and inbox?

codabrink: That's actually a great thing to bring up. Here's the XIP for that https://www.notion.so/ephemerahq/Multi-Wallet-Identity-XIP-ba30e6ddb30c45acb2dc4ff9eccae7ee

Ignore my unique index comment. I think the wallet PK would be good though.

xmtp_mls/src/storage/encrypted_store/mod.rs
codabrink: It would make processing and matching the output data with the input data much easier. It would simplify the logic/processing you would have to do in the caching PR that you have [here](https://github.com/xmtp/libxmtp/pull/1226/files#diff-3c0189371525b6aae3be7e5a7b5b5682204ef341fb2744fc876148c5ecab8642R379-R396) and would avoid having to collect the data into a hashmap.

xmtp_mls/src/client.rs
```diff
@@ -370,13 +371,46 @@ where
...
+
+        let web_results = self.api_client.get_inbox_ids(missing_addresses).await?;
+
+        for (address, inbox_id) in &web_results {
+            results
+                .insert(address.clone(), inbox_id.clone())
+                .unwrap_or_default();
+            let new_entry = WalletEntry {
+                inbox_id: InboxId::from(inbox_id),
+                wallet_address: address.clone(),
```
codabrink, line 406:
```suggestion
                wallet_address: address,
```

insipx (approved):
nice!

xmtp_mls/src/storage/encrypted_store/mod.rs
mchenani: Then I need to do the mapping in the macro itself after getting the data from the DB, and then in the situation, we don't find any data in the DB we return a `[ None ]` but now it's an empty vec, so I think with that we don't gain that much in making the return value as `Vec<Option<model>, StorageError>`!
what do you thing @insipx?

Issue #1111, opened 2024-10-03
Migrate from `ethers-rs` to `alloy`


insipx:
I began to notice that `ethers` is depending on some slightly older versions of libraries, ex: `rustls`. It seems [allow](https://docs.rs/alloy) now has [feature-parity](https://github.com/gakonst/ethers-rs/issues/2667#issuecomment-2173905063) with ethers, so we can migrate using [this guide](https://alloy.rs/migrating-from-ethers/reference.html) as a reference

PR #1254, opened 2024-11-11
Add debug CLI methods

neekolas:
## tl;dr

- Adds a debug section to the CLI
- Adds the first debug method to read raw payloads from the network
- Changes the way we specify envs in the CLI

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,11 +1,13 @@
 [advisories]
+ignore = [
+    { id = "RUSTSEC-2024-0384", reason = "migration to alloy is underway which will resolve this advisory" },
+]
+
 # This rustsec can be added to ignore list if using mls `test_utils` for tests
 # {
 #   id="RUSTSEC-2021-0139",
 #   reason="The maintainer has advised that this crate is deprecated and will not receive any maintenance. https://rustsec.org/advisories/RUSTSEC-2021-0139.html"
 #  },
-ignore = []
-
 [licenses]
 allow = [
     "Apache-2.0",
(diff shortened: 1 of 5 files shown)


neekolas:
* **#1254** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1254?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

examples/cli/Cargo.toml
```diff
@@ -2,35 +2,49 @@
...
-valuable = { version = "0.1", features = ["derive"] }
-valuable-serde = "0.1"
+chrono.workspace = true
 clap = { version = "4.4.6", features = ["derive"] }
+color-eyre = "0.6"
 ethers.workspace = true
 futures.workspace = true
 hex.workspace = true
+kv-log-macro = "1.0.7"
+log = { version = "0.4", features = [
```
insipx, line 23: can remove this, i made everything use `tracing` to be consistent and preserve the `async` context of spans & added K/V pairs to `xmtp_mls`. I made sure the log format was the same. JSON logs can be passed into third party log viewer.

You can use k/v pairs with `=` before the log message, i.e

`info!(key = value, "my message {}", format_arg);`

examples/cli/debug.rs
```diff
@@ -0,0 +1,62 @@
...
+            .extract()
+        {
+            MlsMessageBodyIn::PrivateMessage(message) => message,
+            _ => return Err("Unsupported message type".to_string()),
+        };
+        let timestamp = format_timestamp(msgv1.created_ns);
+        let sequence_id = msgv1.id;
+        let epoch = body.epoch();
+        let content_type = body.content_type();
+        info!("[{timestamp}] [Epoch {epoch}] [Seq {sequence_id}] {content_type:?}");
```
insipx, line 46: it could be useful to make timestamp, epoch and sequence id K/V pairs here

insipx, line 46:
```suggestion
        info!(timestamp, epoch, sequence_id, "[{timestamp}] [Epoch {epoch}] [Seq {sequence_id}] {content_type:?}");
```

neekolas: I did try that. Comes with the unfortunate consequence of making each value take 4 lines instead of 1. 

When you’re looking at 1000 messages that’s a big difference.

deny.toml
```diff
@@ -4,7 +4,7 @@
 #   id="RUSTSEC-2021-0139",
 #   reason="The maintainer has advised that this crate is deprecated and will not receive any maintenance. https://rustsec.org/advisories/RUSTSEC-2021-0139.html"
 #  },
-ignore = []
+ignore = ["instant@0.1.13"]
```
neekolas, line 7: I had to add this. Not sure it wasn't flagging before, since this library was moved to "unmaintained" 6 months ago. But it's still a dependency in ethers...which has itself [moved to unmaintained status](https://github.com/gakonst/ethers-rs) as of a few weeks ago.

insipx: yeah we have to update ethers to alloy anyway, #1111, it should be ok to ignore for now

examples/cli/debug.rs
insipx: That's fair, I got confused with the inclusion of the k-v crate, thought maybe it was something you were looking to do

The K/V are mostly for the JSON output if using some kind of logging software like `lnav`, maybe in a future PR we can just hide the fields for stdout output

deny.toml
neekolas: OK. Fixed

insipx (approved):

mkysel:
This is cool! I will need the other message types too. Hoping to have it done as a follow-up by tomorrow

PR #1260, opened 2024-11-13
Extend the debug capabities of the CLI

mkysel:
Add the other three message types.

Change the parsing of the options to match other commands and allow the user to omit `--group-id` if there is only one named argument required.

Builds on top of #1254 by @neekolas


examples/cli/debug.rs
```diff
@@ -49,6 +65,106 @@ pub async fn debug_group_messages(
...
+
+    let envelopes: Result<Vec<VerifiedKeyPackageV2>, _> = key_package_results
+        .values()
+        .map(|bytes| VerifiedKeyPackageV2::from_bytes(mls_provider.crypto(), bytes.as_slice()))
+        .collect();
+
+    for envelope in envelopes.unwrap() {
+        let inbox_id = envelope.credential.inbox_id;
+        let pkey = hex::encode(envelope.installation_public_key);
+        info!("[InboxId] {inbox_id} [Key Packages] {pkey}");
```
neekolas, line 120: Just a nit, but the formatting of these messages looks different than the others, which enclose the key and the value in a `[ ]`

neekolas (approved):

examples/cli/debug.rs
mkysel: fixed

PR #1261, opened 2024-11-13
Remove Purpose, Add Conversation Type to database. Sync sync groups when syncing all groups too.

codabrink:
- Remove `Purpose` from the db (and the codebase)
- Substitute it for `ConversationType` column in the db
- Add a flag to include sync groups in the find_groups query
- Invoke that flag on the sync all conversations function


xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -228,18 +235,14 @@ impl DbConnection {
...
-            }
+            query = query.filter(groups_dsl::conversation_type.eq(conversation_type));
         }
 
-        query = query.filter(groups_dsl::purpose.eq(Purpose::Conversation));
+        // Were sync groups explicitly asked for? Was the include_sync_groups flag set to true?
+        // Otherwise filter sync groups out by default.
+        if !matches!(conversation_type, Some(ConversationType::Sync)) && !include_sync_groups {
+            query = query.filter(groups_dsl::conversation_type.ne(ConversationType::Sync));
+        }
```
codabrink, line 245: Aside from removing Purpose and adding ConversationType to the db, here's the main change of this PR.
No longer unconditionally filtering out Sync groups, however we still filter them out by default.

cameronvoell (approved):
nice improvement consolidating to conversation_type 👌

PR #1263, opened 2024-11-14
Repro client processing own commit as an external envelope

cameronvoell:
See https://github.com/xmtp/xmtp-react-native/pull/534

** Note this test does not yet repro the issue seen in the RN PR linked above


cameronvoell:
Brainstorming here how parallel member removals an syncs could lead to a race where a client tries to process their own intent without the envelope payload showing in their database yet, and how this can be prevented.

PR #1265, opened 2024-11-14
Lowercase the address being stored in the client on the node bindings

codabrink:


bindings_node/src/client.rs
```diff
@@ -129,6 +129,8 @@ pub async fn create_client(
   history_sync_url: Option<String>,
   log_options: Option<LogOptions>,
 ) -> Result<Client> {
+  let account_address = account_address.to_lowercase();
```
codabrink, line 132: This will lowercase the account_address being stored on line 183. (unchanged in this PR, so I can't comment on it.)

nplasterer:
Are people who previously logged into node possibly going to not be able to log in now due to inboxId mismatch? This seems fine but I wonder if a better additional fix would be wherever the error is being thrown that we do a check on the 3 different ways to create a inboxId from a string?  `ChecksumAddress` `lowercased` maybe `uppercase` and if it finds an inboxId for any of those cases continue otherwise error?

bindings_node/src/client.rs
rygine: curious why this is necessary

Issue #1228, opened 2024-11-06
Remove membership added codec from DMs


nplasterer:
It's weird that the first message of a dm is the membership codec. Lets just not send that message if it's a dm.

cameronvoell:
[Updated in libxmtp](https://github.com/xmtp/libxmtp/pull/1266), just needs to be released in SDKs now

PR #1266, opened 2024-11-14
Filter initial membership change message in dm groups from SDK

cameronvoell:
Closes https://github.com/xmtp/libxmtp/issues/1228

DM groups only ever have one membership change message, and it happens on DM creation, and if it fails DM is not created. You can not leave or add new ids to a DM group. Therefore it should be safe to just filter membership change messages from `find_messages` "queries" in DM groups in order to clear out the noisy user added message at the beginning of all DM groups.


nplasterer (approved):
FYI @rygine since we will probably want to do this in node and wasm as well.

PR #1268, opened 2024-11-14
Check against the inbox id derived against the checksum as well

codabrink:
Fixes https://github.com/xmtp/xmtp-react-native/issues/512

It appears that our SDKs at some point allowed or still allow checksum addresses to create inboxIds. Now we have inboxIds on the network that throw a mismatch id error because we handle most addresses as lowercase.
This is supposed to prevent any checksum inboxIds from being created ever again and help unblock users with inboxIds that were created from checksum addresses.

### DIFF
--- a/xmtp_id/src/associations/association_log.rs
+++ b/xmtp_id/src/associations/association_log.rs
@@ -381,6 +381,7 @@ impl IdentityAction for IdentityUpdate {
         }
 
         let new_state = state.ok_or(AssociationError::NotCreated)?;
+        // Do we need to handle this case for checksum inboxIds?
         if new_state.inbox_id().ne(&self.inbox_id) {
             tracing::error!(
                 "state inbox id mismatch, old: {}, new: {}",
--- a/xmtp_id/src/associations/hashes.rs
+++ b/xmtp_id/src/associations/hashes.rs
@@ -7,8 +7,9 @@ use super::AssociationError;
 
 /// Helper function to generate a SHA256 hash as a hex string.
 fn sha256_string(input: String) -> String {
+    let normalized_input = input.to_lowercase();
     let mut hasher = Sha256::new();
-    hasher.update(input.as_bytes());
+    hasher.update(normalized_input.as_bytes());
     let result = hasher.finalize();
     format!("{:x}", result)
 }
--- a/xmtp_id/src/associations/hashes.rs
+++ b/xmtp_id/src/associations/hashes.rs
@@ -1,3 +1,6 @@
+use std::str::FromStr;
+
+use ethers::{types::H160, utils::to_checksum};
 use sha2::{Digest, Sha256};
 
 use super::AssociationError;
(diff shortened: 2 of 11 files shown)


nplasterer (approved):
I think this makes the most sense to continue to support legacy addresses created from a checksum address. Would be good to know where these checksum inbox Ids are still being created. I'll do another pass on lowercasing in the sdk when I push this fix out through the sdks. And @rygine maybe you can as well whenever you push this out through node and web?

rygine:
@nplasterer i reviewed the WASM/node bindings, and we're already lowercasing where necessary. 🎉

xmtp_id/src/associations/association_log.rs
```diff
@@ -381,6 +381,7 @@ impl IdentityAction for IdentityUpdate {
         }
 
         let new_state = state.ok_or(AssociationError::NotCreated)?;
+        // Do we need to handle this case for checksum inboxIds?
```
nplasterer, line 384: @insipx I could hack some code here to check for mismatched inboxIds from checksums here but it's a lot less straight forward thoughts on this case?

xmtp_id/src/associations/hashes.rs
```diff
@@ -1,11 +1,15 @@
+use std::str::FromStr;
+
+use ethers::{types::H160, utils::to_checksum};
 use sha2::{Digest, Sha256};
 
 use super::AssociationError;
 
 /// Helper function to generate a SHA256 hash as a hex string.
 fn sha256_string(input: String) -> String {
+    let normalized_input = input.to_lowercase();
```
neekolas, line 10: If we want to do this here we should rename the function. Will be a footgun to change the string before hashing, since callers won't expect that.

xmtp_mls/src/identity.rs
```diff
@@ -104,7 +103,8 @@ impl IdentityStrategy {
                 legacy_signed_private_key,
             ) => {
                 if let Some(stored_identity) = stored_identity {
-                    if inbox_id != stored_identity.inbox_id {
+                    let checksum_inbox_id = generate_inbox_id_from_checksum(&address, &nonce)?;
```
neekolas, line 106: If the inbox ID was created with the checksum address, and not the lowercased one, it shouldn't be able to create a valid Create Inbox identity action

neekolas:
How many people do we think this issue is affecting?

I'd much rather force them into creating a new identity and shed tech debt rather than having to support this bug forever.

xmtp_mls/src/identity.rs
neekolas: We check that the inbox ID matches the signer as part of verifying that Identity Action, so it should blow up there. 

So even if your client "works", everyone else would reject you

nplasterer:
The only real person in particular that we know this is effecting is @rygine. But I think a lot of bot accounts and other fresh accounts @humanagent was testing with were also failing but those I think are less important.

PR #1284, opened 2024-11-18
fix(ci): reduce dependabot PR spam

### LINKED ISSUE #1250: Make dependabot group like-kind dependency updates
Dependabot is super noisy b/c it opens a PR for each single dependency update, but dependabot supports grouping. If grouped, dependabot will group dependency updates into a single PR based on a set of rules

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

insipx:
group dependabot updates based upon dev/prod for each dependency grouping (rust/node/wasm)

closes #1250 

rel #236

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,19 +5,38 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      dependencies:
+        dependency-type: "production"
+      dev-dependencies:
+        dependency-type: "development"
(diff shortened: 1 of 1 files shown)


mkysel (approved):

mkysel:
I usually also turn off patch-releases. Nobody cares whether we run 15.0.1 or 15.0.2 so to say :)

PR #1286, opened 2024-11-18
Sign with Installation Key

codabrink:


xmtp_mls/src/identity.rs
```diff
@@ -409,9 +411,23 @@ impl Identity {
...
+
+    pub fn sign_with_public_context<Text: AsRef<str>>(
+        &self,
+        text: Text,
+    ) -> Result<Vec<u8>, IdentityError> {
+        const PUBLIC_SIGNATURE_CONTEXT: &[u8] = b"PUBLIC SIGNATURE CONTEXT";
+
         self.installation_keys
-            .credential_sign(text)
+            .credential_sign(text, Some(PUBLIC_SIGNATURE_CONTEXT))
```
insipx, line 430: You can keep the constants in a central location + leave the context an implementation detail by specializing `CredentialSign` to the public case,

```rust
pub struct PublicInstallationKey;

impl CredentialSign<PublicInstallationKey> for XmtpInstallationCredential {
    const CONTEXT: &[u8] = crate::constants::MY_PUBLIC_CONTEXT_CONST;
    type Error = SignatureError;

    fn credential_sign<S: AsRef<str>>(
        &self,
        text: S,
        context_bytes: Option<&[u8]>,
    ) -> Result<Vec<u8>, Self::Error> {
        let context_bytes = context_bytes.unwrap_or(Self::CONTEXT);

        let mut prehashed: Sha512 = Sha512::new();
        prehashed.update(text.as_ref());
        let context = self.with_context(context_bytes)?;
        let sig = context
            .try_sign_digest(prehashed)
            .map_err(SignatureError::from)?;
        Ok(sig.to_bytes().into())
    }
}
```
then
```rust
self.identity.credential_sign::<PublicInstallationKey>(text)
```

This should be less prone to error since our constants can be in one place and used in one place

codabrink: Yeah, I think you're right. I was initially trying to avoid the duplicated code, but I think I'll just extract that from the implementation body into its own function.

insipx: there's also some workarounds to avoid the fully qualified syntax for this, `HasSome` here: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3e5c322d7408cfc261d1d61d942c3353) but haven't been added yet since this will be the first new context being added

codabrink: Updated

nplasterer (approved):
Looks good to me. Is there any verify signature method to verify that it was made by the installation key?

insipx (approved):

PR #1289, opened 2024-11-19
Restore - Consent Streaming

codabrink:


xmtp_id/src/associations/serialization.rs
```diff
@@ -70,6 +70,8 @@ pub enum DeserializationError {
     InvalidAccountId,
     #[error("Invalid hash (needs to be 32 bytes)")]
     InvalidHash,
+    #[error("A required field is unspecified: {0}")]
+    Unspecified(&'static str),
```
mchenani, line 74: nitpicking: great add, wdyt of something like: `MissingField` or `RequiredFieldMissing`?

codabrink: It's originally there for the shortcomings of protos with unspecified being a required enum field, but I'm happy to generalize it if we want.

nplasterer (approved):
Thanks for fixing this up. 🙏

PR #1290, opened 2024-11-19
Verify signature function

codabrink:
Verify the signatures provided by `sign_with_installation_key`

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -3714,12 +3714,12 @@ mod tests {
 
         // different text should result in an error.
         let result = client.verify_signed_with_installation_key("Hello here.", signature_bytes);
-        assert!(!result.is_ok());
+        assert!(result.is_err());
 
         // different bytes should result in an error
         let signature_bytes = vec![0; 64];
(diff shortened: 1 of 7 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -367,9 +367,28 @@ impl FfiXmtpClient {
 
     pub fn sign_with_installation_key(&self, text: &str) -> Result<Vec<u8>, GenericError> {
         let inner = self.inner_client.as_ref();
-        let context = inner.context().public_sign(text)?;
+        Ok(inner.context().public_sign(text)?)
+    }
+
+    pub fn verify_signed_with_installation_key(
+        &self,
+        signature_text: &str,
```
nplasterer, line 375: Can we just verify without the signature text?

mchenani (approved):
✌️

nplasterer (approved):
This looks fine for a 1 case but I think we will need to also cover the case when your verifying a signature from a different installation and making sure they're from the same client. You can read more about this use case here: https://github.com/ephemeraHQ/converse-backend/discussions/33

Issue #1114, opened 2024-10-03
Streaming Consent Records


nplasterer:
We don't currently have this in V2 but it would be nice in V3 to have the ability to listen to a stream of consent events so that integrators can more accurately display consent.

- [ ] stream consent records

PR #1296, opened 2024-11-20
Stream consent

codabrink:
Consent steaming - in the true sense of the word.

https://github.com/xmtp/libxmtp/issues/1114


nplasterer (approved):
Nice test! 🔥

neekolas:
@codabrink One thing that would be helpful here are some race tests. Our streams don't guarantee ordering, so if you quickly toggle a user from allowed to blocked and back, I'm not sure this streaming setup is going to reliably give you a consistent output.

codabrink:
@neekolas No problem, I can add some in a small follow-up PR.

PR #1300, opened 2024-11-20
Replication Client: Fix Batch message send

mkysel:
In the previous code we would break down a batch message containing N messages to N network calls with 1 message each.

Fix that.


xmtp_api_grpc/src/replication_client.rs
```diff
@@ -481,21 +475,27 @@ impl ClientV4 {
...
+        <T as TryInto<ClientEnvelope>>::Error: std::error::Error + Send + Sync + 'static,
+    {
         let client = &mut self.payer_client.clone();
-        for item in items {
-            let request = item.try_into()?;
-            let res = client.publish_client_envelopes(request).await;
-            if let Err(e) = res {
-                return Err(Error::new(ErrorKind::MlsError).with(e));
-            }
-        }
```
mkysel, line 498: this was the problematic code

richardhuaaa (approved):

Issue #1312, opened 2024-11-21
Update Multi Wallet to just take 1 signature


nplasterer:
Update the way multi wallet add account work to only need 1 signature.

neekolas:
What we would need to do here is sign the signature request with the installation keys inside `identity_updates.rs` in the `associate_wallet` function, and add some tests to verify this all works as expected.

PR #1314, opened 2024-11-21
Simplify Wallet Association By Using Installation Keys

codabrink:
https://github.com/xmtp/libxmtp/issues/1312


xmtp_mls/src/identity_updates.rs
```diff
@@ -285,9 +286,30 @@ where
         let inbox_id = self.inbox_id();
         let builder = SignatureRequestBuilder::new(inbox_id);
 
-        Ok(builder
+        let installation_public_key = self.identity().installation_keys.verifying_key();
+        let mut signature_request = builder
+            .add_association(
+                installation_public_key.as_bytes().to_vec().into(),
```
neekolas, line 292: This looks wrong. You should be adding with the installation public key as the existing member, and the `new_wallet_address` address as the new member

codabrink: Thank you

neekolas (approved):
Nice cleanup. Looks great

nplasterer (approved):

PR #1337, opened 2024-11-25
Cargo deny: bump dependency versions

codabrink:


nplasterer (approved):
I'm assuming this is fine because it's just a .lock change but just want to make sure this doesn't break the android bindings again.

Issue #1316, opened 2024-11-22
Improvement: SyncAll should sync all group messages and welcomes


nplasterer:
syncAllGroups syncs all the groups messages in all groups and syncs all welcome messages

PR #1340, opened 2024-11-26
Sync all conversations syncs all

nplasterer:
Update the sync all conversations to also sync all welcomes. Also adds this to the web and makes sure they are syncing sync groups.

Fixes: https://github.com/xmtp/libxmtp/issues/1316


bindings_ffi/src/mls.rs
```diff
@@ -917,19 +917,30 @@ impl FfiConversations {
...
             groups.len()
         );
 
-        let num_groups_synced: usize = inner.sync_all_groups(groups).await?;
-        // Uniffi does not work with usize, so we need to convert to u32
+        // Call sync operations in parallel
+        let (welcome_result, group_result) = tokio::join!(
+            self.inner_client.sync_welcomes(&conn),
+            inner.sync_all_groups(groups)
+        );
```
richardhuaaa, line 933: If we do these in parallel, is there a chance that we don't sync the messages on a new group?

bindings_ffi/src/mls.rs
```diff
@@ -917,19 +917,30 @@ impl FfiConversations {
 
     pub async fn sync_all_conversations(&self) -> Result<u32, GenericError> {
```
richardhuaaa, line 918: Are there any clients depending on the current behavior? Is it worth making a separate method called `sync_all`, or renaming this one? Don't have a strong opinion

cameronvoell (approved):
nice improvement 👍

bindings_ffi/src/mls.rs
nplasterer: O good call out!

insipx (approved):

PR #1350, opened 2024-11-27
Static association state checking

codabrink:


nplasterer (approved):

xmtp_mls/src/identity_updates.rs
```diff
@@ -528,6 +529,43 @@ async fn verify_updates(
...
+
+    let Some(updates) = updates.remove(inbox_id) else {
+        return Err(ClientError::Generic(
+            "Unable to find provided inbox_id".to_string(),
+        ));
+    };
+    let updates: Vec<_> = updates.into_iter().map(|u| u.update).collect();
+
+    let mut association_state = None;
+    let scw_verifier = RemoteSignatureVerifier::new(api_client.api_client.clone());
```
neekolas, line 555: Would be nice if the caller could specify the verifier, and for this to be the default if they don't.

neekolas (approved):
This looks great. Thanks for getting it together so quickly.

xmtp_mls/src/identity_updates.rs
codabrink: Done

PR #1365, opened 2024-12-03
Start the sync worker in the builder

codabrink:
- Start the sync worker in the builder.
- Have the sync worker wait until the identity is ready before starting.

### DIFF
--- a/xmtp_mls/src/api/test_utils.rs
+++ b/xmtp_mls/src/api/test_utils.rs
@@ -212,3 +212,22 @@ where
     }
     result
 }
+
+pub async fn wait_for_eq<F, Fut, T>(f: F, expected: T)
+where
+    F: Fn() -> Fut,
+    Fut: Future<Output = T>,
+    T: std::fmt::Debug + PartialEq,
+{
+    let start = Instant::now();
+    let mut result = f().await;
+    while start.elapsed() < Duration::from_secs(3) {
+        if result == expected {
+            break;
+        }
+        crate::sleep(Duration::from_millis(100)).await;
+        result = f().await;
+    }
+
+    assert_eq!(expected, result);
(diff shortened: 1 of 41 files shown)


xmtp_mls/src/api/test_utils.rs
```diff
@@ -176,3 +181,34 @@ mod wasm {
         }
     }
 }
+
+pub async fn wait_for_some<F, T>(f: F) -> Option<T>
+where
+    F: Fn() -> Option<T>,
+{
+    let start = Instant::now();
```
insipx, line 189: Try to prefer https://docs.rs/web-time/latest/web_time/ `Instant` fn, so that it works in webassembly too

insipx: #1366 puts all these into a common crate for easier access, but that PR still needs some love to get tests passing

codabrink: Cool, thanks

insipx (approved):

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -113,55 +113,81 @@ impl RetryableError for DeviceSyncError {
...
+        inbox_id = client.inbox_id(),
+        installation_id = hex::encode(client.installation_public_key()),
+        "starting sync worker"
+    );
+    let receiver = client.local_events.subscribe();
+    let sync_stream = receiver.stream_sync_messages();
+
+    // Wait for the identity to be ready before doing anything
+    while !client.identity().is_ready() {
+        crate::sleep(Duration::from_millis(200)).await;
```
insipx, line 163: can also `tokio::yield_now()` to yield back to the executor

PR #1366, opened 2024-12-03
common crate, expose more tests in wasm

insipx:
This PR adds a common crate & some utilities for testing WebAssembly. It mostly moves things around and should not effect functionality other than a few minor bug fixes for tests which previously did not run in web assembly

the bug fixes:
- fixed `is_ready` in register_identity to toggle once the identity signature has been verified with backend, fixing `SequenceId` errors
- fixed some webassembly tests hanging because of `sleep` implementation, replaced with `yield`
- fixed `Notify`/`Delivery` test utilities panicking streams in webassembly environment because of `std::time`, replaced with `xmtp_common::time`

**Common Crate**:
---
- creates common crate that houses:
    - the `rand_xx` fns that were duplicated across a few crates
    - web assembly `time::` shims, that are now accessible via `xmtp_common::time`. It should be a drop-in replacement for `std::time`, and preferable to use instead of `std::time`
    - `xmtp_common::yield_` which delegates to tokio or the browsers executor depending on environment. Should be preferred over `tokio::task::yield`, since a WebAssembly future may not necessarily be running in a tokio executor
    - `retry` moved from `xmtp_mls` to `common`. Retry is general enough that this makes sense, and allows it to be used without including all of `xmtp_mls`.
    - `traced_test` moved from `xmtp_mls` to common. Also general enough to be used w/o mls, and useful elsewhere
    - `xmtp_common` includes a common `logger` initializer toggling on arch, `xmtp_common::logger`
- `assert_x` and `optify` macros moved to `common`
- new test utilities in `store/mod.rs` create triggers to track intent metadata (created/published/deleted) and exposed for tests on `DbConnection`
- replaced some uses of `cfg_attr/cfg_attr` pair, with the more succinct `wasm_bindgen_test(unsupported = tokio::test)` macro


insipx:
* **#1366** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1366?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

xmtp_mls/src/groups/device_sync/consent_sync.rs
```diff
@@ -50,33 +50,41 @@ where
...
+        groups::scoped_client::ScopedGroupClient,
         storage::consent_record::{ConsentState, ConsentType},
         utils::test::wait_for_min_intents,
     };
     use xmtp_cryptography::utils::generate_local_wallet;
     use xmtp_id::InboxOwner;
 
-    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[wasm_bindgen_test(unsupported = tokio::test(flavor = "multi_thread", worker_threads = 1))]
+    #[cfg_attr(target_family = "wasm", ignore)]
```
insipx, line 78: these tests (and all of message history on web) is blocked on https://github.com/xmtp/xmtp-message-history-server/issues/6, then the `ignore` can be removed

xmtp_mls/src/groups/device_sync/consent_sync.rs
```diff
@@ -120,30 +128,25 @@ pub(crate) mod tests {
...
-            if start.elapsed() > Duration::from_secs(3) {
-                panic!("Did not receive sync reply.");
-            }
-        }
+                .unwrap()
+        })
+        .await
+        .unwrap();
 
         // Wait up to 3 seconds for sync to process (typically is almost instant)
```
cameronvoell, line 140: might want to update this comment since `wait_for_eq` looks like it would wait for 60 sec

xmtp_mls/src/storage/mod.rs
```diff
@@ -13,3 +13,95 @@ pub async fn init_sqlite() {
...
 #[cfg(not(target_arch = "wasm32"))]
 pub async fn init_sqlite() {}
+
+#[cfg(any(test, feature = "test-utils"))]
+pub mod test_util {
+    #![allow(clippy::unwrap_used)]
+    use super::*;
+    use diesel::{connection::LoadConnection, deserialize::FromSqlRow, sql_query, RunQueryDsl};
+    impl DbConnection {
+        /// Create a new table and register triggers for tracking column updates
```
cameronvoell, line 23: this is a cool idea, nice 😎

cameronvoell (approved):
awesome problem solving on the browser executor for streaming. and clean up in this pr is fantastic thanks! 🥇

common/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub fn rand_string<const N: usize>() -> String {
+    Alphanumeric.sample_string(&mut crypto_utils::rng(), N)
+}
+
+pub fn rand_array<const N: usize>() -> [u8; N] {
+    let mut buffer = [0u8; N];
+    crypto_utils::rng().fill_bytes(&mut buffer);
+    buffer
+}
```
codabrink, line 35: That's cool. I hadn't thought about using a generic like that before.

common/src/test.rs
```diff
@@ -0,0 +1,183 @@
...
+    .await
+}
+
+pub async fn wait_for_eq<F, Fut, T>(f: F, expected: T) -> Result<(), Expired>
+where
+    F: Fn() -> Fut,
+    Fut: Future<Output = T>,
+    T: std::fmt::Debug + PartialEq,
+{
+    let result = crate::time::timeout(crate::time::Duration::from_secs(60), async {
```
codabrink, line 169: Can we reduce the default timeout to something like 2-3 seconds?

insipx: I can reduce it a bit, but reducing too much causes tests in webassembly to flake out. I figure the common case of test success it will still finish quickly, but we need to give liberal time for wasm. I'll lower it to ~20sec?

PR #1372, opened 2024-12-04
Native Logging with tracing for Android and iOS

insipx:
Replaces `FfiLogger` with native bindings to Android and iOS/Mac `Log` utilities. This removes the need to pass a logger into `xmtpv3` bindings, since it interacts directly with the android/ios libraries.

This also makes it possible to create benchmarks in `bindings_ffi`, and generate flamegraphs profiling the performance (the main reason behind opening this PR now is to profile `create_client`).

 Future work maybe interested in adding a non-blocking [appender](https://docs.rs/tracing-appender/latest/tracing_appender/) layer to write logs to a file and send them somewhere once a fork is detected.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -1,6 +1,5 @@
 pub use crate::inbox_owner::SigningError;
 use crate::logger::init_logger;
-use crate::logger::FfiLogger;
 use crate::{FfiSubscribeError, GenericError};
 use std::{collections::HashMap, convert::TryInto, sync::Arc};
 use tokio::sync::Mutex;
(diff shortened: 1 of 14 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -142,7 +140,6 @@ pub async fn create_client(
 #[allow(unused)]
 #[uniffi::export(async_runtime = "tokio")]
 pub async fn get_inbox_id_for_address(
-    logger: Box<dyn FfiLogger>,
```
nplasterer, line 145: How will this work now for surfacing libxmtp errors in native? https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/libxmtp/XMTPLogger.kt

nplasterer: Ahh sorry just read you description so it should just work now with out this! Nice

nplasterer (approved):

bindings_ffi/src/mls.rs
insipx: Yup just tested with android, we actually get way more info now: 

Screenshot 2024-12-04 at 11 47 46 AM

Issue #422, opened 2024-01-17
Support for is_sender verification for push notifications via HMAC


neekolas:

richardhuaaa:
More info about private preferences: https://community.xmtp.org/t/xip-42-universal-allow-and-block-preferences/544

nplasterer:
We need to make sure the message history server backing this syncs automatically in instances where other installations are online.

PR #1377, opened 2024-12-04
Database migration for HMAC Key Records

nplasterer:
Part of https://github.com/xmtp/libxmtp/issues/422

Adds the database table for hmac key records
Matches the structure currently used for hmac keys in V2 https://github.com/xmtp/proto/blob/main/proto/keystore_api/v1/keystore.proto#L306-L317

### DIFF
--- a/xmtp_mls/migrations/2024-12-04-194513_hmac-keys-records-table/up.sql
+++ b/xmtp_mls/migrations/2024-12-04-194513_hmac-keys-records-table/up.sql
@@ -1,0 +1,11 @@
+CREATE TABLE "hmac_key_records"(
+    -- Group ID that the Hmac keys are associated with
+    "group_id" BLOB NOT NULL,
+    -- Dm ID that the Hmac keys are associated with 
+    "dm_id" TEXT,
+    -- The hmac key
+    "hmac_key" BLOB NOT NULL,
+    -- The number of 30 day periods since epoch
+    "thirty_day_periods_since_epoch" INT NOT NULL,
+    PRIMARY KEY ("group_id", "hmac_key")
+);
--- a/xmtp_mls/src/storage/encrypted_store/hmac_key_record.rs
+++ b/xmtp_mls/src/storage/encrypted_store/hmac_key_record.rs
@@ -1,0 +1,159 @@
+use crate::{impl_store, storage::StorageError};
+
+use super::Sqlite;
+use super::{
+    db_connection::DbConnection,
+    schema::consent_records::{self, dsl},
+};
+use diesel::{
+    backend::Backend,
+    deserialize::{self, FromSql, FromSqlRow},
+    expression::AsExpression,
+    prelude::*,
+    serialize::{self, IsNull, Output, ToSql},
+    sql_types::Integer,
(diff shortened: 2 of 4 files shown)


xmtp_mls/src/storage/encrypted_store/hmac_key_record.rs
```diff
@@ -0,0 +1,159 @@
...
+            thirty_day_periods_since_epoch
+        }
+    }
+}
+
+impl_store!(StoredHmacKeyRecord, hmac_key_records);
+
+impl DbConnection {
+    /// Returns all hmac_key_records for the given group_id
+    pub fn get_hmac_key_records(
```
mchenani, line 53: suggestion: does `impl_fetch_list_with_key!` do the job?

xmtp_mls/src/storage/encrypted_store/hmac_key_record.rs
```diff
@@ -0,0 +1,159 @@
...
+    ) -> Result<Vec<StoredHmacKeyRecord>, StorageError> {
+        Ok(self.raw_query(|conn| -> diesel::QueryResult<_> {
+            dsl::hmac_key_records
+                .filter(dsl::group_id.eq(group_id))
+                .load::<StoredHmacKeyRecord>(conn)
+        })?)
+    }
+
+    /// Insert hmac_key_records without replacing existing ones
+    pub fn insert_hmac_key_records(
```
mchenani, line 65: suggestion: can we use `impl_fetch_list_with_key!`

mchenani:
By mistake update the PR from the main branch, I hope that would be ok!

xmtp_mls/migrations/2024-12-04-194513_hmac-keys-records-table/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE "hmac_key_records"(
+    -- Group ID that the Hmac keys are associated with
+    "group_id" BLOB NOT NULL,
```
nplasterer, line 3: I did this based off of the hmac keys work in V2 but @codabrink mentioned we might be doing this based on inboxId now. So maybe this is irrelevant in V3?

richardhuaaa:
Some high-level thoughts, based on our previous discussion about deriving the hmac key from a user secret rather than a group secret in v3. It might be worth looping @codabrink into this PR too:

1. Right now the schema stores hmac keys for each `thirty_day_periods_since_epoch`. However, I think we need to store the user's root key from which the hmac keys are derived, otherwise we can't derive keys for future epochs.
2. If we are storing the root key, then maybe we don't need to store the derived keys, because we can just derive them when we need them. Then we don't need to worry about storing new records or deleting old records as time progresses.
3. I think there's an opportunity here to generalize this beyond hmacs. In a nutshell, all of a user's installations need to synchronize a root key used to derive hmacs. We could envision using this same system for other things, for example synchronizing a nickname. I'm wondering if what we want is a generic table for user settings/preferences.

nplasterer:
Closing out for now infavor of a more general database long term for these kinds of preferences similar to our consent records DB.

Issue #1380, opened 2024-12-05
HMAC V3


codabrink:
### High level implementation: 

- [x] The HMAC key will be derived from:
    - A random device-synced root secret
    - The message’s respective group’s id
    - The number of 30 day periods since epoch
- [x] The root key will be changed when an installation is revoked.
- [x] The root secret will be stored in a new `user_preferences` table to look something like this:
```sql
CREATE TABLE "user_preferences"(
    -- The latest id is the current preference
    id INTEGER PRIMARY KEY ASC,
    -- HMAC root key
    "hmac_key" BLOB,
    -- Nickname (example column) 
    "nickname" TEXT,
    -- Database export path (example column)
    "default_export_path" TEXT
);
```
- [x] The key will be device-synced via generic proto sync message with a serialized enum as a payload to allow for easy extensibility for syncing new preferences without altering the protos any further.
- [x] This generated KDF secret will be used to sign the encrypted payload, generating 3 signatures (current 30 day epoch, and +/- 1 epoch)
- The notification server will then use these keys to check if it should send a push notification in the same fashion as it did in v2.
- Ideally there will be no changes to the push notification server.
- [x] Add a preference update stream to the FFI so that integrators can know when to resubscribe to push-notifications.

richardhuaaa:
This looks good to me, love how succinct and clear it is. One thing I've been wondering recently is if rotating the hmac root key only on revocation is frequent enough, but would be good to get feedback from @franziskuskiefer.

The other thing that might be good to describe is the algorithm that determines when a key is generated and rotated, and what happens in race conditions (e.g. key is rotated as you're using it, or if two installations try to generate a new key at the same time). These don't need to be handled perfectly, but would be good to enumerate the worst case scenarios.

keks:
On the call it sounded like you were planning to sync the root keys using an MLS group that contains all clients/devices of a user. If that is still the plan, you could also export the root key from the MLS key schedule instead of regularly generating new keys and sending them through the group. That would mean that removing a device from the user's group will immediately produce a new key known to all devices. The mechanism is called "Exporter Secrets" ([RFC], [OpenMLS Docs]). For exporting, you need to provide both a label and a context. You could use something like `"xmtp hmacv3 root"` as the label and use some user id as the context.

[...]

PR #1384, opened 2024-12-05
User Preferences Table

codabrink:
Add the user preferences table for https://github.com/xmtp/libxmtp/issues/1380

### DIFF
--- a/xmtp_mls/src/storage/encrypted_store/schema.rs
+++ b/xmtp_mls/src/storage/encrypted_store/schema.rs
@@ -108,6 +108,13 @@ diesel::table! {
 }
 
 diesel::table! {
+    user_preferences (id) {
+        id -> Nullable<Integer>,
+        hmac_key -> Nullable<Binary>,
+    }
+}
+
+diesel::table! {
     wallet_addresses (wallet_address) {
         inbox_id -> Text,
         wallet_address -> Text,
@@ -129,5 +136,6 @@ diesel::allow_tables_to_appear_in_same_query!(
     openmls_key_store,
     openmls_key_value,
     refresh_state,
+    user_preferences,
     wallet_addresses,
 );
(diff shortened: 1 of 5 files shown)


xmtp_mls/src/storage/encrypted_store/schema.rs
```diff
@@ -107,6 +107,13 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    user_preferences (id) {
+        id -> Nullable<Integer>,
```
mchenani, line 112: Question: is it correct to be nullabe?

xmtp_mls/src/storage/encrypted_store/user_preferences.rs
```diff
@@ -0,0 +1,85 @@
...
+        let mut preferences = Self::load(conn)?;
+        preferences.id = None;
+        preferences.hmac_key = Some(hmac_key);
+
+        preferences.insert(conn)?;
+
+        Ok(())
+    }
+
+    fn insert(&self, conn: &DbConnection) -> Result<(), StorageError> {
```
mchenani, line 37: Suggestion: can't we use existing macros?

xmtp_mls/src/storage/encrypted_store/schema.rs
codabrink: Yes, it's correct. We want the database to provide this field if it's null.

xmtp_mls/src/storage/encrypted_store/user_preferences.rs
codabrink: Yeah, looks like `impl_store!` is a good use here. Thanks for the suggestion. Updated.

xmtp_mls/src/storage/encrypted_store/user_preferences.rs
```diff
@@ -0,0 +1,76 @@
...
+pub struct StoredUserPreferences {
+    /// Primary key - latest key is the "current" preference
+    pub id: Option<i32>,
+    /// Randomly generated hmac key root
+    pub hmac_key: Option<Vec<u8>>,
+}
+impl_store!(StoredUserPreferences, user_preferences);
+
+impl StoredUserPreferences {
+    pub fn load(conn: &DbConnection) -> Result<Self, StorageError> {
```
mchenani, line 21: Suggestion: maybe the fetch macro and the insert macro could be used here, otherwise looks good.

mchenani (approved):

Issue #1144, opened 2024-10-16
Ability to not sync groups that are not consented to in sync all group functionality


nplasterer:
Syncing in batches.
 For example, syncing the 50 most recently messaged consented groups first, and in parallel or afterwards syncing the rest?

nplasterer:
I'll add consent filtering to the groups list as step 1.

PR #1385, opened 2024-12-05
Add the ability to sync based on consent state

nplasterer:
Closes https://github.com/xmtp/libxmtp/issues/1144

### DIFF
--- a/bindings_node/src/conversations.rs
+++ b/bindings_node/src/conversations.rs
@@ -258,7 +258,7 @@ impl Conversations {
 
     let num_groups_synced = self
       .inner_client
-      .sync_all_welcomes_and_groups(&conn)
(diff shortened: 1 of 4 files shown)


bindings_node/src/conversations.rs
```diff
@@ -258,7 +258,7 @@ impl Conversations {
 
     let num_groups_synced = self
       .inner_client
-      .sync_all_welcomes_and_groups(&conn)
+      .sync_all_welcomes_and_groups(&conn, None)
```
nplasterer, line 261: I think a follow up PR needs to be done not only bring the consent filtering to the sync but also to the list functions on the node and browser bindings

codabrink (approved):

PR #1394, opened 2024-12-09
Provide HMAC Keys

codabrink:
This provides the HMAC keys while uploading messages, and also provides the last hmac, current hmac, and next hmac on the FFI.
Issue: https://github.com/xmtp/libxmtp/issues/1380


xmtp_mls/src/constants.rs
```diff
@@ -0,0 +1 @@
+pub(crate) const HMAC_SALT: &[u8] = b"libXMTP HKDF salt!";
```
insipx, line 1: we have a place for constants in `configuration.rs` github.com/xmtp/libxmtp/blob/main/xmtp_mls/src/configuration.rs

xmtp_mls/src/storage/encrypted_store/user_preferences.rs
```diff
@@ -5,30 +5,53 @@ use super::{
...
         let query = dsl::user_preferences.order(dsl::id.desc()).limit(1);
         let mut result = conn.raw_query(|conn| query.load::<StoredUserPreferences>(conn))?;
 
-        Ok(result.pop().unwrap_or_default())
+        let result = match result.pop() {
+            Some(result) => result,
+            None => {
+                // Create a default and store it.
+                let result = Self::default();
+                // TODO: emit an hmac key update event here.
```
insipx, line 41: do we need this in this PR or is it OK to exclude for now?/ is it for streams

insipx (approved):
Looks good other than a couple nits!

xmtp_mls/src/storage/encrypted_store/user_preferences.rs
codabrink: It's the next step. I have a TODO just to mark it to make it easy in the next PR :)

xmtp_mls/src/constants.rs
codabrink: Ahh, cool thanks. I'll move it there!

PR #1400, opened 2024-12-10
Sync All Preference Updates Across Devices

codabrink:
Issue: https://github.com/xmtp/libxmtp/issues/1380
- Harden consent sync ordering by relying on MLS for ordering
- Sync user preferences (includes HMAC keys) across all devices
- Add worker handle so we can be notified when certain jobs complete


bindings_wasm/Cargo.toml
```diff
@@ -7,25 +7,24 @@ version.workspace = true
...
-xmtp_common.workspace = true
 xmtp_proto = { path = "../xmtp_proto", features = ["proto_full"] }
-tracing-web = "0.1"
-tracing.workspace = true
-tracing-subscriber = { workspace = true, features = ["env-filter", "json"] }
-console_error_panic_hook.workspace = true
 
 [dev-dependencies]
 wasm-bindgen-test.workspace = true
-xmtp_mls = { path = "../xmtp_mls", features = ["test-utils", "http-api"] }
```
codabrink, line 31: Was redundant.

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -404,10 +469,10 @@ where
 
         let _message_id = sync_group.prepare_message(&content_bytes, provider, {
             let request = request.clone();
-            move |_time_ns| PlaintextEnvelope {
+            move |now| PlaintextEnvelope {
                 content: Some(Content::V2(V2 {
                     message_type: Some(MessageType::DeviceSyncRequest(request)),
-                    idempotency_key: new_request_id(),
+                    idempotency_key: now.to_string(),
```
insipx, line 475: what's the implication of using `now` vs `request_id`?

xmtp_mls/src/storage/encrypted_store/user_preferences.rs
```diff
@@ -1,67 +1,93 @@
...
         let mut preferences = Self::load(conn)?;
-        // Have the id increment
-        preferences.id = None;
-        preferences.hmac_key = hmac_key;
 
-        preferences.store(conn)?;
-
-        Ok(())
+        let mut hmac_key = vec![0; 32];
+        OsRng.fill_bytes(&mut hmac_key);
```
insipx, line 65: now can use `let hmac_key = xmtp_common::rand_vec::<32>()`

You'll just have to move it from `test.rs` to `lib.rs`, shouldn't need to refactor since test is re-exported at the root

insipx (approved):
nice!

xmtp_mls/src/groups/device_sync.rs
codabrink: It's just for idempotency. I switched to it to conform with how the main group send function.

Issue #1359, opened 2024-12-02
Tests: Write tests showing inbox id login cases after wallet is associated


cameronvoell:
### Is your feature request related to a problem?

Missing tests for the following use cases:

1. once a wallet has been associated with an inbox id, the added wallet can then be used to login to that inbox id on new devices 

2. if wallet B previously was used to create  and log in to inbox B, but then wallet B is added as an associated wallet to inbox A, then wallet B will then fail to be able to login to inbox B afterwards.

PR #1406, opened 2024-12-11
Adds tests showing authn abilities lost after wallet is associated to another inbox_id

cameronvoell:
Closes https://github.com/xmtp/libxmtp/issues/1359

1. Once a wallet/address has been associated to an existing inbox_id, it will fail when trying to create a client with a new inbox_id derived from its own address.
2. For two wallets A and B that created identities with inbox_id A and B, if we successfully add association from wallet B to inbox_id A, then wallet B will no longer be able to create a client for inbox_id B.


insipx (approved):
nice, these tests are useful for making sure our assumptions hold now & in the future

PR #1412, opened 2024-12-13
Docker build for xdbg

insipx:
`xdbg` docker build + binary build to github releases with nix + workflow file
`validation-service` just b/c it was exactly the same as xdbg


insipx:
closed in favor of #1419

PR #1413, opened 2024-12-13
Create a lock struct for the mls groups

codabrink:


xmtp_mls/src/groups/serial.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn lock_blocking<'a>(&'a mut self) -> SerialOpenMlsGroup<'a> {
+        // .clone() is important here so that the outer lock gets dropped
+        let mutex = MLS_COMMIT_LOCK
+            .lock()
+            .entry(self.group_id().to_vec())
+            .or_default()
+            .clone();
+
+        // this may block
+        let lock = mutex.blocking_lock();
```
insipx, line 72: Do we ever use this variant?

I'm surprised the webassembly tests pass, under the hood this uses `futures::block_on` which will panic the WASM executor. I suspect we maybe don't have any webassembly tests that use code touching the lock

xmtp_mls/src/groups/serial.rs
```diff
@@ -0,0 +1,84 @@
...
+        let lock = unsafe {
+            // let the borrow checker know that this guard's mutex is going to be owned by the struct it's returning
+            std::mem::transmute::<MutexGuard<'_, ()>, MutexGuard<'a, ()>>(lock)
```
insipx, line 75: how does this avoid a dangling pointer?

The Mutex creates a reference with `blocking_lock` that is freed at the end of the function, but we reinterpret the reference to be longer than the function.

insipx: AFAIU This would work if `Mutex::blocking_lock` function was annotated with lifetimes like:

```
fn blocking_lock<'a>(&'a self) -> MutexGuard<'a, ()> { }
```

however, the function signature [is different](docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.blocking_lock)

codabrink: The returned struct owns an Arc that owns what the pointer is pointing to.

xmtp_mls/src/groups/serial.rs
codabrink: We do not, it's just there in case it's used in the future.

codabrink:
Going to close my PR in favor of @cameronvoell's PR which seems to fix the root issue https://github.com/xmtp/libxmtp/pull/1416

PR #1416, opened 2024-12-14
Do not allow overriding an intent that is already published

cameronvoell:
Before this change parallel executing threads could calculate two different `payload_hash` for the same `intent id` and the second executing thread could override the first's intent in the DB even though it was already published to the network.

https://github.com/xmtp/libxmtp/blob/4336127769cbb7b5dc13595a465624fa7b123541/xmtp_mls/src/groups/mls_sync.rs#L980-L1025

This was causing an intent that was already on the network to be overwritten locally, causing the incoming message to be interpreted as an external message, which would cause the sender of the intent to be forked from the rest of the group, this was shows in tests on the following React Native PRs:

- https://github.com/xmtp/xmtp-react-native/pull/534
- https://github.com/xmtp/xmtp-react-native/pull/538

After this change the parallel executing commit messages in React Native do not cause the forked state.

### DIFF
--- a/xmtp_mls/src/storage/encrypted_store/group_intent.rs
+++ b/xmtp_mls/src/storage/encrypted_store/group_intent.rs
@@ -197,16 +197,12 @@ impl DbConnection {
         staged_commit: Option<Vec<u8>>,
         published_in_epoch: i64,
     ) -> Result<(), StorageError> {
-        let payload_hash_clone = payload_hash.clone();
         let res = self.raw_query(|conn| {
             diesel::update(dsl::group_intents)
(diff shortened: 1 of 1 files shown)


insipx:
iirc, the change from just `ToPublish` to `ToPublish` and `Published` was because of a different race condition, but I don't remember exactly if it was a real issue, or something just transient [PR](https://github.com/xmtp/libxmtp/pull/923). It seems like maybe a non-issue now

codabrink (approved):
v nice!

insipx (approved):

PR #1426, opened 2024-12-16
MLS Group Locking

codabrink:


xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -368,17 +369,18 @@ where
             id: ref msg_id,
             ..
         } = *envelope;
+        let mut locked_openmls_group = openmls_group.lock().await;
```
mchenani, line 372: kinda we tried this idea in the beginning to see if that fixes the issue, but only as a POC.
the downside here is that other developers easily can forget to lock the group!

xmtp_mls/src/groups/serial.rs
```diff
@@ -0,0 +1,73 @@
+use openmls::prelude::MlsGroup as OpenMlsGroup;
```
mchenani, line 1: This file looks good, will try it later to adapt the changes into my PR and see what would happen!

xmtp_mls/src/groups/mls_sync.rs
codabrink: > the downside here is that other developers easily can forget to lock the group!

Not sure what you mean. That would be the case with any implementation? You have to remember to lock the group no matter how you implement it. The upside of this approach is you can have functions require the `SerialOpenMlsGroup` as a param where locking is critical, so it's actually the opposite that would be true.

Issue #1403, opened 2024-12-11
Tracking - Queryable Content Types in Rust


cameronvoell:
### Is your feature request related to a problem?

- [ ] Move Content Types to new crate
- [ ] Start saving content type to DB
- [ ] Enable Readable only flag on messages FFI functions
- [ ] Add new Reaction Content Type with Reaction proto
- [ ] Add Legacy JSON deserialization for Reaction Content Type
- [ ] Extract queryable content field "parent_msg_id" from message and save to new db field parent id.
- [ ] Add query for returning "child_messages" when returning messages from a group
- [ ] Add messagesWithMetadata funciton to bindings_ffi configuring which metadata you want to return (initial option is reactions)

- [ ] Add Replies support for messagesWithMetadata function
- [ ] Add Read Receipt support for messagesWithMetadata function
- [ ] Move Remote Attachment Download and Decrypt logic to rust

PR #1435, opened 2024-12-19
Save group message content types

cameronvoell:
Part of https://github.com/xmtp/libxmtp/issues/1403

Adds fields to `group_messages` table corresponding to `ContentTypeId` proto and saves them when sending and receiving messages.

`ContentTypeId` for reference: 

https://github.com/xmtp/proto/blob/404a0f41a6dc00f5de5fcfc24856c8b4e417fe59/proto/mls/message_contents/content.proto#L10-L16

```proto
// ContentTypeId is used to identify the type of content stored in a Message.
message ContentTypeId {
  string authority_id = 1;  // authority governing this content type
  string type_id = 2;       // type identifier
  uint32 version_major = 3; // major version of the type
  uint32 version_minor = 4; // minor version of the type
}
```

### DIFF
--- a/xmtp_content_types/src/lib.rs
+++ b/xmtp_content_types/src/lib.rs
@@ -1,3 +1,4 @@
+pub mod attachment;
 pub mod group_updated;
 pub mod membership_change;
 pub mod reaction;
--- a/xmtp_content_types/src/lib.rs
+++ b/xmtp_content_types/src/lib.rs
@@ -2,75 +2,8 @@ pub mod group_updated;
 pub mod membership_change;
 pub mod text;
 
-use diesel::{
-    backend::Backend,
-    deserialize::{self, FromSql, FromSqlRow},
-    expression::AsExpression,
-    serialize::{self, IsNull, Output, ToSql},
-    sql_types::Integer,
-    sqlite::Sqlite,
-};
-use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use xmtp_proto::xmtp::mls::message_contents::{ContentTypeId, EncodedContent};
-
-/// ContentType and their corresponding string representation
-/// are derived from the `ContentTypeId` enum in the xmtp-proto crate
-/// that each content type in this crate establishes for itself
-#[repr(i32)]
-#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, FromSqlRow, AsExpression)]
-#[diesel(sql_type = diesel::sql_types::Integer)]
-pub enum ContentType {
-    Unknown = 0,
-    Text = 1,
-    GroupMembershipChange = 2,
-    GroupUpdated = 3,
-}
-
-impl ContentType {
-    pub fn from_string(type_id: &str) -> Self {
-        match type_id {
-            text::TextCodec::TYPE_ID => Self::Text,
-            membership_change::GroupMembershipChangeCodec::TYPE_ID => Self::GroupMembershipChange,
-            group_updated::GroupUpdatedCodec::TYPE_ID => Self::GroupUpdated,
-            _ => Self::Unknown,
-        }
-    }
-
-    pub fn to_string(&self) -> &'static str {
-        match self {
-            Self::Unknown => "unknown",
-            Self::Text => text::TextCodec::TYPE_ID,
-            Self::GroupMembershipChange => membership_change::GroupMembershipChangeCodec::TYPE_ID,
-            Self::GroupUpdated => group_updated::GroupUpdatedCodec::TYPE_ID,
-        }
-    }
-}
-
-impl ToSql<Integer, Sqlite> for ContentType
-where
-    i32: ToSql<Integer, Sqlite>,
-{
-    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Sqlite>) -> serialize::Result {
-        out.set_value(*self as i32);
-        Ok(IsNull::No)
-    }
-}
-
-impl FromSql<Integer, Sqlite> for ContentType
-where
-    i32: FromSql<Integer, Sqlite>,
-{
-    fn from_sql(bytes: <Sqlite as Backend>::RawValue<'_>) -> deserialize::Result<Self> {
-        match i32::from_sql(bytes)? {
-            0 => Ok(ContentType::Unknown),
-            1 => Ok(ContentType::Text),
-            2 => Ok(ContentType::GroupMembershipChange),
-            3 => Ok(ContentType::GroupUpdated),
-            x => Err(format!("Unrecognized variant {}", x).into()),
-        }
-    }
-}
 
 #[derive(Debug, Error)]
 pub enum CodecError {
--- a/xmtp_content_types/src/lib.rs
+++ b/xmtp_content_types/src/lib.rs
@@ -1,9 +1,9 @@
 pub mod group_updated;
 pub mod membership_change;
 pub mod reaction;
-pub mod reply;
 pub mod read_receipt;
 pub mod remote_attachment;
+pub mod reply;
 pub mod text;
 pub mod transaction_reference;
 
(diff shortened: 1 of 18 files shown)


cameronvoell:
* **#1435** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1435?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

cameronvoell:
ContentTypeId for reference: 

https://github.com/xmtp/proto/blob/404a0f41a6dc00f5de5fcfc24856c8b4e417fe59/proto/mls/message_contents/content.proto#L10-L16

xmtp_content_types/src/lib.rs
```diff
@@ -2,13 +2,74 @@ pub mod group_updated;
...
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use xmtp_proto::xmtp::mls::message_contents::{ContentTypeId, EncodedContent};
 
+/// ContentType and their corresponding string representation
+/// are derived from the `ContentTypeId` enum in the xmtp-proto crate
+/// that each content type in this crate establishes for itself
+#[repr(i32)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, FromSqlRow, AsExpression)]
+#[diesel(sql_type = diesel::sql_types::Integer)]
```
cameronvoell, line 22: Seems a bit off to have the DB specific code in the `xmtp_content_types` crate, but the alternative would be to have a replica of this enum in `storage/encrypted_store/group_message` sort of like this example of `GroupMessageKind`

https://github.com/xmtp/libxmtp/blob/eccce0610585310713951a995200bb6e8dfd8bf6/xmtp_mls/src/storage/encrypted_store/group_message.rs#L49-L55

The benefit of how I have it organized now is that when you add a new content type, you do not have to update the same enum in two different places. But I'm open to feedback here.

cameronvoell: moved to storage/encrypted_store/group_message to simplify

xmtp_mls/src/groups/mod.rs
```diff
@@ -309,6 +309,14 @@ pub enum UpdateAdminListType {
     RemoveSuper,
 }
 
+/// Fields extracted from content of a message that should be stored in the DB
+pub struct QueryableContentFields {
```
insipx, line 313: can impl [Default](https://doc.rust-lang.org/std/default/trait.Default.html) here

xmtp_mls/src/groups/mod.rs
```diff
@@ -706,6 +714,36 @@ impl<ScopedClient: ScopedGroupClient> MlsGroup<ScopedClient> {
...
+    fn extract_queryable_content_fields(message: &[u8]) -> QueryableContentFields {
+        let default = QueryableContentFields {
+            content_type: ContentType::Unknown,
+            version_major: 0,
+            version_minor: 0,
+            authority_id: "unknown".to_string(),
+        };
+
+        // Return early with default if decoding fails or type is missing
+        let content_type_id = match EncodedContent::decode(message)
```
insipx, line 727: Using `Default::default` this can probably be cleaner


you can use [Result::inspect_err](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err) to throw the log message.

then use [Result::unwrap_or_default](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_default)

then `EncodedContent::decode(message).map(|c| c.type_id).unwrap_or_default()` with the `Default` trait.

xmtp_mls/src/groups/mod.rs
```diff
@@ -706,6 +714,36 @@ impl<ScopedClient: ScopedGroupClient> MlsGroup<ScopedClient> {
...
+            .map_err(|e| tracing::debug!("Failed to decode message as EncodedContent: {}", e))
+            .ok()
+            .and_then(|content| content.r#type)
+        {
+            Some(type_id) => type_id,
+            None => {
+                tracing::debug!("Message content type is missing");
```
insipx, line 734: This matching on Some/None is redundant, since its just the transformed error we're already throwing a log on

xmtp_mls/src/storage/encrypted_store/group_message.rs
```diff
@@ -77,6 +86,61 @@ where
...
+    pub fn from_string(type_id: &str) -> Self {
+        match type_id {
+            text::TextCodec::TYPE_ID => Self::Text,
+            membership_change::GroupMembershipChangeCodec::TYPE_ID => Self::GroupMembershipChange,
+            group_updated::GroupUpdatedCodec::TYPE_ID => Self::GroupUpdated,
+            _ => Self::Unknown,
+        }
+    }
+
+    pub fn to_string(&self) -> &'static str {
```
insipx, line 109: It would be more rustlike to `impl std::fmt::Display` here, which provides a `to_string`. 

implementing display allows it to be easily `printed` and logged

xmtp_mls/src/storage/encrypted_store/group_message.rs
```diff
@@ -77,6 +86,61 @@ where
...
+#[diesel(sql_type = diesel::sql_types::Integer)]
+pub enum ContentType {
+    Unknown = 0,
+    Text = 1,
+    GroupMembershipChange = 2,
+    GroupUpdated = 3,
+}
+
+impl ContentType {
+    pub fn from_string(type_id: &str) -> Self {
```
insipx, line 100: Should always prefer `impl From<String>` rather than hand-rolled, so that generic `Foo: From<String>` types work

insipx (changes requested):

xmtp_mls/src/groups/mod.rs
```diff
@@ -706,6 +714,36 @@ impl<ScopedClient: ScopedGroupClient> MlsGroup<ScopedClient> {
...
+            authority_id: "unknown".to_string(),
+        };
+
+        // Return early with default if decoding fails or type is missing
+        let content_type_id = match EncodedContent::decode(message)
+            .map_err(|e| tracing::debug!("Failed to decode message as EncodedContent: {}", e))
+            .ok()
+            .and_then(|content| content.r#type)
+        {
+            Some(type_id) => type_id,
```
insipx, line 732: Any reason we can't `impl TryFrom<EncodedContent> for QueryableContentFields`? Because then you could 

```rust
EncodedContent::decode(msg)
    .inspect_err(|e| tracing::debug!(e))
    .and_then(|content| content.r#type)
    .map(QueryableContentFields::try_from)
    .unwrap_or_default()
```

which is more readable

insipx (changes requested):
updated review

xmtp_mls/src/groups/mod.rs
cameronvoell: added here https://github.com/xmtp/libxmtp/pull/1435/commits/80c7673119b6dcdeef4e4836620b980da934c300

xmtp_mls/src/storage/encrypted_store/group_message.rs
cameronvoell: 🙏  added https://github.com/xmtp/libxmtp/pull/1435/commits/dc6428cd4633aab9f49ff2408921eed1461b714d

xmtp_mls/src/storage/encrypted_store/group_message.rs
cameronvoell: added https://github.com/xmtp/libxmtp/pull/1435/commits/dc6428cd4633aab9f49ff2408921eed1461b714d

xmtp_mls/src/groups/mod.rs
cameronvoell: thanks this was really helpful. updated here with some more tips from clippy to make it even more concise 💯 
https://github.com/xmtp/libxmtp/pull/1435/commits/fbed7ce02c68dbcb639e5d3cee55e35871299fd1

insipx (approved):

Issue #1181, opened 2024-10-23
Join table on group and groupMessages for sorting by last message


nplasterer:
I do this in the sdks but it would be way more performant if I could do it in libxmtp and also allow limit to actually work correctly.

neekolas:
Great idea on the join table.

What's the problem with `limit`?

nplasterer:
if you ask for limit 1 of the list of groups and order by last message. Since ordering by last message currently happens on the sdk side and limit happens in the database you will get the first 10 items that were ordered by created at and then order them by last message. When in reality you want the last 10 items that are ordered by last message. Does that make sense?

PR #1437, opened 2024-12-19
feat(conversations): create conversation list with last message

mchenani:
**Summary:**
Added a SQL view to represent the list of client conversations, including the last message in each conversation.
- If the last message is an application message, its timestamp is used.
- Otherwise, the group creation time determines the order.

**Context:**
Previously, all conversations were loaded, and the last message for each conversation was fetched individually, resulting in inefficiencies.

**Solution:**
Introduced a SQL-level View to streamline fetching the conversation list, with each row including the latest message, improving performance and clarity.


xmtp_mls/src/storage/encrypted_store/group_message.rs
```diff
@@ -290,7 +290,7 @@ pub(crate) mod tests {
     use wasm_bindgen_test::wasm_bindgen_test;
     use xmtp_common::{assert_err, assert_ok, rand_time, rand_vec};
 
-    fn generate_message(
+    pub fn generate_message(
```
codabrink, line 293:
```suggestion
    pub(crate) fn generate_message(
```

bindings_ffi/src/mls.rs
```diff
@@ -1141,6 +1161,12 @@ pub struct FfiConversation {
     inner: MlsGroup<RustXmtpClient>,
 }
 
+#[derive(uniffi::Object)]
+pub struct FfiConversationListItem {
```
mchenani, line 1165: introduced this struct to avoid changing all the current functions!

bindings_ffi/src/mls.rs
```diff
@@ -965,6 +965,26 @@ impl FfiConversations {
         Ok(convo_list)
     }
 
+    pub async fn list_conversations(
```
mchenani, line 968: default order is:
`ORDER BY COALESCE(rm.sent_at_ns, g.created_at_ns) DESC`

insipx (approved):
nice!

mchenani:
closes #1181

mchenani:
resolves #1181

bindings_ffi/src/mls.rs
```diff
@@ -947,6 +947,26 @@ impl FfiConversations {
...
```
nplasterer, line 950: I'm not sure I understand when this is supposed to be used instead of the `list()` function above?

mchenani: What I saw in android, we fetched the conversations and the get the last message one by one and map them together. 
we can replace it with this.

PR #1450, opened 2025-01-02
Provide hmac keys on the napi

codabrink:


bindings_node/src/conversations.rs
```diff
@@ -97,6 +98,21 @@ impl From<ListConversationsOptions> for GroupQueryArgs {
   }
 }
 
+#[napi(object)]
+pub struct NodeHMACKey {
```
rygine, line 102: can you make this just `HmacKey` and alias the import as `XmtpHmacKey`? it makes the generated types usable without having to rename them in the SDK.

codabrink: Done

bindings_node/src/conversations.rs
```diff
@@ -97,6 +98,21 @@ impl From<ListConversationsOptions> for GroupQueryArgs {
...
+  pub key: Vec<u8>,
+  pub epoch: i64,
```
rygine, line 104: `i64` will be typed as a number in javascript. not sure if that's desirable or an issue, just wanted to point it out. while the rust `i64` type supports a range of `−2^63 to 2^63 − 1`, the javascript type only supports `-2^53 + 1 to 2^53 - 1`.

there's a `BigInt` type in NAPI that should provide full precision, if necessary.

codabrink: Oh, the number will never get that large, but I can change it over for correctness. Thanks for letting me know.

rygine: cool, yea, it's not a big deal to keep as is. we can always change in future if it becomes an issue.

codabrink: All good. Swapped it out for BigInt.

rygine (approved):

PR #1467, opened 2025-01-03
Add ability to revoke specific installationIds

nplasterer:
Adds the ability to revoke specific installation ids so we can pass certain installations that may be malformed.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -544,7 +544,7 @@ impl FfiXmtpClient {
     ) -> Result<Arc<FfiSignatureRequest>, GenericError> {
         let installation_id = self.inner_client.installation_public_key();
 
-        if installation_ids.iter().any(|id| id == &installation_id) {
+        if installation_ids.iter().any(|id| id == installation_id) {
             return Err(GenericError::Generic {
                 err: "Cannot revoke the current installation ID.".to_string(),
             });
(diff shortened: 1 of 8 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -534,6 +534,32 @@ impl FfiXmtpClient {
...
+        &self,
+        installation_ids: Vec<Vec<u8>>,
+    ) -> Result<Arc<FfiSignatureRequest>, GenericError> {
+        let installation_id = self.inner_client.installation_public_key();
+
+        if installation_ids.iter().any(|id| id == &installation_id) {
+            return Err(GenericError::Generic {
+                err: "Cannot revoke the current installation ID.".to_string(),
+            });
+        }
```
nplasterer, line 551: @neekolas any preference here should we error if they try to revoke the installation they are currently logged into or should we let users revoke that installation if they want to?

neekolas: I think there should be some way to revoke your current installation, even if it likely will break things in your current app until you create a new client.

neekolas: At the very least, I don't think we should block it at the protocol level even if it's a bad idea in most cases.

bindings_node/src/signatures.rs
```diff
@@ -112,6 +112,33 @@ impl Client {
     Ok(signature_text)
   }
 
+  #[napi]
+  pub async fn revoke_installations_signature_text(
```
neekolas, line 116: I can't remember why we needed a separate function for the signature text instead of using the `SignatureRequest` returned from `revoke_installations`. I'm sure there was a good reason.

neekolas (approved):

PR #1470, opened 2025-01-06
feat(conversations): add filter args to conversation list with last message

mchenani:
Summary:
In [PR #1437](https://github.com/xmtp/libxmtp/pull/1437), we added a new function to fetch conversations along with their last message. This replaces the old functions with the new query.
Also updated the list_conversations function to support filters.
https://github.com/xmtp/libxmtp/issues/1181


xmtp_mls/src/storage/encrypted_store/conversation_list.rs
```diff
@@ -53,17 +57,112 @@ pub struct ConversationListItem {
...
+            include_sync_groups,
+            include_duplicate_dms,
+        } = args.as_ref();
+        let mut query = conversation_list
             .select(conversation_list::all_columns())
+            .filter(conversation_list_dsl::conversation_type.ne(ConversationType::Sync))
             .into_boxed();
-        Ok(self.raw_query(|conn| query.load::<ConversationListItem>(conn))?)
+
+        //question: do we need to do this always? if yes, we can put it in the main view query instead of here!
```
codabrink, line 82: Do you mean is this flag always true?

mchenani: So, what you're saying I can keep it like this, right?

codabrink: If you want to always filter out duplicate DMs. This section of code will filter out duplicates.

mchenani: I will keep it like these to give the control to the sdk and the integrators.

nplasterer: So just for clarity by default the duplicate dms are filtered out correct? Or will it be required to do it on my end?

nplasterer (approved):

PR #1472, opened 2025-01-06
Expose conversation and last message on list

nplasterer:
Add the ability to get the conversation and last message on the list items.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -958,16 +958,18 @@ impl FfiConversations {
     pub async fn list(
         &self,
         opts: FfiListConversationsOptions,
-    ) -> Result<Vec<FfiConversationListItem>, GenericError> {
+    ) -> Result<Vec<Arc<FfiConversationListItem>>, GenericError> {
         let inner = self.inner_client.as_ref();
-        let convo_list: Vec<FfiConversationListItem> = inner
+        let convo_list: Vec<Arc<FfiConversationListItem>> = inner
             .list_conversations(opts.into())?
             .into_iter()
-            .map(|conversation_item| FfiConversationListItem {
-                conversation: conversation_item.group.into(),
-                last_message: conversation_item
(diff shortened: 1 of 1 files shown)


codabrink (approved):

bindings_ffi/src/mls.rs
```diff
@@ -1185,18 +1185,28 @@ impl TryFrom<UserPreferenceUpdate> for FfiPreferenceUpdate {
...
 
-#[derive(uniffi::Object)]
+#[derive(uniffi::Object, Clone)]
 pub struct FfiConversation {
     inner: MlsGroup<RustXmtpClient>,
 }
 
 #[derive(uniffi::Object)]
-#[allow(unused_variables, dead_code)]
+#[allow(dead_code)]
```
mchenani, line 1194: if we do this, won't be enough?
```suggestion
#[uniffi::export]
#[allow(dead_code)]
```

insipx: do we still need the dead_code now that we have `pub fn` to the variables?

mchenani: retracted!

bindings_ffi/src/mls.rs
```diff
@@ -1185,18 +1185,29 @@ impl TryFrom<UserPreferenceUpdate> for FfiPreferenceUpdate {
...
```
insipx, line 1194: do we still need these `dead_code` attributes if we make the functions in the `impl` `pub`?

insipx (changes requested):

bindings_ffi/src/mls.rs
nplasterer: It says you cannot export a struct

bindings_ffi/src/mls.rs
nplasterer: Nice! Excited to get rid of that. 🔥

PR #1473, opened 2025-01-06
Expose installation ID bytes when calling `inboxState` (node + WASM bindings)

rygine:
# Summary

- Added `bytes` field to installation data when calling `inboxState` on clients

Since encoded installation ID strings can't be reversed back into bytes, we need a way to expose installation bytes so that we can use them when revoking specific installations.

### DIFF
--- a/bindings_node/src/conversation.rs
+++ b/bindings_node/src/conversation.rs
@@ -5,7 +5,6 @@ use napi::{
   threadsafe_function::{ErrorStrategy, ThreadsafeFunction, ThreadsafeFunctionCallMode},
   JsFunction,
 };
-use xmtp_cryptography::signature::ed25519_public_key_to_address;
 use xmtp_mls::{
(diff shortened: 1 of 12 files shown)


insipx (approved):

bindings_node/src/conversation.rs
```diff
@@ -231,7 +230,7 @@ impl Conversation {
         installation_ids: member
           .installation_ids
           .into_iter()
-          .map(|id| ed25519_public_key_to_address(id.as_slice()))
```
richardhuaaa, line 234: Double-checking, the ID here is bytes and not string right?

richardhuaaa (approved):

bindings_node/src/conversation.rs
rygine: yep, should be

Issue #1449, opened 2025-01-02
Message retention policy (disappearing messages)


nplasterer:
We need a way to allow admins to set an expiration for messages in a group chat. Once that expiration is met messages should be deleted from all participants local database.

- [x] Add new group metadata field for message expiration that takes an enum with different predefined intervals daily, weekly, monthly, 3 months, 6 months, yearly.
- [x] Add a new permission policy for setting message expiration.
- [x]  Add a job that checks the message expiration and runs on a interval to clean up messages from the local database.

saulmc:
Agreed. The idiomatic approach here is admin-only for groups, both members for 1:1.

nplasterer:
@saulmc how do you feel about these predefined options or would you like more or less options?

- None
- 12 hrs
- 24 hrs
- 7 days
- 30 days
- 90 days
- 1 year

covrter:
WhatsApp's default behavior is that all members can update the Disappearing time. [Link](https://faq.whatsapp.com/673193694148537)

> Anyone can turn disappearing messages on or off in group chats by default. But, group admins can change group settings to allow only admins to do this. Learn how to turn disappearing messages on or off in a group in this [help article](https://faq.whatsapp.com/515705483748188?helpref=faq_content).

So I think, if possible, we'd want to build in flexibility for who can manage retention — in a small group of 3 or 10 friends, "everyone" of course makes sense; in a huge channel … less so.

covrter:
> @saulmc how do you feel about these predefined options or would you like more or less options?
> 
> * None
> * 12 hrs
> * 24 hrs
> * 7 days
> * 30 days
> * 90 days
> * 1 year

I'd suggest ...
- adding some Tiny windows to support the "realtime secrets" use case: 10sec and 1min.
- 8h rather than 12h (just for more difference btwn 12/24)

FWIW I think of like 3 broad use cases here, altho of course they're squishy:
1. "message retention" — long window (30d+)
2. "a chat where you can speak your mind" — medium window (8h, 24h, 7d)
3. "realtime secret/sensitive chat" where "you had to be there" and messages regularly during the convo — tiny retention windows (minutes or seconds)

nplasterer:
@covrter I saw in some mocks that you had a box for custom retention times. In that case I wonder if the database field should be a Long. 🤔 and the enum should just be something we use programmatically. So that we can in the future extend this with custom content types cc @mchenani

mchenani:
[...]

PR #1477, opened 2025-01-07
feat(group): add message expiration to group metadata

mchenani:
This pull request adds support for message expiration policies in group settings. The key updates include:
	•	A new message_expiration_ms field added to group creation options: holds how long after the message was received  gets deleted
	•	A new message_expiration_from_ms field added to group creation options: holds from when the messages get deleted
	•	Permission updates to allow managing message expiration settings.

#1449

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -1727,9 +1727,8 @@ impl FfiConversation {
     }
 
     pub fn update_consent_state(&self, state: FfiConsentState) -> Result<(), GenericError> {
-        let provider = self.inner.mls_provider()?;
         self.inner
-            .update_consent_state(&provider, state.into())
+            .update_consent_state(state.into())
             .map_err(Into::into)
     }
 
--- a/bindings_node/src/conversation.rs
+++ b/bindings_node/src/conversation.rs
@@ -642,10 +642,9 @@ impl Conversation {
       self.group_id.clone(),
       self.created_at_ns,
     );
-    let provider = group.mls_provider().map_err(ErrorWrapper::from)?;
 
     group
-      .update_consent_state(&provider, state.into())
+      .update_consent_state(state.into())
       .map_err(ErrorWrapper::from)?;
 
     Ok(())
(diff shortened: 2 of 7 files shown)


xmtp_mls/src/configuration.rs
```diff
@@ -55,6 +55,7 @@ pub const DEFAULT_GROUP_NAME: &str = "";
 pub const DEFAULT_GROUP_DESCRIPTION: &str = "";
 pub const DEFAULT_GROUP_IMAGE_URL_SQUARE: &str = "";
 pub const DEFAULT_GROUP_PINNED_FRAME_URL: &str = "";
+pub const DEFAULT_MESSAGE_EXPIRATION_MS: i64 = 0;
```
mchenani, line 58: Still thinking of setting this to null instead of 0, if you have another idea, lmk.

nplasterer: @cameronvoell might have thoughts on this.

mchenani: Since its a value in the map, instead of keeping a default value I decided to not insert it at all.

nplasterer: We had issues in the past of if two people are using two different versions of the SDK in the same group chat. We need to make sure that the person on the old version of the sdk does not break. I believe there was some sort of migration that needs to happen but again @cameronvoell would remember better than me how we decided to go about this for new metadata.

cameronvoell: No migration needed here, existing groups will not have message expiration set, but updating the expiration set will be possible with default permission of admin_only (for groups created before this change goes in) => 

https://github.com/xmtp/libxmtp/blob/ab9bb055e6fe60057836f5a631733d2aee2af584/xmtp_mls/src/groups/group_permissions.rs#L1041-L1049

That path of adding new metadata to existing groups was tested with this change here, which we can reference for adding expiration here: https://github.com/xmtp/libxmtp/pull/888/files

bindings_node/src/permissions.rs
```diff
@@ -161,12 +161,13 @@ pub struct PermissionPolicySet {
   pub update_group_description_policy: PermissionPolicy,
   pub update_group_image_url_square_policy: PermissionPolicy,
   pub update_group_pinned_frame_url_policy: PermissionPolicy,
+  pub update_message_expiration_ms_policy: PermissionPolicy,
```
mchenani, line 164: I Added only one policy for both `message_expiration_ms` and `message_expiration_from_ms`. wdyt?

cameronvoell: yea I think that makes sense, doesnt make sense for someone to be able to change one without the other 👍

cameronvoell (approved):

PR #1479, opened 2025-01-08
Backup your account without a second device

codabrink:
Proto: https://github.com/xmtp/proto/pull/239


xmtp_mls/src/groups/device_sync/backup/backup_importer.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    pub async fn insert(&mut self, provider: &XmtpOpenMlsProvider) -> Result<(), DeviceSyncError> {
+        provider
+            .transaction_async(|provider| async move {
+                let conn = provider.conn_ref();
+                while let Some(element) = self
+                    .next_element()
+                    .await
+                    .map_err(|e| StorageError::Generic(e.to_string()))?
```
insipx, line 97: I'd rather stay away from Generic string errors if possible, they can't be matched on and more difficult to process afterwards than normal run of the mill errors. Both these functions return DeviceSyncError, why not use that?

codabrink: I would love for this to return a DeviceSync error, but since it's in a transaction closure, it enforces StorageError. I'll see if I can put some more through towards it to get out of this pattern.

xmtp_mls/src/groups/device_sync/backup/export_stream.rs
```diff
@@ -0,0 +1,139 @@
...
+            .elements
+            .iter()
+            .flat_map(|&e| match e {
+                BackupElementSelection::Consent => {
+                    vec![BackupRecordStreamer::<ConsentSave>::new_stream(
+                        provider, opts,
+                    )]
+                }
+                BackupElementSelection::Messages => vec![
+                    // Order matters here. Don't put groups before messages.
```
insipx, line 34: what happens if groups come before messages?

xmtp_mls/src/groups/device_sync/backup.rs
```diff
@@ -0,0 +1,180 @@
...
+            end_ns: value.end_ns,
+            start_ns: value.start_ns,
+            elements: value.elements.iter().map(|&e| e as i32).collect(),
+            exported_at_ns: now_ns(),
+        }
+    }
+}
+
+impl BackupOptions {
+    #[cfg(not(target_arch = "wasm32"))]
```
insipx, line 43: what's the plan for web support/how would i use this from the web?

xmtp_mls/src/groups/device_sync/backup.rs
```diff
@@ -0,0 +1,180 @@
...
+use backup_exporter::BackupExporter;
+use std::{path::Path, sync::Arc};
+use thiserror::Error;
+use xmtp_common::time::now_ns;
+use xmtp_proto::xmtp::device_sync::{BackupElementSelection, BackupMetadata};
+
+pub use backup_importer::BackupImporter;
+
+// Increment on breaking changes
+const BACKUP_VERSION: u32 = 0;
```
insipx, line 12: what qualifies as a breaking change where we need to increment this?

insipx (commented):
i'll have to give this a deeper look later on, but it looks good, I'm not sure how it will work from webassembly from just this PR, though.

xmtp_mls/src/groups/device_sync/backup/backup_importer.rs
codabrink: Oh, I completely overlooked the fact that `async_transaction` returns an unconstrained generic. It's fixed.

xmtp_mls/src/groups/device_sync/backup/export_stream.rs
codabrink: The insertion order is opposite the order here, and inserting groups before messages will cause a FK constraint error. I could probably reverse the iterator when building the streams to make that more clear..

codabrink: Fixed.

xmtp_mls/src/groups/device_sync/backup.rs
codabrink: Just like how we have a handler that consumes the `AsyncRead` to a file for native [here](https://github.com/xmtp/libxmtp/pull/1479/files#diff-d276d35d0e0e31d58f70cacfa50d236b16e8e5571aaed8b08c2f7d624210c924R11), we'd want to do something similar for wasm files.

xmtp_mls/src/groups/device_sync/backup.rs
codabrink: If we add a new optional field to one of the structs that's exported, we'd want to put some logic on the importer that says something like... `If backup_version < 1, set new field to this default value` or maybe use this separate hydrater entirely. That's just one example. Really this is just a field that I'm sure will be useful to delineate between updates, and make importing old archives possible.

xmtp_mls/src/groups/device_sync/backup/backup_exporter.rs
```diff
@@ -0,0 +1,158 @@
...
+                }
+                Stage::Elements => match this.stream.poll_next_unpin(cx) {
+                    Poll::Ready(Some(element)) => element.encode_to_vec(),
+                    Poll::Pending => return Poll::Pending,
+                    Poll::Ready(None) => {
+                        if !this.encoder_finished {
+                            this.encoder_finished = true;
+                            let fut = this.zstd_encoder.close();
+                            pin_mut!(fut);
+                            let _ = fut.poll(cx)?;
```
insipx, line 122: what happens if this is `Poll::Pending` and we don't handle it within this Futures state, would some elements not get flushed/lead to corrupted state?

xmtp_mls/src/groups/device_sync/backup/backup_exporter.rs
```diff
@@ -0,0 +1,158 @@
...
+                        break;
+                    }
+                },
+            };
+
+            let mut element = this
+                .cipher
+                .encrypt(&this.nonce, &*element)
+                .expect("Encryption should always work");
+            let mut bytes = (element.len() as u32).to_le_bytes().to_vec();
```
insipx, line 133: what's the reasoning for little endian here?

xmtp_mls/src/groups/device_sync/backup/backup_exporter.rs
```diff
@@ -0,0 +1,158 @@
...
+                            let fut = this.zstd_encoder.close();
+                            pin_mut!(fut);
+                            let _ = fut.poll(cx)?;
+                        }
+                        break;
+                    }
+                },
+            };
+
+            let mut element = this
```
insipx, line 129: for my own understanding:

the element is encrypted with the nonce and element from above, then the length of it is pre-pended, then we write it with the encoder.

What happens if the write encoder returns `Pending`? Do we re-do the above/am I missing a part where we are storing the position and encoding from there

xmtp_mls/src/groups/device_sync/backup.rs
insipx: is it possible to get a test that just reads bytes from the exported backup from wasm?

xmtp_mls/src/groups/device_sync/backup/backup_exporter.rs
codabrink: Compatibility across architectures, even though x86 and arm are both little endian, it's nice to be safe.

xmtp_mls/src/groups/device_sync/backup/backup_exporter.rs
codabrink: Looked into it together - this resolves immediately when the inner is a vec.

insipx (approved):

irzhywau:
is this PR have impact on sdk usage? any new API methods we can use to trigger the flow?

nplasterer:
@irzhywau 👋 Hey this is just the first PR in a stack of several PRs coming over the next few weeks. The feature isn't complete just yet. I think we are shooting for mid march or later. We will open an XIP so you can track the progress. 🙏 cc @jhaaaa

PR #1481, opened 2025-01-09
Revert push to `main` 404e996

cameronvoell:
will comment here when branch push rules are fixed 🤦


cameronvoell:
Updated permissions:

image

codabrink (approved):
Interesting lint. What caused that to show up? :thinking:

PR #1488, opened 2025-01-11
json deserialization for legacy reactions

cameronvoell:
Part of https://github.com/xmtp/libxmtp/issues/1403


cameronvoell:
* **#1488** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1488?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xmtp_content_types/src/reaction.rs
```diff
@@ -47,6 +48,41 @@ impl ContentCodec<ReactionV2> for ReactionCodec {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct LegacyReaction {
```
neekolas, line 52: We can't shoe-horn the legacy reaction into the new reaction type? I thought the data was about the same

cameronvoell: Original type had arbitrary String for reaction action and reaction schema ([js sdk used union type to restrict it](https://github.com/xmtp/xmtp-js/blob/0f0ae21a335cbb5b1a401248ce9a080314559f5f/content-types/content-type-reaction/src/Reaction.ts#L14-L37)), I updated[ to enum](https://github.com/xmtp/proto/blob/e295b4666d8f71469b953c7288cd34beea0e40d9/proto/mls/message_contents/content_types/reaction.proto#L18-L31) in the new proto type to be more precise so we wouldnt need to do this type of conversion across the SDKs: https://github.com/xmtp/xmtp-android/blob/51c7b170d53f73d32d3a1d33b0d170d156183e43/library/src/main/java/org/xmtp/android/library/codecs/ReactionCodec.kt#L40-L55

nplasterer:
👋 Sorry for the drive by comment here just getting caught up on my notifications.
When you say legacy reaction type are you referencing the legacy reaction types from our very very first reaction iteration in 2023? We have to support that in V2 because of existing reactions existing on the network before we switched to the new scheme. We switched to the new scheme before V3 was ever supported so there shouldn't be any need to continue to support these in V3. There shouldn't be any legacy reactions in V3. This legacy check here https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/codecs/ReactionCodec.kt#L82-L88 should actually be deleted.

xmtp_content_types/src/reaction.rs
neekolas: Ahh. Too bad. 

Agree the new format is nicer.

cameronvoell:
> 👋 Sorry for the drive by comment here just getting caught up on my notifications. When you say legacy reaction type are you referencing the legacy reaction types from our very very first reaction iteration in 2023? We have to support that in V2 because of existing reactions existing on the network before we switched to the new scheme. We switched to the new scheme before V3 was ever supported so there shouldn't be any need to continue to support these in V3. There shouldn't be any legacy reactions in V3. This legacy check here https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/codecs/ReactionCodec.kt#L82-L88 should actually be deleted.

@nplasterer Not referencing the first 2023 reaction, just the current JSON serialized one. I added a comment here to clarify the source https://github.com/xmtp/libxmtp/pull/1488/commits/4ca67fe7547ce3a0b540330905896125f687141f#diff-24e4e78e8578981034245acc25394ec3f75bb7e0edf8dede5259d1ab82f96354R51. I'll make a note to clean up that old code in the SDks though when I add the new proto serialized version though 👌

nplasterer (approved):

PR #1494, opened 2025-01-13
test(policy): update policies of an empty group

mchenani:
Add a test to update the group permissions on an empty group.


cameronvoell (approved):
we'll make sure CI looks good than this should be good to go. thanks for help with the test 🙏

PR #1511, opened 2025-01-16
Allow creating conversations with inboxIds

nplasterer:
Add methods for creating groups and dms by inboxId instead of address.


insipx (approved):

insipx:
nice!

PR #1517, opened 2025-01-16
Make pool read-only, with a single write connection.

codabrink:


xmtp_mls/src/storage/encrypted_store/native.rs
```diff
@@ -145,7 +155,11 @@ impl NativeDb {
             pool.state().connections
         );
 
-        Ok(pool.get()?)
+        // Turn of writitng by default
+        let mut conn = pool.get()?;
+        conn.batch_execute("PRAGMA query_only = ON;")?;
```
insipx, line 160: this will reset once the connection is out of scope, so you need to put it in the customizer `on_acquire`. the writeconn is OK as long as its not dropped/held on the NativeDb for the duration of execution

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -393,23 +396,23 @@ where
         tracing::debug!("Transaction beginning");
         {
             let connection = self.conn_ref();
-            let mut connection = connection.inner_mut_ref();
+            let mut connection = connection.read_mut_ref();
```
insipx, line 399: Transactions are directly tied to their connection, so there may be trouble by using the read connection interchangeably with the write connection. The write transaction will have started `COMMIT TRANSACTION` but a `BEGIN TRANSACTION` pragma will never have happened because it occurred on the previous read reference

insipx: This does pose another problem though, if we use `begin_transaction` and then release the mutex lock, I think we can still run into a db lock issue

- begin transaction with write conn
- do a write
- tx releases lock
- yield b/c tx is async and does a network call
- begin another tx with write lock somewhere else
    - completely legal because we relinquisheed the mutex after begin_transaction
- db lock even though we're using a mutex, because although the mutex exists in libxmtp, the transaction is open in SQLite with the same connection, which conflicts with another transaction on the same database



If we instead ensure the mutex is held for the entire transaction, we'll just run into a deadlock cycle waiting for access to the RwLock

codabrink: Yeah, you're right. It does deadlock

xmtp_mls/src/client.rs
```diff
@@ -1403,6 +1403,7 @@ pub(crate) mod tests {
         not(target_arch = "wasm32"),
         tokio::test(flavor = "multi_thread", worker_threads = 1)
     )]
+    #[ignore]
```
insipx: how com we need to ignore this test?

xmtp_mls/src/storage/encrypted_store/db_connection.rs
```diff
@@ -18,38 +25,83 @@ pub type DbConnection = DbConnectionPrivate<sqlite_web::connection::WasmSqliteCo
...
 where
     C: diesel::Connection,
 {
+    pub(crate) fn start_transaction<Db: XmtpDb<Connection = C>>(
+        &self,
+    ) -> Result<TransactionGuard, StorageError> {
+        let mut write = self
+            .write
+            .as_ref()
+            .expect("Tried to open transaction on read-only connection")
```
insipx, line 58: we should probably throw an error here instead of panicking

xmtp_mls/src/storage/encrypted_store/db_connection.rs
```diff
@@ -18,38 +25,83 @@ pub type DbConnection = DbConnectionPrivate<sqlite_web::connection::WasmSqliteCo
...
+            .lock();
+        <Db as XmtpDb>::TransactionManager::begin_transaction(&mut *write)?;
+
+        if self.in_transaction.swap(true, Ordering::SeqCst) {
+            panic!("Already in transaction.");
```
insipx: likewise, this would be better as an error

xmtp_mls/src/groups/mod.rs
```diff
@@ -933,6 +933,8 @@ impl<ScopedClient: ScopedGroupClient> MlsGroup<ScopedClient> {
             intent_data.into(),
         )?;
 
+        tracing::warn!("This makes it here?");
```
insipx, line 936: testing log

xmtp_mls/src/storage/encrypted_store/db_connection.rs
```diff
@@ -18,38 +25,83 @@ pub type DbConnection = DbConnectionPrivate<sqlite_web::connection::WasmSqliteCo
 // Do not derive clone here.
 // callers should be able to accomplish everything with one conn/reference.
 #[doc(hidden)]
+#[derive(Clone)]
 pub struct DbConnectionPrivate<C> {
-    inner: Arc<Mutex<C>>,
+    read: Arc<Mutex<C>>,
+    write: Option<Arc<Mutex<C>>>,
```
insipx, line 31: When would the write connection be None?

insipx (changes requested):

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -667,78 +664,4 @@ pub(crate) mod tests {
...
-
-    // get two connections
-    // start a transaction
-    // try to write with second connection
-    // write should fail & rollback
-    // first thread succeeds
-    // wasm does not have threads
-    #[cfg_attr(not(target_arch = "wasm32"), tokio::test)]
-    #[cfg(not(target_arch = "wasm32"))]
-    async fn test_transaction_rollback() {
```
codabrink, line 679: Had to delete this test since the barriers in this test were trying to force something that should be impossible now and was causing a deadlock.

xmtp_mls/src/storage/encrypted_store/db_connection.rs
```diff
@@ -18,38 +25,77 @@ pub type DbConnection = DbConnectionPrivate<sqlite_web::connection::WasmSqliteCo
 // Do not derive clone here.
 // callers should be able to accomplish everything with one conn/reference.
 #[doc(hidden)]
+#[derive(Clone)]
```
insipx, line 28: when do we clone the db connection?

codabrink: Ah, there was a moment where I was cloning it, but that implementation is gone. Let me clean it out.

mchenani (approved):

xmtp_mls/src/storage/encrypted_store/db_connection.rs
```diff
@@ -19,37 +26,75 @@ pub type DbConnection = DbConnectionPrivate<sqlite_web::connection::WasmSqliteCo
...
 }
 
 impl<C> DbConnectionPrivate<C>
 where
     C: diesel::Connection,
 {
+    pub(crate) fn start_transaction<Db: XmtpDb<Connection = C>>(
+        &self,
+    ) -> Result<TransactionGuard<'_>, StorageError> {
+        let guard = self.transaction_lock.lock();
```
insipx, line 63: what can go wrong if we don't have this mutex?

xmtp_mls/src/storage/encrypted_store/mod.rs
```diff
@@ -383,25 +384,21 @@ where
...
     {
         tracing::debug!("Transaction beginning");
-        {
-            let connection = self.conn_ref();
-            let mut connection = connection.inner_mut_ref();
-            <Db as XmtpDb>::TransactionManager::begin_transaction(&mut *connection)?;
-        }
 
         let conn = self.conn_ref();
+        let _guard = conn.start_transaction::<Db>()?;
```
insipx, line 389: now that async transaction are gone and our connection is a mutex, why can't we use diesels [transaction](https://docs.rs/diesel/latest/diesel/connection/trait.Connection.html#method.transaction) function?

xmtp_mls/src/storage/encrypted_store/db_connection.rs
codabrink: Transactions will get nested.

insipx: do we have a test that fails with nested transactions -- shouldn't we avoid nested transactions altogether now that we no longer have async? 

Or, if we have nested transaction in sync code, is this Mutex just covering up for a suboptimal code design somewhere else in libxmtp that we need to investigate

codabrink: Transactions or other writes could get nested if another provider tries to start a transaction before this transaction closes.

codabrink: > shouldn't we avoid nested transactions altogether now that we no longer have async?

With the lib itself being async a worker could try to start a write while this thread is still writing. This mutex will block that.

xmtp_mls/src/storage/encrypted_store/db_connection.rs
```diff
@@ -19,37 +26,81 @@ pub type DbConnection = DbConnectionPrivate<sqlite_web::connection::WasmSqliteCo
...
-        self.inner.lock()
+    /// Do a scoped query with a mutable [`diesel::Connection`]
+    /// reference
+    pub(crate) fn raw_query_write<T, E, F>(&self, fun: F) -> Result<T, E>
+    where
+        F: FnOnce(&mut C) -> Result<T, E>,
+    {
+        let _guard;
+        // If this connection is not in a transaction
+        if !self.in_transaction.load(Ordering::SeqCst) {
```
insipx, line 99: what happens if we _are_ in a transaction, and a worker/thread not in a transaction tries to write with the connection that started a transaction

codabrink: This thread should own the guard belonging to the lock on 101, and will have to wait until we're done.

xmtp_mls/src/storage/encrypted_store/db_connection.rs
```diff
@@ -19,37 +26,81 @@ pub type DbConnection = DbConnectionPrivate<sqlite_web::connection::WasmSqliteCo
 // callers should be able to accomplish everything with one conn/reference.
 #[doc(hidden)]
 pub struct DbConnectionPrivate<C> {
-    inner: Arc<Mutex<C>>,
+    // Connection with read-only privileges
+    read: Option<Arc<Mutex<C>>>,
+    // Connection with write privileges
+    write: Arc<Mutex<C>>,
+    // Is any connection (possibly this one) currently in a transaction?
+    transaction_lock: Arc<Mutex<()>>,
```
insipx, line 34: nit: it would be helpful at least for me to explicitly state that this lock is shared among all transactions on a particular DB

insipx: iirc, this transaction lock seems like a global write lock. 
- if the current connection is in a transaction, then we can access the write connection. 
- If there are no transactions, then we can access the write connection. 
- if there is another transaction, we will block until we can access the write connection.

Is that correct?

codabrink: Yes, I would love to just keep the mutex open on the write connection itself and was my original goal, but with transaction closures taking a provider and getting connection references on their own, that route was impossible without severe lifetime leakage. I had to use a double-mutex pattern to avoid putting a lifetime on anything that implemented a dbconnection.

xmtp_mls/src/storage/encrypted_store/db_connection.rs
codabrink: > do we have a test that fails with nested transactions

We have a test that checks for nested transactions here: https://github.com/xmtp/libxmtp/blob/17a8d804ff8fee4464cb7db71911cdbf7e4142c5/xmtp_mls/src/groups/mod.rs#L3801

insipx (approved):

xmtp_mls/src/storage/encrypted_store/mod.rs
insipx: right, because the barriers would never progress since the transaction never finishes. So we would only deadlock IRL if we somehow had a transaction that never finishes, which shouldn't be possible w/o async

insipx:
Looks good!

PR #1532, opened 2025-01-23
Remove transaction_async

codabrink:


xmtp_mls/src/groups/mls_ext/welcome_ext.rs
```diff
@@ -0,0 +1,68 @@
...
+pub(crate) fn build_group_join_config() -> MlsGroupJoinConfig {
+    MlsGroupJoinConfig::builder()
+        .wire_format_policy(WireFormatPolicy::default())
+        .max_past_epochs(MAX_PAST_EPOCHS)
+        .use_ratchet_tree_extension(true)
+        .build()
+}
+
+fn deserialize_welcome(welcome_bytes: &Vec<u8>) -> Result<Welcome, ClientError> {
+    // let welcome_proto = WelcomeMessageProto::decode(&mut welcome_bytes.as_slice())?;
```
mchenani, line 60: Q: do we need this?

Cargo.toml
```diff
@@ -41,10 +41,10 @@ futures = { version = "0.3.30", default-features = false }
...
 hex = "0.4.3"
 hkdf = "0.12.3"
-openmls = { git = "https://github.com/xmtp/openmls", rev = "043b347cb18d528647df36f500725ab57c41c7db", default-features = false }
-openmls_basic_credential = { git = "https://github.com/xmtp/openmls", rev = "043b347cb18d528647df36f500725ab57c41c7db" }
-openmls_rust_crypto = { git = "https://github.com/xmtp/openmls", rev = "043b347cb18d528647df36f500725ab57c41c7db" }
-openmls_traits = { git = "https://github.com/xmtp/openmls", rev = "043b347cb18d528647df36f500725ab57c41c7db" }
+openmls = { git = "https://github.com/xmtp/openmls", branch = "coda/expose-some-apis", default-features = false }
+openmls_basic_credential = { git = "https://github.com/xmtp/openmls", branch = "coda/expose-some-apis" }
+openmls_rust_crypto = { git = "https://github.com/xmtp/openmls", branch = "coda/expose-some-apis" }
+openmls_traits = { git = "https://github.com/xmtp/openmls", branch = "coda/expose-some-apis" }
```
nplasterer, line 47: Lets merge the PR and point to a commit instead 
https://github.com/xmtp/openmls/pull/39#pullrequestreview-2579325061

Also is this something we should commit back to openMLS?

codabrink: Ah yes, good catch. I need to change this back to a rev.
And yes, I think it'd be good to get that into the core openmls. We talked with them during the last meeting about relying on that data from the staged welcome and they said it should be fine.

xmtp_mls/src/identity.rs
```diff
@@ -533,16 +533,16 @@ impl Identity {
             return Ok(());
         }
 
-        self.rotate_key_package(provider, api_client).await?;
+        let kp_bytes = self.rotate_key_package(provider)?;
```
mchenani, line 536: Q: why we don't use the function from here:
https://github.com/xmtp/libxmtp/pull/1532/files?diff=split&w=1#diff-3c0189371525b6aae3be7e5a7b5b5682204ef341fb2744fc876148c5ecab8642R750

Since seems to has the same two lines we're having here. wdyt?

xmtp_mls/src/client.rs
```diff
@@ -752,17 +747,8 @@ where
...
-                let provider = &provider;
-                async {
-                    self.identity()
-                        .rotate_key_package(provider, &self.api_client)
-                        .await?;
-                    Ok::<_, IdentityError>(())
-                }
-            })
-            .await?;
+        let kp_bytes = self.identity().rotate_key_package(provider)?;
```
mchenani, line 750: Q: wdyt of renaming the function to reflect the updated logic? suggestion: `rotate_and_upload_key_pacakge`

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -351,286 +351,371 @@ where
...
+        envelope: &GroupMessageV1,
+        cursor: Option<i64>,
+    ) -> Result<(), GroupMessageProcessingError> {
+        let GroupMessageV1 {
+            created_ns: envelope_timestamp_ns,
+            id: ref msg_id,
+            ..
+        } = *envelope;
+
+        let processed_message = mls_group.process_message(provider, message)?;
```
neekolas, line 530: If this isn't in a transaction, won't this function immediately delete the keys used to decrypt the message?

What happens if the transaction below rolls back? How do we get those keys back to retry?

codabrink: You're right. I missed this when I was doing read only connection checks. Let me move it into the transaction.

xmtp_mls/src/client.rs
```diff
@@ -753,17 +748,8 @@ where
...
+        self.api_client.upload_key_package(kp_bytes, true).await?;
```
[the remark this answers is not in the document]
insipx: Although, i'm not sure if this was in a transaction previously so it might be OK

xmtp_mls/src/client.rs
```diff
@@ -861,46 +847,48 @@ where
...
-                let cursor = welcome.id;
-                let is_updated = provider.conn_ref().update_cursor(
-                    self.installation_public_key(),
-                    EntityKind::Welcome,
-                    welcome.id as i64,
-                )?;
-                if !is_updated {
-                    return Err(ProcessIntentError::AlreadyProcessed(cursor).into());
+        let cursor = welcome.id;
+        let is_updated = provider.conn_ref().update_cursor(
```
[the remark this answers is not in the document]
codabrink: Thanks for calling this out. Moved it down to the transaction.

xmtp_mls/src/client.rs
codabrink: Love it. Just changed.

xmtp_mls/src/identity.rs
codabrink: The generic constraints are more lax here than where the rotate_and_upload_key_package function is defined on. Definitely a cleanup item for sure, but outside the scope of this PR I think, and this PR is already big enough :sweat_smile:

xmtp_mls/src/groups/mls_sync.rs
codabrink: This is resolved.

xmtp_mls/src/client.rs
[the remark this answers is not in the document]
codabrink: This should be handled well now

xmtp_mls/src/identity.rs
codabrink: Ended up rewriting this to provide better data integrity. Reuses the function now.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -351,286 +356,429 @@ where
...
+        let GroupMessageV1 {
+            created_ns: envelope_timestamp_ns,
+            id: ref msg_id,
+            ..
+        } = *envelope;
+
+        // We need to process the message twice to avoid an async transaction.
+        // We'll process for the first time, get the processed message,
+        // and roll the transaction back, so we can fetch updates from the server before
+        // being ready to process the message for a second time.
```
neekolas, line 550: Would be nice as a follow-up to see how this impacts our benchmarks

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -655,72 +803,67 @@ where
...
+            ProcessedMessageContent::ProposalMessage(_proposal_ptr) => {
+                // intentionally left blank.
+            }
+            ProcessedMessageContent::ExternalJoinProposalMessage(_external_proposal_ptr) => {
+                // intentionally left blank.
+            }
+            ProcessedMessageContent::StagedCommitMessage(staged_commit) => {
+                let staged_commit = *staged_commit;
+                let validated_commit =
+                    validated_commit.expect("Needs to be present when this is a staged commit");
```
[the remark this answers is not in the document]
codabrink: Just cleaned this part up in the most recent commit, actually. :) Refresh the page and take another look.

xmtp_mls/src/groups/mod.rs
```diff
@@ -582,108 +582,87 @@ impl<ScopedClient: ScopedGroupClient> MlsGroup<ScopedClient> {
...
         welcome_id: i64,
+        cursor: Option<i64>,
     ) -> Result<Self, GroupError>
     where
         ScopedClient: Clone,
     {
         tracing::info!("Creating from welcome");
-        let mls_welcome =
+        let staged_welcome =
             StagedWelcome::new_from_welcome(provider, &build_group_join_config(), welcome, None)?;
```
insipx, line 599: This is where we re-form the staged welcome from the decrypted welcome parts. Maybe we could just pass the decrypted welcome directly, and store the staged welcome on the decrypted welcome?

xmtp_mls/src/groups/mls_ext/mls_ext_welcome.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+        let staged_welcome = processed_welcome.into_staged_welcome(provider, None)?;
+
+        let added_by_node = staged_welcome.welcome_sender()?;
+
+        let added_by_credential = BasicCredential::try_from(added_by_node.credential().clone())?;
+        let added_by_inbox_id = parse_credential(added_by_credential.identity())?;
+
+        Ok(DecryptedWelcome {
+            welcome,
```
insipx, line 45: nit: 
i think we're being redundant here, we turn the welcome into a staged welcome on 37, then return the un-staged welcome. But where we use DecryptedWelcome, we immediately turn it back into a staged welcome

insipx: Its a leftover from what the code was before, but would be nice to fix nonetheless

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1658,8 +1790,8 @@ fn get_and_clear_pending_commit(
     Ok(commit)
 }
 
-fn decode_staged_commit(data: Vec<u8>) -> Result<StagedCommit, GroupMessageProcessingError> {
-    Ok(db_deserialize(&data)?)
+fn decode_staged_commit(data: &[u8]) -> Result<StagedCommit, GroupMessageProcessingError> {
```
insipx, line 1793: nice

mchenani (approved):

PR #1534, opened 2025-01-23
Remove pinned frames url from metadata

nplasterer:
Talked with product and it doesn't currently feel like there is any value in leaving a url field in the metadata at all so lets just remove it entirely.

THIS IS A BREAKING GROUP CHANGE


codabrink (approved):

cameronvoell (approved):
this change looks good. Even in rare cases where a user on an old version tried to update this field on a new group that doesnt have the field, we would not have a fork. 👌 see metadata update eval logic for more details: https://github.com/xmtp/libxmtp/blob/3b01e70d28550e327f91be40f7cb3489d6aed174/xmtp_mls/src/groups/group_permissions.rs#L1023-L1064

Issue #1526, opened 2025-01-20
Update character length for name and description


nplasterer:
We need to take another pass on metadata

- [ ] All metadata fields should have a character limit to safe guard against extremely large payloads

name should have a character limit of 100
description should have a character limit of 300
imageUrl should have a character limit 2048

PR #1561, opened 2025-01-28
Set metadata character limits

nplasterer:
Closes https://github.com/xmtp/libxmtp/issues/1526

Adds a character limit to Name, Description, and ImageURL

Not a breaking change will only break on update of existing fields if they are trying to update to a larger than allowed size.


codabrink (approved):

mchenani (approved):
Q: what would happen if an old group hold data larger than these thresholds? I would expect that we don't allow the user to modify anything else; is that right?

cameronvoell (approved):

Issue #1594, opened 2025-02-04
Bug: Conflicting Uniffi Headers


nplasterer:
### Describe the bug

For apps that have there own uniffi bindings our headers conflict. The problem is that both libraries generate same headers etc which shouldn't happen. The one from libxmtp should have it's own paths so clients don't have issues 

### Expected behavior

Add this to header

```
PATH_SUBDIR = ./out/bindings/include/libxmtp

ios-package: ios-build
	mkdir -p ${PATH_SUBDIR}
	cp ${PATH_HEADER} ${PATH_SUBDIR}/
	cp ${PATH_MODULEMAP} ${PATH_SUBDIR}/module.modulemap
```

PR #1596, opened 2025-02-05
Make iOS uniffi path more unique

nplasterer:
Fixes: https://github.com/xmtp/libxmtp/issues/1594

Makes the path more unique so that it does not conflict with other uniffi bindings.


nplasterer:
Pushed through the iOS SDK and confirmed it still works as expected. I think this should be safe.

neekolas (approved):

PR #1597, opened 2025-02-05
fix: start disappearing messages worker when building the client

mchenani:
Initialize the disappearing message deleter in the client builder so it runs automatically.


codabrink:
Since the app can be backgrounded, you're going to want to restart in the reconnect_db too https://github.com/xmtp/libxmtp/blob/ae4eb9cf5dd09b3ce60818bd5abef61e92d46bda/xmtp_mls/src/client.rs#L267

codabrink (approved):

PR #1598, opened 2025-02-05
contextual logging, remove optional cursor

insipx:
- get cursor from message envelope rather than having be optional
- this PR adds contextual logging, which outputs logs grouped by context in a tree format based on tracing spans.
    - enable contextual logging with `CONTEXTUAL=1`, i.e `CONTEXTUAL=1 RUST_LOG=xmtp_mls::subscriptions=debug,xmtp_mls::groups::mls_sync=debug cargo test`
    - contextual logging is only available in tests and does not effect production binary

### DIFF
--- a/bindings_ffi/src/logger.rs
+++ b/bindings_ffi/src/logger.rs
@@ -46,47 +46,7 @@ mod other {
     where
         S: Subscriber + for<'a> LookupSpan<'a>,
     {
-        use tracing_subscriber::{
-            fmt::{self, format},
-            EnvFilter, Layer,
-        };
-        let structured = std::env::var("STRUCTURED");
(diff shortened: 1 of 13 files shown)


codabrink (approved):

bindings_ffi/src/logger.rs
```diff
@@ -46,47 +46,7 @@ mod other {
...
-                                    write!(writer, "{:?}", value)?;
-                                }
-                                Ok(())
-                            })
-                        })
-                        .with_filter(filter())
-                })
-                .boxed(),
-        ]
+        xmtp_common::logger_layer()
```
cameronvoell, line 49: looks like this broke swift bindings build @insipx 

See https://github.com/xmtp/libxmtp/actions/runs/13182727625/job/36797319367

PR #1601, opened 2025-02-05
Error handling for rate limits

insipx:
- moved api to `xmtp_api` crate to separate from xmtp_mls/xmtp_id
- added Error associated type to Api Traits + Added Error trait that can indicate gRPC error code if it exists
- removed old ErrorKind/Error xmtp_proto errors
- simplified ClientBuilder
- added timed limits to exponential backoff strategy

`RetryableError` has a new 'needs_cooldown' function (default to false) that can be set on certain error types. If set, globally sets the cooldown for that `Retry` type based on the Backoff strategy. We currently only have one backoff strategy which is exponential.


bindings_ffi/src/mls.rs
```diff
@@ -166,7 +171,16 @@ pub async fn get_inbox_id_for_address(
     api: Arc<XmtpApiClient>,
     account_address: String,
 ) -> Result<Option<String>, GenericError> {
-    let api = ApiClientWrapper::new(Arc::new(api.0.clone()), Default::default());
+    let cooldown = xmtp_common::ExponentialBackoff::builder()
+        .duration(std::time::Duration::from_secs(3))
```
insipx, line 175: this is where the rate limit cooloff is configured. Wasn't completely sure about the values here, but 3 seconds that multiplies by three for every consecutive rate limit sounds reasonable for now.

mchenani: nit: I suggest to extract the value as a value somewhere else, like in the configuration file? Also since the rate limit on the BE configured for 5 minutes, that's still a huge gap, what we should do about it? prob a question for @neekolas

bindings_node/src/inbox_id.rs
```diff
@@ -3,27 +3,44 @@ use napi::bindgen_prelude::Result;
...
+use xmtp_api::ApiClientWrapper;
 use xmtp_api_grpc::grpc_api_helper::Client as TonicApiClient;
 use xmtp_common::retry::Retry;
+use xmtp_common::ExponentialBackoff;
 use xmtp_id::associations::generate_inbox_id as xmtp_id_generate_inbox_id;
 use xmtp_id::associations::MemberIdentifier;
-use xmtp_mls::api::ApiClientWrapper;
+use xmtp_proto::api_client::ApiBuilder;
+
+fn retry_strategy() -> Retry<ExponentialBackoff, ExponentialBackoff> {
```
mchenani, line 14: Q: so this function set it's own values? regardless of what we set in:
https://github.com/xmtp/libxmtp/pull/1601/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8fR174

insipx: bindings_node/bindings_ffi/bindings_wasm all configure separately. We can expose a function from xmtp_mls to have a unified cooldown strategy and/or just make it a default in the ApiClientWrapper too

bindings_ffi/src/mls.rs
insipx: Extracting to a place in xmtp_api is reasonable, will update for that

xmtp_mls/src/client.rs
```diff
@@ -1073,9 +1071,13 @@ pub(crate) mod tests {
     async fn test_group_member_recovery() {
         let amal = ClientBuilder::new_test_client(&generate_local_wallet()).await;
         let bola_wallet = generate_local_wallet();
+        tracing::info!("amal = {}", amal.inbox_id());
```
mchenani, line 1074: nit: do we still need these?-- I know they're in the test, but still just wanted to make sure you want them.

insipx: yeah i'll remove them to make test logging less noisy

xmtp_mls/src/storage/errors.rs
```diff
@@ -54,6 +46,10 @@ pub enum StorageError {
     Generic(String),
     #[error("Transaction was intentionally rolled back")]
     IntentionalRollback,
+    #[error("failed to deserialize from db")]
+    DbDeserialize,
+    #[error("failed to serialize from db")]
```
mchenani, line 51: 🤔 not sure
```suggestion
    #[error("failed to serialize for db")]
```

xmtp_proto/src/error.rs
```diff
@@ -1,147 +1,258 @@
...
-            f.field(source);
+impl std::fmt::Display for ApiEndpoint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        use ApiEndpoint::*;
+        match self {
+            Publish => write!(f, "publish"),
+            SubscribeGroupMessages => write!(f, "subscribe_group_messages"),
+            SubscribeWelcomes => write!(f, "subscribe_welcomes"),
+            UploadKeyPackage => write!(f, "upload_key_package"),
+            FetchKeyPackages => write!(f, "fetch_key_package"),
```
mchenani, line 193: nit: to make it aligned with others
```suggestion
            FetchKeyPackages => write!(f, "fetch_key_packages"),
```

xmtp_proto/src/error.rs
```diff
@@ -1,147 +1,258 @@
...
-    }
+/// General Error types for use when converting/deserializing From/To Protos
+/// Loosely Modeled after serdes [error](https://docs.rs/serde/latest/serde/de/value/struct.Error.html) type.
+/// This general error type avoid circular hard-dependencies on crates further up the tree
+/// (xmtp_id/xmtp_mls) if they had defined the error themselves.
+#[derive(thiserror::Error, Debug)]
+pub enum ConversionError {
+    #[error("missing field {} of type {} during conversion from protobuf", .item, .r#type)]
+    Missing {
+        item: &'static str,
```
mchenani, line 216:
```suggestion
        /// the item being converted
        item: &'static str,
```

xmtp_proto/src/error.rs
```diff
@@ -1,147 +1,258 @@
...
+        r#type: &'static str,
+    },
+    #[error("type {} has invalid length. expected {} got {}", .item, .expected, .got)]
+    InvalidLength {
+        item: &'static str,
```
mchenani, line 221:
```suggestion
        /// the item being converted
        item: &'static str,
```

xmtp_proto/src/error.rs
```diff
@@ -1,147 +1,258 @@
...
+        expected: usize,
```
mchenani, line 222:
```suggestion
        /// expected length of the item being converted
        expected: usize,
```

xmtp_proto/src/error.rs
```diff
@@ -1,147 +1,258 @@
...
+        got: usize,
```
mchenani, line 223:
```suggestion
        /// the length of the received item
        got: usize,
```

bindings_node/src/inbox_id.rs
mchenani: Yeah, make sense, we can have the default and let each change the configuration based on their prefs.

mchenani (approved):
Great improvements all around! Thanks @insipx
Amazing work 🎉🔥🧹

common/src/retry.rs
```diff
@@ -16,62 +16,275 @@
...
-    // The amount to multiply the duration on each subsequent attempt
-    multiplier: u32,
-    max_jitter_ms: usize,
+    strategy: S,
+    /// global cooldown for this retry strategy
+    cooldown: C,
+    /// whether we are currently in a cooldown period
+    is_cooling: Arc<AtomicBool>,
+    /// since when have we been cooling
+    cooling_since: Arc<ArcSwap<crate::time::Instant>>,
```
codabrink, line 74: Seeing this `ArcSwap` lib for the first time. That's useful.

common/src/retry.rs
```diff
@@ -171,21 +404,34 @@ macro_rules! retry_async {
...
             let span = span.clone();
             #[allow(clippy::redundant_closure_call)]
+            $retry.cooldown().await;
             let res = $code.instrument(span).await;
             match res {
                 Ok(v) => break Ok(v),
                 Err(e) => {
+                    $retry.last_err(&e);
+                    if (&e).needs_cooldown() {
+                        tracing::warn!("Hit {}, cooling down", e);
```
cameronvoell, line 419: If an app thinks they might be getting rate limited, the error we could look for in the logs would be "Hit {Error}, cooling down"? Is that right? would the `ResourceExhausted` error bubble up here, and be logged, or would the cooling down warnings be the main hint?

insipx: that's correct, we would see ResourceExhausted bubbled from the GRPC call, the grpc code/message, and 'cooling down'

common/src/retry.rs
codabrink: Nit: would probably be better to use SystemTime here and just store the NS since epoch in an atomic.

PR #1606, opened 2025-02-06
fix(group): not fail on unverified KeyPackages and return failed to add installations when adding members to a conversation

### LINKED ISSUE #1507: Soft fail on group installation add for expired or bad key packages
We need a way to successfully add a member to a group even if all there installations are not valid.
We need a way to retry when a new installation comes back online with a valid key package that was not previously added

- Fixes https://github.com/xmtp/libxmtp/issues/1489 for empty key packages from 6 months ago
- Fixes expired key packages in a decentralized network

Notes:

Let’s say you have two installations RichA and RichB. RichB has an expired key package. You are on sequence ID 5.

I try and add you to group. RichA can be added successfully, RichB cannot. OK, fine, we just add RichA.

Now RichB comes back online and uploads a new key package. Two problems:

1. No one knows this happened because you are still on sequence_id 5. When other clients check for new members to add they won’t see any reason to look back and retry your missing installation.
2. Even if someone did try to add your missing installation, the commit wouldn’t match the expected diff from the GroupMembership structs before/after the commit.

We could loosen the requirements for 2 and allow for installations to be added who should be in the group, even if they weren’t added in the most recent diff. But it still doesn’t help us with 1.

Thinking about this a bit more - in theory the two listed problems can be solved by adding more info to the GroupMembership state. For example, a list of installations that we were not able to add.

HOWEVER - I think this might actually fall under a broader problem that goes beyond message expiries. Even with our current implementation:

1) When someone adds an installation to a group, what if they accidentally or deliberately don’t send the accompanying Welcome messages?

2) When someone adds an installation to a group, what if they accidentally or deliberately use a LeafNode/encryption_key from a very old key package that the installation no longer has?

The broader issue could be phrased as - what if an installation is considered a valid member of the group, but it didn’t actually receive the information it needed in order to join? Any solution that solves for these would seemingly also solve the original problem with key package expiry. Not that we need to solve these today, but it would imply that we should just go ahead with 6 month expiries seeing as we need to solve that issue anyway. Would love for this logic to be double-checked!

mchenani:
When an installation has a bad KeyPackage or a KeyPackage with an invalid lifetime, we currently fail to add other members as well.
This PR updates the logic to add all healthy installations while returning a list of installations that failed.
closes #1507


xmtp_mls/src/client.rs
```diff
@@ -814,14 +814,30 @@ where
...
-            .collect::<Result<_, _>>()?)
+
+        let results: HashMap<Vec<u8>, Result<VerifiedKeyPackageV2, KeyPackageVerificationError>> =
+            installation_ids
+                .iter()
+                .zip(key_package_results.values())
+                .map(|(installation_id, bytes)| {
+                    let parsed_result =
+                        VerifiedKeyPackageV2::from_bytes(&crypto_provider, bytes.as_slice())
+                            .map_err(|e| e);
```
codabrink, line 835: doesn't seem to do anything

xmtp_mls/src/groups/mod.rs
```diff
@@ -979,7 +992,10 @@ impl<ScopedClient: ScopedGroupClient> MlsGroup<ScopedClient> {
             intent_data.into(),
         )?;
 
-        self.sync_until_intent_resolved(provider, intent.id).await
+        match self.sync_until_intent_resolved(provider, intent.id).await {
+            Ok(_) => ok_result,
+            Err(error) => Err(error),
+        }
```
codabrink, line 998:
```suggestion
        Ok(self.sync_until_intent_resolved(provider, intent.id).await?)
```

codabrink, line 998:
```suggestion
        self.sync_until_intent_resolved(provider, intent.id).await?;
        ok_result
```

xmtp_mls/src/client.rs
mchenani: left out from the forced error return, good catch!

xmtp_mls/src/client.rs
```diff
@@ -1489,14 +1509,19 @@ pub(crate) mod tests {
...
             .get_key_packages_for_installation_ids(vec![installation_id.as_ref().to_vec()])
             .await
-            .unwrap();
-        let kp = kps.first().unwrap();
+            .map_err(|_| IdentityError::NewIdentity("fetched malformed keypackage".to_string()))?;
+
+        let kp_result = kps_map
+            .get(installation_id.as_ref())
+            .ok_or(IdentityError::NewIdentity(
+                "fetched malformed keypackage".to_string(),
```
neekolas, line 1521: Ideally we would persist this failure somewhere on the group so that someone can attempt to add the failed installation later. Otherwise we will update the sequence_id for the member in the `GroupMembership` and we'll never try to add this installation again.

xmtp_mls/src/client.rs
```diff
@@ -1489,14 +1509,19 @@ pub(crate) mod tests {
...
+            ))?;
 
-        serialize_key_package_hash_ref(&kp.inner, &client.mls_provider().unwrap()).unwrap()
+        serialize_key_package_hash_ref(&kp_result.clone().unwrap().inner, &client.mls_provider()?)
```
neekolas, line 1524: Won't this unwrap panic if there is a bad key package?

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1682,63 +1712,91 @@ fn extract_message_sender(
...
+                Err(_) => failed_installations.push(installation_id.clone()),
+            }
         }
     }
 
+    Ok(MembershipDiffWithKeyPackages::new(
+        new_installations,
+        new_key_packages,
+        installation_diff.removed_installations,
+        failed_installations,
```
neekolas, line 1769: Does this get persisted somewhere on the group? Ideally we would keep track of the failed installations, so that they can be retried at some point later. Otherwise once this fails it will be very difficult for us to add retrying later, because we won't know who is missing.

mchenani: I would say it’s not our responsibility to re-add failed installations. There are two cases to consider:

1. All of a user’s installations fail – In this case, we inform the user that we can’t add them. They need to resolve the issue with their account before we can proceed.
2. At least one of the user’s installations fails, but not all – In this case, the successful installations should handle adding any missing installations.

With this approach, we place the responsibility on the user rather than the admin to retry. This is because there’s a chance that the problematic installation may never be fixed.

neekolas: I'm thinking about retrying at a much later date. For example, if you have an expired key package that got dropped from the network and then come back online and upload a new key package.

If we don't store the failed installations it requires a very expensive diffing operation to figure out who is missing from a large group.

mchenani: I'm still thinking of if we distribute the load on the clients of each user it would be better, then we can show their healthy clients some alerts that you have a bad installation go and fix it, or here you can revoke that installation!
Just showing the admin that we can't add that user bc they don't have a healthy client, then the admin can add them later rather than doing it in the background.

neekolas: I don't think we have a mechanism for adding the unhealthy client later if we don't store it.

The inbox's `sequence_id` will be set to the latest. Uploading a new key package won't change your `sequence_id`. So even if you try to add the user again it won't see any installations in the diff

mchenani: Yes, we need to implement it. this is my idea:

- the client fetch the list of all installations from either the server or the sync_group
- on every welcome message the client check if all of installations are in the group, if not, the tries to add
- on every membership change, we can do like what we do on welcome messages, just to make sure we have our installations always are updated
- based on the installations list, the client will fetch all the keypackages for other installations, if something wrong, it will inform the user-- we can provide a list of all the user installations and their status

mchenani: just to be clear, I 100% agree of saving the failed installations, just wanted to put the load on other clients.
Thanks for the comments Nick, as always insightful. will cover this part in the PR too 🙌

neekolas: FWIW, I don't think you need to handle the retrying now. We just need to know how it will work to make sure the right data is persisted to the group so we can retry it later.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1556,13 +1560,39 @@ where
...
+            .await?;
+
+            // If we fail to fetch or verify all the added members' KeyPackage, return an error.
+            // skip if the inbox ids is 0 from the beginning
+            //todo: get the installations for all!
+            if !inbox_ids_to_add.is_empty()
+                && changes_with_kps.failed_installations.len() == inbox_ids_to_add.len()
+            {
+                return Err(GroupError::Generic(
+                    "Failed to add all installations".to_string(),
```
codabrink, line 1585: Would it be more accurate to say "Failed to verify all installations" instead? Just thinking about reading this error in future logs, and what would make it easier to comprehend.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1713,85 +1743,138 @@ fn extract_message_sender(
...
+                } else {
+                    key_packages.clone()
+                }
+            }
+
+            #[cfg(not(any(test, feature = "test-utils")))]
+            {
+                key_packages.clone()
+            }
+        };
```
codabrink, line 1805:
```suggestion
        #[cfg(any(test, feature = "test-utils"))]
        let key_packages = {
            if is_test_mode_upload_malformed_keypackage() {
                let malformed_installations = get_test_mode_malformed_installations();
                failed_installations.extend(malformed_installations.clone());

                // Return only valid key packages (excluding malformed)
                key_packages
                    .clone()
                    .into_iter()
                    .filter(|(id, _)| !malformed_installations.contains(id))
                    .collect::<HashMap<_, _>>()
            } else {
                key_packages
            }
        };
```

xmtp_mls/src/utils/test/mod.rs
```diff
@@ -336,3 +338,56 @@ pub async fn wait_for_min_intents(conn: &DbConnection, n: usize) {
...
+
+#[cfg(any(test, feature = "test-utils"))]
+/// Retrieves and decodes malformed installations from the environment variable.
+/// Returns an empty list if test mode is not enabled.
+pub fn get_test_mode_malformed_installations() -> Vec<Vec<u8>> {
+    if !is_test_mode_upload_malformed_keypackage() {
+        return Vec::new();
+    }
+
+    env::var("TEST_MODE_MALFORMED_INSTALLATIONS")
```
insipx, line 382: how do these environment variables work? What am I expected to set here

mchenani: this is only for dynamic testing purposes, see here:
https://github.com/xmtp/libxmtp/pull/1606/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3R2428

made this to manipulate the logic during the tests to simulate certain conditions.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1713,85 +1743,130 @@ fn extract_message_sender(
...
 
-        for key_package in key_packages {
-            // Add a proposal to add the member to the local proposal queue
-            new_installations.push(Installation::from_verified_key_package(&key_package));
-            new_key_packages.push(key_package.inner);
+        tracing::info!("trying to validate keypackages");
+
+        #[cfg(any(test, feature = "test-utils"))]
+        let key_packages = {
+            if is_test_mode_upload_malformed_keypackage() {
```
insipx, line 1785: i'm not super keen on adding test code directly into our business logic, it seems too easy to make a mistake or accidentally introduce a bug.


This seems like it can be mocked using [mockall](https://docs.rs/mockall/latest/mockall/) ([example](https://github.com/xmtp/libxmtp/blob/656e039222e410062e97ee29ab3f938a4122997b/xmtp_api/src/mls.rs#L395)) by setting up an ApiClientWrapper with the expected malformed key packages. Is there a blocker to using that here?

mchenani: Tried that tbh, not a fan of my current approach and I'm on the same page with you on it. 
Since only one api needs to be mocked and the rest need to work as normal apis, mixing that couldn't spend more time to figure that out. But we can sync on it maybe you have a better idea.

PR #1607, opened 2025-02-06
feat(group): add is_conversation_message_disappearing_enabled to groups

mchenani:
Add a function to get the state of disappearing messages

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -5063,11 +5063,9 @@ mod tests {
             group_from_db.unwrap().message_disappear_in_ns.unwrap(),
             disappearing_settings.in_ns
         );
-        assert!(
-            alix_group
-                .is_conversation_message_disappearing_enabled()
-                .unwrap()
-        );
+        assert!(alix_group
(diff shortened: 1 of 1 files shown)


nplasterer (approved):

bindings_ffi/src/mls.rs
```diff
@@ -1866,6 +1866,12 @@ impl FfiConversation {
         ))
     }
 
+    pub fn is_conversation_message_disappearing_enabled(&self) -> Result<bool, GenericError> {
+        let settings = self.conversation_message_disappearing_settings()?;
+
+        Ok(settings.from_ns > 0 && settings.in_ns > 0)
```
nplasterer, line 1872: @mchenani So I think this logic is wrong because what you actually care about is if they are both equal to 0 or not. If one is zero and the other is 10 then it would not be disabled

mchenani: Having both zero, doesn't make sense, bc means messages from the beginning and live 0ns! At the end we let them live 1 second! Bc the worker run on 1 second cycle.

PR #1611, opened 2025-02-07
fix(disappearing-messages): run disappearing messages cleaner worker every second

mchenani:
Add a loop to keep the disappearing messages cleaner worker running.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -1866,6 +1866,12 @@ impl FfiConversation {
         ))
     }
 
+    pub fn is_conversation_message_disappearing_enabled(&self) -> Result<bool, GenericError> {
+        let settings = self.conversation_message_disappearing_settings()?;
+
+        Ok(settings.from_ns > 0 && settings.in_ns > 0)
+    }
+
     pub fn admin_list(&self) -> Result<Vec<String>, GenericError> {
         let provider = self.inner.mls_provider()?;
         self.inner.admin_list(&provider).map_err(Into::into)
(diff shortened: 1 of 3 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -5080,10 +5080,6 @@ mod tests {
             .unwrap();
         let msg_counts_before_cleanup = alix_messages.len();
 
-        // Step 7: Start cleanup worker and delete expired messages
```
mchenani, line 5083: no need to do this, the worker is working in the background!

xmtp_mls/src/groups/disappearing_messages.rs
```diff
@@ -80,9 +91,22 @@ where
...
+        let mut interval = time::interval(INTERVAL_DURATION);
+
+        loop {
+            #[cfg(target_arch = "wasm32")]
+            interval.next().await;
+
+            #[cfg(not(target_arch = "wasm32"))]
+            interval.tick().await;
+
+            if let Err(err) = self.delete_expired_messages().await {
```
insipx, line 107: this error doesn't need to be printed here, we can use `?` and propagate it to the caller and handle it with the rest of the errors in `spawn_worker`.

xmtp_mls/src/groups/disappearing_messages.rs
```diff
@@ -80,9 +91,22 @@ where
...
+
+        #[cfg(not(target_arch = "wasm32"))]
+        let mut interval = time::interval(INTERVAL_DURATION);
+
+        loop {
+            #[cfg(target_arch = "wasm32")]
+            interval.next().await;
+
+            #[cfg(not(target_arch = "wasm32"))]
+            interval.tick().await;
```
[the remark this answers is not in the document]
codabrink: Here's the reference https://github.com/xmtp/libxmtp/blob/6a8c03a80fe062373b09818faf156b81dcc666b5/common/src/time.rs#L74

codabrink (approved):
The interval struct suggestion is just a nit. Looks good :+1:

PR #1613, opened 2025-02-07
fix broadcast group stream bug

insipx:


cameronvoell (approved):
Verified fixes failing test on android for streaming your self groups +dms
🚀

Issue #1126, opened 2024-10-09
Caching SCW Verifier for server-side use


neekolas:
Follow up to https://github.com/xmtp/libxmtp/issues/1070

One thing that would be really nice for the server is a SmartContractSignatureVerifier that caches the results for a given input in memory.

If you imagine a user getting added to a 200 person group, all 200 members are going to have to go and verify that same signature. Will be both faster and cheaper if we can cache the result.

Since the results are immutable we can theoretically cache forever, but we should give it a bounded size so we can never run out of memory getting spammed. Maybe a LRU cache.

PR #1625, opened 2025-02-12
feat(cache): add in-memory lru cache for signature verification

fbac:
Closes https://github.com/xmtp/libxmtp/issues/1126


mls_validation_service/src/config.rs
```diff
@@ -18,4 +18,8 @@ pub(crate) struct Args {
     // A path to a json file in the same format as chain_urls_default.json in the codebase.
     #[arg(long)]
     pub(crate) chain_urls: Option<String>,
+
+    // The size of the cache to use for the smart contract signature verifier.
+    #[arg(long, default_value_t = 20)]
```
codabrink, line 23: Probably want this to be more like a few thousand at least, right? How big is a cache item?

xmtp_id/Cargo.toml
```diff
@@ -18,6 +18,7 @@ ed25519-dalek = { workspace = true, features = ["digest"] }
 ethers = { workspace = true, features = ["rustls"] }
 futures.workspace = true
 hex.workspace = true
+lru = "0.13.0"
```
codabrink, line 21: Let's make this optional, and add an "lru" feature down below in this file that the validation service can turn on.
```suggestion
lru = { version = "0.13.0", optional = true }
```

codabrink: And down below under features, add
```suggestion
scw_cache = ["lru"]
```

mls_validation_service/Cargo.toml
```diff
@@ -34,9 +34,9 @@ xmtp_proto = { path = "../xmtp_proto", features = ["proto_full", "convert"] }
 anyhow.workspace = true
 ethers.workspace = true
 rand = { workspace = true }
+xmtp_common = { workspace = true, features = ["test-utils"] }
 xmtp_id = { workspace = true, features = ["test-utils"] }
```
codabrink, line 38: With the new feature we added below.
```suggestion
xmtp_id = { workspace = true, features = ["test-utils", "scw_cache"] }
```

xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -1,19 +1,21 @@
+mod cache_key;
```
codabrink, line 1:
```suggestion
#[cfg(feature = "scw_cache")]
mod cache_key;
```

xmtp_id/Cargo.toml
codabrink: On the fence about it though. Curious what others think

insipx: if this is only used in `mls_validation_service`, i think the rustier way to do this would be to prefer [composition](https://www.hackingwithrust.net/2023/10/23/a-composite-pattern-in-rust/)

Something like:

- leave xmtp_id alone
- create a new mod in mls_validation_service `scw_cache` or something
- create a new struct to contain the MultiSmartContract Verifier + the cache:
- `struct CachedScwVerifier { inner: MultiSmartContractSignatureVerifier, cache: /* */}`
- impl the caching on `CachedScwVerifier` by implementing the trait `SmartContractSignatureVerifier`on `CachedScwVerifier`

This has the nice property of being composable. We can now choose whether we want a cached version of "MultiSmartContractVerifier" or a non-cached version. Further, if we decide that there are other consumers of the cached version, we can easily copy paste  it into a more core crate like xmtp_id.

mls_validation_service/src/config.rs
fbac: 120B at minimum, if I'm correct:
```
pub struct CacheKey {
    pub address: String,
    pub chain_id: String,
    pub hash: [u8; 32],
    pub signature: Vec<u8>,
    pub block_number: Option<u64>,
}
````

On the other hand, probably it's okay if we just cache hash+signature+block_number.

File: mls_validation_service/src/cached_signature_verifier.rs [hunk 5]
```diff
@@ -0,0 +1,187 @@
...
+    cache: Mutex<LruCache<CacheKey, ValidationResponse>>,
+}
+
+impl CachedSmartContractSignatureVerifier {
+    pub fn new(
+        verifier: impl SmartContractSignatureVerifier + 'static,
+        cache_size: usize,
+    ) -> Result<Self, VerifierError> {
+        if cache_size == 0 {
+            return Err(VerifierError::InvalidCacheSize(cache_size));
```
codabrink, line 54: Rather than creating this special error case, you can change `cache_size` to be a `NonZeroUsize`

mls_validation_service/src/config.rs
codabrink: Yeah, you could probably up this to 10k at least

codabrink:
Really good so far, very close!

mls_validation_service/src/config.rs
fbac: Removed the whole logic regarding CacheKey, instead now a CacheKey is just a hash `type CacheKey = [u8; 32];`. 32 bytes gives us enough uniqueness.

Default cache size is 10000.

PR #1628, opened 2025-02-13
API prototype

insipx:
d14n API in `xmtp_api_d14n`

- add `xmtp_api_d14n`
- add d14n methods, get_inbox_ids, query_envelopes, publish_client_envelopes


xmtp_api_d14n/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "xmtp_api_v2"
```
mchenani, line 2:
```suggestion
name = "xmtp_api_d14n"
```

insipx: good catch

PR #1630, opened 2025-02-14
Create DM & Group with InboxIds WASM Node

nplasterer:
Create groups and dms with inboxIds


rygine:
resolved by #1643 #1650 and #1651

PR #1648, opened 2025-02-19
feat(d14n): d14n api wrapper

### LINKED ISSUE #1640: Port Identity from replication_client to xmtp_api_d14n

mchenani:
Port Identity and MLS APIs for d14n APIWrapper
closes #1640 #1639


xmtp_api_d14n/src/compat/d14n.rs
```diff
@@ -41,80 +48,318 @@ where
     E: std::error::Error + RetryableError + Send + Sync + 'static,
     P: Send + Sync + Client,
     C: Send + Sync + Client,
+    ApiError<E>: From<ApiError<<P as Client>::Error>>
+        + From<ApiError<<C as Client>::Error>>
+        + Send
+        + Sync
+        + 'static,
+    ApiError<E>: From<GrpcError>,
```
insipx, line 56: we can't depend on GrpcError inside this crate, since we want to use it for both grpc and http apis. You need to create a new error variant to cover the cases for this.

if the conversions are covered in replication client, they should be moved to xmtp_api_d14n, and the error type should bev replaced with something generic over http or grpc. 


This shouldn't be a big deal, since the conversions here dont depend on http or grpc, but instead are plain type conversion, which usually error on missing fields or something

mchenani: Certainly, just wanted to get rid of build issues now. Will fix it.

xmtp_api_d14n/src/compat/d14n.rs
```diff
@@ -41,80 +48,320 @@ where
...
+                                created_ns: unsigned_originator_envelope.originator_ns as u64,
+                                installation_key: request.installation_key.clone(),
+                                data: v1_welcome_message.data,
+                                hpke_public_key: v1_welcome_message.hpke_public_key,
+                            })),
+                        }));
+                    }
+                }
+
+                Some(Err(ApiError::<E>::ProtoError(ProtoError::NotFound(
```
neekolas, line 249: Do we want to halt on a single bad payload, or should we ignore it?

mchenani: Good catch, makes sense, will fix it.

xmtp_proto/src/convert.rs
```diff
@@ -47,6 +52,67 @@ impl TryFrom<UploadKeyPackageRequest> for ClientEnvelope {
     }
 }
 
+impl TryFrom<OriginatorEnvelope> for KeyPackage {
+    type Error = crate::ProtoError;
```
insipx, line 56: Probably more apt to make this a [conversion error](https://github.com/xmtp/libxmtp/blob/9f0dd2cfcfa484f95e19fd8b3ec7c3217567ba84/xmtp_proto/src/error.rs#L213) which is also defined in proto, but allows more information about lost types in the error

xmtp_api_d14n/src/compat/d14n.rs
```diff
@@ -41,80 +48,315 @@ where
...
         &self,
         request: UploadKeyPackageRequest,
     ) -> Result<(), Self::Error> {
-        todo!()
+        let envelope: ClientEnvelope = request.try_into()?;
+
+        PublishClientEnvelopes::builder()
+            .envelopes(vec![envelope])
+            .build()
+            .unwrap()
```
insipx, line 68: ideally we don't unwrap at all, but can do this in a followup or create an issue for this

insipx (approved):

PR #1651, opened 2025-02-20
Update node bindings

rygine:
# Summary

- Added new methods to create groups by inbox ID
- Added consent states option to `sync_all_conversations`
- Updated list conversations options to include `consent_states` and `include_duplicate_dms`
- Removed automatic message filtering from DM groups
- Added disappearing messages methods to conversations
- Updated conversations list methods to return conversations and their last message

### DIFF
--- a/bindings_node/src/conversation.rs
+++ b/bindings_node/src/conversation.rs
@@ -12,10 +12,7 @@ use xmtp_mls::{
     intents::PermissionUpdateType as XmtpPermissionUpdateType,
     members::PermissionLevel as XmtpPermissionLevel, MlsGroup, UpdateAdminListType,
   },
-  storage::{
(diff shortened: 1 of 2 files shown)


neekolas (approved):

rygine:
@insipx i need some help here with `ConversationListItem`. can you take a look when you get a chance?

insipx:
Seems like when something is a Class, it is owned by Javascript, so we can't operate with it directly in rust (with `napi(object)`. Turning `ConversationListItem` into a class itself works since then both classes are in Javascript. So, turned that into a class with `#[napi]` and put getters for each the fields

I also found a workaround for the UserPreferencesEnum via serialization. Fortunately napi-3.0.0 will support structured enums, so it should be temporary

PR #1656, opened 2025-02-24
Add dev client performance tests

nplasterer:


bindings_ffi/src/mls.rs
```diff
@@ -7059,4 +7089,131 @@ mod tests {
...
+                bo_client.account_address.clone(),
+                FfiCreateDMOptions::default(),
+            )
+            .await
+            .expect("DM creation should succeed");
+        let elapsed = start.elapsed();
+
+        println!("created a DM in: {}ms", elapsed.as_millis());
+        assert!(
+            elapsed < Duration::from_millis(200),
```
nplasterer, line 7111: We can increase this number to be something more reasonable but have it low for now cause this is what passes on Local.

PR #1666, opened 2025-02-25
XDBG Streaming + Stream test

insipx:
Culmination of larger scale tests I started Friday. This adds streaming + diagnostics to xdbg, allowing large-scale stream tests on xmtp networks. 

The eventual goal is to run stream tests across all versions of libxmtp as well.

There are still some issues surrounding mass sending

adds `dev/test-stream` to run the test

### DIFF
--- a/xmtp_mls/src/storage/mod.rs
+++ b/xmtp_mls/src/storage/mod.rs
@@ -76,6 +76,14 @@ pub mod test_util {
             }
         }
 
+        /// Disable sqlcipher memory security
+        pub fn disable_memory_security(&self) {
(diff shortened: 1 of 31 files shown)


xmtp_mls/src/storage/mod.rs
```diff
@@ -76,6 +76,14 @@ pub mod test_util {
             }
         }
 
+        /// Disable sqlcipher memory security
+        pub fn disable_memory_security(&self) {
```
insipx, line 80: this is only for cfg(test)/test-utils

insipx:
stale

PR #1671, opened 2025-02-26
use proto in http client

insipx:
serialize with protobuf rather than json in the http-client. this should make web a bit faster while reducing incompatibilities when serializing/deserializing http responses/requests

streams still use json for now

### DIFF
--- a/xmtp_mls/src/storage/mod.rs
+++ b/xmtp_mls/src/storage/mod.rs
@@ -76,6 +76,14 @@ pub mod test_util {
             }
         }
 
+        /// Disable sqlcipher memory security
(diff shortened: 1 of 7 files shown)


xmtp_mls/src/storage/mod.rs
```diff
@@ -76,6 +76,14 @@ pub mod test_util {
             }
         }
 
+        /// Disable sqlcipher memory security
+        pub fn disable_memory_security(&self) {
```
insipx, line 80: this is only for tests

codabrink (approved):

PR #1676, opened 2025-02-27
Use opfs-sahpool from diesel, remove sqlite-web

insipx:


insipx:
screwed up a merge re-opened in https://github.com/xmtp/libxmtp/pull/1690

PR #1694, opened 2025-03-04
legacy interop tests

codabrink:


legacy_interop_tests/.gitignore
```diff
@@ -0,0 +1,175 @@
+# Based on https://raw.githubusercontent.com/github/gitignore/main/Node.gitignore
```
insipx, line 1: should we add this to the root .gitignore?

insipx (approved):

PR #1695, opened 2025-03-04
feat(d14n): http d14n client

### LINKED ISSUE #1641: http d14n client

mchenani:
closes #1641


xmtp_api_http/src/http_client.rs
```diff
@@ -0,0 +1,59 @@
...
+        &self,
+        request: http::request::Builder,
+        body: Vec<u8>,
+    ) -> Result<http::Response<Bytes>, ApiError<Self::Error>> {
+        let request = request.body(body.clone())?;
+        let (parts, _) = request.into_parts();
+
+        let req = self
+            .http_client
+            .request(parts.method.clone(), parts.uri.to_string());
```
insipx, line 27: the method for grpc requests will always be `POST`. I think we can hard code that and not worry about it in the `Query` trait implementation

xmtp_api_d14n/src/endpoints/d14n/query_envelopes.rs
```diff
@@ -25,7 +25,7 @@ impl Endpoint for QueryEnvelope {
     type Output = QueryEnvelopesResponse;
 
     fn http_endpoint(&self) -> Cow<'static, str> {
-        todo!()
+        Cow::from("/xmtp.xmtpv4.message_api.ReplicationApi/QueryEnvelopes")
```
insipx, line 28: I think these endpoints are wrong, should be `mls/v2/query_envelopes` according to the [proto](https://github.com/xmtp/proto/blob/d1da51b50118819fbd12c31ac57a1759f27ad990/proto/xmtpv4/message_api/message_api.proto#L81) ('post' field)

xmtp_api_http/src/http_client.rs
insipx, line 27: We need to build the URI from the [host_url](https://github.com/xmtp/libxmtp/blob/6219b8aaa3fcc05c4e4753e282a46f7522495299/xmtp_api_http/src/lib.rs#L49) defined on the `Client`. The full URI should be something like `host_url + parts.uri.to_string()`. The request builder only contains the endpoint path w/o the base path so far

There are probably also combinators on the [Uri](https://docs.rs/http/latest/http/uri/struct.Uri.html) type that could help you out here, and you could directly set the `host`

xmtp_api_http/src/http_client.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        let response = req
+            .body(body)
```
insipx, line 30: we also have headers on the request that were set in the `Query` portion we want to make sure to send. Since most of the request is already crafted, you could also use the [execute](https://docs.rs/reqwest/latest/reqwest/struct.Client.html#method.execute) method after setting the body on the Request, but both ways work

insipx:
Looks good so far! Left a bunch of comments with some things that should be fixed.

It could be worth setting up just one test with one of the v3 endpoints against an HTTP client just to verify the client is working as expected (sends a request, gets protobuf encoded, receives a response, and properly deserializes from protobuf)

insipx (changes requested):

insipx:
I think you need to run `cargo update` at the root and commit that lockfile

xmtp_api_http/src/http_client.rs
```diff
@@ -0,0 +1,67 @@
...
+impl Client for XmtpHttpApiClient {
+    type Error = HttpClientError;
+    type Stream = Pin<Box<dyn futures::Stream<Item = Result<Bytes, HttpClientError>> + Send>>;
+    async fn request(
+        &self,
+        request: http::request::Builder,
+        body: Vec<u8>,
+    ) -> Result<http::Response<Bytes>, ApiError<Self::Error>> {
+        let request = request.body(body.clone())?;
+        println!("## {:?}", request);
```
insipx, line 24: should remove these printlns

insipx:
Looks good! Just need to remove the printlns in the http_client

xmtp_api_http/src/http_client.rs
mchenani: Good catch. Forgot them.

PR #1707, opened 2025-03-05
Correctly set shouldPush for messages

nplasterer:
We need the shouldPush method passed through on messages so that we correctly hide and show messages from the server.


xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1708,6 +1712,7 @@ where
                 version: Some(GroupMessageInputVersion::V1(GroupMessageInputV1 {
                     data: payload.to_vec(),
                     sender_hmac: sender_hmac.into_bytes().to_vec(),
+                    should_push: Some(false),
```
nplasterer, line 1715: Not really sure how to get the should push in this case? Or what case this is...

xmtp_mls/src/groups/mod.rs
```diff
@@ -376,6 +378,7 @@ impl TryFrom<EncodedContent> for QueryableContentFields {
             version_minor: content_type_id.version_minor as i32,
             authority_id: content_type_id.authority_id.to_string(),
             reference_id,
+            should_push: content.should_push(), // TODO I think this needs to come from the protos?
```
nplasterer, line 381: @cameronvoell almost feels like I need to set should push on the encodedContent proto as well? Or I guess how might I be able to get this from the content type in this case 🤔

cameronvoell: my recommended way to do this:

https://github.com/xmtp/libxmtp/pull/1715

if we set it in encoded content we would need to add this shouldPush mapping to each SDK, think this is a good case for centralizing in libxmtp and making it easier to read, and update, all in one place.

xmtp_mls/src/groups/mls_sync.rs
nplasterer: @codabrink do you know what case this is?

xmtp_mls/src/groups/mod.rs
nplasterer: This means all custom content types will be shouldPush false by default. But we are okay with this trade off for now.

nplasterer: In the future we will write some code to allow getting the shouldPush state from the encoded content

xmtp_mls/src/groups/mls_sync.rs
codabrink: I think we should put should_push in the group_intent table rather than the group_messages.

xmtp_mls/src/groups/mod.rs
neekolas: What would be the point of including `should_push` on the `EncodedContent`? The data in there is all encrypted, so the push server will never see it.

cameronvoell: > What would be the point of including should_push on the EncodedContent? The data in there is all encrypted, so the push server will never see it.

@neekolas we just need a convenient place to put a `shouldPush` boolean for custom content types in the SDKs that can be read by libxmtp. we're already decoding to encoded content type for messages in libxmtp, so I think it would work fine. we'd read it out of encoded content in libxmtp and then set it on the unencrypted api message.

neekolas: A little awkward (we're storing the same data twice), but I get it. In a perfect world we wouldn't store it twice on the wire but I guess an extra byte per message (Protobuf stores booleans as an 8 bit number) isn't the end of the world.

cameronvoell (approved):
this is looking good from my perspective. if we can test it against xmtp-android and confirmed push notifications are going off as expected even better.

adding shouldPush configuration for custom content types via encoded content, or some new parameter that follows messages from SDK to libxmtp sounds like it can be added a bit later 👍

PR #1708, opened 2025-03-05
Pause groups when min libxmtp version exceeds current

cameronvoell:
See https://github.com/xmtp/XIPs/pull/99


nplasterer (approved):
Looks good to me. There is a integrator communication component of this we will need to push. We need to surface that they will need to catch this error and respond appropriately.

bindings_ffi/src/mls.rs
```diff
@@ -7171,3 +7177,14 @@ mod tests {
         }
     }
 }
+
+#[napi]
```
nplasterer, line 7181: Think this was just a bad copy paste

cameronvoell: fixed, thanks

insipx:
As far as the lint for wasm goes, you just need to add the `version_info` function tot he wasm version of the trait [here](https://github.com/xmtp/libxmtp/blob/74a00e5e0290ffbec4af217ec270876b7726f6e2/xmtp_mls/src/groups/scoped_client.rs#L112)

insipx (approved):

PR #1716, opened 2025-03-05
Passkey Signature Validation

codabrink:


bindings_node/src/identity.rs
```diff
@@ -8,7 +8,7 @@ use xmtp_id::associations::{ident, Identifier as XmtpIdentifier};
 pub struct Identifier {
   pub identifier: String,
   pub identifier_kind: IdentifierKind,
-  pub relying_partner: Option<String>,
+  pub relying_party: Option<String>,
```
neekolas, line 11: 👍

xmtp_id/src/associations/verified_signature.rs
```diff
@@ -100,6 +107,53 @@ impl VerifiedSignature {
...
+        let client_data_hash = hasher.finalize();
+
+        // 4. Construct the verification data (authenticator_data + client_data_hash)
+        let mut verification_data =
+            Vec::with_capacity(authenticator_data.len() + client_data_hash.len());
+        verification_data.extend_from_slice(authenticator_data);
+        verification_data.extend_from_slice(&client_data_hash);
+
+        // 5. Verify the signature
+        verifying_key.verify(&verification_data, &signature)?;
```
neekolas, line 144: With all of our other signature verifications we use the verification text that was generated from the `SignatureRequest`.

It seems like the caller can pass in whatever `authenticator_data` and `client_data_json` it wants.

What is stopping someone from taking a valid passkey signature of some _totally unrelated_ payload (or another user's Identity Update) and passing it off as their own?

It seems like we need to verify that the `client_data_json` matches the expected verification text or re-generate our own `client_data_json` based on the `SignatureRequest` and have that passed in here.

xmtp_id/src/associations/unverified.rs
```diff
@@ -278,6 +279,12 @@ impl UnverifiedSignature {
                 &sig.legacy_key_signature.signature_bytes,
                 sig.signed_public_key_proto.clone(),
             ),
+            UnverifiedSignature::Passkey(sig) => VerifiedSignature::from_passkey(
```
neekolas, line 282: This needs to pass in the `signature_text` and use it to verify that the signer was actually signing the right thing.

codabrink: Done

xmtp_id/src/associations/verified_signature.rs
codabrink: Got it.

xmtp_id/src/associations/verified_signature.rs
```diff
@@ -100,6 +110,59 @@ impl VerifiedSignature {
...
+        // 2. Parse the signature
+        let signature = match Signature::from_der(signature) {
+            Ok(sig) => sig,
+            Err(_) => return Err(SignatureError::Invalid),
+        };
+
+        // 3. Hash the client data
+        let mut hasher = Sha256::new();
+        hasher.update(client_data_json);
+        let client_data_hash = hasher.finalize();
```
neekolas, line 144: We don't have some utility function to make this a one-liner?

codabrink: We have a function that will output a string https://github.com/xmtp/libxmtp/blob/04b6bb8197d05b079a26f79128fb0113bdc3cac3/xmtp_id/src/associations/member.rs#L387-L391

I can make one.

xmtp_id/src/associations/verified_signature.rs
```diff
@@ -100,6 +110,59 @@ impl VerifiedSignature {
...
+        }
+
+        // 1. Parse the public key from raw bytes
+        let verifying_key = match VerifyingKey::from_sec1_bytes(public_key) {
+            Ok(key) => key,
+            Err(_err) => return Err(SignatureError::InvalidPublicKey),
+        };
+
+        // 2. Parse the signature
+        let signature = match Signature::from_der(signature) {
```
neekolas, line 136: This would be nicer/more readable as `let signature = Signature::from_der(signature).ok_or(SignatureError::InvalidPublicKey)?`

neekolas (approved):

PR #1729, opened 2025-03-10
[d14n] http/grpc test consolidation

insipx:
GRPC d14n endpoints + http confirmed working with grpc-gateway


xmtp_api_d14n/src/compat/d14n.rs
```diff
@@ -165,11 +158,6 @@ where
         &self,
         request: QueryGroupMessagesRequest,
     ) -> Result<QueryGroupMessagesResponse, Self::Error> {
-        self.message_client
```
insipx, line 168: going to remove stats for d14n for now, i think there's a way to do it directly in the client since we have endpoint information in Query and Client request

codabrink: Yeah, that seems fine.

codabrink (approved):

neekolas (approved):

PR #1737, opened 2025-03-12
Update mls-validator-service registration flow

codabrink:


xmtp_id/src/associations/association_log.rs
```diff
@@ -87,7 +87,7 @@ impl IdentityAction for CreateInbox {
         }
 
         allowed_signature_for_kind(
-            &MemberKind::Ethereum,
+            &self.account_identifier.kind(),
```
codabrink, line 90: The fix

bindings_ffi/src/mls.rs
```diff
@@ -3326,7 +3326,68 @@ mod tests {
 
         // TODO: I'll uncomment this as soon as PR 1733 goes live.
         // Tested locally with this uncommented and it works.
-        // alex.apply_signature_request(sig_request).await.unwrap();
+        alex.apply_signature_request(sig_request).await.unwrap();
```
insipx, line 3329: should this be commented?

PR #1744, opened 2025-03-13
Api Stats

codabrink:
Lay the groundwork to create a test to ensure that we don't accidentally DDOS ourselves in a future release.
When the app is not doing anything, it should not be sending messages in a loop.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -3173,7 +3173,6 @@ mod tests {
 
         // One identity update pushed. Zero interaction with groups.
         assert_eq!(ident_stats.publish_identity_update.get_count(), 1);
-        // Why is this 2?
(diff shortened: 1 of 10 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -3146,6 +3147,47 @@ mod tests {
 
     use xmtp_cryptography::utils::generate_local_wallet;
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn radio_silence() {
+        let alex = new_test_client().await;
```
insipx, line 3152: itwould be nice to put this test closer to the code in `xmtp_api` or `xmtp_mls,` so it can run in wasm too

insipx (approved):

PR #1756, opened 2025-03-17
Sync Stitched DMs

codabrink:
When syncing a DM, the DM you're syncing may not be the DM that the other address is sending from. This is an issue if both DMs are created at the same time.

**The fix:** This PR will also sync the hidden "stitched" DMs (if any) on an MLS group sync.

### DIFF
--- a/xmtp_mls/src/groups/mls_sync.rs
+++ b/xmtp_mls/src/groups/mls_sync.rs
@@ -8,11 +8,11 @@ use super::{
     validated_commit::{extract_group_membership, CommitValidationError, LibXMTPVersion},
     GroupError, HmacKey, MlsGroup, ScopedGroupClient,
 };
+use crate::configuration::sync_update_installations_interval_ns;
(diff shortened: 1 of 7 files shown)


xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -199,8 +201,18 @@ where
             self.client.inbox_id(),
             epoch
         );
-        self.maybe_update_installations(&mls_provider, None).await?;
 
+        // Also sync the "stitched DMs", if any...
+        for other_dm in conn.other_dms(&self.group_id)? {
+            let other_dm =
+                Self::new_from_arc(self.client.clone(), other_dm.id, other_dm.created_at_ns);
```
codabrink, line 208: I don't think this should cause an issue with @mchenani 's global MLS group lock.

cameronvoell (approved):
nice test 👍

PR #1760, opened 2025-03-17
DM Stitching - Tweak triggers, stitch direct group id find

codabrink:
- Changed the DM stitching trigger to respect the message's `sent_at_ns` rather than inserted at time to behave more predictably in tests.
- Stitch group fetch by id


xmtp_mls/src/storage/encrypted_store/group.rs
```diff
@@ -87,7 +86,7 @@ impl StoredGroup {
             welcome_id: Some(welcome_id),
             rotated_at_ns: 0,
             dm_id: dm_members.map(String::from),
-            last_message_ns: Some(now_ns()),
+            last_message_ns: None,
```
codabrink, line 89: Most of our problems came from this :facepalm:

nplasterer (approved):

PR #1767, opened 2025-03-18
Device Sync V2

codabrink:
### Device Sync V2
- Syncing with V1 clients will be supported until the next hammer release in device_sync_legacy.rs
- Hardened syncing by giving the authority of sync event ordering over to the group
- Removed Content::V2 for syncing, and reverted back to Content::V1 for simplicity and maintainability
- Sync groups are now created even when the sync server url is not present.
  - This means that old installations will add new devices to existing groups much sooner.
  - This also means that consent and hmac keys will sync without needing a server url to be present.
- Removed sync group syncing from sync_all_conversations and moved it into it's own function.
- Worker analytics are now collected in the worker handle.
- Added a "mode" option to the sync worker, allowing the worker to be turned off when creating a client.
- Added a sync cursor column to the database to track what sync messages have been processed.


bindings_ffi/src/mls.rs
```diff
@@ -120,7 +123,7 @@ pub async fn create_client(
     account_identifier: FfiIdentifier,
     nonce: u64,
     legacy_signed_private_key_proto: Option<Vec<u8>>,
-    history_sync_url: Option<String>,
+    device_sync_server_url: Option<String>,
```
nplasterer, line 126: We probably don't need to change the low level but on the SDK side and the bindings we should call this `history_transfer_server_url` can be done in a follow up but just noting.

bindings_ffi/src/mls.rs
```diff
@@ -550,10 +550,10 @@ impl FfiXmtpClient {
     }
 
     /// Manually trigger a device sync request to sync records from another active device on this account.
-    pub async fn send_sync_request(&self, kind: FfiDeviceSyncKind) -> Result<(), GenericError> {
+    pub async fn send_sync_request(&self) -> Result<(), GenericError> {
```
nplasterer, line 553: Per my document early I think we need 4 different functions here
- send_history_transfer_request - (sends a request to get history)
- approve_history_transfer_request - (approves a request to give history and sends the history)
- fetch_history - (checks to see if history is present to receive and syncs if it is)
- sync_state - (syncs the sync group for any new consent or hmac records *does not do anything with history transfer requests if called)

nplasterer (commented):
I know this PR is very large so maybe this is more of a history sync refactor and then in a follow up PR you can address the items for history sync v2 which would include the things outlined in this document https://www.notion.so/ephemerahq/History-Sync-V2-1ca30823ce928090b8fbdfb0ae283b2c?showMoveTo=true&saveParent=true
I just want to make sure we are progressing forward here and not just preserving a confusing flow.

bindings_ffi/src/mls.rs
codabrink: Sounds great, let's address those specifically in a follow-up PR! Should be quick and easy.

bindings_ffi/src/mls.rs
```diff
@@ -2989,49 +2995,80 @@ mod tests {
...
                     .unwrap(),
             )
             .await?;
+
+        client.register_identity(signature_request).await?;
+
+        Ok(())
+    }
+
+    async fn register_client_with_passkey(passkey: &PasskeyUser, client: &FfiXmtpClient) {
```
insipx, line 3023: what's the advantage of the `no_panic` version?

Also the convention for this is to name the function `register_client_with_passkey_unchecked`

xmtp_db/src/traits.rs
```diff
@@ -2,12 +2,14 @@ use crate::StorageError;
...
 pub trait Store<StorageConnection> {
-    fn store(&self, into: &StorageConnection) -> Result<(), StorageError>;
+    type Output;
+    fn store(&self, into: &StorageConnection) -> Result<Self::Output, StorageError>;
 }
 
 /// Inserts a model to the underlying data store, silent no-op on unique constraint violations
 pub trait StoreOrIgnore<StorageConnection> {
-    fn store_or_ignore(&self, into: &StorageConnection) -> Result<(), StorageError>;
+    type Output;
```
insipx, line 11: nice!

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -204,111 +174,105 @@ where
...
 
         while let Some(event) = self.stream.next().await {
             let event = event?;
-            match event {
-                LocalEvents::SyncMessage(msg) => match msg {
-                    SyncMessage::Reply { message_id } => {
+
+            if let LocalEvents::SyncEvent(msg) = event {
+                match msg {
+                    SyncEvent::NewSyncGroupFromWelcome => {
```
insipx, line 184: matching & doing the work all in one function just creates a really long function. WDYT about splitting into parts

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -365,16 +323,17 @@ where
...
-            handle: std::sync::Arc::new(Default::default()),
+            handle: Arc::new(WorkerHandle::new()),
         }
     }
 
     fn spawn_worker(mut self) {
         xmtp_common::spawn(None, async move {
             let inbox_id = self.client.inbox_id().to_string();
             let installation_id = hex::encode(self.client.installation_public_key());
+            // let span = info_span!("\x1b[35mDevice sync: ");
```
insipx, line 334:
```suggestion
```

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -365,16 +323,17 @@ where
...
+            // let _guard = span.enter();
```
insipx, line 335:
```suggestion
```

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -399,540 +358,514 @@ where
...
-
-    // encrypt the payload and append to the result
-    result.append(&mut cipher.encrypt(nonce_array, &*payload)?);
-
-    Ok((result, enc_key))
+impl ZipContent<StoredGroupMessage, DeviceSyncContent> for Vec<StoredGroupMessage> {
+    fn iter_with_content(self) -> impl Iterator<Item = (StoredGroupMessage, DeviceSyncContent)> {
+        self.into_iter().filter_map(|msg| {
+            let content: DeviceSyncContent =
+                serde_json::from_slice(&msg.decrypted_message_bytes).ok()?;
```
insipx, line 867: using `.ok()` here could create a situation where we're de-serializing something incorrectly but ignoring the item in an iterator, resulting in something looking like we're missing a message, which could be difficult do debug. We should prefer using inspect() and at the very least trace/debug logging a warning that something failed to deserialize, rather then ignore the option

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -399,540 +358,514 @@ where
...
+    SyncGroupPresence,
+    Request { request_id: String },
 }
 
-fn encrypt_syncables(
-    syncables: &[Vec<Syncable>],
-) -> Result<(Vec<u8>, DeviceSyncKeyType), DeviceSyncError> {
-    let enc_key = DeviceSyncKeyType::new_aes_256_gcm_key();
-    encrypt_syncables_with_key(syncables, enc_key)
+pub trait ZipContent<A, B> {
```
insipx, line 859: how come we can't use [zip](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.zip)

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -399,540 +358,514 @@ where
...
+        let body = {
+            // 2. A compat layer to have futures AsyncRead play nice with tokio's AsyncRead
+            let exporter_compat = tokio_util::compat::FuturesAsyncReadCompatExt::compat(exporter);
+            // 3. Add a stream layer over the async read
+            let stream = tokio_util::io::ReaderStream::new(exporter_compat);
+            // 4. Pipe that stream as the body to the request to the history server
+            reqwest::Body::wrap_stream(stream)
+        };
+        #[cfg(target_arch = "wasm32")]
+        let body = {
```
insipx, line 612: This is probably fine for now, but it would be nice to separate platform-specific code out into traits, for instance a `DeviceSyncBody` trait that can create a Wasm body or a native body. that way as our experience & understanding of WebAssembly improves, we can modify the body as needed w/o modifying the device sync code. For instance, i could see us improving this code by using web_sys apis to get something resembling a stream in the future. 

its also nice not to worry about #[cfg] in code itself which makes it a bit less readable imo

a native vs wasm request body could also be useful elsewhere, like to other potential workers or something

would be a good followup for the libxmtp board

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -399,540 +358,514 @@ where
...
+            // Make exporter mutable
+            let mut exporter = exporter;
+
+            // Wasm does not support stream uploads. So we'll just consume the stream into a vec.
+            let mut buffer = Vec::new();
+            exporter.read_to_end(&mut buffer).await?;
+            buffer
+        };
+
+        // 5. Make the request
```
insipx, line 623: these steps are really great. Wondering if it would make sense to split it off into its own Struct type with each step as a function

something like

```
let send = DeviceSyncSend::new(/* .. */);
send.build()
send.stream()
send.request()
send.reply()
```

etc.

This would be similar to modeling as state machine, and has a few benefits:
- easier to test each individual step
- observe state before/after each step for logging etc.
- readability (each function would do only one thing)
- separation of concerns/concepts

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -399,540 +358,514 @@ where
...
+        self.send_device_sync_message(&provider, content, retry)
+            .await?;
+
+        handle.increment_metric(SyncMetric::PayloadSent);
+
+        Ok(())
     }
 
-    async fn insert_encrypted_syncables(
+    fn did_this_installation_ask_for_this_reply(
```
insipx, line 669: IMO 'this' is a bit ambigious at first glance, i would personally prefer something like `is_reply_requested_by_installation` or simply `is_reply_requested` since a single installation should be implied by the nature of being part of `Client`. that's just a preference though so minor nit/feel free to ignore

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -399,540 +358,514 @@ where
...
-        tracing::error!(
-            "Failed to download file. Status code: {} Response: {:?}",
-            response.status(),
-            response
-        );
-        response.error_for_status()?;
-        unreachable!("Checked for error");
-    }
+        #[cfg(not(target_arch = "wasm32"))]
+        let reader = {
```
insipx, line 730: same wasm/native trait could be here for future ref/ticket

xmtp_mls/src/client/test_utils.rs
```diff
@@ -0,0 +1,34 @@
...
+// Please ensure that all public functions defined in this module
+// start with `test_`
+impl<ApiClient, V> Client<ApiClient, V>
+where
+    ApiClient: XmtpApi,
+    V: SmartContractSignatureVerifier,
+{
+    /// Creates a DM with the other client, sends a message, and ensures delivery,
+    /// returning the created dm and sent message contents
+    pub(crate) async fn test_talk_in_dm_with(
```
insipx, line 12: very nice, love all the testing improvements

xmtp_mls/src/groups/device_sync/backup.rs
```diff
@@ -156,115 +132,111 @@ mod tests {
         assert_eq!(messages.len(), 1);
     }
 
-    #[tokio::test]
+    #[xmtp_common::test(unwrap_try = "true")]
```
insipx, line 135: wasn't sure how i felt about this, since i'm a bit of a purist and transforming a syntax/language feature in-place just feels wrong, but it being opt-in makes alot of sense IMO and i'm definitely for it if it increases test readability

xmtp_mls/src/groups/test_utils.rs
```diff
@@ -0,0 +1,20 @@
+use super::{scoped_client::ScopedGroupClient, GroupError, MlsGroup};
+use xmtp_db::group_message::MsgQueryArgs;
+
+impl<Client: ScopedGroupClient> MlsGroup<Client> {
```
insipx, line 4: `maybe` we can have a test/ or put these alongside the existing test_utils somewhere folder or something with the rest of `test_` utilities so they're easier to find, wdyt?

xmtp_mls/src/groups/device_sync.rs
codabrink: All for it, this was bothering me a bit too

xmtp_mls/src/groups/device_sync.rs
codabrink: For sure. Makes much more sense to start with `is_` when it returns a bool.

bindings_ffi/src/mls.rs
codabrink: This is in `mod tests`

xmtp_db/src/encrypted_store/user_preferences.rs
```diff
@@ -1,38 +1,52 @@
...
+)]
 #[diesel(table_name = user_preferences)]
 #[diesel(primary_key(id))]
 pub struct StoredUserPreferences {
-    /// Primary key - latest key is the "current" preference
     pub id: i32,
     /// Randomly generated hmac key root
     pub hmac_key: Option<Vec<u8>>,
+    // Sync cursor: sync_group_id:cursor
+    pub sync_cursor: Option<String>,
```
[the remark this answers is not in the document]
codabrink: I think it's better to keep them together. The format is extremely simple; formatted as `group_id:msg_offset`, and the two pieces of data are so tightly bound, the one column approach makes it hard to mistakenly update one column without the other, which is what I was more concerned about when I wrote it.

xmtp_mls/src/groups/device_sync.rs
codabrink: Yeah, this should be logged.

xmtp_mls/src/groups/test_utils.rs
codabrink: I think I'd be fine with that. Though I still think we should probably keep them in separate files. We have a bad habit of lumping a bunch of separate concerns into huge files here, and I'd like to start splitting them up into hierarchies lol

I'll move the files.

xmtp_mls/src/groups/device_sync.rs
codabrink: I don't think using zip would be an improvement. I could create a new iter that borrowed the data structure we're zipping with, but for what benefit? It's possible, but it feels like we'd be using zip for the sake of using zip. I can rename the trait to be `IterWithContent`

xmtp_db/src/encrypted_store/user_preferences.rs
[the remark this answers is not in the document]
codabrink: Addressed.

xmtp_db/src/encrypted_store/user_preferences.rs
```diff
@@ -43,34 +57,82 @@ impl Store<DbConnection> for StoredUserPreferences {
...
+        let cursor = SyncCursor::load(&sync_cursor);
+        let cursor = cursor
+            .and_then(|c| (c.group_id == group_id).then_some(c))
+            .unwrap_or_else(default);
+        Ok(cursor)
+    }
+
+    pub fn store_sync_cursor(conn: &DbConnection, cursor: &SyncCursor) -> Result<(), StorageError> {
+        let mut pref = Self::load(conn)?;
+        pref.sync_cursor = Some(format!("{cursor}"));
```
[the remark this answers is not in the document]
codabrink: Addressed

codabrink:
Moved to https://github.com/xmtp/libxmtp/pull/1877

Issue #1763, opened 2025-03-18
Dm id should update to dominate dm when stitched


nplasterer:
After calling sync and stitching a dm the dominate dms id should now become the dms id.

Issue #1762, opened 2025-03-18
Dms with no messages should still stitch correctly


nplasterer:
Even if a dm has no messages inside it. It should still stitch correctly with any other existing dms.

Issue #1784, opened 2025-03-21
When added to a new group show the group membership add as the first message


nplasterer:
It would be nice for people getting added to groups to see the group membership add as the first message so they don't see a blank group.

PR #1790, opened 2025-03-24
No message dms should still stitch

nplasterer:
Fixes https://github.com/xmtp/libxmtp/issues/1762
Fixes https://github.com/xmtp/libxmtp/issues/1763

DMs that have no messages in them should still stitch together. Selecting the group that was created last on the network not locally.
Once DMs are stitched their id should be updated in the local database.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -7782,4 +7782,46 @@ mod tests {
         assert_eq!(final_bo_messages.len(), 5, "Bo should see 5 messages");
         assert_eq!(final_alix_messages.len(), 5, "Alix should see 5 messages");
     }
+
+    #[tokio::test]
+    async fn test_can_successfully_thread_dms_with_no_messages() {
+        // Create two test users
+        let wallet_bo = generate_local_wallet();
+        let wallet_alix = generate_local_wallet();
+
+        let client_bo = new_test_client_with_wallet(wallet_bo).await;
+        let client_alix = new_test_client_with_wallet(wallet_alix).await;
+
+        let bo_provider = client_bo.inner_client.mls_provider().unwrap();
+        let bo_conn = bo_provider.conn_ref();
+        let alix_provider = client_alix.inner_client.mls_provider().unwrap();
+        let alix_conn = alix_provider.conn_ref();
+
+        // Find or create DM conversations
+        let convo_bo = client_bo
+            .conversations()
+            .find_or_create_dm_by_inbox_id(client_alix.inbox_id(), FfiCreateDMOptions::default())
+            .await
+            .unwrap();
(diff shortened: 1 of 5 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -2181,7 +2181,10 @@ impl FfiConversation {
 #[uniffi::export]
 impl FfiConversation {
     pub fn id(&self) -> Vec<u8> {
-        self.inner.group_id.clone()
+        match self.inner.client.stitched_group(&self.inner.group_id) {
+            Ok(group) => group.group_id.clone(),
+            Err(_) => self.inner.group_id.clone(),
+        }
```
nplasterer, line 2187: This fixes a lot of weirdness with dm stitching letting dms now always return the correct id after they've been synced.

bindings_ffi/src/mls.rs
```diff
@@ -7782,4 +7786,41 @@ mod tests {
         assert_eq!(final_bo_messages.len(), 5, "Bo should see 5 messages");
         assert_eq!(final_alix_messages.len(), 5, "Alix should see 5 messages");
     }
+
+    #[tokio::test]
+    async fn test_can_successfully_thread_dms_with_no_messages() {
```
nplasterer, line 7791: Here is the scenario and why this fails
Alix has a group that has 1 message
Bo has a group that has 1 message

Alix can not see bos 1 message and bo can not see alixs 1 message

when alix syncs with the network she sees bos group with no messages so now she has 2 groups 1 that is hers with 1 message and 1 that is his with no messages so it always picks hers

when bo syncs with the network he sees alixs group with no messages so now he has 2 groups 1 that is his with 1 message and 1 that is hers with no messages so it always picks his 

Because of this their dms are not stitched correctly and they always see their own dm instead of the others.

Maybe this is an edge case because the second a message gets sent in either group it will stitch correctly so as long as the groups are both empty that is when the issue occurs.

nplasterer: This would not longer be an issue if we decide to implement this request https://github.com/xmtp/libxmtp/issues/1784

nplasterer: cc @codabrink for visibility

codabrink: A possible solution could be to set the group's `last_message_ns` to the welcome's `created_ns` when creating the DM from a welcome.

bindings_ffi/src/mls.rs
nplasterer: Wanted to confirm that there was no real performance impacts for this change
Previously it .000000s
now its 0.000029s

codabrink (approved):

PR #1791, opened 2025-03-24
fix: don't delete KeyPackages if processing the welcome messages fail

### LINKED ISSUE #1786: Welcome message fails to decrypt
installation cannot decrypt the welcome
https://github.com/xmtp/libxmtp/blob/718510f5cdbb155b94c062ae7db82a149bf59298/xmtp_mls/src/hpke.rs#L62

https://github.com/xmtp/libxmtp/blob/718510f5cdbb155b94c062ae7db82a149bf59298/xmtp_mls/src/groups/mls_ext/mls_ext_welcome.rs#L24

The hash ref for the welcome is missing from my SQLite database, so i can't decrypt the welcome, and i think its actually a protocol bug. 

Right here: https://github.com/xmtp/libxmtp/blob/718510f5cdbb155b94c062ae7db82a149bf59298/xmtp_mls/src/groups/mod.rs#L648

We dont check whether the welcome decrypted successfully. if the hash ref is missing we're supposed to load identity updates/fetch key packages, but we don't so the error just loops and i can never get into the group (edited) 

Image

mchenani:
Fix not deleting the KeyPackages if processing welcome messages failed on some of the received ones.
closes #1786

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -908,7 +908,7 @@ where
             .await;
 
         // If processed groups equal to the number of envelopes we received, then delete old kps and rotate the keys
-        if num_envelopes == groups.len() {
+        if num_envelopes > 0 && num_envelopes == groups.len() {
             self.rotate_and_upload_key_package(provider).await?;
         }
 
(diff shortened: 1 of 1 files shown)


codabrink (approved):
Woop!

PR #1797, opened 2025-03-26
Reproduce invalid key package

nplasterer:
WIP

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -7807,4 +7807,35 @@ mod tests {
         assert_eq!(final_bo_messages.len(), 5, "Bo should see 5 messages");
         assert_eq!(final_alix_messages.len(), 5, "Alix should see 5 messages");
     }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 5)]
+    async fn test_can_add_installation_with_a_bad_key_package() {
+        let alix_client = new_dev_test_client().await;
+        let bo_client = new_dev_test_client().await;
+        let group = alix_client
+            .conversations()
+            .create_group_with_inbox_ids(
+                vec![
+                    "f87420435131ea1b911ad66fbe4b626b107f81955da023d049f8aef6636b8e1b".to_string(),
+                ],
+                FfiCreateGroupOptions::default(),
+            )
+            .await
+            .unwrap();
+
+        group
+            .add_members(vec![bo_client.account_identifier.clone()])
+            .await
+            .unwrap();
+        bo_client
+            .conversations()
+            .sync_all_conversations(None)
+            .await
+            .unwrap();
+        let bo_groups = bo_client
+            .conversations()
+            .list_groups(FfiListConversationsOptions::default())
+            .unwrap();
+        assert_eq!(bo_groups.len(), 1);
+    }
 }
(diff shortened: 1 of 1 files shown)


xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1870,45 +1875,11 @@ async fn calculate_membership_changes_with_keypackages<'a>(
...
-    if is_test_mode_upload_malformed_keypackage() {
-        let malformed_installations = get_test_mode_malformed_installations();
-        key_packages.retain(|id, _| !malformed_installations.contains(id));
-        failed_installations.extend(malformed_installations);
-    }
-
-    Ok(key_packages)
-}
 #[allow(unused_variables, dead_code)]
-#[cfg(not(any(test, feature = "test-utils")))]
```
neekolas, line 1907: Was this switch on purpose?

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1830,7 +1836,7 @@ async fn calculate_membership_changes_with_keypackages<'a>(
 ) -> Result<MembershipDiffWithKeyPackages, GroupError> {
     let membership_diff = old_group_membership.diff(new_group_membership);
 
-    let installation_diff = client
+    let mut installation_diff = client
```
neekolas, line 1839: Does this need to be mutable?

xmtp_mls/src/groups/mls_sync.rs
nplasterer: It looks intentional since the two functions look nearly identical now?

bindings_ffi/src/mls.rs
```diff
@@ -7787,6 +7812,37 @@ mod tests {
         assert_eq!(final_alix_messages.len(), 5, "Alix should see 5 messages");
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 5)]
+    async fn test_can_add_installation_with_a_bad_key_package() {
```
nplasterer: I think we should delete this test before we merge it

xmtp_mls/src/groups/mls_sync.rs
nplasterer: CI was broken so I brought it back

bindings_ffi/src/mls.rs
nplasterer: Before I removed this test I also ran it with the history sync server enabled and everything worked as expected. 🎉

mchenani: the test was flaky due to issue of kps order. now it should be safe to have the tests back.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1957,7 +1963,11 @@ async fn apply_update_group_membership_intent(
 
     // Update the extensions to have the new GroupMembership
     let mut new_extensions = extensions.clone();
-    new_group_membership.failed_installations = changes_with_kps.failed_installations;
+    new_group_membership.failed_installations = [
```
codabrink, line 1966: Just a high level question - are we considering the failed installations that were in this field before this line?

mchenani: yes, we're merging the new failed_installations and the old ones.

mchenani: fyi: if we have before two failed installations, and now one new failed_installation, we need to keep the track of both.

codabrink (approved):

PR #1803, opened 2025-03-28
debug logging for convos - long running actions

cameronvoell:


insipx:
there should also be create client benchmarks for ffi [here](https://github.com/xmtp/libxmtp/blob/main/bindings_ffi/benches/create_client.rs)

PR #1813, opened 2025-04-01
Add key package status to bindings

cameronvoell:


xmtp_mls/src/verified_key_package_v2.rs
```diff
@@ -64,6 +80,21 @@ impl VerifiedKeyPackageV2 {
     pub fn hpke_init_key(&self) -> Vec<u8> {
         self.inner.hpke_init_key().as_slice().to_vec()
     }
+
+    pub fn life_time(&self) -> Option<VerifiedLifetime> {
+        let lifetime_result = panic::catch_unwind(AssertUnwindSafe(|| {
+            self.inner.life_time() // This might panic
```
cameronvoell, line 86: given our uncertainty over whether lifetimes were set, and the existence of [this issue,](https://github.com/openmls/openmls/issues/1663) I added this panic catch to be safe. This way if we make lifetimes optional at some point in the future this would not panic in older versions. happy to remove if this seems excessive.

insipx: great catch

bindings_node/src/inbox_state.rs
```diff
@@ -74,4 +107,46 @@ impl Client {
...
+
+  /**
+   * Get key package statuses for a list of installation IDs.
+   *
+   * Returns a JavaScript Object mapping installation ID strings to KeyPackageStatus objects.
+   */
+  #[napi]
+  pub async fn get_key_package_statuses_for_installation_ids(
+    &self,
+    installation_ids: Vec<Uint8Array>,
```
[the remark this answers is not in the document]
cameronvoell: updated! https://github.com/xmtp/libxmtp/pull/1813/commits/14584e821f4447fc6fdbfeaed3043ff938ea9d51

rygine (approved):

PR #1816, opened 2025-04-02
fix: mark published and processed intents and processed instead of deleting them

mchenani:
### Modify group intent handling to mark intents as processed instead of deleting them to improve debugging capabilities
* Adds new `Processed` state (value=5) to `IntentState` enum in [group_intent.rs](https://github.com/xmtp/libxmtp/pull/1816/files#diff-17a09cbc9886e242f0ff0189d4750dbfd8226be9344e7b41814967745cfe0a0c) and implements `set_group_intent_processed()` method
* Updates message processing logic in [mls_sync.rs](https://github.com/xmtp/libxmtp/pull/1816/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e) to mark intents as processed rather than deleting them
* Adds comprehensive error handling tests in [mls.rs](https://github.com/xmtp/libxmtp/pull/1816/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f) and [stream_all.rs](https://github.com/xmtp/libxmtp/pull/1816/files#diff-4a3d32ed6005867d963b0eae023dc9c1fd8ffd9531413ba572f4287210a18cc1) for message streaming
* Improves error logging in [stream_messages.rs](https://github.com/xmtp/libxmtp/pull/1816/files#diff-ee8c2d1c90650351a2245875b62af09f2c152c29643200276da5743cd59424e7) with more detailed cursor and group information

#### 📍Where to Start
Start with the `IntentState` enum and `set_group_intent_processed()` method in [group_intent.rs](https://github.com/xmtp/libxmtp/pull/1816/files#diff-17a09cbc9886e242f0ff0189d4750dbfd8226be9344e7b41814967745cfe0a0c), which defines the new processed state, then move to [mls_sync.rs](https://github.com/xmtp/libxmtp/pull/1816/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e) to see how this state is used in message processing.

----

_[Macroscope](https://app.macroscope.com) summarized 165f7f4._


bindings_ffi/src/mls.rs
```diff
@@ -5391,6 +5391,40 @@ mod tests {
         assert!(stream.is_closed());
     }
 
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 5)]
+    async fn test_stream_all_messages_error() {
```
insipx, line 5396: if you write the tests in the style of tests in https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/src/subscriptions/stream_all.rs, does it still reproduce?

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1381,8 +1396,7 @@ where
                             installation_id = %self.client.installation_id(),
                             "Skipping intent because no publish data returned"
                         );
-                        let deleter: &dyn Delete<StoredGroupIntent, Key = i32> = provider.conn_ref();
-                        deleter.delete(intent.id)?;
+                        provider.conn_ref().set_group_intent_processed(intent.id)?
```
neekolas, line 1399: The one thing we should keep in mind is that the old workflow was idempotent, and the new one will return an error if the intent was already deleted.

How confident are we that this won't cause issues with concurrency/retries?

mchenani: Great question, as far as I checked the code we have two places that we get the intents to process them and we already handle the processed case there.

neekolas (approved):

cameronvoell (approved):

PR #1817, opened 2025-04-02
file appender

insipx:
introduces `enter_debug_writer` and `exit_debug_writer` to uniffi bindings.

Calling `enter_debug_writer` with a directory path will create a file with a timestamp in the format `libxmtp.log.2025-04-02` in the directory specified. that directory will hold a max of 10 log files.

`exit_debug_writer` flushes the loglines to the file and shuts down the log writer thread. this should be done before the program exits to ensure all loglines are present in the file on-disk

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7822,6 +7822,7 @@ dependencies = [
  "thiserror 2.0.11",
  "tokio",
  "tracing",
+ "tracing-appender",
  "tracing-oslog",
  "tracing-subscriber",
  "tracing_android_trace",
--- a/bindings_ffi/Cargo.toml
+++ b/bindings_ffi/Cargo.toml
@@ -15,6 +15,7 @@ prost.workspace = true
 thiserror.workspace = true
 tokio = { workspace = true, features = ["macros"] }
 tracing.workspace = true
+tracing-appender = "0.2"
 tracing-subscriber = { workspace = true, features = [
   "registry",
   "env-filter",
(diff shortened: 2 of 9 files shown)


cameronvoell:
@insipx think the datetime suffix only gets appended if you set a rotation policy. See https://github.com/xmtp/libxmtp/pull/1819/files#diff-4bbabc172ad0589d5364e814aa0311a1792f06663618c2bbb82fe06fe62e418aR306

I added some options for passing in rotation policy, and log level let e know if you have any feedback in the [PR here](https://github.com/xmtp/libxmtp/pull/1819). Testing on Android worked pretty well. One thing I noticed is that when you kill the app, you do need to re-call `enter_debug_writer` on next app start up to get the appender going again. I'm thinking I'll add to ios/android sdks the ability to save whether the writer should be active and if so call `enter_debug_writer` before Clients are built. Probably some helper functions for returning the list of log files, and clearing the log files. 

Just being able to have this finer control and to have log files that stretch through multiple client build actions is awesome though 🎉

insipx:
While working on this yesterday, (mostly b/c i don't love how everything with this implementation is global) I started wondering whether a better design would be to keep a tracing instance per-client, and somehow scope logs to belong to only that client. I think this is possible to too for tracing, and could be done in a future iteration.

I would also like to port this kind of logging to node, since this sort of log file rotation seems very nice for long-running bots too

PR #1820, opened 2025-04-03
dynamic OPFS capacity, better OPFS error handling

insipx:
### Make database initialization asynchronous and improve OPFS error handling in WebAssembly environments
* Converts all database initialization operations to be asynchronous across native and WebAssembly targets in [encrypted_store/mod.rs](https://github.com/xmtp/libxmtp/pull/1820/files#diff-7c11a42d3a8d1e571eea6a833dfe388063c036317c4d7a6fbce207fbb3beb387)
* Implements dynamic OPFS pool resizing and enhanced error handling for WebAssembly storage in [encrypted_store/wasm.rs](https://github.com/xmtp/libxmtp/pull/1820/files#diff-19c6744af0481bdbb202f3e10379f94040d4d9348d5f2397ed79047f7658117a)
* Updates all database connection calls with `.await` across bindings, examples, and test files
* Adds WebAssembly-specific configuration constants and error types in [configuration.rs](https://github.com/xmtp/libxmtp/pull/1820/files#diff-7b56ec5258ef9c988a6390a3387f34819eb83f404965daea96d59fd7074fce8d) and [errors.rs](https://github.com/xmtp/libxmtp/pull/1820/files#diff-c28554f6385bb1e08c2ed9656a3618cd1f29d58e5b8dc8729b083e84078261c8)
* Introduces builder pattern for `NewGroupIntent` struct in [group_intent.rs](https://github.com/xmtp/libxmtp/pull/1820/files#diff-17a09cbc9886e242f0ff0189d4750dbfd8226be9344e7b41814967745cfe0a0c)

#### 📍Where to Start
Start with the core changes in [encrypted_store/mod.rs](https://github.com/xmtp/libxmtp/pull/1820/files#diff-7c11a42d3a8d1e571eea6a833dfe388063c036317c4d7a6fbce207fbb3beb387) where the database initialization methods `new`, `new_unencrypted`, and `new_database` are made asynchronous, as this is the foundation for all other changes.

----

_[Macroscope](https://app.macroscope.com) summarized f512974._

PR #1823, opened 2025-04-03
Patch release for dynamic OPFS capacity, better OPFS error handling

nplasterer:
Pulls in the https://github.com/xmtp/libxmtp/pull/1820


nplasterer:
Thinking this might not be a good fit for a patch release. Going to move this to a 1.2 dev release instead

PR #1839, opened 2025-04-09
replace cooldown with governor rate limiting

insipx:
The rate limit in `retry.rs` worked but it makes sense to leave ratelimiting to the lower level client implementation. at the time, i also didn't realize tonic had rate-limiting built-in. This uses governor for reqwest/wasm ratelimiting. 

I ignored the rate limiting test (since it takes a minimum of a minute) but verified the limits work locally

rate limiting at the higher-level was making it very difficult to work with error types between d14n and v3, since all the error types would need to be the same. The motive for this was to make it easier to handle those types. This rate limiting implementation is also probably way more robust than whatever was there before

the default is 1900/minute, or roughly 9500/5min


insipx:
* **#1839** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1839?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/1839?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xmtp_api_http/src/lib.rs
```diff
@@ -57,26 +66,16 @@ pub struct XmtpHttpApiClient {
...
-            http_client: client,
-            host_url,
-            app_version,
-            libxmtp_version,
-            stats: ApiStats::default(),
-            identity_stats: IdentityStats::default(),
-        })
+    /// Wait for any rate limit
+    async fn wait_for_ready(&self) {
+        let jitter = Jitter::up_to(Duration::from_secs(5));
```
neekolas, line 77: What are the impacts of the jitter here?

neekolas (approved):

xmtp_api_http/src/lib.rs
insipx: from the docs:

```
When polled, the returned future either resolves immediately (in the case where the rate limiter allows it), or else triggers an asynchronous delay, after which the rate limiter is polled again. This means that the future might resolve at some later time (depending on what other measurements are made on the rate limiter).

This method allows for a randomized additional delay between polls of the rate limiter, which can help reduce the likelihood of thundering herd effects if multiple tasks try to wait on the same rate limiter.
```

so it seems like another randomized value of up to 5 seconds is added to the delay, which is one minute

neekolas: I guess what I'm wondering is if it can add a delay to calls when you aren't already rate limited

insipx: ah good catch, i didn't know but checking the implementation [here](https://docs.rs/governor/latest/src/governor/state/direct/future.rs.html#51) seems like the delay is only added if the limiter check fails/we are being rate limited

PR #1854, opened 2025-04-11
refactor GroupIntent Creation with Trait

insipx:
**This PR is apart of a stack of PRS to add is_commit to d14n payloads**

### Context

d14n needs to know whether each message we're sending out to the network is a commit or not. We could deserialize the data again, but this is redundant and we already deserialize/serialize alot. Instead, the plan is to:
- encapsulate GroupIntent Creation
- add additional fields to PublishIntent w/o serialization
- introduce a Message type that describes things about a message (is_commit etc.), and also how to get a GroupMessageInput from it
- Use this generic message type instead of the serialized protobuf types.

### This PR
 creates a `GroupIntent` trait and encapsulated the logic for creating publish data from an intent in a separate file. This has the benefit of reducing the cognitive load needed to parse the monolithic `mls_sync` file. This will setup up nicely to store more information on `PublishIntentData` that we can pass around, as well as provide a formalized framework for creating Group Intent types.

- `GroupIntent`: Group intent actions, includes a method for getting `PublishIntentData` from an Intent, as well as building the extensions for that intent.
- moves some standalone functions to a `MlsGroupExt` and `MlsExtensionExt` trait implemented on an OpenMlsGroup & Extension from the `openmls` crate
    - these are methods that build on the mls types but add xmtp-specific things like group membership/inbox ids/etc.

### Future PR/s (max 2) in the stack:

- change the `PublishIntentData` to not serialize the commit/application payload immediately, rather store the type on the struct
- `Message`: A generic message type which describes how to 'prepare' the message into a payload/protocol envelope
- a generic `impl Message`, `Box<dyn Message>` or `E: Message` can be passed around, eventually landing in `send_message`
- the function signature of `send_messages` in xmtp_api will change to accommodate a type which implements `Message`
    - the `Message` type must implement a function to produce the protobuf `GroupMessageInput`
    - the message type must provide a way to 'prepare' the message
        - we prepare messages currently:
            - [with hmacs](https://github.com/xmtp/libxmtp/blob/3120239d01ae383e9e52fc2baf9c2fb502342274/xmtp_mls/src/groups/mls_sync.rs#L1779)
            - psuedo-generically w/ a callback [here](https://github.com/xmtp/libxmtp/blob/3120239d01ae383e9e52fc2baf9c2fb502342274/xmtp_mls/src/groups/mod.rs#L905)
     - in reality we have a few types of messages that warrant differences in their preparation, which are currently prepared ad-hoc with a callback or otherwise:
         - UserPreferenceUpdate Messages
         - Sync Group Messages
         - Device Sync Messages
         - Application Messages/Optimistic Messages
         - Publish Intent Messages
         - Force Action Messages (testing) / Force Add Member
   
centralizing message formation/giving it a trait while exposing underlying MLS commit/ProtocolMessage information will make it easier to work with our code, as well as provide a framework for extending messages in a backwards-compatible way with more information in the future.


xmtp_mls/src/groups/mls_ext/mls_ext_group.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn membership(&self) -> Result<GroupMembership, CommitValidationError> {
+        self.extensions().group_membership()
+    }
+}
+
+pub trait MlsExtensionsExt {
+    fn group_membership(&self) -> Result<GroupMembership, CommitValidationError>;
+}
+
+impl MlsExtensionsExt for Extensions {
```
neekolas, line 118: This is so much nicer

xmtp_mls/src/groups/mls_ext/intent.rs
```diff
@@ -0,0 +1,91 @@
...
+            Ok(Box::new(intent))
+        }
+        IntentKind::UpdatePermission => {
+            let intent = UpdatePermissionIntentData::try_from(data.to_vec())?;
+            Ok(Box::new(intent))
+        }
+    }
+}
+
+#[derive(Debug, derive_builder::Builder)]
```
neekolas, line 51: Is the goal to use this `derive_builder` and builder pattern in more places? Seems pretty neat.

insipx: absolutely! It makes construction of many of our type easier, and lets us share more fields between common constructions

insipx:
split into #1943 and #1944

PR #1857, opened 2025-04-13
Sync the sync group outside of syncAll

nplasterer:
This adds the ability to sync all of the sync groups instead of calling syncAll
This also will fetch and new welcomes incase the sync group has not been fetched yet.


xmtp_mls/src/client.rs
```diff
@@ -1032,6 +1032,22 @@ where
         Ok(active_groups_count)
     }
 
+    pub async fn sync_all_welcomes_and_history_sync_groups(
+        &self,
+        provider: &XmtpOpenMlsProvider,
+    ) -> Result<usize, ClientError> {
+        self.sync_welcomes(provider).await?;
```
nplasterer, line 1039: I have this syncing the welcomes as well since the sync group may not be present in the local database? but maybe thats too heavy and not necessary. And people should just call sync conversations instead?

mchenani (approved):

xmtp_mls/src/client.rs
cameronvoell: I think the thinking for adding this is correct. We want a call that just works, and this wont work if we dont have the sync group, and sync_welcomes is the only way to ensure that we have it (not sure if we could make a call that adds the sync group without syncing welcomes?).

Perhaps the happy medium is to have an Optional boolean parameter `skipSyncWelcomes` that defaults to false, in case more advanced integrators with their own sync welcomes flow want to skip that call in this function.

nplasterer: I think advanced use case optionality sounds like a good future improvement. Would be easy to add later. But for debugging purposes it may be easier for not to just give them one option. Thank you.

PR #1860, opened 2025-04-14
fix: adjust retryable settings for database errors

### LINKED ISSUE #1850: Adjust retryable settings for database errors
I noticed that there were some types of `StorageError` that are currently treated as "not retryable" in libxmtp, but may occur randomly due to database load or even lock contention.

If one of those non-retryable errors was hit while processing a valid commit, the client would move on to the next commit and throw away the message leading to a fork.

This task would be to audit all of the storage-related errors and make sure that the retryability is correct, and then release it into a client SDK version that we can use to stress test forking with a real app like Convos.

Image

mchenani:
closes #1850


neekolas (approved):
Maybe we use chaos monkey tests as the way to really test these. Very hard to hit in the right place otherwise.

PR #1862, opened 2025-04-15
add fork repro for cursor update from stream moved passed commit 

cameronvoell:
Temporarily adding this PR to show the failing fork repo in CI


neekolas:
Love it. 

Feels like there is room for some more tooling to make tests like these easier to write. Not quite sure what that looks like, but if we had it would be very valuable

PR #1865, opened 2025-04-16
Fix Sync Behavior

codabrink:
### Refactor sync behavior to use provider-based access and add metric tracking for sync operations in XMTP MLS client
* Introduces a new metrics tracking system through the `WorkerHandle` in [handle.rs](https://github.com/xmtp/libxmtp/pull/1865/files#diff-316ee7c6adf5e43532329e72195ecd03a4378035231738f55fe1e1a888711f5c) that monitors sync events and provides waiting mechanisms
* Changes `sync_preferences` return type from `u32` to `u64` in [mls.rs](https://github.com/xmtp/libxmtp/pull/1865/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f) to prevent data loss
* Modifies sync group access to use provider-based approach instead of direct database connections across multiple files
* Updates test framework in [tester.rs](https://github.com/xmtp/libxmtp/pull/1865/files#diff-e248276c2fffaa7ac565757538c70a86a548e6d8369d4be1e61d122ab98a4061) to support metric-based wait conditions instead of time-based waits
* Adds instrumentation to sync operations in [device_sync.rs](https://github.com/xmtp/libxmtp/pull/1865/files#diff-fe691b22a6268a0ff6ae46b00e693850b30ea7ced6afe483ee9c97602aaddde2) for tracking various sync events

#### 📍Where to Start
Start with the new metrics tracking implementation in [handle.rs](https://github.com/xmtp/libxmtp/pull/1865/files#diff-316ee7c6adf5e43532329e72195ecd03a4378035231738f55fe1e1a888711f5c), which introduces the core functionality for monitoring sync operations.

----

_[Macroscope](https://app.macroscope.com) summarized 5ce3d9a._

### DIFF
--- a/xmtp_mls/src/groups/device_sync/consent_sync.rs
+++ b/xmtp_mls/src/groups/device_sync/consent_sync.rs
@@ -25,12 +25,11 @@ pub(crate) mod tests {
     #[cfg(target_arch = "wasm32")]
     wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_dedicated_worker);
 
-    const HISTORY_SERVER_HOST: &str = "localhost";
-    const HISTORY_SERVER_PORT: u16 = 5558;
-
-    use super::*;
-    use crate::{groups::scoped_client::ScopedGroupClient, utils::tester::Tester};
-    use xmtp_db::consent_record::{ConsentState, ConsentType};
+    use crate::{
+        groups::{device_sync::handle::SyncMetric, scoped_client::ScopedGroupClient},
+        utils::tester::Tester,
+    };
+    use xmtp_db::consent_record::{ConsentState, ConsentType, StoredConsentRecord};
 
     use xmtp_cryptography::utils::generate_local_wallet;
     use xmtp_id::associations::test_utils::WalletTestExt;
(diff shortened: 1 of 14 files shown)


xmtp_mls/src/groups/device_sync/consent_sync.rs
```diff
@@ -60,68 +50,58 @@ pub(crate) mod tests {
...
         // Ensure that consent record now exists.
-        let syncable_consent_records = amal_a.syncable_consent_records(amal_a_conn).unwrap();
+        let syncable_consent_records = amal_a
+            .syncable_consent_records(amal_a.provider.conn_ref())
+            .unwrap();
         assert_eq!(syncable_consent_records.len(), 1);
 
         // Create a second installation for amal with sync.
-        let amal_b = ClientBuilder::new_test_client_with_history(&wallet, &history_sync_url).await;
+        let amal_b = amal_a.clone().await;
```
insipx, line 59: is the clone an implicit new installation?

if i were writing a test with `Tester` I'm not sure I would have caught that, what if i'm sending amal_a to a separate thread and accidentally create another installation?

I feel like a function like `create_installation`/new_installation/init_installation or something along those lines would be more self-documenting

insipx (approved):

insipx:
looks good overall!

xmtp_mls/src/groups/device_sync/consent_sync.rs
codabrink: Yeah, I agree - one of those things I would have eventually renamed. I'll think of a better name.

neekolas: Wait. Cloning a client creates a new installation?

codabrink: No, cloning a Tester creates a new installation. It's a helper struct for testing that wraps a client.

PR #1877, opened 2025-04-18
Device Sync V2

codabrink:


xmtp_mls/src/utils/test/tester_utils.rs
```diff
@@ -45,24 +46,62 @@ where
...
-    async fn new_passkey() -> Tester<PasskeyUser, FullXmtpClient>;
-    fn builder() -> TesterBuilder<LocalWallet>;
+#[macro_export]
+macro_rules! tester {
+    ($name:ident) => {
+        let $name = {
+            use tracing::Instrument;
+            use $crate::utils::Tester;
+            let span = tracing::info_span!(stringify!($name));
+            Tester::new().instrument(span).await
```
insipx, line 56: nice, would this output the tester under their own span? Seems like its useful

codabrink: I'm open to suggestions and tweaks. I know you're far more familiar with the tracing crate than I am. I added it to easily differentiate between the different workers.

richardhuaaa (approved):

PR #1882, opened 2025-04-21
init logger when getting identifier

insipx:


richardhuaaa (approved):
Is `init_logger()` idempotent? Why do we need it here - are we accommodating for an entry point that doesn't go through `create_client()`?

PR #1887, opened 2025-04-21
Stable patch release 1.1.5 with fork fixes

cameronvoell:
* on stream welcome call sync instead of create from welcome

* remove implementation from wasm trait

* Update xmtp_mls/src/groups/scoped_client.rs

* still return vec of mls groups from sync (#1884)

---------

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -437,7 +437,7 @@ dependencies = [
 
 [[package]]
 name = "bindings_node"
-version = "1.1.4"
+version = "1.1.5"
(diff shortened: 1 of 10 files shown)


neekolas (approved):

neekolas (approved):
Preapproving

cameronvoell:
looks like wasm failures showed on previous commit to release/1.1 branch as well. wasm is building stable on a separate release branch with latest fork fixes for now. so we can go ahead and merge this

PR #1889, opened 2025-04-22
feat(group): add fork detection flag

mchenani:
This PR adds logic to detect and handle a potential group fork when a message fails to decrypt because it comes from a future epoch.

When this happens during message processing, the group is marked as forked to help surface the issue early and prevent further inconsistencies.

It also introduces a new FutureEpoch error type to clearly represent this case in the message processing flow.


xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -587,6 +593,17 @@ where
         // and roll the transaction back, so we can fetch updates from the server before
         // being ready to process the message for a second time.
         let mut processed_message = None;
+        if is_test_mode_aead_msg() {
```
mchenani, line 596: @antews @cameronvoell we need to mimic the failure somehow for tests, I can extract the whole test section to another function for tests but like we did here https://github.com/xmtp/libxmtp/blob/4ab7ed92d8c3526a66d8ed5156417bde0495bbad/xmtp_mls/src/groups/mls_sync.rs#L1974

mchenani: https://github.com/xmtp/libxmtp/issues/1903

xmtp_db/migrations/2025-04-22-121314_groups_add_fork_status/up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE groups
+    ADD COLUMN fork_state BOOLEAN DEFAULT FALSE;
```
richardhuaaa, line 2: Given this is a boolean rather than a blob or text, would `maybe_forked` be better?

richardhuaaa: Also, given that we have a `DEFAULT` specified, could we add a `NOT NULL` constraint here, and make the resulting Rust code a simple `bool` rather than an `Option<bool>`? Should simplify the surrounding code.

I think we should consider doing a similar thing with `fork_details` too (`NOT NULL DEFAULT ''`), so that future Rust developers don't need to figure out whether the empty state is NULL or empty string

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -588,6 +607,16 @@ where
...
+            return Err(OpenMlsProcessMessage(ProcessMessageError::ValidationError(
+                ValidationError::UnableToDecrypt(AeadError),
+            )));
+        }
+
+        if is_test_mode_future_wrong_epoch() {
+            //this is just a mock for testing, the whole function only appears in tests
```
richardhuaaa, line 617: I'm not sure where we landed on the mocking - are we keeping this forever, or do we have an idea for the future that's just too costly right now?

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -700,6 +729,141 @@ where
         Ok(())
     }
 
+    #[allow(unused_variables, dead_code)]
+    #[cfg(not(any(test, feature = "test-utils")))]
```
richardhuaaa, line 733: I don't think we should keep two copies of the same function, one for testing and one for not testing. It means all future changes have to be duplicated, and adds a footgun where someone makes a change in one but not the other (resulting in code not making it into prod, or prod code not being tested).

What if we shrinked the area of duplication, and added a much smaller function called something like `maybe_mock_decryption_error_for_tests()`? Then we could call it from within `validate_and_process_external_message` and add `#[cfg(any(test, feature = "test-utils"))]` to this new function instead. A benefit of this approach is that we also hide away the details of exactly which errors are being mocked inside this function, whether that's `aead_message` or `wrong_epoch`.

Another thing I'm not sure if Rust supports or not is whether you can conditionally compile statements. If that's possible, you also wouldn't have to define this new function twice, you could do the following (maybe).

```
async fn validate_and_process_external_message(...) {
   #[cfg(any(test, feature = "test-utils"))]
   maybe_mock_decryption_error_for_tests()?;
   
   <rest of function>
}
```

richardhuaaa (commented):
Overall looks good and is a very important addition! Mainly just want to avoid the duplication

xmtp_db/migrations/2025-04-22-121314_groups_add_fork_status/up.sql
mchenani: Thanks for the comment, code is updated!

xmtp_mls/src/groups/mls_sync.rs
mchenani: here is the issue to use a mocking tool later and refactor the code to make it more mocking-friendly, atm it requires lot of refactoring.
https://github.com/xmtp/libxmtp/issues/1903

xmtp_mls/src/groups/mls_sync.rs
mchenani: Thanks for the comment and great point about modifying the code later would be hard, just adjusted the code in a way you've suggested; on `#[cfg(any(test, feature = "test-utils"))]` for specific block or line of code, yes rust supports that.

xmtp_mls/src/utils/test/mod.rs
```diff
@@ -373,3 +377,37 @@ pub fn get_test_mode_malformed_installations() -> Vec<Vec<u8>> {
         })
         .collect()
 }
+
+#[cfg(any(test, feature = "test-utils"))]
+#[warn(dead_code)]
```
richardhuaaa, line 382: I think we can remove this now that it's no longer dead code

xmtp_mls/src/utils/test/mod.rs
```diff
@@ -373,3 +377,37 @@ pub fn get_test_mode_malformed_installations() -> Vec<Vec<u8>> {
...
+pub fn maybe_mock_wrong_epoch_for_tests() -> Result<(), GroupMessageProcessingError> {
+    if is_test_mode_future_wrong_epoch() {
+        return Err(OpenMlsProcessMessage(ProcessMessageError::ValidationError(
+            WrongEpoch,
+        )));
+    }
+    Ok(())
+}
+
+pub fn maybe_mock_future_epoch_for_tests() -> Result<(), GroupMessageProcessingError> {
```
richardhuaaa, line 408: These two seem very similar - is there a reason why we need to mock them in two different locations? Also, are we no longer handling the scenario where a message is in the right epoch, but undecryptable?

mchenani: Yes, but they return two different errors for two different places, one is when we process a message and just get a wrong epoch, and one in upper layers where we get future epoch.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -150,6 +156,8 @@ impl RetryableError for GroupMessageProcessingError {
             Self::CommitValidation(err) => err.is_retryable(),
             Self::ClearPendingCommit(err) => err.is_retryable(),
             Self::Client(err) => err.is_retryable(),
+            Self::FutureEpoch(_, _) => true,
+            Self::OldEpoch(_, _) => true,
```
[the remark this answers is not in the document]
mchenani: Just checked, and we don't have any retryable flag for wrong epoch.
for the FutureEpoch I'm sure we can set it to false, a little bit hesitant about where we can recover from OldEpoch. But for now in my code we only propagate the future epoch, will set both to false and check if anything breaks.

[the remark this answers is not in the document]
mchenani: Checked again, I changed them to true due to another issue and I fixed it afterwards, now returned them to false and seems working fine!

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1323,6 +1351,71 @@ where
...
+    ) -> Result<(), GroupMessageProcessingError> {
+        let group_id = message.group_id.clone();
+
+        if let OpenMlsProcessMessage(ProcessMessageError::ValidationError(
+            ValidationError::WrongEpoch,
+        )) = error
+        {
+            let group_epoch = match self.epoch(provider).await {
+                Ok(epoch) => epoch,
+                Err(_) => {
```
insipx, line 1368: why would epoch calculation fail here, would it be worth returning the error or nesting the error into another type so we maintain the context here? Could also just print the error out with the log by turning it into a string, `e.to_string()`

Also, what's the reasoning behind failing silently here -- if we can't deserialize a message the library user should know about it, and/or there's a bug that we need to fix since serialization shouldn't be changing like that

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1323,6 +1351,71 @@ where
...
+                    tracing::info!(
+                        "WrongEpoch encountered but group_epoch could not be calculated"
+                    );
+                    return Ok(());
+                }
+            };
+
+            let mls_message_in = match MlsMessageIn::tls_deserialize_exact(&message.data) {
```
insipx, line 1376: same thing here, why not just bubble the error up with `?`

If a message deserializes incorrectly it would probably point to issues other than forking, and we would need the context of the error to diagnose, and its valid to return it form this function since it still indicates an error

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -577,6 +596,12 @@ where
         envelope: &GroupMessageV1,
         allow_cursor_increment: bool,
     ) -> Result<(), GroupMessageProcessingError> {
+        #[cfg(any(test, feature = "test-utils"))]
```
insipx, line 599: can probably reduce this to a nice two-liner:
```suggestion
        #[cfg(any(test, feature = "test-utils"))]
        crate::utils::maybe_mock_wrong_epoch_for_tests()?;
```

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1323,6 +1351,71 @@ where
...
+    async fn process_group_message_error_for_fork_detection(
+        &self,
+        provider: &XmtpOpenMlsProvider,
+        message: &GroupMessageV1,
+        error: &GroupMessageProcessingError,
+    ) -> Result<(), GroupMessageProcessingError> {
+        let group_id = message.group_id.clone();
+
+        if let OpenMlsProcessMessage(ProcessMessageError::ValidationError(
+            ValidationError::WrongEpoch,
```
insipx, line 1363: we're catching wrong epoch here, but not decryption error for the right epoch. Were we able to figure out a place to catch those areas as well, so that we can log it? Is it possible to do in this function too

insipx (changes requested):

xmtp_mls/src/groups/mls_sync.rs
mchenani: Basically if we reach here means the message was processed, these error handlings are just for not crashing of something happens afterward and we couldn't extract the proper data of the message.

xmtp_mls/src/groups/mls_sync.rs
mchenani: same reason as above, I would say we can follow your advice to print the error in details.

xmtp_mls/src/groups/mls_sync.rs
insipx: OK, lets create an issue to cover the AEAD fork errors so we don't lose track of it as well

insipx (approved):

PR #1894, opened 2025-04-22
Update build/test instructions

richardhuaaa:
I ran into a few issues getting the repo back up and running on my machine, seems like a lot has changed. This updates the READMEs and default workspace members for the next person!

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,16 +29,25 @@ Start Docker Desktop.
 - To run tests:
 
   ```
-  dev/test
+  cargo test
   ```
 
 - To run WebAssembly tests headless:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -29,16 +29,25 @@ Start Docker Desktop.
 - To run tests:
 
   ```
-  dev/test
```
richardhuaaa, line 32: Looks like this script was removed (it only ran `cargo test` previously). Updated the README to reflect that, but also happy to add back a script if we want one

insipx (approved):
Nice!

PR #1906, opened 2025-04-25
Message sync test

nplasterer:
Reproduce message history test


bindings_ffi/src/mls.rs
```diff
@@ -7960,4 +7960,50 @@ mod tests {
...
+        let initial_messages = alix_group.find_messages(FfiListMessagesOptions::default()).await.unwrap();
+        assert_eq!(initial_messages.len(), 2);
+
+        let alix2 = alix.builder.build().await;
+
+        let state = alix2.inbox_state(true).await.unwrap();
+        assert_eq!(state.installations.len(), 2);
+        
+        // Doesn't fail on the not finding the group if you send a message after the new installation is created.
+        // alix_group.send_text("hi").await.unwrap();
```
nplasterer, line 7990: This piece doesn't make sense to me. Seems like sending a message in a group will automatically add the installation? But syncing doesn't seem to be enough?

PR #1937, opened 2025-05-01
Stream All Messages conversation filtering

nplasterer:
Adds the ability to filter stream all messages by an array of consents. 
Defaults to allowed conversations only.

### DIFF
--- a/xmtp_mls/src/subscriptions/stream_all.rs
+++ b/xmtp_mls/src/subscriptions/stream_all.rs
@@ -11,6 +11,7 @@ use crate::{
 
 use futures::stream::Stream;
 use xmtp_db::{
+    consent_record::ConsentState,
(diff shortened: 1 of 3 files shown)


xmtp_mls/src/subscriptions/stream_all.rs
```diff
@@ -452,4 +462,77 @@ mod tests {
         tracing::info!("Total Messages: {}", messages.len());
         assert_eq!(messages.len(), 5);
     }
+
+    #[rstest::rstest]
+    #[case(ConsentState::Allowed, "msg in allowed")]
+    #[case(ConsentState::Denied, "msg in denied")]
+    #[case(ConsentState::Unknown, "msg in unknown")]
```
insipx, line 469: Nice! I think this is the first time anyone is using these

insipx (approved):

Issue #1499, opened 2025-01-13
HMAC & DM stitching follow ups


nplasterer:
I want the ability to get the hmac keys for a single conversation
I want the ability to get a list of topics for a single conversation

If I get a new welcome I want to be able to subscribe to all the topics for that conversation
If I get a new welcome I want to be able to subscribe to the hmac keys for that conversation.

Example

``` kotlin
// Push notification received for a new welcome message for a DM
                
                val subscription = dm.allTopics().map { topic ->
                    val hmacKeysResult = dm.getHmacKeys()
                    val hmacKeys = hmacKeysResult.hmacKeysMap
                    val result = hmacKeys[topic]?.valuesList?.map { hmacKey ->
                        Service.Subscription.HmacKey.newBuilder().also { sub_key ->
                            sub_key.key = hmacKey.hmacKey
                            sub_key.thirtyDayPeriodsSinceEpoch = hmacKey.thirtyDayPeriodsSinceEpoch
                        }.build()
                    }

                    Service.Subscription.newBuilder().also { sub ->
                        sub.addAllHmacKeys(result)
                        sub.topic = topic
                        sub.isSilent = false
                    }.build()
                }
                PushNotificationTokenManager.xmtpPush.subscribeWithMetadata(subscription)

// Push notification received for a new welcome message for a GROUP
                
                    val hmacKeysResult = group.getHmacKeys()
                    val hmacKeys = hmacKeysResult.hmacKeysMap
                    val result = hmacKeys[group.topic]?.valuesList?.map { hmacKey ->
                        Service.Subscription.HmacKey.newBuilder().also { sub_key ->
                            sub_key.key = hmacKey.hmacKey
                            sub_key.thirtyDayPeriodsSinceEpoch = hmacKey.thirtyDayPeriodsSinceEpoch
                        }.build()
                    }

                   val subscription = Service.Subscription.newBuilder().also { sub ->
                        sub.addAllHmacKeys(result)
                        sub.topic = group.topic
                        sub.isSilent = false
                    }.build()
                
                PushNotificationTokenManager.xmtpPush.subscribeWithMetadata(subscription)

}
```

PR #1953, opened 2025-05-05
Get all duplicate dms for a dm

nplasterer:
Enables the ability to get all the topics for a dm.

Part of https://github.com/xmtp/libxmtp/issues/1499

### DIFF
--- a/xmtp_db/src/encrypted_store/group.rs
+++ b/xmtp_db/src/encrypted_store/group.rs
@@ -555,21 +555,31 @@ impl DbConnection {
         Ok(())
     }
 
-    pub fn find_duplicate_dms(&self) -> Result<Vec<StoredGroup>, StorageError> {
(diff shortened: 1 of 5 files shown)


xmtp_db/src/encrypted_store/group.rs
```diff
@@ -554,6 +554,33 @@ impl DbConnection {
 
         Ok(())
     }
+
+    pub fn find_duplicate_dms_for_group(
```
codabrink, line 558: Just to put on your radar, something very similar [exists](https://github.com/xmtp/libxmtp/blob/7cefaa267df9970ab4ff471b1eaf2bcfdecb745f/xmtp_db/src/encrypted_store/group/dms.rs#L45-L66), the only difference is it excludes the id of the group you provided.

codabrink (approved):

PR #1957, opened 2025-05-05
Reduce max group size

richardhuaaa:
Putting up an arbitrary number to provoke discussion. If we are advising SDK consumers not to create groups of a certain size due to performance risks, how about we document that as the limit and enforce it programmatically?


richardhuaaa:
We've had a lot of discussion, and we think 250 is heuristically a conservative number to set for this release, but we are in the process of rehauling our performance testing, which could reveal that a much higher number is acceptable.

codabrink (approved):

richardhuaaa:
Reduced it further to 220 to be maximally conservative, but can definitely be increased in the future based on feedback and metrics. We also have some streaming improvements coming that should hopefully help with performance also.

nplasterer (approved):
@jhaaaa we need to update the docs fyi

Issue #1948, opened 2025-05-01
Ability to filter streamAllMessages by conversation consent


nplasterer:
We should not be streaming all messages for conversations that are denied
But give integrators the ability to pass in an array of consent to filter.

nplasterer:
Everywhere we filter consent it should only be `allowed and unknown`.

PR #1958, opened 2025-05-05
Changes consent filtering defaults to always filter out denied

nplasterer:
Closes https://github.com/xmtp/libxmtp/issues/1948

### DIFF
--- a/xmtp_db/src/encrypted_store/group.rs
+++ b/xmtp_db/src/encrypted_store/group.rs
@@ -253,52 +253,44 @@ impl DbConnection {
             query = query.filter(groups_dsl::conversation_type.eq(conversation_type));
         }
 
-        let mut groups = if let Some(consent_states) = consent_states {
-            if consent_states
-                .iter()
-                .any(|state| *state == ConsentState::Unknown)
-            {
-                // Include both `Unknown`, `null`, and other specified states
-                let query = query
-                    .left_join(
-                        consent_dsl::consent_records
-                            .on(sql::<diesel::sql_types::Text>("lower(hex(groups.id))")
-                                .eq(consent_dsl::entity)),
-                    )
-                    .filter(
-                        consent_dsl::state
-                            .is_null()
-                            .or(consent_dsl::state.eq(ConsentState::Unknown))
-                            .or(consent_dsl::state.eq_any(
-                                consent_states
-                                    .iter()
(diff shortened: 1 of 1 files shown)


neekolas:
I think this makes sense as a default. We should be sure to communicate this clearly to integrators in the release notes and docs.

While I don't normally need to see denied conversations, people block things by accident all the time. The app needs some way to un-block, which is now going to require explicitly looking for the denied conversations or changing your filters when you get the full conversation list.

xmtp_db/src/encrypted_store/conversation_list.rs
```diff
@@ -114,53 +114,48 @@ impl DbConnection {
...
+        } else {
+            // INNER JOIN: strict match only to specific states (no Unknown or NULL)
+            let inner_joined_query = query
+                .inner_join(consent_dsl::consent_records.on(
+                    sql::<diesel::sql_types::Text>("lower(hex(conversation_list.id))").eq(consent_dsl::entity),
+                ))
+                .filter(consent_dsl::state.eq_any(filtered_states.clone()))
+                .select(conversation_list::all_columns());
+
+            self.raw_query_read(|conn| inner_joined_query.load::<ConversationListItem>(conn))?
```
nplasterer, line 158: I'm really at a loss here for how to preserve order while filtering for consent. I feel like I've tried a million different queries @insipx @mchenani do either of you have ideas? In my head this should just work? I don't know why the order is getting moved and the test fails but if I run it 5 times in a row it will pass so it's for sure ordering randomly.

richardhuaaa: Do we need a `.order(conversation_list_dsl::created_at_ns.asc())`?

nplasterer: I added that in and the tests still failed. Because this function in particular from my understanding orders the conversations my message last sent at ns?

nplasterer: This was the fix!
`.order(sql::<BigInt>("COALESCE(sent_at_ns, created_at_ns) DESC"))`

cameronvoell (approved):
>This was the fix!
.order(sql::<BigInt>("COALESCE(sent_at_ns, created_at_ns) DESC"))

nice find!!

PR #1961, opened 2025-05-06
Device Sync Streaming

codabrink:
### Add automatic message streaming to device sync by implementing StreamAllMessages in xmtp_mls client
* Implements automatic message streaming for device sync in `StreamAllMessages` by adding sync group tracking and notification events in [stream_all.rs](https://github.com/xmtp/libxmtp/pull/1961/files#diff-4a3d32ed6005867d963b0eae023dc9c1fd8ffd9531413ba572f4287210a18cc1)
* Refactors test utilities to support automatic message streaming by adding stream handling capabilities to `Tester` and `TesterBuilder` in [tester_utils.rs](https://github.com/xmtp/libxmtp/pull/1961/files#diff-727ea71bc8afdfdf6f0b1d9813b2e6e46f6fdd647abbc9b56be80bd873034c7d)
* Updates test code across multiple files to use simplified tester macro syntax and new streaming functionality
* Updates `deranged` dependency from version 0.4.0 to 0.4.1

#### 📍Where to Start
Start with the `StreamAllMessages` struct implementation in [stream_all.rs](https://github.com/xmtp/libxmtp/pull/1961/files#diff-4a3d32ed6005867d963b0eae023dc9c1fd8ffd9531413ba572f4287210a18cc1) which contains the core streaming functionality changes.

#### Changes since #1961 opened

- Added test skip configuration in `xmtp_mls` package [86c3f99]
- Modified sync group message handling in `xmtp_mls` streaming implementation [8f85dc4]
- Converted database query type in `xmtp_db` message processing [8f85dc4]
----

_[Macroscope](https://app.macroscope.com) summarized 8f85dc4._

### DIFF
--- a/xmtp_mls/src/subscriptions/stream_all.rs
+++ b/xmtp_mls/src/subscriptions/stream_all.rs
@@ -494,6 +494,7 @@ mod tests {
     #[xmtp_common::test]
     #[timeout(Duration::from_secs(20))]
     #[cfg_attr(target_arch = "wasm32", ignore)]
+    #[ignore]
     async fn test_stream_all_messages_filters_by_consent_state(
         #[case] filter: ConsentState,
         #[case] expected_message: &str,
--- a/xmtp_mls/src/subscriptions/stream_all.rs
+++ b/xmtp_mls/src/subscriptions/stream_all.rs
@@ -60,8 +60,9 @@ where
                 conversation_type,
                 consent_states,
                 include_duplicate_dms: true,
-                include_sync_groups: conversation_type.is_none(),
-
+                include_sync_groups: conversation_type
+                    .map(|ct| matches!(ct, ConversationType::Sync))
+                    .unwrap_or(true),
                 ..Default::default()
             })?;
 
@@ -127,6 +128,7 @@ where
                         .send(LocalEvents::SyncWorkerEvent(
                             SyncWorkerEvent::NewSyncGroupMsg,
                         ));
+                    return self.poll_next(cx);
                 }
             };
 
--- a/xmtp_mls/src/subscriptions/stream_all.rs
+++ b/xmtp_mls/src/subscriptions/stream_all.rs
@@ -1,9 +1,12 @@
 use std::{
+    collections::HashSet,
     pin::Pin,
     task::{ready, Context, Poll},
 };
 
-use crate::subscriptions::stream_messages::MessagesApiSubscription;
+use crate::subscriptions::{
+    stream_messages::MessagesApiSubscription, LocalEvents, SyncWorkerEvent,
+};
 use crate::{
     groups::{scoped_client::ScopedGroupClient, MlsGroup},
     Client,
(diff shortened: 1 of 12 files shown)


xmtp_mls/src/subscriptions/stream_all.rs
```diff
@@ -52,23 +52,36 @@ where
...
-                    conversation_type,
-                    consent_states,
-                    include_duplicate_dms: true,
-                    ..Default::default()
-                })?
+            let groups = provider.conn_ref().find_groups(GroupQueryArgs {
+                conversation_type,
+                consent_states,
+                include_duplicate_dms: true,
+                include_sync_groups: conversation_type.is_none(),
```
insipx, line 63: shouldn't it be `conversation_type.map(|c| c == ConversationType::Sync).unwrap_or(false)`?

If I set conversation type and consent states both to `None` I expect all messages, but also if i set conversation type to Sync I also expect just sync messages.

xmtp_mls/src/subscriptions/stream_all.rs
```diff
@@ -105,6 +119,17 @@ where
         let mut this = self.as_mut().project();
 
         if let Ready(msg) = this.messages.as_mut().poll_next(cx) {
+            if let Some(Ok(msg)) = &msg {
+                if self.sync_groups.contains(&msg.group_id) {
```
insipx, line 123: Couple questions:

instead of piping to an intermediary channel here, why not just call stream all messages with the filter set to only sync groups?

Also, if there are multiple streams running, how do we ensure the sync worker isn't just getting a bunch of duplicate messages


This stream will also return the sync group message since its `return Ready(msg)`, could that be possibly confusing if someone streams all messages, and now they're getting sync messages in two places.

insipx (changes requested):

xmtp_mls/src/subscriptions/stream_all.rs
codabrink: Yes, but with `.unwrap_or(true)` I believe. Thanks for that catch. Updating.

xmtp_mls/src/subscriptions/stream_all.rs
codabrink: That's a good concern. I'm a little out of context with streaming, but thinking about this a bit more, the integrator should never receive sync group messages. I've adjusted the handling to filter out sync messages and just notify the worker [here](https://github.com/xmtp/libxmtp/pull/1961/commits/8f85dc469645cf42aad552a5c85fa815b4815269#diff-4a3d32ed6005867d963b0eae023dc9c1fd8ffd9531413ba572f4287210a18cc1R131).

If the worker receives multiple pings from multiple streams, that's fine. The worker will receive the ping, see that there are no new unprocessed messages and go back to sleep.

xmtp_mls/src/subscriptions/stream_all.rs
```diff
@@ -105,6 +120,18 @@ where
...
+                    let _ = self
+                        .client
+                        .local_events()
+                        .send(LocalEvents::SyncWorkerEvent(
+                            SyncWorkerEvent::NewSyncGroupMsg,
```
insipx, line 129: looks good, jw about this part:
when this is sent, what happens in the sync worker? Would it be useful to send the message itself in the channel, or how does the worker get the actual message

insipx (approved):

PR #1964, opened 2025-05-06
Add optimistic group creation

rygine:
# Summary

Added `create_group_optimistic` to all bindings

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -1157,10 +1157,12 @@ impl From<&FfiMetadataField> for MetadataField {
 
 #[uniffi::export(async_runtime = "tokio")]
 impl FfiConversations {
-    pub async fn create_group_optimistic(
+    pub fn create_group_optimistic(
         &self,
         opts: FfiCreateGroupOptions,
     ) -> Result<Arc<FfiConversation>, GenericError> {
+        log::info!("creating optimistic group");
+
         if let Some(FfiGroupPermissionsOptions::CustomPolicy) = opts.permissions {
             if opts.custom_permission_policy_set.is_none() {
                 return Err(GenericError::Generic {
@@ -1213,13 +1215,13 @@ impl FfiConversations {
                 .join(", ")
         );
 
-        let convo = self.create_group_optimistic(opts).await?;
+        let convo = self.create_group_optimistic(opts)?;
 
         if !account_identities.is_empty() {
             convo.add_members(account_identities).await?;
-        };
-
-        convo.sync().await?;
+        } else {
+            convo.sync().await?;
+        }
 
         Ok(convo)
     }
@@ -1234,13 +1236,13 @@ impl FfiConversations {
             inbox_ids.join(", ")
         );
 
-        let convo = self.create_group_optimistic(opts).await?;
+        let convo = self.create_group_optimistic(opts)?;
 
         if !inbox_ids.is_empty() {
             convo.add_members_by_inbox_id(inbox_ids).await?;
+        } else {
+            convo.sync().await?;
         };
-
-        convo.sync().await?;
 
         Ok(convo)
     }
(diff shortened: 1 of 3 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -1208,21 +1192,38 @@ impl FfiConversations {
...
-            group.sync().await?;
-            group
-        } else {
-            self.inner_client
-                .create_group_with_members(&account_identities, group_permissions, metadata_options)
-                .await?
-        };
+        let convo = self
+            .inner_client
+            .create_group(group_permissions, metadata_options)?;
```
nplasterer, line 1197: Does adding metadata options not create a commit? cc @cameronvoell

bindings_ffi/src/mls.rs
```diff
@@ -1157,26 +1157,10 @@ impl From<&FfiMetadataField> for MetadataField {
...
-        account_identities: Vec<FfiIdentifier>,
         opts: FfiCreateGroupOptions,
     ) -> Result<Arc<FfiConversation>, GenericError> {
-        let account_identities: Result<Vec<Identifier>, _> = account_identities
-            .into_iter()
-            .map(|ident| ident.try_into())
-            .collect();
-        let account_identities = account_identities?;
-
-        log::info!(
```
richardhuaaa, line 1171: Could we still have a log here, something like 'creating optimistic group'?

bindings_ffi/src/mls.rs
richardhuaaa: It's probably okay either way. Regardless of if a commit was queued or not, what we really want is for the method to return immediately so that integrators have a group ID. Given the lack of `await`, the network requests are probably only going out when `sync()` is being called.

We should give clear guidance to integrators on what to do though, and validate it matches their intent. When a group is created in the UI, this method can be called to return a group ID synchronously, and either `add_members()` or `sync()` can be called afterwards to add members or sync your own installations respectively, and the UI shouldn't block on the second method returning.

bindings_ffi/src/mls.rs
```diff
@@ -1208,21 +1192,38 @@ impl FfiConversations {
...
+                .join(", ")
+        );
+
+        let convo = self.create_group_optimistic(opts).await?;
+
+        if !account_identities.is_empty() {
+            convo.add_members(account_identities).await?;
+        };
+
+        convo.sync().await?;
```
richardhuaaa, line 1222: I don't think we need to call `sync()` here, if and only if `add_members` was called, because that method already syncs

bindings_ffi/src/mls.rs
```diff
@@ -1233,50 +1234,15 @@ impl FfiConversations {
...
-                .inner_client
-                .create_group(group_permissions, metadata_options)?;
-            group.sync().await?;
-            group
-        } else {
-            self.inner_client
-                .create_group_with_inbox_ids(&inbox_ids, group_permissions, metadata_options)
-                .await?
-        };
+        convo.sync().await?;
```
richardhuaaa, line 1243: Same comment here

bindings_ffi/src/mls.rs
```diff
@@ -1157,26 +1157,10 @@ impl From<&FfiMetadataField> for MetadataField {
 
 #[uniffi::export(async_runtime = "tokio")]
 impl FfiConversations {
-    pub async fn create_group(
+    pub async fn create_group_optimistic(
```
richardhuaaa, line 1160: Does this need to be `async`? That may give integrators the wrong impression

richardhuaaa (approved):
Really appreciate the deduplication between methods! I only reviewed the `mls.rs` file, I imagine the others will be similar

bindings_ffi/src/mls.rs
rygine: good catch, will fix!

bindings_ffi/src/mls.rs
```diff
@@ -1157,25 +1157,11 @@ impl From<&FfiMetadataField> for MetadataField {
 
 #[uniffi::export(async_runtime = "tokio")]
 impl FfiConversations {
-    pub async fn create_group(
+    pub fn create_group_optimistic(
```
nplasterer, line 1160: One thing I did think about is this won't work for dms. Anyway we could make this work for dms? 🤔

PR #1986, opened 2025-05-09
ConnectionExt suggestion

codabrink:


insipx:
got a fix along the lines of the default type in `DbConnection`, but the straight default type didn't work with PRs further up the stack, so i went a different route that also gets rid of the generic types in `raw_query{read|write}`, but doesn't introduce a default type

PR #1989, opened 2025-05-09
Rename Backup to Archive

codabrink:
The term "backup" tends to give integrators the wrong impression of what this feature does. This PR renames the `Backup` feature to `Archive` to reduce potential confusion.


bindings_ffi/src/mls.rs
```diff
@@ -771,13 +771,13 @@ impl From<BackupMetadata> for FfiBackupMetadata {
...
 #[derive(uniffi::Record)]
-pub struct FfiBackupOptions {
+pub struct FfiArchiveOptions {
     start_ns: Option<i64>,
     end_ns: Option<i64>,
     elements: Vec<FfiBackupElementSelection>,
 }
-impl From<FfiBackupOptions> for BackupOptions {
-    fn from(value: FfiBackupOptions) -> Self {
+impl From<FfiArchiveOptions> for BackupOptions {
```
mchenani, line 779: Do we need to change this too?
```suggestion
impl From<FfiArchiveOptions> for ArchiveOptions {
```

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -4,7 +4,7 @@ use crate::{
     subscriptions::{LocalEvents, SubscribeError, SyncWorkerEvent},
     Client,
 };
-use backup::BackupError;
+use archive::BackupError;
```
mchenani, line 7: How about the BackupError?

mchenani (approved):
Looks good, some nits to consider.

bindings_ffi/src/mls.rs
codabrink: It's a Proto. I'll update those later.

xmtp_mls/src/groups/device_sync.rs
codabrink: Good point, let's change that too.

PR #1996, opened 2025-05-10
fix: ensure WASM bindings return consistent values for debug_info

### LINKED ISSUE #1983: WASM bindings `debugInfo` returns different values under normal conditions
when calling `conversation.debugInfo()` in the WASM bindings, it's returning:

```
{
  epoch: 1n,
  maybeForked: undefined,
  forkDetails: undefined,
}
```

in the node bindings, i see this:

```
{
  epoch: 1n,
  maybeForked: false,
  forkDetails: "",
}
```

crStiv:
### Problem
`conversation.debugInfo()` returns `undefined` values for `maybeForked` and `forkDetails` fields in WASM bindings, while Node bindings return explicit `false` and empty string values.

### Solution
Added serde default value annotations to the ConversationDebugInfo struct to ensure consistent serialization behavior across platforms:
- `#[serde(default = "bool_false")]` for `maybe_forked` field
- `#[serde(default = "string_empty")]` for `fork_details` field

### Testing
- Added unit test `test_conversation_debug_info_serialization` to verify serialization behavior
- Verified using `dev/check-wasm` and `dev/lint-rust` that the code compiles and passes style checks

Fixes #1983

### DIFF
--- a/bindings_wasm/src/conversation.rs
+++ b/bindings_wasm/src/conversation.rs
@@ -657,7 +657,14 @@ impl Conversation {
       .await
       .map_err(|e| JsError::new(&format!("{e}")))?;
 
-    Ok(crate::to_value(&ConversationDebugInfo::new(debug_info))?)
+    // Create a new ConversationDebugInfo with explicit maybe_forked and fork_details
+    let conversation_debug_info = ConversationDebugInfo {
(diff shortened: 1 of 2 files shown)


rygine:
@insipx this is a bot

Issue #1985, opened 2025-05-09
Get clear on device sync versus preference sync


nplasterer:
Clean up the bindings to do 1 or the other not both.

PR #2001, opened 2025-05-12
Remove sync_device_sync from FFI

codabrink:


nplasterer (approved):
Can you do this for the node and wasm bindings as well. They also have duplicates. Also closes https://github.com/xmtp/libxmtp/issues/1985

Issue #1988, opened 2025-05-09
16KB Align Android FFI bindings


nplasterer:
<img width="895" alt="Image" src="https://github.com/user-attachments/assets/f9754b91-0e02-4475-acad-c0700a33772e" />

PR #2009, opened 2025-05-13
16 KB Alignment in Native Libraries

nplasterer:
Fixes https://github.com/xmtp/libxmtp/issues/1988

When compiling a .so library you need to configure the build process to use 16 KB page alignment. Android now flags native libraries that are only aligned to the older 4 KB boundary.


nplasterer:
Confirmed on Android everything is now 16KB!
Screenshot 2025-05-13 at 12 19 08 PM

mchenani (approved):

PR #2017, opened 2025-05-19
Remove circular dependency on client in MlsGroup

insipx:
The circular dependency on 'Client' in MlsGroup makes it very difficult to split code into digestible sections, while also making it difficult to unit test. XmtpDb and XmtpApi on their own are easily mockable. Together, however, it is very difficult to mock with all the function on client/scoped group client.

This Pr:

- moves most of Client into XmtpMlsLocalContext
- removes client from MlsGroup
- removes ScopedGroupClient
- creates an XmtpContextProvider trait
- all external dependencies may be accessed through context
- split up some functions out of Client into their own modules (DeviceSyncClient, MlsStore, WelcomeService)

These were extracted out of Client and Group so that both may use them. They can be constructed by passing the context to the structs, i.e
`
MlsStore::new(self.context.clone()).query_group_messages()`

`MlsStore`
    - basically higher-level cached queries which return Mls types instead of Database types (MlsGroup etc)

`WelcomeService`
    - syncing welcomes
 
`DeviceSyncClient`
    - all device sync methods defined here
 
`IdentityUpdates`
    - AssociationState/identity update actions

Some methods on client that should have been on group were moved, i.e
`disappearing_settings`/`find_duplicate_dms`

The `for_group` methods still exist on `Client`, however they use the `MlsGroup::new_cached()` method, which builds an MlsGroup from the database by its GroupId. This previously existed as `new_validated`

This PR is almost done, requires finishing touches to the node/wasm bindings to finishes

With the circular dependency gone, we can easily mock the api/db in stream tests. This way we can use the OpenMls test suite to generate lots of messages and ensure our stream can process them, as well as mock missed messages, out of order messages, etc in unit tests w/o the network. This speeds up the feedback loop and allows us to create much more targeted, succinct tests.

All current stream tests will be moved to be integration tests


insipx:
* **#2017** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2017?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2017?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xmtp_db/src/encrypted_store/database/native/sqlcipher_connection.rs
```diff
@@ -66,7 +66,7 @@ impl EncryptedConnection {
                             db_pathbuf.display(),
                             salt_path.display(),
                         );
-                        let file = File::open(salt_path)?;
+                        let file = BufReader::new(File::open(salt_path)?);
```
cameronvoell, line 69: is this addressing any specific issue? seems unrelated to rest of the PR.

cameronvoell (approved):
nice improvement for organization and testability 🙇

examples/cli/serializable.rs
```diff
@@ -6,6 +6,8 @@ use xmtp_db::group_message::StoredGroupMessage;
 use xmtp_mls::{groups::MlsGroup, XmtpApi};
 use xmtp_proto::xmtp::mls::message_contents::EncodedContent;
 
+// use crate::RustMlsGroup;
```
mchenani, line 9: nit: do we need this?

mchenani (approved):
🔥

xmtp_db/src/encrypted_store/database/native/sqlcipher_connection.rs
insipx: Clippy complained that opening without a buffer is slow

PR #2019, opened 2025-05-20
Event Tracking Pt. 1, move Archives into it's own crate

codabrink:
### Move archive functionality into dedicated `xmtp_archive` crate and reorganize MLS common types into `xmtp_mls_common` crate
The changes reorganize the codebase by:

* Creating new `xmtp_archive` crate containing archive functionality previously in `xmtp_mls`, including import/export streams, encryption, and backup metadata
* Creating new `xmtp_mls_common` crate containing shared MLS types and constants previously in `xmtp_mls`
* Adding client event tracking system in `xmtp_db` with new `client_events` table and tracking for client builds, intent queuing, and group welcomes
* Moving time constants to `xmtp_common` crate's new `const.rs` module
* Updating import paths across multiple crates to reference types from their new locations

#### 📍Where to Start
Start with the new `xmtp_archive` crate's main entry point in [xmtp_archive/src/lib.rs](https://github.com/xmtp/libxmtp/pull/2019/files#diff-3dd8422e592fea2b7df162cc4d08f9ef775b5030b1d451e1b5ece24142c71598) which defines the public API for the archive functionality.

----

_[Macroscope](https://app.macroscope.com) summarized 1dc8329._


nplasterer (approved):
In favor of pulling archives into it's own crate. Just want to make sure this events table isn't going to slow down any of our calls? You mind bench marking something to make sure writing to this table on certain actions doesn't cause any performance issues?

xmtp_db/src/encrypted_store/client_events.rs
```diff
@@ -0,0 +1,93 @@
...
+#[diesel(primary_key(created_at_ns))]
+pub struct ClientEvents {
+    pub created_at_ns: i64,
+    pub details: serde_json::Value,
+}
+
+impl_store!(ClientEvents, client_events);
+
+impl ClientEvents {
+    pub fn track<C: ConnectionExt>(db: &DbConnection<C>, event: impl AsRef<ClientEvent>) {
```
insipx, line 21: Are these used anywhere yet? 

It would be nice to have a configuration option on the client builder or somewhere where we can enable/disable the stats feature. Sort of like the ApiDebugWrapper for the ApiClient. It probably won't have a huge impact on performance, but it would be nice to have that safety switch that could be useful for benchmarks/etc.

insipx:
more crates good

xmtp_db/src/encrypted_store/client_events.rs
codabrink: Yeah, great callout.  I thought about that last night too. I have a follow-up PR that will add a config.

We track in 3 places right now, client creation, welcomes, and queue intent. Just a start.

insipx (approved):

xmtp_archive/src/importer.rs
```diff
@@ -0,0 +1,101 @@
...
+            if element_len != 0 && this.decoded.len() >= element_len {
+                let decrypted = this
+                    .cipher
+                    .decrypt(&this.nonce, &this.decoded[..element_len])?;
+                let element = BackupElement::decode(&*decrypted);
+                this.decoded.drain(..element_len);
+                return Poll::Ready(Some(element.map_err(ArchiveError::from)));
+            }
+
+            if amount == 0 && this.decoded.is_empty() {
```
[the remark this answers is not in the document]
codabrink: No, we need decoded to also be empty before ending the stream.

Issue #2022, opened 2025-05-20
Do not call network when passing inboxId to client create


nplasterer:
When building a client do not call the network for identity updates if an inboxId is passed.

PR #2027, opened 2025-05-22
Do not call the network when building a clinet

nplasterer:
Fixes https://github.com/xmtp/libxmtp/issues/2022

Give the ability to bypass a network request on client build for offline usage.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -139,6 +139,7 @@ pub async fn create_client(
     legacy_signed_private_key_proto: Option<Vec<u8>>,
     device_sync_server_url: Option<String>,
     device_sync_mode: Option<FfiSyncWorkerMode>,
+    allow_offline: Option<bool>,
 ) -> Result<Arc<FfiXmtpClient>, GenericError> {
     let ident = account_identifier.clone();
     init_logger();
@@ -180,6 +181,7 @@ pub async fn create_client(
         .api_client(Arc::unwrap_or_clone(api).0)
         .enable_api_debug_wrapper()?
         .with_remote_verifier()?
+        .with_allow_offline(allow_offline)?
         .store(store);
 
     if let Some(sync_worker_mode) = device_sync_mode {
--- a/xmtp_mls/src/builder.rs
+++ b/xmtp_mls/src/builder.rs
@@ -48,6 +48,7 @@ pub struct ClientBuilder<ApiClient, Db = xmtp_db::DefaultStore> {
     device_sync_server_url: Option<String>,
     device_sync_worker_mode: SyncWorkerMode,
     version_info: VersionInfo,
+    allow_offline: bool,
 }
 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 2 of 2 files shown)


xmtp_mls/src/builder.rs
```diff
@@ -153,13 +157,15 @@ impl<ApiClient, Db> ClientBuilder<ApiClient, Db> {
...
-        .await?;
+        if !allow_offline {
+            // get sequence_id from identity updates and loaded into the DB
+            load_identity_updates(
+                &api_client,
+                provider.db(),
+                vec![identity.inbox_id.as_str()].as_slice(),
+            )
+            .await?;
+        }
```
nplasterer, line 168: Really want to make sure there aren't any downstream consequences for bypassing this identity update load some times when a client is loading an already locally stored client. cc @richardhuaaa

bindings_ffi/src/mls.rs
```diff
@@ -139,6 +139,7 @@ pub async fn create_client(
     legacy_signed_private_key_proto: Option<Vec<u8>>,
     device_sync_server_url: Option<String>,
     device_sync_mode: Option<FfiSyncWorkerMode>,
+    allow_offline: Option<bool>,
```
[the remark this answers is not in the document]
nplasterer: Maybe a dumb rust question but how do you set a default if you don't make it nullable? This should really always be false unless someone explicitly sets it to true in very rare cases

xmtp_mls/src/builder.rs
richardhuaaa: This is the original PR that added this here, which doesn't add much context sadly: https://github.com/xmtp/libxmtp/pull/737/files

We should do some testing, but specifically in the case that this is an existing client/database that was closed and is launching again, I think it's likely okay to skip this. My logic: this should be no different to the case where the existing client never closed and re-launched - we are picking back up from exactly where we were previously. The rest of the code in this method doesn't seem to depend on the identity being up-to-date, and the identity will be refreshed soon enough when `add_missing_installations` is called.

I don't fully understand the context behind #2022 though - don't we make other network requests aside from this one? Why does the integrator need this?

nplasterer: It's a little confusing because libxmtp looks a little different from the abstraction at the SDK level. At the SDK level we have two methods create and build. At the time of create the inboxId is unknown so it makes a network request to find or create an inboxId from the given identity. On create most likely the user does not have a local database so it will also request a signature requiring network requests. The assumption with build is that the local database already exists and no signature will be required so build does not take a signing key and if an inboxId is provided instead of an address it will do the create directly without making a request to find the inboxId. That leaves this spot in particular is the only place in a true build that makes a network request.

I have 1 test here but now that I have a little more time I'm going to beef it up a bit and can build android bindings for extra validatation.

The integrator request here is that when someone opens the messaging inbox that it loads instantly since it's all stored locally. This is the case for the list view however there is a delay of up to a second or more of creating the client because of this network request which makes the inbox appear delayed as no methods can be called on client until the client is built.

richardhuaaa: Ahh gotcha, both of those pieces make perfect sense. For the latter, I guess when the user already has an account, the app should be able to load and query existing groups/messages from the libxmtp database even with no internet connection.

richardhuaaa (approved):
It could be nice to modify an existing unit test inside of the mls crate to just see what happens when you make one of these clients and do anything with it, but it's probably fine either way!

nplasterer:
Built on android and validated that the only thing that calls the network is the device sync worker on build which does not block the build from completing.

PR #2044, opened 2025-05-27
fix hpke error by rotating keys gracefully

mchenani:


xmtp_db/migrations/2025-05-27-201031_add_delete_at_and_rotate_in_to_keys/up.sql
```diff
@@ -0,0 +1,7 @@
+-- Add a column to schedule deletion of key packages
+ALTER TABLE key_package_history
+    ADD COLUMN delete_in BIGINT;
+
+-- Add a column to schedule rotation of key packages
+ALTER TABLE key_package_history
+    ADD COLUMN rotate_in BIGINT;
```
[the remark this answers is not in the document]
insipx: maybe we can rename the identity table to `client_metadata`, which is just a bunch of rows of metadata like identity and the table is constrained to having 1 row

xmtp_db/migrations/2025-05-20-195531_create_stats_table/up.sql
```diff
@@ -1,7 +1,7 @@
 CREATE TABLE events (
     created_at_ns BIGINT NOT NULL,
     group_id BLOB,
-    event String NOT NULL,
+    event TEXT NOT NULL,
```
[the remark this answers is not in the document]
codabrink: This is a typo from an earlier PR that I made that needed to be fixed for migrations to run. It was invalid syntax, so we can be sure nobody ran it.

xmtp_db/migrations/2025-05-27-201031_add_delete_at_and_rotate_in_to_keys/up.sql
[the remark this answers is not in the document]
codabrink: `user_preferences` is more or less the same thing. We could probably merge the two. (In another PR)

xmtp_mls/src/groups/key_package_cleaner_worker.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    /// Check if we need to rotate the keys and upload new keypackage if the las one rotate in has passed
+    async fn rotate_last_key_package_if_needed(&mut self) -> Result<(), KeyPackagesCleanerError> {
+        self.client.rotate_and_upload_key_package(true).await?;
+        Ok(())
+    }
+    
+    async fn run(&mut self) -> Result<(), KeyPackagesCleanerError> {
+        let mut intervals = xmtp_common::time::interval_stream(INTERVAL_DURATION);
+        while (intervals.next().await).is_some() {
```
codabrink, line 153: Can we slow the interval to 5s or maybe 10s?

xmtp_mls/src/groups/key_package_cleaner_worker.rs
```diff
@@ -0,0 +1,159 @@
...
+        Self {
+            client,
+            init: OnceCell::new(),
+        }
+    }
+    pub(crate) fn spawn_worker(mut self) {
+        xmtp_common::spawn(None, async move {
+            let inbox_id = self.client.inbox_id().to_string();
+            let installation_id = hex::encode(self.client.installation_public_key());
+            while let Err(err) = self.run().await {
```
codabrink, line 74: We should abstract this out to a worker struct / trait implementation now that we have the same pattern in 3 places.

xmtp_db/migrations/2025-05-20-195531_create_stats_table/up.sql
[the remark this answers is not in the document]
richardhuaaa: Curious how it passed through CI in that case! But sounds good

xmtp_mls/src/identity.rs
```diff
@@ -572,26 +574,27 @@ impl Identity {
...
+            return Ok(());
+        }
 
         let kp = self.new_key_package(provider)?;
         let kp_bytes = kp.tls_serialize_detached()?;
         let hash_ref = serialize_key_package_hash_ref(&kp, provider.crypto())?;
         let history_id = conn.store_key_package_history_entry(hash_ref.clone())?.id;
+        
         match api_client.upload_key_package(kp_bytes, true).await {
             Ok(()) => {
```
richardhuaaa, line 594: We might need to clear the `rotate_in` here, or we will keep trying to rotate every time

mchenani: I did it in the Ok state, in case of error I think we should keep the rotate_in value to make sure at the next check the worker catch it and rotate the keys.

xmtp_mls/src/groups/key_package_cleaner_worker.rs
codabrink: A window of 5-10s should be fine, yeah? If the user closes the app and comes back in 30 minutes, the kp won't be deleted for 30 minutes.

xmtp_mls/src/groups/key_package_cleaner_worker.rs
```diff
@@ -0,0 +1,165 @@
...
+                tracing::info!("Running worker..");
+                if err.db_needs_connection() {
+                    tracing::warn!(
+                        inbox_id,
+                        installation_id,
+                        "Pool disconnected. task will restart on reconnect"
+                    );
```
richardhuaaa, line 81: @codabrink, what causes the task to restart on reconnect? I notice we break out of the retry loop in this branch. Are we re-creating the client from scratch in that scenario?

richardhuaaa (approved):
So clean and well-tested, well done!

Only some nitpicks left, I'll start working on moving the cursor on HPKE error as a separate PR in parallel to this.

PR #2045, opened 2025-05-28
Prevent concurrent use of identity in xdbg

mkysel:
### Implement thread-safe identity locking mechanism to prevent concurrent use of identity in xdbg message generation
Implements a global locking mechanism using `lazy_static` and `Arc<Mutex<HashMap>>` to ensure each identity can only be used by one worker thread at a time during message generation. The changes include:

• Added `lazy_static` dependency to enable static variable initialization
• Created `IDENTITY_LOCKS` global variable storing locks keyed by `inbox_id` 
• Implemented `get_identity_lock` function to retrieve or create locks for specific identities
• Modified `send_many_messages` method in [xmtp_debug/src/app/generate/messages.rs](https://github.com/xmtp/libxmtp/pull/2045/files#diff-eeaf742c78c261e05257950d12daa2ec07c1dfe459a732f2ee1484f184da9371) to acquire identity locks before usage
• Added new `dbg` build profile with debugging configurations

#### 📍Where to Start
Start with the `send_many_messages` method in [xmtp_debug/src/app/generate/messages.rs](https://github.com/xmtp/libxmtp/pull/2045/files#diff-eeaf742c78c261e05257950d12daa2ec07c1dfe459a732f2ee1484f184da9371) to see how the identity locking mechanism is implemented.

#### Changes since #2045 opened

- Replaced `lazy_static` crate with `std::sync::OnceLock` for thread-safe lazy initialization [fa40a85]
----

_[Macroscope](https://app.macroscope.com) summarized fa40a85._

### DIFF
--- a/xmtp_debug/Cargo.toml
+++ b/xmtp_debug/Cargo.toml
@@ -22,6 +22,7 @@ fs_extra = "1.3"
 hex.workspace = true
 indicatif = "0.17"
 k256 = "0.13"
+lazy_static = "1.5.0"
 lipsum = "0.9"
 miniserde = "0.1"
 openmls.workspace = true
@@ -54,4 +55,3 @@ xmtp_id.workspace = true
 xmtp_mls.workspace = true
 xmtp_proto.workspace = true
 xxhash-rust = { version = "0.8", features = ["xxh3"] }
-lazy_static = "1.5.0"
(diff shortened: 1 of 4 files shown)


xmtp_debug/Cargo.toml
```diff
@@ -22,6 +22,7 @@ fs_extra = "1.3"
 hex.workspace = true
 indicatif = "0.17"
 k256 = "0.13"
+lazy_static = "1.5.0"
```
insipx, line 25: rust standard library has a [LazyCell](https://doc.rust-lang.org/std/cell/struct.LazyCell.html) and [LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) that replaces uses of lazy static

insipx (approved):

Issue #1371, opened 2024-12-04
Formalize Workers into Trait


insipx:
Device + Consent sync operate with `LocalEvents`, which are events global to a single client instance. Workers are spawned on client instantiation and do work for the life of the client. Right now we only have a single worker for sync, but its reasonable to foresee other possible workers in the future. 

If this pattern becomes useful, it would be great to formalize workers into a trait, where workers can be dynamically registered to the client and spawned on instantiation together. This worker pattern so far is useful for separating concerns for tasks that don't have strict deadlines (like syncing)

insipx:
completed by @codabrink in #2058

PR #2058, opened 2025-06-02
Worker Abstraction

codabrink:
There are 3 workers with very similar boilerplate. This PR moves them into a single, easily manageable abstraction.

### DIFF
--- a/xmtp_mls/src/groups/device_sync/worker.rs
+++ b/xmtp_mls/src/groups/device_sync/worker.rs
@@ -1,11 +1,9 @@
 use super::{
-    handle::{SyncMetric, WorkerHandle},
     preference_sync::{store_preference_updates, PreferenceUpdate},
     DeviceSyncClient, DeviceSyncError, IterWithContent,
 };
 use crate::{
     client::ClientError,
-    configuration::WORKER_RESTART_DELAY,
     context::{XmtpContextProvider, XmtpMlsLocalContext},
     groups::{
         device_sync::{archive::insert_importer, default_archive_options},
@@ -13,15 +11,14 @@ use crate::{
         GroupError,
     },
     subscriptions::{LocalEvents, StreamMessages, SubscribeError, SyncWorkerEvent},
-    utils::worker::WorkerCore,
+    worker::{metrics::WorkerMetrics, WorkerCore},
 };
 use futures::{Stream, StreamExt};
 use std::{pin::Pin, sync::Arc};
 use tokio::sync::OnceCell;
 #[cfg(not(target_arch = "wasm32"))]
 use tokio_util::compat::TokioAsyncReadCompatExt;
-use tracing::{info_span, instrument, Instrument};
-use xmtp_api::test_utils::ApiClient;
+use tracing::instrument;
 use xmtp_archive::{exporter::ArchiveExporter, ArchiveImporter};
 use xmtp_db::{
     group_message::{MsgQueryArgs, StoredGroupMessage},
@@ -51,7 +48,7 @@ pub struct SyncWorker<ApiClient, Db> {
     stream: Pin<Box<dyn Stream<Item = Result<LocalEvents, SubscribeError>> + Send + Sync>>,
     init: OnceCell<()>,
 
-    handle: Arc<WorkerHandle<SyncMetric>>,
+    handle: Arc<WorkerMetrics<SyncMetric>>,
 }
 
 #[async_trait::async_trait]
@@ -115,7 +112,7 @@ where
             client,
             stream,
             init: OnceCell::new(),
-            handle: Arc::new(WorkerHandle::new()),
+            handle: Arc::new(WorkerMetrics::new()),
         }
     }
 }
@@ -125,7 +122,7 @@ where
     ApiClient: XmtpApi + 'static,
     Db: XmtpDb + 'static,
 {
-    pub(super) fn handle(&self) -> &Arc<WorkerHandle<SyncMetric>> {
+    pub(super) fn handle(&self) -> &Arc<WorkerMetrics<SyncMetric>> {
         &self.handle
     }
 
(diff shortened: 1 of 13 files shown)


insipx (approved):

insipx:
closes #1371

xmtp_mls/src/worker.rs
```diff
@@ -0,0 +1,130 @@
...
+impl<Core, Metric> WorkerManager for WorkerRunner<Core, Metric>
+where
+    Core: Send + Sync,
+    Metric: Send + Sync + 'static,
+{
+    fn sync_metrics(&self) -> Option<Arc<WorkerMetrics<SyncMetric>>> {
+        if std::any::TypeId::of::<Metric>() == std::any::TypeId::of::<SyncMetric>() {
+            self.metrics.as_ref().lock().clone().map(|arc| {
+                // This is safe because we verified Metric == SyncMetric
+                unsafe { std::mem::transmute::<Arc<WorkerMetrics<Metric>>, Arc<WorkerMetrics<SyncMetric>>>(arc.clone()) }
```
insipx, line 31: This feels like a shortcut/smell that shouldn't be necessary/anti-pattern. Why can't we define what a metric should return and enforce that, rather than do an unsafe cast to a concrete type?

Or, just enforce a concrete type on the trait, rather than an opaque type. Could we come up with a metric type generic enough to work for all workers? What metrics are useful here

I already approved so its too late for this PR, but we should aim to avoid unsafe, even if it looks OK on paper. This could fail down the road in ways we don't expect, and lead to some very difficult to debug issues. I would argue its just too fragile to trust unsafe

xmtp_mls/src/worker.rs
```diff
@@ -0,0 +1,130 @@
...
+        Core: Worker<ApiClient, Db, Metric> + 'static,
+    {
+        let metrics = Arc::new(Mutex::default());
+        let runner = WorkerRunner {
+            metrics: metrics.clone(),
+            spawn_fn: Box::new({
+                let client = client.clone();
+                let metrics = metrics.clone();
+                move || {
+                    Self::spawn_worker_internal(&client, &metrics);
```
insipx, line 66: do we need the entire Client here or could we get away with just an `Arc<XmtpMlsLocalContext<ApiClient, Db>>`

xmtp_mls/src/worker.rs
```diff
@@ -0,0 +1,130 @@
...
+where
+    Self: Send + Sync,
+    ApiClient: XmtpApi,
+    Db: xmtp_db::XmtpDb,
+    Metric: PartialEq + Hash + Send + Sync + 'static,
+{
+    type Error: NeedsDbReconnect + Debug + Send;
+
+    fn kind() -> WorkerKind;
+    fn init(client: &Client<ApiClient, Db>) -> Self;
```
insipx, line 117: I think it could be useful to refactor a bit here

Why can't the init method accept a `&self` parameter?

Then in the Worker registration, you could accept a `W: Worker`. Then, kind would also be able to accept self by reference, which would allow this Worker Trait to be Boxable

This would further allow you to unravel the metrics, since you could define a monolithic metric type or even include the metrics as an associated type parameter on the Worker trait


Each worker could hold a reference to the XmtpMlsLocalContext, which would allow them to use any of the methods they need on the client or other structs


Since the type generics exist on the concrete type your implementing on, this would allow you to elide all the generics on the trait itself as well

xmtp_mls/src/groups/device_sync/worker.rs
```diff
@@ -49,73 +49,45 @@ pub struct SyncWorker<ApiClient, Db> {
...
+    fn kind() -> WorkerKind {
+        WorkerKind::DeviceSync
+    }
+
+    fn metrics(&self) -> Option<Arc<WorkerMetrics<SyncMetric>>> {
+        Some(self.metrics.clone())
+    }
+
+    fn init(client: &Client<ApiClient, Db>) -> Self {
+        let context = &client.context;
```
insipx, line 72: client.context is available on `self\.client`

xmtp_mls/src/groups/device_sync/worker.rs
```diff
@@ -49,73 +49,45 @@ pub struct SyncWorker<ApiClient, Db> {
...
 where
-    ApiClient: XmtpApi + Send + Sync + 'static,
-    Db: XmtpDb + Send + Sync + 'static,
+    ApiClient: XmtpApi + 'static,
+    Db: XmtpDb + 'static,
 {
-    pub(super) fn new(context: Arc<XmtpMlsLocalContext<ApiClient, Db>>) -> Self {
+    type Error = DeviceSyncError;
+
+    fn kind() -> WorkerKind {
```
insipx, line 63: making this self would make worker Object-Safe

xmtp_mls/src/worker.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub(crate) fn spawn_worker_internal<ApiClient, Db>(
+        client: &Client<ApiClient, Db>,
+        metrics: &Arc<Mutex<Option<Arc<WorkerMetrics<Metric>>>>>,
+    ) where
+        ApiClient: XmtpApi + 'static,
+        Db: XmtpDb + Send + Sync + 'static,
+        Core: Worker<ApiClient, Db, Metric> + 'static,
+    {
+        let mut core = Core::init(client);
+        *metrics.lock() = core.metrics();
```
insipx, line 87: are we overwriting the metrics here/I'm having a hard time understanding how the metrics connect back to each worker here

xmtp_mls/src/worker.rs
codabrink: The device sync worker relies on the context, but the other two workers still rely on the client, and I didn't want to make refactoring them a part of this PR. Definitely something we can make a PR for as a followup.

PR #2067, opened 2025-06-03
Add GroupUpdated message for members added to group

rygine:
Closes https://github.com/xmtp/libxmtp/issues/1784

### Group Join Behavior Matrix

| Scenario                                                      | Should Send Group Update? |
|---------------------------------------------------------------|---------------------------|
| New Inbox                                                     | Yes                         |
| Active Installations                                          | Yes                      |
| New Installations while you're in the group (no history sync) | Yes                        |
| New Installations while you're in the group (with history sync) | Yes but should be improved in the future                      |
| New Group but a duplicate DM                                  | Yes but thread should only show 1  |
| Creator of the group with active installation                 | Yes                        |

---

**Notes:**

- Matches how WhatsApp handles group updates.


nplasterer:
Some interesting edge cases being caught here in the tests that we should think about. It feels like you probably don't want to see this message if you are an installation being added to a group only if you are a new member. So you will probably need to add a check somewhere to make sure of that. 🤔 

I imagine this is going to come with a lot of test failures but I do still think it would be useful to write a well thought out test that makes sure the person being added sees the message and that a new installation processing a welcome does not.

nplasterer:
Built bindings and ran your branch on my android tests. A few things that I saw of concern. When I was stitching dms together the stitched dms would have multiple groupUpdate codecs. Something should probably write a test for to catch in libxmtp. https://github.com/xmtp/xmtp-android/blob/main/library/src/androidTest/java/org/xmtp/android/library/DmTest.kt#L95-L96 maybe dm stitching shouldn't stitch group updated codecs 🤔 

I also saw that new installations where getting the group update codec. I think they should not. If I log into my iPad I don't expect to see a Naomi added Naomi to the group message. I think I need to write a better test for this but I believe that was what I was seeing when new installations joined. I think we only want to send it when a new inbox joins. 🤔

xmtp_mls/src/groups/device_sync.rs
```diff
@@ -255,7 +255,7 @@ where
     pub async fn add_new_installation_to_groups(&self) -> Result<(), DeviceSyncError> {
         let groups = self.mls_store.find_groups(GroupQueryArgs {
             activity_after_ns: Some(now_ns() - NS_IN_DAY * 90),
-            consent_states: Some(vec![ConsentState::Allowed]),
+            consent_states: Some(vec![ConsentState::Allowed, ConsentState::Unknown]),
```
nplasterer, line 258: I added this from a comment I left to Dakota. https://github.com/xmtp/libxmtp/pull/2053/files#r2132902595

Can revert but I do think we want this.

bindings_ffi/src/mls.rs
```diff
@@ -4704,8 +4704,8 @@ mod tests {
             .find_messages(FfiListMessagesOptions::default())
             .await
             .unwrap();
-        assert_eq!(bo_messages1.len(), 0);
-        assert_eq!(bo_messages5.len(), 0);
+        assert_eq!(bo_messages1.len(), 1);
+        assert_eq!(bo_messages5.len(), 1);
```
nplasterer, line 4787: Alix created the group but bo should now see the groupUpdate message as well.

bindings_ffi/src/mls.rs
```diff
@@ -4845,7 +4845,7 @@ mod tests {
             .find_messages(FfiListMessagesOptions::default())
             .await
             .unwrap();
-        assert_eq!(bo_messages.len(), 9);
+        assert_eq!(bo_messages.len(), 10);
```
nplasterer, line 4927: Alix created the group but bo should now see the groupUpdate message as well.

bindings_ffi/src/mls.rs
```diff
@@ -5127,9 +5127,9 @@ mod tests {
             .await
             .unwrap();
 
-        assert_eq!(caro_messages.len(), 5);
+        assert_eq!(caro_messages.len(), 6);
         assert_eq!(alix_messages.len(), 6);
-        assert_eq!(bo_messages.len(), 5);
+        assert_eq!(bo_messages.len(), 6);
         // Bo 2 only sees three messages since it joined after the first 2 were sent
         assert_eq!(bo2_messages.len(), 3);
```
nplasterer: Bo2 should not get a group update change so this should not increase

bindings_ffi/src/mls.rs
```diff
@@ -7975,10 +7976,10 @@ mod tests {
...
             .await
             .unwrap();
-        assert_eq!(bo_messages.len(), 3, "Bo should see 3 messages after sync");
+        assert_eq!(bo_messages.len(), 4, "Bo should see 3 messages after sync");
         assert_eq!(
             alix_messages.len(),
-            3,
+            4,
             "Alix should see 3 messages after sync"
         );
```
nplasterer: I think this is still a bug we shouldn't be seeing multiple group update messages in stitched dms.

bindings_ffi/src/mls.rs
```diff
@@ -8057,8 +8058,8 @@ mod tests {
             .await
             .unwrap();
 
-        assert_eq!(final_bo_messages.len(), 5, "Bo should see 5 messages");
-        assert_eq!(final_alix_messages.len(), 5, "Alix should see 5 messages");
+        assert_eq!(final_bo_messages.len(), 6, "Bo should see 5 messages");
+        assert_eq!(final_alix_messages.len(), 6, "Alix should see 5 messages");
```
nplasterer: Same here I think this should still be 5. We should not stitch the group update message.

xmtp_mls/src/groups/mod.rs
```diff
@@ -693,6 +698,77 @@ where
...
             );
 
+            // Create a GroupUpdated payload
+            let current_credential = context.credential();
+            let other_installations_with_same_inbox = mls_group
+                .members()
+                .filter(|member| member.credential == current_credential)
+                .count();
+
+            if other_installations_with_same_inbox <= 1 && metadata.creator_inbox_id != context.inbox_id() {
```
nplasterer: If this is the first installation in the group and it is not the creator of the group. Then add a groupUpdated codec.

xmtp_db/src/encrypted_store/group_message.rs
```diff
@@ -317,7 +317,45 @@ impl<C: ConnectionExt> DbConnection<C> {
...
+
+        // Check if this is a DM group
+        let is_dm = self.raw_query_read(|conn| {
+            groups_dsl::groups
+                .filter(groups_dsl::id.eq(group_id))
+                .select(groups_dsl::conversation_type)
+                .first::<ConversationType>(conn)
+        })? == ConversationType::Dm;
+
+        // If DM, retain only one GroupUpdated message (the oldest)
```
nplasterer, line 330: When stitching dms only keep one group updated message.

xmtp_mls/src/groups/mod.rs
nplasterer: Updated member -> installation - don't send a message
added member -> just got added - send message
removed member -> don't need

nplasterer: validate_initial_group_membership

xmtp_db/src/encrypted_store/group_message.rs
nplasterer: We should insert this as the first message not the last.

xmtp_mls/src/groups/device_sync.rs
nplasterer: If we don't do this we run the risk of an installation not being added to a group it should be added to. One that maybe was in request and would be approved eventually.

xmtp_db/src/encrypted_store/group_message.rs
```diff
@@ -317,7 +317,48 @@ impl<C: ConnectionExt> DbConnection<C> {
...
+        let messages = if is_dm && args.limit.is_none() {
```
mchenani: Q: if the limit is not none, but the groups holds 10 messages, we kinda still gonna show the GroupUpdate message multiple times, right?

mchenani: Nit: since this is on the query messages code, it would be calculated every time, for now I would say it's ok, but we need to think about this kind of complexities to not make message querying slower.

mchenani (approved):
LGTM, just added a nit and question, but not blockers!

xmtp_db/src/encrypted_store/group_message.rs
nplasterer: 💯 I thought about this as well. Probably something we need to think about eventually

PR #1852, opened 2025-04-10
Decrypt welcomes with different ciphersuites

neekolas:

PR #2068, opened 2025-06-04
Add more context to welcome and key package logs

richardhuaaa:
Adds more context to logs so that we can debug any hpke or cleaner worker errors more effectively.

This depends on https://github.com/xmtp/libxmtp/pull/1852 landing first. There's one failing unit test here which is also failing in the base branch.


richardhuaaa:
Added as part of fork fixes

PR #2077, opened 2025-06-05
Auto-consent to own group

codabrink:
### Add automatic consent to groups created by the user's own inbox in MlsGroup.process_welcome method
The `MlsGroup.process_welcome` method now automatically sets consent state to 'Allowed' when processing welcome messages for groups created by the user's own inbox. A new `quietly_update_consent_state` method is added to [xmtp_mls/src/groups/mod.rs](https://github.com/xmtp/libxmtp/pull/2077/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3) that updates consent state without broadcasting changes, and the existing `update_consent_state` method is refactored to use this new method internally. Test coverage is added in [xmtp_mls/src/groups/tests/test_consent.rs](https://github.com/xmtp/libxmtp/pull/2077/files#diff-0de642dbf701a4bac69a30e4a8637303617871b4986fb0d561299467551d2b7c) to verify the auto-consent functionality works correctly for self-created groups while maintaining existing behavior for groups created by others.

#### 📍Where to Start
Start with the `process_welcome` method in [xmtp_mls/src/groups/mod.rs](https://github.com/xmtp/libxmtp/pull/2077/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3) to see how the auto-consent logic is implemented when processing welcome messages.

----

_[Macroscope](https://app.macroscope.com) summarized 68973b8._

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -2272,8 +2272,8 @@ impl FfiConversation {
         self.inner.added_by_inbox_id().map_err(Into::into)
     }
 
-    pub fn group_metadata(&self) -> Result<Arc<FfiConversationMetadata>, GenericError> {
-        let metadata = self.inner.metadata()?;
+    pub async fn group_metadata(&self) -> Result<Arc<FfiConversationMetadata>, GenericError> {
+        let metadata = self.inner.metadata().await?;
         Ok(Arc::new(FfiConversationMetadata {
             inner: Arc::new(metadata),
         }))
@@ -4129,7 +4129,7 @@ mod tests {
         message_callbacks.wait_for_delivery(None).await.unwrap();
         assert_eq!(bo.provider.db().intents_published(), 4);
 
-        assert_eq!(message_callbacks.message_count(), 6);
+        assert_eq!(message_callbacks.message_count(), 5);
 
         stream_messages.end_and_wait().await.unwrap();
         assert!(stream_messages.is_closed());
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -1492,7 +1492,7 @@ pub(crate) mod tests {
             .unwrap();
 
         // Verify DM was created with correct properties
-        let metadata = dm1.metadata().unwrap();
+        let metadata = dm1.metadata().await.unwrap();
         assert_eq!(
             metadata.dm_members.clone().unwrap().member_one_inbox_id,
             client1.inbox_id()
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -709,7 +709,7 @@ where
             );
 
             // If this group is created by us - auto-consent to it.
-            if context.inbox_id() == group.metadata()?.creator_inbox_id {
+            if context.inbox_id() == metadata.creator_inbox_id {
                 group.quietly_update_consent_state(ConsentState::Allowed)?;
             }
 
@@ -1073,7 +1073,7 @@ where
                 length: MAX_GROUP_NAME_LENGTH,
             });
         }
-        if self.metadata()?.conversation_type == ConversationType::Dm {
+        if self.metadata().await?.conversation_type == ConversationType::Dm {
             return Err(MetadataPermissionsError::DmGroupMetadataForbidden.into());
         }
         let intent_data: Vec<u8> =
@@ -1117,7 +1117,7 @@ where
     ) -> Result<(), GroupError> {
         self.ensure_not_paused().await?;
 
-        if self.metadata()?.conversation_type == ConversationType::Dm {
+        if self.metadata().await?.conversation_type == ConversationType::Dm {
             return Err(MetadataPermissionsError::DmGroupMetadataForbidden.into());
         }
         if permission_update_type == PermissionUpdateType::UpdateMetadata
@@ -1167,7 +1167,7 @@ where
             });
         }
 
-        if self.metadata()?.conversation_type == ConversationType::Dm {
+        if self.metadata().await?.conversation_type == ConversationType::Dm {
             return Err(MetadataPermissionsError::DmGroupMetadataForbidden.into());
         }
         let intent_data: Vec<u8> =
@@ -1204,7 +1204,7 @@ where
             });
         }
 
-        if self.metadata()?.conversation_type == ConversationType::Dm {
+        if self.metadata().await?.conversation_type == ConversationType::Dm {
             return Err(MetadataPermissionsError::DmGroupMetadataForbidden.into());
         }
         let intent_data: Vec<u8> =
@@ -1359,7 +1359,7 @@ where
 
     /// Retrieves the conversation type of the group from the group's metadata extension.
     pub async fn conversation_type(&self) -> Result<ConversationType, GroupError> {
-        let metadata = self.metadata()?;
+        let metadata = self.metadata().await?;
         Ok(metadata.conversation_type)
     }
 
@@ -1369,7 +1369,7 @@ where
         action_type: UpdateAdminListType,
         inbox_id: String,
     ) -> Result<(), GroupError> {
-        if self.metadata()?.conversation_type == ConversationType::Dm {
+        if self.metadata().await?.conversation_type == ConversationType::Dm {
             return Err(MetadataPermissionsError::DmGroupMetadataForbidden.into());
         }
         let intent_action_type = match action_type {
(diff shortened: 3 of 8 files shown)


xmtp_mls/src/groups/mod.rs
```diff
@@ -700,12 +701,19 @@ where
...
+            let group = Self::new(
+                context.clone(),
                 stored_group.id,
                 stored_group.dm_id,
                 stored_group.created_at_ns,
-            ))
+            );
+
+            // If this group is from our own inbox - auto-consent to it.
+            if context.inbox_id() == stored_group.added_by_inbox_id {
```
nplasterer: I think we should check `creator_inbox_id` instead right? I don't think we necessarily want to consent to every group you add yourself to?

xmtp_mls/src/client.rs
```diff
@@ -1075,7 +1075,7 @@ pub(crate) mod tests {
             })
             .unwrap();
 
-        assert_eq!(messages.len(), 3);
+        assert_eq!(messages.len(), 4);
```
nplasterer: Was this just a flaky test?

codabrink: No, `update_consent_state` is emitting a new message to the sync group because we're updating the consent the conventional high-level way, so this is consent being streamed. It's technically a harmless message, but we can avoid the message if we just store the record rather than using the helper, which thinking about it a bit more, is probably worth doing.

codabrink: Just added a quiet update consent function so this no longer broadcasts.

xmtp_mls/src/groups/mod.rs
codabrink: I would think we would want to auto-consent if we added ourselves yeah? A self-add is an implicit consent?

xmtp_mls/src/groups/mod.rs
```diff
@@ -1491,15 +1507,13 @@ where
...
 
     /// Get the `GroupMetadata` of the group.
-    pub async fn metadata(&self) -> Result<GroupMetadata, GroupError> {
-        self.load_mls_group_with_lock_async(|mls_group| {
-            futures::future::ready(
-                extract_group_metadata(&mls_group)
-                    .map_err(MetadataPermissionsError::from)
-                    .map_err(Into::into),
-            )
+    pub fn metadata(&self) -> Result<GroupMetadata, GroupError> {
```
codabrink: `metadata` was needlessly async. Cleaned that up.

bindings_ffi/src/mls.rs
```diff
@@ -4129,7 +4129,7 @@ mod tests {
         message_callbacks.wait_for_delivery(None).await.unwrap();
         assert_eq!(bo.provider.db().intents_published(), 4);
 
-        assert_eq!(message_callbacks.message_count(), 5);
+        assert_eq!(message_callbacks.message_count(), 6);
```
codabrink: That was a weird quirk. Making `.metadata()` non-async increased this to 6.. strangely enough. It appears that with the thread no longer yielding, the thread retains control until the 6th message is processed.

xmtp_mls/src/groups/mod.rs
nplasterer: O this is great! Means I can clean up my sdks as well. Might be some places in the FFI bindings we can clean up 🤔

bindings_ffi/src/mls.rs
```diff
@@ -2273,7 +2273,7 @@ impl FfiConversation {
     }
 
     pub async fn group_metadata(&self) -> Result<Arc<FfiConversationMetadata>, GenericError> {
```
nplasterer: Can we make this no longer async?

xmtp_mls/src/groups/tests/test_consent.rs
```diff
@@ -0,0 +1,32 @@
...
+        .create_group_with_inbox_ids(&[alix1.inbox_id()], None, None)
+        .await?;
+
+    alix1.sync_welcomes().await?;
+    let unwanted = alix1.group(&unwanted_bo.group_id)?;
+    // We were added by ourselves, but we did not consent to the group. The group should remain unconsented.
+    assert_eq!(unwanted.consent_state()?, ConsentState::Unknown);
+
+    tester!(alix2, from: alix1);
+    unwanted.send_message(b"hi unwanted group").await?;
```
nplasterer: I thought that sending a message in a group auto consented to the group... did I make that up? Was that a V2 only thing. 👀

codabrink: It does, but the sync worker is not enabled in the test, so `alix2` will never know that. I just need someone to call `.maybe_update_installations`. I can make bo the one to send the message and call `.maybe_update_installations`.

bindings_ffi/src/mls.rs
codabrink: Done

bindings_ffi/src/mls.rs
```diff
@@ -2272,8 +2272,8 @@ impl FfiConversation {
         self.inner.added_by_inbox_id().map_err(Into::into)
     }
 
-    pub async fn group_metadata(&self) -> Result<Arc<FfiConversationMetadata>, GenericError> {
-        let metadata = self.inner.metadata().await?;
+    pub fn group_metadata(&self) -> Result<Arc<FfiConversationMetadata>, GenericError> {
+        let metadata = self.inner.metadata()?;
```
nplasterer: @mchenani mind gut checking this. I know you were the one who made metadata async awhile back maybe for forking?

xmtp_mls/src/groups/mod.rs
```diff
@@ -700,12 +701,19 @@ where
...
                 stored_group.id,
                 stored_group.dm_id,
                 stored_group.created_at_ns,
-            ))
+            );
+
+            // If this group is created by us - auto-consent to it.
+            if context.inbox_id() == group.metadata()?.creator_inbox_id {
+                group.quietly_update_consent_state(ConsentState::Allowed)?;
+            }
```
nplasterer: @richardhuaaa you mind gut checking me here. What we are saying is that anytime your installation is added to a group that you created your self that you would like to mark it as allowed. So the case we are trying to cover is I create a group on my phone and it auto consents to allowed because I created. However whenever I open my iPad it shows in the security line because the consent state is unknown when I processed the welcome. In Convos in particular the conversation will usually move to allowed after a refresh or the consent preference is streamed. But it is a odd glitch that the conversation shows in the security line for a period.

The edge case here that this could cause is if I create a large group and then it changes and I decide to block the group when a my new installation gets added to it in the future it will show as allowed in that inbox when it processes the welcome. This feels rare and maybe not unexpected as you were the creator of the group and can block it again.
The other edge case is when you add all your old conversations on a new installation on initially login. It's possible if the consent doesn't get transferred because the history serve is turned off or fails for those conversations on initially load you could say a weird experience were some of your conversations are allowed and some are not that you might expect to be. You'd get all your created conversations correctly but all the ones others created would show in your security line. Seems like an edge case but something to think about.

bindings_ffi/src/mls.rs
mchenani: The reason was we didn't have groups sync lock, since now we have, that would be ok.

nplasterer (approved):
I feel good about this just might consider this comment https://github.com/xmtp/libxmtp/pull/2053/files#r2132902595 might be better to start less restrictive and then get more restrictive with time. Hopefully people are better at blocking conversations and spam doesn't stay in unknown for long.

xmtp_mls/src/groups/mod.rs
richardhuaaa: > So the case we are trying to cover is I create a group on my phone and it auto consents to allowed because I created. However whenever I open my iPad it shows in the security line because the consent state is unknown when I processed the welcome.

So this is a fix for a specific scenario in the more general problem of delays in consent sync, which also covers other situations, such as consenting to a group created by someone else. Does this perceived delay in consent sync affect other developers other than Convos? If there is a delay, is it easy to improve it directly within Convos for example, or is it an issue with the protocol? It doesn't feel like there's any theoretical reason that the delay from consent sync can't be extremely short. I'm not opposed to low-cost workaround fixes to larger problems, but just checking that it's worth it.

> The edge case here that this could cause is if I create a large group and then it changes and I decide to block the group when a my new installation gets added to it in the future it will show as allowed in that inbox when it processes the welcome. 
> The other edge case is when you add all your old conversations on a new installation on initially login...

We're sort of adding a 'heuristic' to the SDK with this change - "If I created a group, it will always be allowed". There's a way to solve the first edge case - you could just check your user preferences at the moment you're about to apply the heuristic, and allow any previous explicit consent update to 'win'. Would that be an easy change? If that becomes complex, then yeah it would be worth double-checking if this heuristic is truly worth adding or not.

xmtp_mls/src/groups/mod.rs
richardhuaaa: Just double-checking, does changing the lock from async to sync mean it's going to synchronously wait, in the event that something else is holding the lock? We're not introducing risk of deadlock or perf issues right?

insipx: The async is needed here to hold onto the lock. If the sync version is used the group lock throws an error instead of waiting until the lock is free. We can't wait in a sync context because it would block the executor. Throwing an error w/o waiting might lead to more forks ( would need to verify). It was part / close to a past fork fix

bindings_ffi/src/mls.rs
insipx: returning 6 on main as well, sans sync metadata, definitely weird. Feels like maybe another race condition/deadlock somewhere, which were independently fixed in two ways 

https://github.com/xmtp/libxmtp/blob/main/bindings_ffi/src/mls.rs#L4151

insipx (changes requested):
The async lock is needed so the lock is actually acquired. if the lock isn't ready. the sync version throws an error instead of holding the lock, which might cause forks/issues with streaming

codabrink:
Yeap, hadn't merged it yet because I had the same thought. I'll update.

PR #2081, opened 2025-06-06
Make Events Good

codabrink:
### Redesign event system to use string-based event names with severity levels and disable events by default across all client bindings
- Replaces the `Event` enum with string-based event names and introduces an `EventLevel` enum with variants `None`, `Success`, `Warn`, `Error`, and `Fault` in [xmtp_db/src/encrypted_store/events.rs](https://github.com/xmtp/libxmtp/pull/2081/files#diff-fd762ac4955f0d9cb365948d8a28c84fe2d7c2eb4cd9cd51eb85a4c9d881499c)
- Changes default event tracking from enabled to disabled across all client creation functions in FFI, Node.js, and WebAssembly bindings
- Implements a new worker management system using `WorkerRunner` instead of `HashMap` for managing workers in [xmtp_mls/src/worker.rs](https://github.com/xmtp/libxmtp/pull/2081/files#diff-ea560e63ca633c03d1189957b9efca26de6aa33294f42c116a574e5bf259ce23)
- Adds `level` (INTEGER NOT NULL DEFAULT 0) and `icon` (TEXT) columns to the events table through database migration in [xmtp_db/migrations/2025-06-06-161143_add_level_to_event/up.sql](https://github.com/xmtp/libxmtp/pull/2081/files#diff-a350c4c919564ddb94dd579b74a57d8cc873e9bfbc8491e4b0a0b83e349aad22)
- Replaces direct `Events::track` calls with a `track!` macro throughout the codebase for event tracking
- Renames `device_sync()` method to `device_sync_client()` across all client implementations
- Reduces event retention period from 30 days to 15 days and removes automatic clearing of old events on client build
- Removes `PreferenceSyncService` and moves its functionality directly into `DeviceSyncClient`
- Simplifies envelope protocol buffer definition by removing `OriginatorPasskeySignature` variant and renaming `OriginatorEcdsaSignature` to `OriginatorSignature` in [xmtp_proto/src/gen/xmtp.xmtpv4.envelopes.rs](https://github.com/xmtp/libxmtp/pull/2081/files#diff-b1916f1e362dc674259fe8ead0047f6d60358508fbb3843286fc075b17ff623c)

#### 📍Where to Start
Start with the event system redesign in [xmtp_db/src/encrypted_store/events.rs](https://github.com/xmtp/libxmtp/pull/2081/files#diff-fd762ac4955f0d9cb365948d8a28c84fe2d7c2eb4cd9cd51eb85a4c9d881499c) to understand the core changes to event tracking, then examine the client builder changes in [xmtp_mls/src/builder.rs](https://github.com/xmtp/libxmtp/pull/2081/files#diff-2df8df9acb5405ca5c2dde2487e92ffc9a7daf6be38dd075a2a2d4878cc35ce6) to see how events are now disabled by default and how the new worker system is initialized.

----

_[Macroscope](https://app.macroscope.com) summarized 83ca119._


xmtp_mls/src/utils/events.rs
```diff
@@ -1,10 +1,296 @@
...
+///     label: "api_timeout",
+///     group_id: &group_id
+/// );
+/// ```
+///
+/// The error is stored in the event details as `{"error": "Debug representation"}`.
+/// See the `track!` macro documentation for all available parameters and options.
+#[macro_export]
+macro_rules! track_err {
+    ($result:expr, label: $label:expr, $(, $k:ident $(: $v:expr)?)*) => {
```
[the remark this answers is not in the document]
codabrink: That's the point.

xmtp_mls/src/builder.rs
```diff
@@ -144,10 +145,6 @@ impl<ApiClient, Db> ClientBuilder<ApiClient, Db> {
             disable_local_telemetry: disable_events,
```
nplasterer: I don't see this exposed out through the FFI Bindings? How does this ever get set to true?

codabrink: It's exposed now

xmtp_db/src/encrypted_store/events.rs
```diff
@@ -16,19 +26,65 @@ pub struct Events {
...
+            4 => Ok(EventLevel::Fault),
+            x => Err(format!("Unrecognized variant {}", x).into()),
+        }
+    }
 }
 
 impl_store!(Events, events);
 
-pub static EVENTS_ENABLED: AtomicBool = AtomicBool::new(true);
+pub static EVENTS_ENABLED: AtomicBool = AtomicBool::new(false);
```
nplasterer, line 78: Is there a reason we are defaulting this to false?

codabrink: It gets enabled on build here: https://github.com/xmtp/libxmtp/blob/2217e8cfddca3a29666f0fea65bcd90d7d1acc24/xmtp_mls/src/builder.rs#L217

It doesn't really matter what this value is initially set to here, the builder will set it to what it's configured to on build, it just needs an initial value.

xmtp_mls/src/utils/events.rs
```diff
@@ -1,10 +1,300 @@
...
+            Self::Client(s) => s.db_needs_connection(),
+        }
+    }
+}
+
+static EVENT_TX: LazyLock<Mutex<Option<mpsc::Sender<Events>>>> = LazyLock::new(Mutex::default);
+
+pub(crate) struct EventBuilder<'a, E, D> {
+    pub event: E,
+    pub details: D,
```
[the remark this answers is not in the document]
codabrink: I want it to be overwritten. The worker on the other end of the channel can be restarted.

nplasterer (approved):
I just want to repeat my comment about making sure this table doesn't get super large on accident. Might be worth putting a cap on the number rows this table can have instead of the arbitrary 30 days. Possible that forks might happen for people who don't open the app frequently outside of 30 days. Also possible power users could make this table very large very quickly.

PR #2087, opened 2025-06-09
make messages stream more readable

insipx:
### Refactor message streaming code to improve readability by extracting types and tests into separate modules
- Extracts `MessageStreamError`, `MessagePosition`, and new `GroupList` type from [stream_messages.rs](https://github.com/xmtp/libxmtp/pull/2087/files#diff-ee8c2d1c90650351a2245875b62af09f2c152c29643200276da5743cd59424e7) into [types.rs](https://github.com/xmtp/libxmtp/pull/2087/files#diff-6cce397f3eaea2a0d0532094908d527d2096e0e04c855000ec4d083ae7dc369a) module
- Moves test code from [stream_all.rs](https://github.com/xmtp/libxmtp/pull/2087/files#diff-4a3d32ed6005867d963b0eae023dc9c1fd8ffd9531413ba572f4287210a18cc1) into separate [tests.rs](https://github.com/xmtp/libxmtp/pull/2087/files#diff-da36cd09e25a94ff061dc4a491c213ce2a588ec230d68dc46575f927063176c7) module
- Replaces direct `HashMap<GroupId, MessagePosition>` usage in `StreamGroupMessages` with new `GroupList` abstraction
- Renames `MessagePosition::pos()` method to `last_streamed()` for clarity
- Adds custom `Debug` implementation for `GroupId` that displays bytes in hexadecimal format
- Updates Nix build configuration to add WebAssembly toolchain support and remove framework dependencies

#### 📍Where to Start
Start with the new `GroupList` type in [types.rs](https://github.com/xmtp/libxmtp/pull/2087/files#diff-6cce397f3eaea2a0d0532094908d527d2096e0e04c855000ec4d083ae7dc369a) to understand the extracted group tracking logic, then review how it's used in the `StreamGroupMessages` struct in [stream_messages.rs](https://github.com/xmtp/libxmtp/pull/2087/files#diff-ee8c2d1c90650351a2245875b62af09f2c152c29643200276da5743cd59424e7).

----

_[Macroscope](https://app.macroscope.com) summarized ec895ac._


insipx:
* **#2105** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2105?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2087** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2087?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2087?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

insipx:
Quick PR to increase readability of MessageStreaming module. Introduces a 'types' mod and encapsulates some more logic w.r.t getting the right cursor ids

xmtp_mls/src/subscriptions/stream_messages/types.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+}
+
+pub(super) trait Api {
+    /// get the latest message for a cursor
+    async fn query_latest_position(
+        &self,
+        group: &GroupId,
+    ) -> Result<MessagePosition, SubscribeError>;
+}
```
richardhuaaa, line 86: Out of curiosity, why do we need this as a separate trait?

xmtp_mls/src/subscriptions/stream_messages/types.rs
```diff
@@ -0,0 +1,153 @@
...
+#[derive(Clone, Debug, Default)]
+pub(super) struct GroupList {
+    list: HashMap<GroupId, MessagePosition>,
+}
+
+impl GroupList {
+    pub(super) async fn new(list: Vec<GroupId>, api: &impl Api) -> Result<Self, SubscribeError> {
+        let list = stream::iter(list)
+            .map(|group| async {
+                let position = api.query_latest_position(&group).await?;
```
richardhuaaa, line 114: Just understanding better, why do we need the latest position in each group, and why fetch this from the server instead of from local DB?

xmtp_mls/src/subscriptions/stream_messages.rs
```diff
@@ -322,14 +222,14 @@ where
...
+    #[tracing::instrument(level = "trace", skip(context, new_group), fields(new_group = hex::encode(&new_group)))]
     async fn subscribe(
-        client: &'a Arc<XmtpMlsLocalContext<ApiClient, Db>>,
+        context: &'a Arc<XmtpMlsLocalContext<ApiClient, Db>>,
         filters: Vec<GroupFilter>,
         new_group: Vec<u8>,
     ) -> Result<(MessagesApiSubscription<'a, ApiClient>, Vec<u8>, Option<u64>)> {
         // get the last synced cursor
-        let stream = client.api().subscribe_group_messages(filters).await?;
+        let stream = context.api().subscribe_group_messages(filters).await?;
```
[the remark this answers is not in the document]
richardhuaaa: @insipx not super important for this PR, but do we need the `new_group` param?

xmtp_mls/src/subscriptions/stream_messages.rs
```diff
@@ -521,9 +405,8 @@ where
             hex::encode(&group.group_id)
         );
         let this = self.as_mut().project();
-        this.group_list
-            .insert(group.group_id.clone().into(), MessagePosition::new(1, 1));
-        let future = Self::subscribe(self.context, self.filters(), group.group_id);
+        this.groups.add(&group.group_id, MessagePosition::new(1, 1));
```
richardhuaaa, line 408: Why is the message position `(1, 1)` here?

xmtp_mls/src/subscriptions/stream_messages/types.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    pub fn add(&mut self, group: impl AsRef<[u8]>, position: MessagePosition) {
+        self.list.insert(group.as_ref().to_vec().into(), position);
+    }
+
+    pub fn set(&mut self, group: impl AsRef<[u8]>, cursor: u64) {
+        if let Some(position) = self.list.get_mut(group.as_ref()) {
+            position.set(cursor)
+        }
```
richardhuaaa: Should we log something if the group is not found?

xmtp_mls/src/subscriptions/stream_messages.rs
```diff
@@ -452,43 +344,35 @@ where
...
-            .copied();
-        if let Some(m) = cursor {
-            // If we get a message, and its cursor is greater than what the stream started at
-            // but our last synced position is greater, try to get it from the database rather
-            // then starting process future
-            // otherwise, skip the message
-            if m.pos() > envelope.id && m.started() > envelope.id {
+        let cursor = self.groups.position(&next_msg.group_id);
+        if let Some(position) = cursor {
+            if position.last_streamed() > next_msg.id && position.started() > next_msg.id {
```
richardhuaaa, line 356: Not important for this PR, which doesn't/shouldn't have logical changes, but is `last_streamed` always bigger than `started`? I wonder if it would be easier to understand if this first conditional was simplified to `next_msg.id <= position.started()`, and the next one to `else if next_msg.id <= position.last_streamed()`. Also I'm wondering if these comparisons should be inclusive rather than exclusive

xmtp_mls/src/subscriptions/stream_messages.rs
```diff
@@ -452,43 +344,35 @@ where
...
-                    envelope.id,
+                    next_msg.id,
                 );
-                envelope = ready!(self.as_mut().skip(cx, envelope))?;
-            // we got a message with a sequence_iud greater than a message we already processed
+                next_msg = ready!(self.as_mut().skip(cx, next_msg))?;
+            // we got a message with a sequence_id greater than a message we already processed
             // so it must be present in the database
-            } else if m.pos() > envelope.id && m.started() < envelope.id {
+            } else if position.last_streamed() > next_msg.id && position.started() < next_msg.id {
```
richardhuaaa, line 369: Another comment unrelated to the refactor, but maybe we can go over the logic together? I'm possibly misunderstanding some details here. Without totally ordered streams, does it make sense to track `last_streamed` at all? It seems to me that regardless of if the incoming message is bigger or smaller than last_streamed, we want to process it if it's not in the DB. And that when the stream is initialized, `started` should be fetched from the DB rather than from the network.

xmtp_mls/src/subscriptions/stream_messages/types.rs
insipx: In the recent past, we didn't keep sequence ID on group messages, we only kept the last commit synced in the refresh state table. In order to get the latest message for a group we needed to query the network.

Recently we added sequenceIDs with the multi-cursor migration, but its still technically possible that some clients have a null sequence ID for their latest message. So I kept this in to be safe maybe until the next hammer

xmtp_mls/src/subscriptions/stream_messages/types.rs
insipx: I want to experiment with using small, focused traits private to a module in order to create more focused code and facilitate unit testing. Traits make it very easy to create quick mocks of data and test small pieces of logic. I'm not sure if this is how it will stay yet, but this module is small enough that its easy to revert.

This trait in particular removes a dependency on the monolithic local context from the module -- So, externally we don't need to go searching which method is depended on, because the trait lets us know exactly the functionality required

xmtp_mls/src/subscriptions/stream_messages.rs
insipx: yeah this logic needs to be re-scrutinized. I'm working on another pr to split the stream messages module further into digestible bits, moving unrelated things like next_message, resubscribing and processing futures into their own parts while encapsulating the state update logic. This should make the focus of stream messages on the critical bits in `on_waiting`, and allow scrutinizing the unrelated bits on their own (and hopefully make them easier to understand, too, would love your feedback on that). Maybe we can go over after that PR? Should be ready today

xmtp_mls/src/subscriptions/stream_messages/types.rs
insipx: We don't necessarily need the latest position, but before we had a sequence_id on group messages, there was no good way to know where a client left off with messages. During the initial design of this, my thinking went something like this,

If we wanted to avoid duplicates: 

- add the cursor id to the group messages (which is done now), & start from there
- query the latest from the network and start from there
- start the stream from the cursor in the refresh state table. 
    Starting from the cursor in the refresh state table increases complexity, since we would have to filter out duplicate messages by timestamp & group_id, which seems fragile & prone to mistake

Querying the latest seemed like a natural choice since I guess a stream implies new messages

Since we have sequence id now, I can make a followup to query for SequenceID, and only query the network if its None

xmtp_mls/src/subscriptions/stream_messages.rs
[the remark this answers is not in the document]
insipx: Its used once this future finishes, to correctly set the cursor on the group. Without passing the group ID here, we don't have a good way to know which group was added when. The logic for resubscribes is also being refactored to hopefully make more sense in a future PR

xmtp_mls/src/subscriptions/stream_messages.rs
insipx: last_streamed should never be less then started, but could be equal. Also, I totally agree, these conditionals are definitely confusing and need to be re-thought. I intentionally don't want to modify them in this PR, since it has the most likelihood of introducing bugs, so ideally I'd like to do it in a PR where its the focus & maybe includes some unit tests to verify the logic and validate it going forward

xmtp_mls/src/subscriptions/stream_messages.rs
richardhuaaa: Sounds great!

xmtp_mls/src/subscriptions/stream_messages.rs
insipx: TBH I don't remember exactly why. IIRC, setting a cursor of 1 the backend is _supposed_ to get all historical messages from that group so far. Setting the cursor to one here has been carried over from the very first impl of streams for v3 I think.

It's not really consistent with the stream creation, which gets the latest message from a stream. I think its assumed here that the add function is only used in stream_all, which would only add on a new welcome & getting messages from the first historical is desirable. Could probably change the naming around to make that clear

1,1 sets last streamed and started to 1

I think this is changing with reliable streams very soon

insipx:
### Merge activity

* **Jun 13, 4:56 PM UTC**: @insipx merged this pull request with [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2087).

PR #2088, opened 2025-06-10
Add a Cursor to Welcome

codabrink:
### Include an encrypted message cursor in welcome processing and add a `limit` parameter to `xmtp_mls::groups::Group::receive` and `xmtp_api::mls::ApiClientWrapper::query_group_messages`, clamped by `MAX_PAGE_SIZE` (100 prod, 20 test)
- Attach encrypted `WelcomeMetadata` carrying a `message_cursor` in `xmtp_mls::groups::Group::send_welcomes`, decrypt it in `xmtp_mls::groups::mls_ext::DecryptedWelcome::from_encrypted_bytes`, and persist the group cursor in `xmtp_mls::groups::welcomes::xmtp_welcome::process_with_tx` ([xmtp_mls/src/groups/mls_sync.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e), [xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-9825b13d685491fae6f92c0a816a683e6cdd5e91b0d39963afd2c1247b295cb2), [xmtp_mls/src/groups/welcomes/xmtp_welcome.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-a8d5a5fab05f9277d06cce92de1f0806f13f6d3298cf178dcd67f8599f94220e)).
- Add per-call paging `limit` to `xmtp_api::mls::ApiClientWrapper::query_group_messages` and propagate through `xmtp_mls::mls_store::MlsStore::query_group_messages` and `xmtp_mls::groups::Group::receive`, clamped to `xmtp_configuration::MAX_PAGE_SIZE`; replace local constants and adjust pagination checks ([xmtp_api/src/mls.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-8d4d4d46a198cd86023448bbca9e16b9f3362affb961b655082166f1a6095476), [xmtp_mls/src/mls_store.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-d203196a40247d60627de69f661c4f0cf106d56de1da51c33e203a286b6e220a), [xmtp_mls/src/groups/mls_sync.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e), [xmtp_configuration/src/prod/api.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-83ac7e897002dc0e43d73d6be82a9206fb6bf9044430a7d9c78cef4ecdb80e1c), [xmtp_configuration/src/test/api.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-53a66feee19a8c24142cfd2aa5330f53fdbb6e13099daf2ebce27cd51922be83)).
- Persist `sequence_id` when committing intents by adding a nullable `sequence_id` column and requiring `set_group_intent_committed(intent_id, sequence_id)`; include migrations ([xmtp_db/src/encrypted_store/group_intent.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-17a09cbc9886e242f0ff0189d4750dbfd8226be9344e7b41814967745cfe0a0c), [xmtp_db/migrations/2025-08-11-185550_add_sequence_id_to_intents/up.sql](https://github.com/xmtp/libxmtp/pull/2088/files#diff-27628d5b1a8b4995b3cbb0d042a96d2aa2f545c498c8886c87cc6c5685941d25), [xmtp_db/migrations/2025-08-11-185550_add_sequence_id_to_intents/down.sql](https://github.com/xmtp/libxmtp/pull/2088/files#diff-9f7549491fa3afc44c9c67646075026d0d539aa5560cac904698209f3bb8e267)).
- Change `RefreshState` updates to upsert on missing rows in `update_cursor` ([xmtp_db/src/encrypted_store/refresh_state.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-3eb320843309b9973b2ceaf8240120758bd1f2524b263d9a6b869b33266804ea)).
- Update tests, examples, and commit-log paging to the new signatures and constants ([xmtp_mls/src/groups/tests](https://github.com/xmtp/libxmtp/pull/2088/files#diff-0e04ea7656c794f4db32d120dba3e53abd1afacd29961f6ea09068baaae49ab6), [xmtp_mls/src/groups/commit_log.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-c25dc4d36a6249375b1001e9c4b66e50034909d2fc05025a45476e115429dc3a), [examples/cli/debug.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-aa20314d6a1788db316e20673489bf517e77c9bee991e34391c1bff1693448b7)).

#### 📍Where to Start
Start with `ApiClientWrapper::query_group_messages` in [xmtp_api/src/mls.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-8d4d4d46a198cd86023448bbca9e16b9f3362affb961b655082166f1a6095476) to see the new `limit` paging flow, then follow through `MlsStore::query_group_messages` in [xmtp_mls/src/mls_store.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-d203196a40247d60627de69f661c4f0cf106d56de1da51c33e203a286b6e220a) and `Group::receive` in [xmtp_mls/src/groups/mls_sync.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e); review welcome cursor handling in `xmtp_welcome::process_with_tx` and `DecryptedWelcome::from_encrypted_bytes` in [xmtp_mls/src/groups/welcomes/xmtp_welcome.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-a8d5a5fab05f9277d06cce92de1f0806f13f6d3298cf178dcd67f8599f94220e) and [xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs](https://github.com/xmtp/libxmtp/pull/2088/files#diff-9825b13d685491fae6f92c0a816a683e6cdd5e91b0d39963afd2c1247b295cb2).

#### Changes since #2088 opened

- Refactored database operation in QueryRefreshState.update_cursor method from two-step flow to single upsert [decd4d5]
- Updated major dependency versions in Cargo.lock [1044f5a]
- Consolidated regex-related dependency references [1044f5a]
- Migrated Windows-specific dependencies [1044f5a]
- Removed obsolete package dependency [1044f5a]
- Reworked chunk size calculation logic in `sync` method to prevent division-by-zero errors and ensure proper welcome message batching [ad937aa]
- Removed limit parameters from core query and receive methods [44892d5]
- Updated all method callers and tests to use new signatures [44892d5]
- Modified cursor validation logic in sync method to handle negative cursors [8f31ef9]
- Added InvalidNegativeCursor error variant and classification [8f31ef9]
----

_[Macroscope](https://app.macroscope.com) summarized 8f31ef9._


xmtp_db/src/encrypted_store/group_intent.rs
```diff
@@ -73,6 +73,7 @@ pub struct StoredGroupIntent {
     pub staged_commit: Option<Vec<u8>>,
     pub published_in_epoch: Option<i64>,
     pub should_push: bool,
+    pub sequence_id: Option<i64>,
```
insipx, line 76: will every intent type have a sequence_id? 

If so, where does the sequence id come from for each intent type

if not, which intents have a sequenceId and which don't

nplasterer:
I think there is a follow up here to syncAll so that you can specify how many messages you want to sync from all the groups you are syncing.

xmtp_db/src/encrypted_store/group_intent.rs
neekolas: Anything that gets published to the network will come back with a `sequence_id`. So, everything should at the time it reaches the state of `committed`

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
```diff
@@ -35,16 +37,31 @@ impl DecryptedWelcome {
...
+
+        let welcome_bytes = unwrap_welcome(
+            encrypted_welcome_bytes,
+            &private_key,
+            wrapper_ciphersuite.clone(),
+        )?;
         let welcome = deserialize_welcome(&welcome_bytes)?;
 
+        if !encrypted_welcome_metadata_bytes.is_empty() {
+            let metadata_bytes = unwrap_welcome(
```
neekolas: @richardhuaaa @mchenani curious what you think. Given that this metadata is strictly optional, should we fail open here rather than returning on error?

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -2130,14 +2131,28 @@ where
...
+    ) -> Result<(), GroupError> {
         let welcomes = action
             .installations
             .into_iter()
             .map(
                 |installation| -> Result<WelcomeMessageInput, WrapWelcomeError> {
                     let installation_key = installation.installation_key;
                     let algorithm = installation.welcome_wrapper_algorithm;
+
+                    let welcome_metadadta = WelcomeMetadata {
```
neekolas: Typo here 
```suggestion
                    let welcome_metadata = WelcomeMetadata {
```

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
```diff
@@ -35,16 +37,31 @@ impl DecryptedWelcome {
         provider: &XmtpOpenMlsProvider<C>,
         hpke_public_key: &[u8],
         encrypted_welcome_bytes: &[u8],
+        encrypted_welcome_metadata_bytes: &[u8],
         wrapper_ciphersuite: WrapperAlgorithm,
-    ) -> Result<DecryptedWelcome, GroupError> {
+    ) -> Result<(DecryptedWelcome, Option<WelcomeMetadata>), GroupError> {
```
neekolas: Why wouldn't we put the `Option<WelcomeMetadata>` into the `DecryptedWelcome` struct? We control the struct and can put whatever we want in there

xmtp_mls/src/groups/mod.rs
```diff
@@ -805,6 +810,10 @@ where
                 group.quietly_update_consent_state(ConsentState::Allowed)?;
             }
 
+            // Set the message cursor
+            let cursor = welcome_metadata.map(|m| m.message_cursor as i64).unwrap_or_default();
+            provider.db().update_cursor(&group.group_id, EntityKind::Group, cursor)?;
```
neekolas: This is the scary part. We've had multiple forking incidents around how Welcomes and cursors interact. Let's really think through all the cases.

1. You are being added to a group for the first time
2. You have been removed from a group, and are being added back
3. You have received multiple concurrent adds to the same group
4. Are there more?

`1` feels safe to write here.

`2` is a little dicey. What if you haven't synced the fact that you were removed yet? Now you might fast forward past your own removal. The code above should protect against that by throwing an already processed error, right?

`3` is questionable. I _think_ it's safe because the cursor will match the Welcome that is actually being applied. But are we 100% sure?

The more conservative approach would be to only update the cursor if there was no cursor already, and never overwrite. 

/cc @mchenani @richardhuaaa

xmtp_api/src/mls.rs
```diff
@@ -113,6 +114,12 @@ where
                 break;
             }
 
+            if let Some(limit) = limit {
+                if out.len() >= limit as usize {
+                    break;
+                }
+            }
```
richardhuaaa: I think this logic breaks if the limit is greater than the page size. For example with a max_page_size of 5, a limit of 6, and 20 entries on the server, I believe this method would return 10 entries instead of 6. It might make sense to add unit tests for this change, if it is actually needed.

bindings_ffi/src/mls.rs
```diff
@@ -2011,6 +2011,11 @@ impl FfiConversation {
         Ok(())
     }
 
+    pub async fn receive(&self, limit: Option<u32>) -> Result<(), GenericError> {
+        self.inner.receive(limit).await?;
+        Ok(())
+    }
```
richardhuaaa: Is this change needed, and what is the context behind it? It seems unrelated to adding a welcome cursor

xmtp_db/src/encrypted_store/refresh_state.rs
```diff
@@ -109,17 +105,41 @@ impl<C: ConnectionExt> DbConnection<C> {
...
+            }
+
+            // If no update, try to insert
+            match diesel::insert_into(dsl::refresh_state)
+                .values((
+                    dsl::entity_id.eq(&entity_id_bytes),
+                    dsl::entity_kind.eq(entity_kind),
+                    dsl::cursor.eq(cursor),
+                ))
                 .execute(conn)
```
richardhuaaa: I'm unsure if there is a race condition here, depending on if we're running this in a transaction, and the isolation level of the transaction. Let's say thread 1 is setting the cursor to 5, and thread 2 is setting it to 6. Let's say the row does not exist and both thread 1 and thread 2 reach L125, upon which thread 1 inserts the record first. Then thread 2 would not set the cursor, even though it's greater, and that would not be reported as an error.

It seems like it would be safer to reverse these two operations - try to insert it, and if it fails, then try to update it. Also, it would be nice to understand the context behind this change - if there's a bug we've encountered, or this is simply a nice-to-have.

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
richardhuaaa: I guess the question is whether or not we should throw away the whole welcome in this situation. It seems like it would be fine to log an error message, but process the rest of the welcome without aborting with an error. Probably either behavior is fine, because the welcome is technically malformed

xmtp_mls/src/groups/mod.rs
richardhuaaa: So I would argue these are questions of whether the encryption state from the Welcome should be applied at all. If the encryption state from the Welcome is applied at all, then probably the cursor corresponding to that encryption state should also be applied (whether that fast-forwards you, rewinds you, etc).

Which brings us to the `AlreadyProcessed` detection on L601. It protects you from (3), but it does not protect you from being fast-forwarded past your own removal in (2). It seems like there's two ways to address this issue - either calling `sync()` as part of Welcome processing, or storing the Welcome in a holding area in the DB and applying it after you've been removed. I feel like we might be able to split that into a separate PR.

WDYT?

bindings_ffi/src/mls.rs
codabrink: Yes, adding a receive function to query the first n-messages was what prompted adding a cursor to welcomes.

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
mchenani: I decided to just log this and continue processing the welcome here, since it can be sent from an older version.

xmtp_mls/src/groups/mod.rs
mchenani: created this issue for later: https://github.com/xmtp/libxmtp/issues/2209

bindings_ffi/src/mls.rs
nplasterer: I think we can push this into a follow up PR. I also have an issue for adding ability to syncAll N messages because I think that most people will be calling syncAll for unknown consent states and if the group is unknown then they don't want all the messages. But that can be decoupled from this work and probably should be for clarity.

xmtp_db/src/encrypted_store/refresh_state.rs
mchenani: After looking at this, the reason seems to be: we don’t have a cursor row until we fetch a group/conversation’s messages for the first time. On that first pass, the update is a no-op and we insert the cursor; once the group exists, updates happen normally.

On the race condition part: I think that could happen if we fetch messages for that and we already removed, and we receive the welcome to be added again at the same time.
wdyt?

deny.toml
```diff
@@ -2,7 +2,6 @@
 ignore = [
   { id = "RUSTSEC-2024-0384", reason = "need to upgrade openmls" },
   { id = "RUSTSEC-2024-0436", reason = "required by uniffi" },
-  { id = "RUSTSEC-2024-0370", reason = "proc-macro-error is a dependency of libcrux" },
```
mchenani, line 5: fyi: cargo deny says this is not needed and not related to any dependencies.

xmtp_mls/src/groups/welcome_sync.rs
```diff
@@ -275,36 +275,54 @@ mod tests {
...
+        message_cursor: Option<u64>,
     ) -> welcome_message::V1 {
         let w = wrap_welcome(
             &welcome.tls_serialize_detached().unwrap(),
             &public_key,
             &WrapperAlgorithm::Curve25519,
         )
         .unwrap();
+
+        let wrapped_welcome_metadata: Vec<u8> = if let Some(cursor) = message_cursor {
```
mchenani, line 303: fyi: intentionally did this to test welcomes with empty welcome_metadata

xmtp_db/src/encrypted_store/refresh_state.rs
richardhuaaa: Makes sense. So could we reverse the order (try to insert first), or even better, do a sqlite [upsert](https://sqlite.org/lang_upsert.html), which is an `insert... on conflict...`? Just want to be careful here, because this affects anything that touches a cursor, which is pretty much every feature

cameronvoell:
Validated the first issue flagged by claude with this new test [here](https://github.com/xmtp/libxmtp/pull/2088/commits/504506d47c33fcb34a959bb10f8f914e6aa14f28).

Not sure if there are ramifications across the rest of the PR. mind taking a look @mchenani 🙏

xmtp_db/src/encrypted_store/refresh_state.rs
insipx: actually ran into this same issue during d14n migration. since different originator ids correspond to different rows, its possible that updating a cursor will try to get a non-existent row

this is what update cursor method looks like with the upsert i've added. it was a bit annoying to figure out b/c diesel docs & errors with conflicting traits.

```rust
fn update_cursor<Id: AsRef<[u8]>>(
        &self,
        entity_id: Id,
        entity_kind: EntityKind,
        cursor: Cursor,
    ) -> Result<bool, StorageError> {
        use super::schema::refresh_state::dsl;
        let state: RefreshState = self
            .get_refresh_state(&entity_id, entity_kind, Some(cursor.originator_id))?
            .unwrap_or(RefreshState {
                entity_id: entity_id.as_ref().to_vec(),
                entity_kind,
                cursor: cursor.sequence_id as i64,
                originator_id: cursor.originator_id as i32,
            });
        let num_updated = self.raw_query_write(|conn| {
            let incomplete_query = diesel::insert_into(dsl::refresh_state)
                .values(&state)
                .on_conflict((dsl::entity_id, dsl::entity_kind, dsl::originator_id))
                .do_update()
                .set(dsl::cursor.eq(excluded(dsl::cursor)));
            query_dsl::methods::FilterDsl::filter(
                incomplete_query,
                dsl::cursor.lt(excluded(dsl::cursor)),
            )
            .execute(conn)
        })?;
        Ok(num_updated >= 1)
    }

```


for some reason i had to specify FilterDsl, since the filter method conflicted with another specifically for upsert.

richardhuaaa: That looks great! I took a stab at simplifying it a tiny bit more, I happen to also be working on a different upsert statement at the moment:

```
fn update_cursor<Id: AsRef<[u8]>>(
        &self,
        entity_id: Id,
        entity_kind: EntityKind,
        cursor: Cursor,
    ) -> Result<bool, StorageError> {
        use super::schema::refresh_state::dsl;
        use diesel::query_dsl::methods::FilterDsl;

        let new_cursor = cursor.sequence_id as i64;
        let num_updated = self.raw_query_write(|conn| {
            diesel::insert_into(dsl::refresh_state)
                .values(&RefreshState {
                    entity_id: entity_id.as_ref().to_vec(),
                    entity_kind,
                    originator_id: cursor.originator_id as i32,
                    cursor: new_cursor,
                })
                .on_conflict((dsl::entity_id, dsl::entity_kind, dsl::originator_id))
                .do_update()
                .filter(dsl::cursor.lt(new_cursor)
                .set(dsl::cursor.eq(new_cursor));
                .execute(conn)
        })?;
        Ok(num_updated >= 1)
    }
```

EDIT: added FilterDsl import

cameronvoell (approved):
thanks for pushing this through @mchenani !

PR #2094, opened 2025-06-10
Add WalletSendCalls content type

rygine:
# Summary

- Generated protos
- Added WalletSendCalls content type


nplasterer:
I know we are making some changes to the wallet send call so just want to make sure we address those before merging this. Once we land this in libxmtp it will be harder to change I think.

rygine:
added by #2352

PR #2100, opened 2025-06-11
Fix: workers are starting before identity is ready

codabrink:


codabrink:
Need to move the wait inside of the spawn when I get a bit of time.

PR #2106, opened 2025-06-12
Store the local commit log

codabrink:


xmtp_db/migrations/2025-06-09-225910_create_local_commit_log/up.sql
```diff
@@ -13,13 +13,14 @@ CREATE TABLE remote_commit_log (
 );
 
 CREATE TABLE local_commit_log (
-    "sequence_id" BIGINT NOT NULL PRIMARY KEY,
+    "id" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    "sequence_id" BIGINT,
```
richardhuaaa: This one should be NOT NULL

xmtp_mls/src/groups/mls_ext/mls_ext_merge_staged_commit.rs
```diff
@@ -0,0 +1,48 @@
...
+impl MergeStagedCommitAndLog for MlsGroup {
+    fn merge_staged_commit_and_log<Db: ConnectionExt>(
+        &mut self,
+        provider: &XmtpOpenMlsProvider<Db>,
+        staged_commit: StagedCommit,
+        validated_commit: &ValidatedCommit,
+        sequence_id: Option<i64>,
+    ) -> Result<(), GroupMessageProcessingError> {
+        let mut log = NewLocalCommitLog {
+            epoch_authenticator: Some(self.epoch_authenticator().as_slice().to_vec()),
```
richardhuaaa: We might actually need both the epoch authenticator before the commit was merged (as `last_epoch_authenticator`) and the epoch authenticator after the commit was merged (as `epoch_authenticator`)

xmtp_mls/src/groups/mls_ext/mls_ext_merge_staged_commit.rs
```diff
@@ -0,0 +1,48 @@
...
+            sender_inbox_id: Some(validated_commit.actor_inbox_id()),
+            sender_installation_id: Some(validated_commit.actor_installation_id()),
+        };
+
+        if let Err(err) = self.merge_staged_commit(&provider, staged_commit) {
+            tracing::error!("Error merging commit: {err}");
+            log.result = CommitResult::Invalid;
+            log.store(provider.db())?;
+            return Err(err)?;
+        }
```
richardhuaaa: When there is an error in merging a commit, I think there is retry logic in the outer functions (for example, on retriable errors like a failure to get a DB connection). In the event that there are multiple retries, what we want is the final result, rather than each attempt. Have we accounted for that?

xmtp_mls/src/groups/mls_ext/mls_ext_merge_staged_commit.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        log.store(provider.db())?;
```
richardhuaaa: Double-checking, is all of this happening in a transaction in the outer scope? If so then all good. We don't want one of the writes in this function to succeed without the others also succeeding.

richardhuaaa: Confirmed that both call-trees happen within a transaction - one initiated at `validate_and_process_external_message()`, another initiated at `process_message()`

xmtp_mls/src/groups/mls_ext/mls_ext_merge_staged_commit.rs
richardhuaaa: So good news is I was able to move this error logging to an outer function, so that all non-retriable error types can be caught. And we can simply not log retriable errors, which is half the battle.

Bad news is even non-retriable errors could be retried. Unfortunately in today's codebase we don't move the cursor forward for non-retriable errors - we rely on processing a subsequent payload to update the cursor. So in theory, two different calls to `sync()` can throw the same non-retriable error twice. I figured out a way to handle it, though - only insert an error entry if the last row on the group isn't identical to the entry we want to insert.

richardhuaaa: Either way, we should consider having the fork detection logic only take into account successful commits, and ignore errors.

There is also an extremely rare scenario where a commit's error log could be dropped - if there is a really well-timed storage error.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1369,8 +1383,23 @@ where
...
+                {
+                    tracing::error!(
+                        "Error inserting commit entry for failed commit: {}",
+                        accounting_error
+                    );
+                }
+                if let Err(accounting_error) = self
+                    .process_group_message_error_for_fork_detection(msgv1, &e)
+                    .await
+                {
```
richardhuaaa, line 1445: Before this change, this would have returned the accounting error instead of the message processing error. I'm not sure if the older code would have messed up the logs, and there's also a risk of returning a retriable error instead of a non-retriable one, or vice versa (causing forks?)

richardhuaaa:
Still to do:
- Populate newly added 'error_message' field on commit log - just serialize the error
- Unit tests

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1515,6 +1544,64 @@ where
...
+                    CommitResult::Undecryptable
+                }
+                _ => CommitResult::Unknown,
+            };
+            NewLocalCommitLog {
+                group_id: self.group_id.to_vec(),
+                commit_sequence_id: message_cursor as i64,
+                last_epoch_authenticator,
+                commit_result,
+                applied_epoch_number: Some(message.epoch().as_u64() as i64), // For debugging purposes
```
richardhuaaa, line 1644: I'm not sure if this is the epoch the commit is being applied *on*, or the epoch the commit transitions the group *to*. Worth double checking

richardhuaaa:
Still to do in follow-up:
- Log group creation
- Log welcomes
- Populate epoch_authenticator and epoch number more accurately on errors, and add unit test for errors
- Fix intent publishing retries in poor network conditions (independent to this feature)

xmtp_mls/src/groups/mls_sync.rs
richardhuaaa: I will fix the epoch authenticator and epoch number as part of the next PR

richardhuaaa (approved):

richardhuaaa:
The web assembly tests have been failing on main for a while now, this can be fixed separately

PR #2107, opened 2025-06-14
Fix diesel migration code generation

richardhuaaa:
Relying on a `schema.patch` to update the code-genned `schema_gen.rs` file is too brittle here - the line numbers change anytime anything new is added to the schema. Currently in main, anytime the migrations are code-genned via `cargo run --bin update-schema --features update-schema`, it will fail due to the patch being unable to apply.

I moved the change we need from the `schema.patch` to the `schema.rs` file, which is not codegenned, instead.


codabrink (approved):
Thanks for fixing that!

PR #2114, opened 2025-06-16
forked groups investigations

mchenani:


xmtp_mls/src/groups/mls_ext/reload.rs
```diff
@@ -0,0 +1,22 @@
...
+use crate::groups::mls_sync::GroupMessageProcessingError;
+
+pub trait MlsGroupReload {
+    fn reload<C: ConnectionExt>(
+        &mut self,
+        provider: &XmtpOpenMlsProvider<C>,
+    ) -> Result<(), GroupMessageProcessingError>;
+}
+
+impl MlsGroupReload for OpenMlsGroup {
```
mchenani, line 13: Nice!

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -712,9 +717,28 @@ where
 
         let validated_commit = match &processed_message.content() {
             ProcessedMessageContent::StagedCommitMessage(staged_commit) => {
-                let validated_commit =
+                let result =
                     ValidatedCommit::from_staged_commit(&self.context, staged_commit, mls_group)
-                        .await?;
+                        .await;
+
+                let validated_commit = match result {
```
mchenani, line 724: @richardhuaaa could you please write you comment here?

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1099,6 +1149,9 @@ where
             Some(intent) => {
                 let mut identifier = MessageIdentifierBuilder::from(envelope);
                 identifier.intent_kind(intent.kind);
+                if intent.state == IntentState::Error {
```
mchenani:
```suggestion
               if intent.state != IntentState::Published{
```

mchenani: cc: @richardhuaaa

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1084,6 +1121,19 @@ where
...
+
+        let group_cursor = provider
+            .db()
+            .get_last_cursor_for_id(&self.group_id, EntityKind::Group)?;
+        if group_cursor >= cursor as i64 {
+            // early return if the message is already procesed
+            // _NOTE_: Not early returning and re-processing a message that
+            // has already been processed, has the potential to result in forks.
+            return MessageIdentifierBuilder::from(envelope)
+                .previously_processed(true)
```
insipx, line 1133: is this field new/effect anything? if so, we should set it in the other places where we flag messages as previously processed

mchenani: It's not new and we already doing it.

xmtp_mls/src/groups/mls_sync.rs
richardhuaaa: I'm not sure why we increment the cursor specifically for `ProtocolVersionTooLow` here and not other non-retryable errors. If the goal is to reduce the amount of retries, it could make sense to do this for all non-retriable errors, or we could leave it for later as a bigger refactor, as there are also other non-retriable errors we wouldn't be able to catch here. TL;DR: I think it's okay to leave this change out for now

xmtp_mls/src/groups/mls_sync.rs
richardhuaaa: I feel like this change could cause a stuck cursor. It doesn't look like we've incremented the cursor here, so it seems like this function could just get retried over and over.

Note that when we do the same thing on L1133, it's because the cursor has already advanced past the current payload.

@mchenani was this an important change to get in? Any idea in what case this could happen? Note that with the change on L1091 (locking the group) and L1133 (checking the cursor), we already have some degree of concurrency protection.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1889,6 +1970,8 @@ where
         update_interval_ns: Option<i64>,
     ) -> Result<(), GroupError> {
         let provider = self.mls_provider();
+
+        //todo: check for the current inbox installations
```
richardhuaaa: Could we clarify what this means, or remove it?

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1926,6 +2009,7 @@ where
      * and the group has not been updated to include it.
      */
     pub(super) async fn add_missing_installations(&self) -> Result<(), GroupError> {
+        //todo-> make sure we only take care of our own installations not everyone
```
richardhuaaa: Same here

richardhuaaa:
It looks good other than the comments I left! I think it's worth considering whether some of the fixes are safer to leave out, as we have coverage with the other fixes

xmtp_mls/src/groups/mls_sync.rs
mchenani: @richardhuaaa i agree, with L1133, seems that we don't hit the issue here.

xmtp_mls/src/groups/mls_sync.rs
mchenani: that's unnecessary

xmtp_mls/src/groups/mls_sync.rs
mchenani: here too, unnecessary.

xmtp_mls/src/groups/mls_sync.rs
codabrink: Looks like you misread @richardhuaaa 
This increments the cursor for all non-retryable errors except for ProtocolVersionTooLow

PR #2117, opened 2025-06-17
Limit Installations to 5

nplasterer:
WIP


xmtp_mls/src/configuration.rs
```diff
@@ -29,7 +29,9 @@ const SYNC_UPDATE_INSTALLATIONS_INTERVAL_NS: i64 = NS_IN_HOUR / 2; // 30 min
 
 pub const SEND_MESSAGE_UPDATE_INSTALLATIONS_INTERVAL_NS: i64 = 5 * NS_IN_SEC;
 
-pub const MAX_GROUP_SIZE: usize = 220;
+pub const MAX_GROUP_SIZE: usize = 250;
```
nplasterer, line 32: Bump this up to 250 so it's a more square number and matches other protocols limits.

xmtp_mls/src/identity_updates.rs
```diff
@@ -119,44 +164,17 @@ where
     /// If no `to_sequence_id` is provided, use the latest value in the database
     pub async fn get_association_state(
         &self,
-        conn: &DbConnection<<Db as XmtpDb>::Connection>,
+        _conn: &DbConnection<<Db as XmtpDb>::Connection>,
```
nplasterer: I could remove this from all the places it's used instead of marking it as _. Just want to confirm first that this is the route we want to go.

xmtp_mls/src/identity_updates.rs
```diff
@@ -77,6 +77,51 @@ impl<ApiClient, Db> IdentityUpdates<ApiClient, Db> {
     }
 }
 
+/// Get the association state for a given inbox_id up to the (and inclusive of) the `to_sequence_id`
+/// If no `to_sequence_id` is provided, use the latest value in the database
+pub async fn get_association_state_with_verifier(
+    provider: impl MlsProviderExt,
```
insipx: since this provider is immediately turned into a DbCoonnection, could accept the DbConnection instead, and then that db connection does not have to be `_conn`. The provider then doesn't need to be passed, at least I don't think its being used for anything here

xmtp_mls/src/identity_updates.rs
insipx: could just replace the call to mls_provider with dbconnection, then this would be used (see other comment as well)

xmtp_mls/src/configuration.rs
```diff
@@ -29,7 +29,9 @@ const SYNC_UPDATE_INSTALLATIONS_INTERVAL_NS: i64 = NS_IN_HOUR / 2; // 30 min
...
+
+pub const MAX_INSTALLATIONS_PER_INBOX: usize = 5;
```
nplasterer, line 35: cc @jhaaaa here are two changes that the docs should update

PR #2123, opened 2025-06-20
Static Revocation

nplasterer:
Adds the ability to revoke an installation statically without having access to the installation.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -52,6 +52,8 @@ use xmtp_mls::groups::device_sync::archive::BackupMetadata;
 use xmtp_mls::groups::device_sync::DeviceSyncError;
 use xmtp_mls::groups::device_sync_legacy::ENC_KEY_SIZE;
 use xmtp_mls::groups::ConversationDebugInfo;
+use xmtp_mls::identity_updates::apply_signature_request_with_verifier;
+use xmtp_mls::identity_updates::revoke_installations_with_verifier;
 use xmtp_mls::utils::events::upload_debug_archive;
 use xmtp_mls::verified_key_package_v2::{VerifiedKeyPackageV2, VerifiedLifetime};
 use xmtp_mls::{
@@ -145,7 +147,8 @@ pub async fn static_revoke_installations(
     let scw_verifier = api.api_client.context.scw_verifier.clone();
 
     let signature_request =
-        revoke_installations_with_verifier(store, inbox_id, installation_ids, scw_verifier).await?;
+        revoke_installations_with_verifier(&store.db(), inbox_id, installation_ids, scw_verifier)
+            .await?;
 
     Ok(Arc::new(FfiSignatureRequest {
         inner: Arc::new(tokio::sync::Mutex::new(signature_request)),
@@ -185,8 +188,13 @@ pub async fn static_apply_signature_request(
     let signature_request = signature_request.inner.lock().await;
     let scw_verifier = api.api_client.context.scw_verifier.clone();
 
-    apply_signature_request_with_verifier(api, store, signature_request.clone(), scw_verifier)
-        .await?;
+    apply_signature_request_with_verifier(
+        &api,
+        &store.db(),
+        signature_request.clone(),
+        scw_verifier,
+    )
+    .await?;
 
     Ok(())
 }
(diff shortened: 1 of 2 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -112,6 +112,22 @@ pub async fn connect_to_backend(
...
 
+/**
+ * Static revoke a list of installations
+ */
+#[uniffi::export(async_runtime = "tokio")]
+pub async fn static_revoke_installations(
+    inbox_id: &InboxId,
+    installation_ids: Vec<Vec<u8>>,
+) -> Result<Arc<FfiSignatureRequest>, GenericError> {
+    let signature_request = revoke_installations_with_verifier(???DB???, inbox_id, installation_ids, ???SCWVerifier???).await?;
```
nplasterer: How do I get the DB and SCW verifier without the client? 🤔

bindings_ffi/src/mls.rs
```diff
@@ -112,6 +115,96 @@ pub async fn connect_to_backend(
...
+                .try_into()
+                .map_err(|_| "Malformed 32 byte encryption key".to_string())?;
+            let db = NativeDb::new(&storage_option, key)?;
+            EncryptedMessageStore::new(db)?
+        }
+        None => {
+            let db = NativeDb::new_unencrypted(&storage_option)?;
+            EncryptedMessageStore::new(db)?
+        }
+    };
```
nplasterer: If someone blows up there local database they likely wont still have access to their encryption key. Can we still do this revoke on an ephemeral database? I tried that but it errored.

bindings_ffi/src/mls.rs
```diff
@@ -112,6 +115,51 @@ pub async fn connect_to_backend(
     Ok(Arc::new(XmtpApiClient(api_client)))
 }
 
+/**
+ * Static revoke a list of installations
+ */
+#[uniffi::export(async_runtime = "tokio")]
+pub async fn static_revoke_installations(
```
richardhuaaa: Seeing as this is part of the public API, could we just call these `revoke_installations` and `apply_signature_request` respectively?

I could also imagine these and other methods being grouped under an `IdentityManager` that only needs to be initialized with the `api` once, but no need to get too fancy right now

bindings_ffi/src/mls.rs
```diff
@@ -112,6 +115,51 @@ pub async fn connect_to_backend(
...
+    api: Arc<XmtpApiClient>,
+    recovery_identifier: FfiIdentifier,
+    inbox_id: &InboxId,
+    installation_ids: Vec<Vec<u8>>,
+) -> Result<Arc<FfiSignatureRequest>, GenericError> {
+    let api: ApiClientWrapper<Arc<TonicApiClient>> =
+        ApiClientWrapper::new(Arc::new(api.0.clone()), strategies::exponential_cooldown());
+    let scw_verifier = Arc::new(
+        Box::new(RemoteSignatureVerifier::new(api)) as Box<dyn SmartContractSignatureVerifier>
+    );
```
richardhuaaa, line 132: Seeing as we do this in multiple places, these could potentially be placed into helper methods in an intuitive location, but up to you

xmtp_mls/src/identity_updates.rs
```diff
@@ -121,6 +121,48 @@ pub async fn get_association_state_with_verifier<C: ConnectionExt>(
...
+pub async fn revoke_installations_with_verifier(
+    identifier: &Identifier,
+    inbox_id: &str,
+    installation_ids: Vec<Vec<u8>>,
+) -> Result<SignatureRequest, ClientError> {
+    let mut builder = SignatureRequestBuilder::new(inbox_id);
+
+    for installation_id in installation_ids {
+        builder = builder.revoke_association(
+            identifier.clone().into(),
```
richardhuaaa, line 134: Ah, I just realized that the signature request *does* specify the [address it wants a signature from](https://github.com/xmtp/libxmtp/blob/92185c33d9a90252f0273060d1e0042498eb73a6/xmtp_id/src/associations/builder.rs#L107-L108). So ideally we would fetch the recovery address from the network and populate it here, so that there's no confusion about what to sign with, but having the integrator specify it for now is fine

nplasterer: They have to pass a signer to the function in the sdk so I will have the identity anyways.

PR #2130, opened 2025-06-22
Static Revocation Node & Browser

nplasterer:
Adds the ability to revoke an installation statically without having access to the installation.
This is more complicated in the node and browser sdk since they don't use a sign request object.

### DIFF
--- a/bindings_node/src/signatures.rs
+++ b/bindings_node/src/signatures.rs
@@ -4,11 +4,21 @@ use crate::ErrorWrapper;
 use napi::bindgen_prelude::{BigInt, Error, Result, Uint8Array};
 use napi_derive::napi;
 use std::ops::Deref;
+use std::sync::Arc;
+use xmtp_api::{strategies, ApiClientWrapper};
+use xmtp_api_grpc::grpc_api_helper::Client as TonicApiClient;
 use xmtp_id::associations::Identifier as XmtpIdentifier;
+use xmtp_id::associations::MemberIdentifier;
 use xmtp_id::associations::{
   unverified::{NewUnverifiedSmartContractWalletSignature, UnverifiedSignature},
   verify_signed_with_public_context, AccountId,
 };
+use xmtp_id::scw_verifier::RemoteSignatureVerifier;
+use xmtp_id::scw_verifier::SmartContractSignatureVerifier;
+use xmtp_mls::identity_updates::apply_signature_request_with_verifier;
+
+static SIGNATURE_REQUESTS: Lazy<Mutex<HashMap<SignatureRequestType, SignatureRequest>>> =
+  Lazy::new(|| Mutex::new(HashMap::new()));
 
 #[napi]
 pub fn verify_signed_with_public_key(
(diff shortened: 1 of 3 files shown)


bindings_node/Cargo.toml
```diff
@@ -36,6 +36,7 @@ xmtp_db.workspace = true
 xmtp_id.workspace = true
 xmtp_mls.workspace = true
 xmtp_proto = { workspace = true, features = ["proto_full"] }
+once_cell = "1.18"
```
codabrink: Once cell is standardized now, no need to add a new package.
`use std::cell::OnceCell`

bindings_node/src/signatures.rs
```diff
@@ -3,12 +3,26 @@ use crate::identity::{Identifier, IdentifierKind};
 use crate::ErrorWrapper;
 use napi::bindgen_prelude::{BigInt, Error, Result, Uint8Array};
 use napi_derive::napi;
+use once_cell::sync::Lazy;
```
codabrink: Use a lazy lock instead.
`use std::sync::LazyLock`

bindings_node/src/signatures.rs
```diff
@@ -3,12 +3,26 @@ use crate::identity::{Identifier, IdentifierKind};
...
 use xmtp_id::associations::{
   unverified::{NewUnverifiedSmartContractWalletSignature, UnverifiedSignature},
   verify_signed_with_public_context, AccountId,
 };
+use xmtp_id::scw_verifier::RemoteSignatureVerifier;
+use xmtp_id::scw_verifier::SmartContractSignatureVerifier;
+use xmtp_mls::identity_updates::apply_signature_request_with_verifier;
+use xmtp_mls::identity_updates::revoke_installations_with_verifier;
+
+static SIGNATURE_REQUESTS: Lazy<Mutex<HashMap<SignatureRequestType, SignatureRequest>>> =
```
codabrink: This is going to cause tests to break, since all tests will be contesting for the same lock.

bindings_node/src/signatures.rs
```diff
@@ -146,143 +280,42 @@ impl Client {
...
-    Ok(())
+    Ok(SignatureRequestHandle {
+      inner: Arc::new(tokio::sync::Mutex::new(signature_request)),
+      scw_verifier: self.inner_client().scw_verifier().clone(),
+    })
   }
 
   #[napi]
-  pub async fn add_passkey_signature(
+  pub async fn apply_signature_requests(
```
rygine: `apply_signature_requests` => `apply_signature_request`

since it's only applying 1 signature

bindings_node/src/signatures.rs
```diff
@@ -33,13 +46,132 @@ pub fn verify_signed_with_public_key(
...
+    .map_err(ErrorWrapper::from)?;
+
+  Ok(SignatureRequestHandle {
+    inner: Arc::new(tokio::sync::Mutex::new(signature_request)),
+    scw_verifier: scw_verifier.clone(),
+  })
+}
+
+#[napi]
+pub async fn apply_signature_request(
```
rygine, line 80: i don't understand this fn. does it mutate the passed `signature_request`?

nplasterer: It verifies that the signature is a valid signature from the correct recovery address before applying the identity change.
So you initiate a revoke for installations
Then you sign that revoke
Then you have to make sure that the signature was the signature you were expecting. And this will fail if the signature is from a different address or whatever. 
The identity update is marked as done so you can't do it again with the same request.

rygine: got it! i'm just confused on the use of `inner.clone()` when calling `apply_signature_request_with_verifier`. thought that fn would take ownership of the passed `signature_request` instead of a copy of it.

codabrink (approved):

rygine (approved):

Issue #1752, opened 2025-03-14
Feature request: Need a way to preserve order or distinguish between non recovery address identifiers in generated bindings


cameronvoell:
### Is your feature request related to a problem?

Currently inbox `AssociateState` stores Associated Address is an un-ordered HashMap of <MemberIdentifier, Member>. In rust we can still derive order because [Member](https://github.com/xmtp/libxmtp/blob/954f68c7ee6e3a9c093a38ec2a322e642ec88f7d/xmtp_id/src/associations/member.rs#L335-L342) struct has a `client_timestamp_ns` field.

In wasm/node/ffi bindings we translate that map to an array, but we do not utilize the timestamp to establish any order for the associations. We also do not pass on the `client_timestamp_ns` field. 

### Describe the solution to the problem

Ideally when SDKs read the AssociationState of an inboxId they have some way of distinguishing priority, or ordering of associated identifiers. This will help set norms for which address should be used for UI display purposes in cases where multiple identifiers are associated.

PR #2135, opened 2025-06-24
Order AssociationState by client timestamp

nplasterer:
Fixes https://github.com/xmtp/libxmtp/issues/1752

Ideally when SDKs read the AssociationState of an inboxId they have some way of distinguishing priority, or ordering of associated identifiers. This will help set norms for which address should be used for UI display purposes in cases where multiple identifiers are associated.

So "address to display" resolution for now should be
1. If there is only one address returned in the [inboxState functions](https://github.com/xmtp/xmtp-js/blob/d70e7535ec0635662bda0a6da8c36cd607936736/sdks/node-sdk/src/Client.ts#L516-L522) identifiers Array, use that as the display address. (we expect there always to be at least one, which is the recovery address)
2. If there are two addresses use the one that does not match the recoveryIdentifier
3. If there is more than 2 - we can't distinguish them yet - but we should be able to get in a fix for that soon for you. [Issue created here](https://github.com/xmtp/libxmtp/issues/1752).

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -986,24 +986,34 @@ impl From<VerifiedKeyPackageV2> for FfiKeyPackageStatus {
 
 impl From<AssociationState> for FfiInboxState {
     fn from(state: AssociationState) -> Self {
+        let mut sorted_members: Vec<_> = state.members().values().cloned().collect();
+        sorted_members.sort_by_key(|m| m.client_timestamp_ns.unwrap_or(u64::MAX));
+
         Self {
             inbox_id: state.inbox_id().to_string(),
             recovery_identity: state.recovery_identifier().clone().into(),
-            installations: state
-                .members()
-                .into_iter()
+            installations: sorted_members
+                .iter()
                 .filter_map(|m| match m.identifier {
-                    MemberIdentifier::Ethereum(_) => None,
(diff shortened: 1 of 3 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -986,24 +986,34 @@ impl From<VerifiedKeyPackageV2> for FfiKeyPackageStatus {
 
 impl From<AssociationState> for FfiInboxState {
     fn from(state: AssociationState) -> Self {
+        let mut sorted_members: Vec<_> = state.members().values().cloned().collect();
```
neekolas: Why not sort this at xmtp_id/xmtp_mls level so we don’t have to maintain the same sorting code across all our bindings?

xmtp_id/src/associations/state.rs
```diff
@@ -179,9 +181,22 @@ impl AssociationState {
     }
 
     pub fn identifiers(&self) -> Vec<Identifier> {
-        self.members
+        let mut address_members: Vec<_> = self
+            .members
             .values()
+            .filter(|member| {
+                matches!(
```
codabrink: Can we invert this to `!matches!(member.identifier, MemberIdentifier::Installation(_))`?

codabrink (approved):
Looks good! Just a nit

PR #2142, opened 2025-06-25
Release 1.2.5

nplasterer:
Included in this release are fixes for bugs that could cause forking

- [x] https://github.com/xmtp/libxmtp/pull/2143


xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
```diff
@@ -0,0 +1,130 @@
+use openmls::{
```
richardhuaaa: I don't think this file should be here

xmtp_mls/src/worker.rs
```diff
@@ -0,0 +1,187 @@
+use crate::{
```
richardhuaaa: This one shouldn't be here either, I think it's another refactor that happened in main

richardhuaaa (approved):
It looks like the only errors left are lint errors, although I'm not sure what '[Check iOS & Android Bindings / check-swift]' does.

The changes look good to me - if the tests are passing then should be fine.

PR #2143, opened 2025-06-26
Fork fixes

richardhuaaa:
Cherry picks fixes from @mchenani's forking investigation, as well as welcome logging from https://github.com/xmtp/libxmtp/pull/2068.

Changes:

1. Do not allow syncing to proceed on inactive groups (for example, groups received from device sync, or groups that we are no longer a member of)
2. Discard welcomes for groups that we are already active in (for example, if another member of the group sends us a welcome for the group that comes from a past epoch)
3. When syncing, call `maybe_update_installations()` after the sync is complete rather than before - this doesn't 'fix' forks necessarily, but significantly reduces race conditions
4. Additional logging, especially around welcome processing

Changes have been manually tested via Node.js integration tests, as well as existing Rust unit tests. I also tried to write a Rust unit test for (2) but was unable to reproduce - I think this happens only in very special race conditions that I haven't been able to pin down.

### DIFF
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -577,7 +577,7 @@ where
         // in the `GroupMembership` extension.
         validate_initial_group_membership(&context, &staged_welcome).await?;
         let group_id = staged_welcome.public_group().group_id();
-        if let Some(_) = provider.db().find_group(group_id.as_slice())? {
+        if provider.db().find_group(group_id.as_slice())?.is_some() {
             // Fetch the original MLS group, rather than the one from the welcome
             let (group, _) = MlsGroup::new_cached(context.clone(), group_id.as_slice())?;
             // Check the group epoch as well, because we may not have synced the latest is_active state
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -579,23 +579,31 @@ where
         let group_id = staged_welcome.public_group().group_id();
         if provider.db().find_group(group_id.as_slice())?.is_some() {
             // Fetch the original MLS group, rather than the one from the welcome
-            let (group, _) = MlsGroup::new_cached(context.clone(), group_id.as_slice())?;
-            // Check the group epoch as well, because we may not have synced the latest is_active state
-            // TODO(rich): Design a better way to detect if incoming welcomes are valid
-            if group.is_active()?
-                && staged_welcome
-                    .public_group()
-                    .group_context()
-                    .epoch()
-                    .as_u64()
-                    <= group.epoch().await?
-            {
+            let result = MlsGroup::new_cached(context.clone(), group_id.as_slice());
+            if result.is_err() {
                 tracing::error!(
-                    "Skipping welcome {} because we are already in group {}",
-                    welcome.id,
-                    hex::encode(group_id.as_slice())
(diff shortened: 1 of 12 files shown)


xmtp_mls/src/groups/mod.rs
```diff
@@ -614,6 +615,13 @@ where
             let disappearing_settings = mutable_metadata.clone().and_then(|metadata| {
                 Self::conversation_message_disappearing_settings_from_extensions(metadata).ok()
             });
+
+            if let Some(group) = provider.db().find_group(group_id.as_slice())?{
+                if group.membership_state == GroupMembershipState::Allowed {
```
neekolas: Are we sure the `GroupMembershipState` is the thing we want to be using to make this decision? If we have moved the group to "rejected" or the membership state is still "pending", from an MLS perspective nothing is different.

I worry this is going to be a footgun.

xmtp_mls/src/subscriptions/stream_conversations.rs
```diff
@@ -620,13 +620,18 @@ mod test {
             .await
             .unwrap();
         futures::pin_mut!(stream);
+        // TODO(rich): The next welcome to re-add Bo will currently be discarded, unless
+        // we sync first and detect that Bo has been removed from the group.
+        bo.sync_all_welcomes_and_groups(None).await.unwrap();
```
richardhuaaa: This unit test illustrates a tradeoff we are making with this fix - being re-added to a group you've been removed from may fail if you have not synced the group up to the point of your removal, because the welcome will be discarded.

I did try to sync during the welcome processing errors, but that led to stack overflow errors and other unit test failures, which we don't have time to address for the moment.

richardhuaaa: I decided to check the epoch on the welcome instead, which should address this issue more cleanly

richardhuaaa:
There are lint failures unrelated to this changeset that I want to ignore to keep the changeset small and easy to cherry-pick into the release branch

mchenani (approved):
🥳

xmtp_mls/src/groups/mod.rs
```diff
@@ -551,31 +551,53 @@ where
...
-        let DecryptedWelcome { staged_welcome, .. } = decrypted_welcome.expect("Set to some")?;
-
+        let DecryptedWelcome { staged_welcome, .. } = decrypt_result?;
         // Ensure that the list of members in the group's MLS tree matches the list of inboxes specified
         // in the `GroupMembership` extension.
         validate_initial_group_membership(&context, &staged_welcome).await?;
+        let group_id = staged_welcome.public_group().group_id();
+        if provider.db().find_group(group_id.as_slice())?.is_some() {
+            // Fetch the original MLS group, rather than the one from the welcome
+            let (group, _) = MlsGroup::new_cached(context.clone(), group_id.as_slice())?;
```
neekolas: We are sure that in the case that the group came in from DeviceSync this will always at least return an `Ok` result?

Does the DeviceSync MLS group think that it is active? Is it in epoch 0?

richardhuaaa: The DeviceSyncMLS group thinks it is active and in epoch 0. But it's a good point - given the sensitivity of this region, and our limited ability to verify the result right now or protect against future changes, I can just swallow all errors here

PR #2144, opened 2025-06-27
Add Welcomes crate

richardhuaaa:


xmtp_mls/src/welcomes/decrypt.rs
```diff
@@ -6,27 +8,55 @@ use openmls::{
...
+    ) -> Result<DecryptedWelcome, GroupError>
+    where
+        ApiClient: XmtpApi,
+        Db: XmtpDb,
+    {
+        let provider = context.mls_provider();
+        let mut decrypt_result: Result<DecryptedWelcome, GroupError> =
+            Err(GroupError::UninitializedResult);
+        let transaction_result = provider.transaction(|provider| {
+            decrypt_result = DecryptedWelcome::from_encrypted_bytes(
```
insipx, line 45: At some point we spoke with cryspen about the issues where Staged versions of Messages/Welcomes were hitting the DB and it was fixed it at some point, so that we could avoid this intentional rollback thing. It would be nice to create a test to verify this and then remove the rollback if its not writing anything

created an issue for it: https://github.com/xmtp/libxmtp/issues/2199

I think there was a similar fix for processed messages, but can't find that specific PR currently

insipx:
always in favor of more focused modules!

PR #2146, opened 2025-06-27
Fix more lint warnings

richardhuaaa:
I'm mystified why the lint CI passed on the last PR but then failed afterwards. In any case, there are some more lints to fix here.

### DIFF
--- a/common/src/retry.rs
+++ b/common/src/retry.rs
@@ -397,7 +397,7 @@ pub(crate) mod tests {
     }
 
     fn retryable_with_args(foo: usize, name: String, list: &Vec<String>) -> Result<(), SomeError> {
-        println!("I am {} of {} with items {:?}", foo, name, list);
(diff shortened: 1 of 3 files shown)


richardhuaaa:
So `uninlined-format-args` seems to be newly checked - we have quite a lot of historical code that violates this rule. It also doesn't show up when running the clippy command locally unless explicitly specified. The lint failures also don't show up all at once, instead coming in batches when the previous batch is fixed, making it super annoying to fix. I'll try just disabling the rule seeing as it wasn't checked in the past.

richardhuaaa:
It seems that the root cause is Rust 1.88 released last week which introduced new lints to clippy. So if we update cargo locally then the local behavior should match the remote behavior.

richardhuaaa:
I've fixed the lints for some of the packages, and added ignore rules individually to all the other packages, so we can fix them one by one if we want to.

Issue #2148, opened 2025-06-28
SyncAll is slow when inbox has many groups


nplasterer:
On a inbox with 74 convos and groups. No requests. It can take up to 5 seconds to see new messages on initial sync after reopening.

Potential improvements to syncAll may be a solution
Allow passing of groupIds to sync? Paging syncs?
Possibly the slow down is actually the RN bridge and not libxmtp and we should start thinking about a path to get rid of the bridge

nplasterer:
Possibly related issue https://github.com/xmtp/libxmtp/issues/1558

Issue #1558, opened 2025-01-28
Bug: syncAllConversations may overwhelm servers when executed from accounts with many conversations


cameronvoell:
### Describe the bug

Regardless of server side constraints, we need to ensure that we are not attempting to sync all conversations concurrently. 

### Expected behavior

Absent API improvements (batching updates for several groups), we should look into some client side batching to make sure that we are not trying to sync many groups and update installations for many groups concurrently.

cameronvoell:
not sure why github linked this issue to #1577 🤷

cameronvoell:
Think we can mark this resolved by https://github.com/xmtp/libxmtp/pull/2156

PR #2156, opened 2025-07-01
Sync All Performance

nplasterer:
Closes https://github.com/xmtp/libxmtp/issues/2148

Does several things
- Syncs conversations in the order of latest message
- Adds batching to conversation syncing
- Does not fail if one sync conversation fails
- Does not fail if sync welcomes fails


xmtp_mls/src/groups/welcome_sync.rs
```diff
@@ -148,42 +148,108 @@ where
...
+        let groups = provider
+            .db()
+            .all_sync_groups()?
+            .into_iter()
+            .map(|g| MlsGroup::new(self.context.clone(), g.id, g.dm_id, g.created_at_ns))
+            .collect();
+        let active_groups_count = self.sync_all_groups(groups).await?;
+
+        Ok(active_groups_count)
+    }
```
nplasterer, line 163: Just ordered this one shouldn't be any change here.

xmtp_mls/src/groups/welcome_sync.rs
```diff
@@ -148,42 +148,108 @@ where
...
     pub async fn sync_all_welcomes_and_groups(
         &self,
         consent_states: Option<Vec<ConsentState>>,
     ) -> Result<usize, GroupError> {
         let provider = self.context.mls_provider();
-        self.sync_welcomes().await?;
+
+        if let Err(err) = self.sync_welcomes().await {
+            tracing::warn!(?err, "sync_welcomes failed, continuing with group sync");
+        }
```
nplasterer, line 175: Don't error and end the sync if syncing welcomes fails. Throw a warning instead? Is this bad? Will this hide bugs? Should we be swallowing a failed sync_welcome?

xmtp_mls/src/groups/welcome_sync.rs
```diff
@@ -148,42 +148,108 @@ where
...
             consent_states,
             include_duplicate_dms: true,
             include_sync_groups: true,
             ..GroupQueryArgs::default()
         };
-        let groups = provider
-            .db()
-            .find_groups(query_args)?
+
+        let conversations = provider.db().fetch_conversation_list(query_args)?;
```
nplasterer, line 184: `find_groups` sorts by created_at
`fetch_conversation_list` sorts by last message

This will sync groups by last message first.

xmtp_mls/src/groups/welcome_sync.rs
```diff
@@ -148,42 +148,108 @@ where
...
+
+        let groups: Vec<MlsGroup<Api, Db>> = conversations
             .into_iter()
-            .map(|g| MlsGroup::new(self.context.clone(), g.id, g.dm_id, g.created_at_ns))
+            .map(|c| MlsGroup::new(self.context.clone(), c.id, c.dm_id, c.created_at_ns))
             .collect();
-        let active_groups_count = self.sync_all_groups(groups).await?;
 
-        Ok(active_groups_count)
+        let success_count = self.sync_groups_in_batches(groups, 10).await?;
```
nplasterer, line 191: Could allow passing a batch size. But I think 10 is pretty good for now.

xmtp_mls/src/groups/welcome_sync.rs
```diff
@@ -148,42 +148,108 @@ where
...
+    /// Sync groups in batches to limit concurrency. Returns success count.
+    pub async fn sync_groups_in_batches(
+        &self,
+        groups: Vec<MlsGroup<Api, Db>>,
+        batch_size: usize,
+    ) -> Result<usize, GroupError> {
+        let active_group_count = Arc::new(AtomicUsize::new(0));
+        let failed_group_count = Arc::new(AtomicUsize::new(0));
+
+        for chunk in groups.chunks(batch_size) {
```
nplasterer: This should address Camerons concerns here https://github.com/xmtp/libxmtp/issues/1558 by batching instead of running them all concurrently at once.

xmtp_mls/src/client.rs
```diff
@@ -1278,6 +1278,65 @@ pub(crate) mod tests {
...
+            group
+                .add_members_by_inbox_id(&[bo.inbox_id()])
+                .await
+                .unwrap();
+            groups.push(group);
+        }
+
+        xmtp_common::time::sleep(Duration::from_millis(100)).await;
+
+        let start = std::time::Instant::now();
```
insipx: should use `xmtp_common::time::Instant::now()` so that this works in both native and webassembly

insipx:
```suggestion
        let start = xmtp_common::time::Instant::now();
```

xmtp_mls/src/client.rs
```diff
@@ -1278,6 +1278,65 @@ pub(crate) mod tests {
...
+        );
+
+        println!(
+            "Synced {} groups in {:?} (avg per group: {:?})",
+            group_count,
+            elapsed,
+            elapsed / group_count as u32
+        );
+
+        let start = std::time::Instant::now();
```
insipx:
```suggestion
        let start = xmtp_common::time::Instant::now();
```

xmtp_mls/src/groups/welcome_sync.rs
insipx: it could be more elegant to use [try_for_each_concurrent](https://docs.rs/futures/latest/futures/prelude/stream/trait.TryStreamExt.html#method.try_for_each_concurrent) or just [for_each_concurrent](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.for_each_concurrent) here, after converting the vector of groups to a stream with `stream::iter` and passing a limit to the function. the example in the docs is very similar to this use case

PR #2167, opened 2025-07-07
Add retry logic for tcp connection

nplasterer:
We are seeing an increase in 
`tcp connect error`
&
`transport error`

This is likely due to users being on poor connections.
I think our timeouts and backoffs are pretty generous at the moment. Lets add retry logic to keep trying to connect when issues arise.


xmtp_api_grpc/src/lib.rs
```diff
@@ -13,39 +13,56 @@ use std::time::Duration;
...
+            // Functionality: When set to true, this option ensures that periodic pings are sent on an idle connection to keep it alive and detect if the server is still responsive.
+            // Impact: This helps maintain active connections, particularly through NATs, load balancers, and other middleboxes that might drop idle connections. It helps ensure that the connection is promptly usable when new requests need to be sent.
+            .keep_alive_while_idle(true)
+            // Purpose: Sets the maximum amount of time the client will wait for a connection to be established.
+            // Functionality: If a connection cannot be established within the specified duration, the attempt is aborted and an error is returned.
+            // Impact: This setting prevents the client from waiting indefinitely for a connection to be established, which is crucial in scenarios where rapid failure detection is necessary to maintain responsiveness or to quickly fallback to alternative services or retry logic.
+            .connect_timeout(Duration::from_secs(10))
```
nplasterer, line 45: Should we lower this to 5?

neekolas: That's probably fine. 5 seconds is a lot to just do the TLS handshake and connect. Although on a spotty mobile network, maybe not enough. Wouldn't want to create a situation where a client consistently takes 6 seconds to connect and gets boxed out entirely.

IDK. I'd say leave it for now.

nplasterer:
I'm not sure if this is going to give us anything of real value. May close in favor of my sdk caching refresh check.

PR #2172, opened 2025-07-08
Is Connected Check

nplasterer:
Add a `isConnectedCheck` to the grpc api client so that we can force a reconnection if the cached client is broken.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -116,10 +116,9 @@ pub async fn connect_to_backend(
     Ok(Arc::new(XmtpApiClient(api_client)))
 }
 
-#[uniffi::export]
-pub fn is_connected(api: Arc<XmtpApiClient>) -> bool {
-    let mut api = Arc::unwrap_or_clone(api);
-    api.0.is_connected()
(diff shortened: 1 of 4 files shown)


insipx (approved):

bindings_ffi/src/mls.rs
```diff
@@ -116,9 +116,16 @@ pub async fn connect_to_backend(
     Ok(Arc::new(XmtpApiClient(api_client)))
 }
 
+#[uniffi::export]
+pub fn is_connected(api: Arc<XmtpApiClient>) -> bool {
```
insipx: is it possible this returns false negatives, for instance if the channel has just used up all its permits b/c its sending a bunch of requests at once? in that case, we'd have lots of requests pending but this might return not connected as a result

nplasterer: Nice catch I was just actually looking into this. But I think you're right. Gonna push up a fix.

nplasterer: Okay I think I resolved that in https://github.com/xmtp/libxmtp/pull/2172/commits/de416eaee74838c8719a1580155bf39a4cbf1066 by making it async and removing the now or never check

PR #2175, opened 2025-07-08
Make `applied_epoch_number` and `applied_epoch_authenticator` non-optional

richardhuaaa:
Making these fields non-optional simplifies comparison between the local commit log and the remote commit log, as you now only need to find the latest sequence ID they have in common and verify that the epoch_authenticator matches.

Additionally, there was a direct modification to the DB migration script in #2106, but that PR will not be landing in the same release that the migration script was added as originally intended. I moved the schema change from that PR, as well as the additions in this PR, to a separate migration script.


xmtp_db/migrations/2025-06-14-004145_create_commit_log/up.sql
```diff
@@ -1,35 +1,21 @@
...
     "commit_result" INT NOT NULL,
-    "error_message" TEXT,
-    -- Items below this line are only set if the commit was applied
     "applied_epoch_number" BIGINT,
     "applied_epoch_authenticator" BLOB,
-    -- Items below this line are for debugging purposes
     "sender_inbox_id" TEXT,
     "sender_installation_id" BLOB,
-    "commit_type" TEXT
+    "commit_type" INT
```
richardhuaaa, line 11: Simply undoing the changes from #2106, and removing the comments

nplasterer (approved):

PR #2177, opened 2025-07-08
Fix keepalives and use all network settings in local

neekolas:
### TL;DR

Improved gRPC connection reliability by refactoring channel options and adding keepalive settings.

### What changed?

- Extracted common channel configuration into a new `apply_channel_options` function
- Reduced keep-alive timeout from 25 seconds to 5 seconds
- Added HTTP/2 keep-alive interval of 15 seconds
- Added a test case to verify reconnection behavior when connections are interrupted

### How to test?

Run the new test case `should_reconnect` which verifies that clients can recover from connection failures:
- The test simulates a network interruption using a proxy
- It verifies that the client detects the failure and can establish a new connection
- It confirms that streaming operations resume successfully after reconnection

### Why make this change?

This change improves the reliability of gRPC connections, especially in unstable network environments. By reducing the keep-alive timeout and adding regular keep-alive intervals, the client can detect connection failures more quickly and initiate reconnection, resulting in better resilience against network disruptions.

PR #2182, opened 2025-07-09
node test iwhth toxiproxy

insipx:


insipx:
* **#2189** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2189?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2182** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2182?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2182?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2177** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

xmtp_api_grpc/src/streams.rs
```diff
@@ -0,0 +1,122 @@
...
+    ) -> std::task::Poll<Option<Self::Item>> {
+        // if we are broken, do not attempt to poll
+        // the inner stream anymore
+        if self.is_broken {
+            return Poll::Ready(None);
+        }
+        let mut this = self.as_mut().project();
+        let item = ready!(this.inner.as_mut().poll_next(cx));
+        match item {
+            Some(Err(e)) => {
```
neekolas, line 60: I want to make sure I understand the decision tree here.

- If we get a broken pipe error: we mark the stream as broken and will return `Poll::Ready(None)` on the next poll. 
- If we get another type of error we return it and continue.

What is the expectation of the caller when they get the error returned to them? Do we know what other kinds of errors can come down these streams? Are we sure all forms of disconnect ultimately turn into a broken pipe?

xmtp_mls/src/subscriptions/stream_conversations.rs
```diff
@@ -131,6 +131,86 @@ where
...
+                    return Poll::Ready(None);
+                } else {
+                    cx.waker().wake_by_ref();
+                    return Poll::Pending;
+                }
+            }
+            Poll::Pending => (),
+        };
+
+        match this.s2.as_mut().poll_next(cx) {
```
neekolas: Why is the business logic simpler for s2 than s1? Is there an implicit assumption that S2 is of a different type than S1?

insipx: Since S1 is polled first, S2 can be simpler -- S1 would have returned Poll::Ready with an item before ever reaching S2. Therefore, we can assume that if we got to S2, S1 is pending and doesn't have anything sitting in its queue

Whereas if S1 finishes, it's possible S1 had a bunch of items in the pipeline before it ended and was starving S2 of emitting its events the whole time.

A better implementation would round robin both streams to keep it fair + require both streams to implement FusedStream so as to separate polling and termination check steps. But, I was going for speed. Plus, for the purposes of welcome messages it will work ok, since the broadcast stream are local events and we never expect a huge amount of welcome messages at once


Ideally i'd like to unit test this with this in mind + implement the round robin. Unfortunately,this sort of multiplex is required since the previous [select](https://docs.rs/futures/latest/futures/future/fn.select.html) thing wouldn't work b/c of the broadcast stream

xmtp_mls/src/subscriptions/stream_conversations.rs
```diff
@@ -131,6 +131,86 @@ where
...
+impl<S1, S2, T> Stream for MultiplexedStream<S1, S2, T>
+where
+    S1: Stream<Item = T>,
+    S2: Stream<Item = T>,
+{
+    type Item = T;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut this = self.as_mut().project();
+        if *this.terminated {
```
insipx: there's definitely a bug here where the `FusedStream` implementation will return terminated as `true`, while its still possible that the stream has items sitting in `queue` and therefore still expects to be polled which goes against the rust Future contract (should not be polled once ended). Not as bad as it could be since we implement `FusedStream` ourselves, so no panic can occur.

xmtp_api_grpc/src/streams.rs
insipx: The decision tree is correct. Further, we guarantee the next poll through calling `wake_by_ref()`

> Do we know what other kinds of errors can come down these streams?

these streams can get any of the tonic errors: https://docs.rs/tonic/latest/tonic/struct.Status.html

If its an http error rather than gRPC, it comes up in gRPC as 'Unknown', and AFAIU can be any one of the [hyper errors](https://docs.rs/hyper/latest/hyper/struct.Error.html) and then [h2 error](docs.rs/h2/latest/h2/struct.Error.html) and then [std::io::Error](https://doc.rust-lang.org/nightly/std/io/enum.ErrorKind.html)

We do have the ability to check for specific errors, but I'm not knowledgeable enough on the internals of hyper/h2 to know for certain what could happen, im assuming anything that can go wrong with http/2 or grpc can go wrong here

> Are we sure all forms of disconnect ultimately turn into a broken pipe?

on this I still don't know, it could be useful to ask the tonic team what they think of our implementation here, they could maybe have more input on best practices. not sure how fast we could get a response, but I'll try their discord in the morning

It seems Fabri already found a case where the network goes out, but stream does not close b/c the error is a DNS error rather than the BrokenPipe. I guess there's a hope that HTTP/2 keepalive will always result in broken pipe after the interval, but not sure if that holds in all cases.

> What is the expectation of the caller when they get the error returned to them?

I didn't think about this too hard, but it felt like a good choice to return the error we're ending the stream on, since the general behavior of the stream before this was to pass all errors to the caller. If we instead silently close the stream on this error without returning to the caller, it could technically be a breaking change if a developer using this sdk is already handling the broken pipe error in a different way. 

But, streams closing on broken pipe would be a breaking change anyway. 


the other thought I had was w.r.t logging, it lets the developer see the exact error closing the stream

i don't feel too strongly leaving the BrokenPipe error out of the stream though

xmtp_mls/src/subscriptions/stream_conversations.rs
insipx: some improvements & unit tests: https://github.com/xmtp/libxmtp/pull/2182/commits/c63bab91ed1336b75716241acb10423c748546af

yesterdays rendition definitely still had a starvation issue

xmtp_mls/src/subscriptions/stream_conversations.rs
insipx: this is fixed

xmtp_mls/src/subscriptions/stream_utils/multiplexed.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl<S1, S2> MultiplexedStream<S1, S2>
+where
+    S1: Stream<Item = S2::Item>,
+    S2: Stream,
+{
+    fn poll_s2(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<S2::Item>> {
+        let mut this = self.as_mut().project();
+        if let Poll::Ready(Some(item)) = this.s2.as_mut().poll_next(cx) {
```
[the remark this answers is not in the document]
insipx: this is intentional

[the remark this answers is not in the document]
insipx: The multiplex stream type takes two streams, as soon as the first stream ends, we want to kill the multiplex stream type. We try to be fair by draining S2s items, but only as long as S2 has items which are Ready. Otherwise, we want to close the stream asap.

nplasterer (approved):

insipx:
### Merge activity

* **Jul 10, 5:51 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2182).
* **Jul 10, 5:51 PM UTC**: @insipx merged this pull request with [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2182).

PR #2180, opened 2025-07-09
Add welcomes to local commit log

richardhuaaa:
When Welcomes have a cursor pointing to where in the group the newly added installation should sync from (#2088), we should also be able to use this as the `commit_sequence_id` in the local/remote commit log. This would allow a newly added installation to immediately check if they were added into a fork, tackling the viral nature of forks today.

Because we don't have this information yet, we can instead store it with a sequence_id of 0 for local debugging purposes for now. Anything with a sequence_id of 0 won't be uploaded to the remote commit log, and is effectively ignored for the purposes of fork detection.

PR #2179, opened 2025-07-08
Add group creation to local commit log

richardhuaaa:
This adds group creation to the local commit log.

This should help with debugging only, not automated fork recovery. We don't need to upload these to the remote commit log, because no other installation is going to be able to see this encryption state (all other installations will receive a welcome that points to a commit made after group creation). I've updated the XIP to explicitly state that log entries with a sequence ID of 0 should not be uploaded to the remote commit log, and should be ignored when read from the remote commit log.

PR #2185, opened 2025-07-09
Improved commit log error handling

richardhuaaa:
Changes include:
1. After joining a group from a Welcome message, do not add errored commit log entries for commits sent before the installation joined the group. When we get an epoch error, we do this by checking if the last entry was a Welcome, or if there is no previous entry. Worth noting:
     1. Even if we implement logic to add a cursor to Welcomes and only sync from that cursor onwards, we cannot rely on old clients populating this cursor when they send us a Welcome
     1. When we get an epoch error, we can check if the last commit log entry is a Welcome to determine whether to ignore it, but we can't rely on that entry being present - the client may have received a Welcome on an older version, and not have written that entry to the log.
2. Populate the `epoch_number` and `epoch_authenticator` for error commit logs from the actual group state, rather than inferring it from the most recent entry

We also need to be careful about how we handle Welcomes that are readds (for example, as part of automated fork recovery, or after leaving and re-joining the group). I chose not to clear the local commit log on a readd, so that we have a clearer audit trail for debugging. Instead, when comparing the local commit log with the remote commit log, we can stop comparing when we reach the most recent Welcome. I've updated the XIP to describe this.

Next PR will add more unit tests for errors, and have some light refactoring.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -23,6 +23,7 @@
     "tokio": ["test"],
     "async-stream": ["stream", "try_stream"]
   },
+  "gitlens.blame.ignoreWhitespace": true,
   "[toml]": {
     "editor.defaultFormatter": "tamasfe.even-better-toml"
   },
--- a/xmtp_mls/src/groups/mls_ext/commit_log_storer.rs
+++ b/xmtp_mls/src/groups/mls_ext/commit_log_storer.rs
@@ -1,10 +1,13 @@
 use crate::groups::GroupError;
 use crate::groups::{mls_sync::GroupMessageProcessingError, validated_commit::ValidatedCommit};
 use crate::identity::Identity;
+use crate::StorageError;
 use openmls::group::{MlsGroup, MlsGroupCreateConfig, StagedCommit};
 use openmls::prelude::CredentialWithKey;
+use openmls::prelude::GroupEpoch;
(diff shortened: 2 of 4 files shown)


richardhuaaa:
* **#2223** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2223?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2213** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2185** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2185?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2185?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2180** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2180?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2179** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2179?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

cameronvoell (approved):

PR #2196, opened 2025-07-11
[1/2] Refactor intent validation for clarity

richardhuaaa:
I need to do some refactoring to support the automated fork recovery feature. As part of that refactor there are two functional changes listed in the PR comments, and hopefully no other change.

Before this PR, the `stage_and_validate_intent()` method returned a:
```
Result<
        Option<(StagedCommit, ValidatedCommit)>,
        Result<IntentState, GroupMessageProcessingError>,
    >
```

Specifically, there are four states here:

1. If the intent was a `SendMessage` and it was valid, we return `Ok(None)`
2. If the intent was a commit and it was valid, we return `Ok((StagedCommit, ValidatedCommit))`
3. If the function ran successfully but the intent was not valid, it would return either `Err(Ok(IntentState::ToPublish))` or `Err(Ok(IntentState::Error))` depending on what we wanted to update the intent state to.
4. If the function did not run successfully, it would return an `Err(Err())`, but not update the intent state.

This PR tries to simplify (3) and (4) and remove the result nesting. As you'll see in the comments, the important distinguishing factor is not whether the function ran successfully or not, but whether the resulting error is retriable or not.

### DIFF
--- a/xmtp_mls/src/groups/mls_sync.rs
+++ b/xmtp_mls/src/groups/mls_sync.rs
@@ -209,6 +209,26 @@ impl RetryableError for GroupMessageProcessingError {
     }
 }
 
+#[derive(Debug, Error)]
+pub struct IntentValidationError {
+    processing_error: GroupMessageProcessingError,
+    // The next intent state to transition to, if the error is non-retriable.
+    // Should not be used for retryable errors.
+    next_intent_state: IntentState,
+}
+
+impl std::fmt::Display for IntentValidationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "IntentValidationError: {}", self.processing_error)
+    }
+}
+
+impl RetryableError for IntentValidationError {
+    fn is_retryable(&self) -> bool {
+        self.processing_error.is_retryable()
+    }
+}
+
 #[derive(Debug)]
 struct PublishIntentData {
     staged_commit: Option<Vec<u8>>,
@@ -492,10 +512,7 @@ where
         intent: &StoredGroupIntent,
         message: &ProtocolMessage,
         envelope: &GroupMessageV1,
-    ) -> Result<
-        Option<(StagedCommit, ValidatedCommit)>,
-        Result<IntentState, GroupMessageProcessingError>,
-    > {
+    ) -> Result<Option<(StagedCommit, ValidatedCommit)>, IntentValidationError> {
         let GroupMessageV1 {
             created_ns: envelope_timestamp_ns,
             id: ref cursor,
@@ -514,6 +531,7 @@ where
                 if let Some(published_in_epoch) = intent.published_in_epoch {
                     let group_epoch = group_epoch.as_u64() as i64;
 
+                    // TODO(rich): Merge into validate_message_epoch()
                     if published_in_epoch != group_epoch {
                         tracing::warn!(
                             inbox_id = self.context.inbox_id(),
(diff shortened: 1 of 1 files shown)


richardhuaaa:
* **#2201** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2196** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2196?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2196?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2212** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -525,18 +528,49 @@ where
...
+                                group_id = hex::encode(&self.group_id),
+                                cursor = cursor,
+                                intent_id = intent.id,
+                                intent.kind = %intent.kind,
+                                "Error decoding staged commit for intent, now may be forked: {err:?}",
+                            );
+                            IntentValidationError {
+                                processing_error: err,
+                                next_intent_state: IntentState::Error,
+                            }
```
richardhuaaa, line 588: This is one of only two functional changes in this PR as far as I can tell. Previously, an error here would return an `Err(err)`, which will get bubbled up by the caller on L1245 of the old code.

Now, because the error is non-retriable, the caller will swallow it on L1279 of the new code and update the intent state to `Error`.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1184,95 +1224,89 @@ where
...
+                        // cursor, but actually cannot process the message.
+                        identifier.previously_processed(true);
+                        return Ok(());
+                    }
+                    let (intent_state, internal_message_id) = match maybe_validated_commit {
+                        Err(err) => {
+                            if err.processing_error.is_retryable() {
+                                // Rollback the transaction so that we can retry
+                                return Err(err.processing_error);
                             }
```
richardhuaaa, line 1289: We are processing the error from above here. I checked and don't see any retriable errors returned from `stage_and_validate_intent()` - note there are no DB or networking calls.

Even though there's no functional difference I chose to include this logic anyway, in case `stage_and_validate_intent()` is modified to return a retriable error in the future - in that case, we should be bubbling up the error here and rolling everything back so we can try again.

richardhuaaa:
The tests are currently failing due to an unrelated failure on main - other than that, all existing tests pass.

insipx (approved):
nice, glad to cleanup that doubly nested error

Flaky test fix is up here https://github.com/xmtp/libxmtp/pull/2198

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -563,9 +597,12 @@ where
...
                             );
-                            // Return before merging commit since it does not pass validation
-                            // Return OK so that the group intent update is still written to the DB
-                            return Err(Ok(IntentState::Error));
+                            return Err(IntentValidationError {
+                                processing_error: GroupMessageProcessingError::CommitValidation(
+                                    err,
+                                ),
+                                next_intent_state: IntentState::Error,
+                            });
```
richardhuaaa, line 621: This is another functional change - `CommitValidation` could return a retriable error (as it makes DB calls that could lock). Before, the intent would be marked as errored and we wouldn't apply the commit, which could cause a fork. Now, the caller will see that it is a retriable error and bubble it up so that it can be retried.

richardhuaaa: I'm wrong that there are no retriable errors returned - `CommitValidationError` might return a retriable error above. In that case the change here might actually help prevent forks.

xmtp_mls/src/groups/mls_sync.rs
richardhuaaa: I'm wrong that there are no retriable errors returned - `CommitValidationError` might return a retriable error above, I've added a PR comment explaining. In that case the change here might actually help prevent forks.

richardhuaaa:
`test_disappearing_messages_deletion` is flaky after this change. Because it's very time-consuming to validate these PR's don't regress anything, I'm going to pause on this for now and prioritize work on fork recovery, before coming back to this stack. I do think there are fork root causes being identified/addressed in this PR and the one stacked on top though.

Primarily, when you are processing your own intent, in several cases we're just transitioning the intent state rather than checking if the error is retriable or not (e.g. if the DB is locked).

xmtp_mls/src/groups/mls_sync.rs
neekolas: This seems like a big net positive

xmtp_mls/src/groups/mls_sync.rs
neekolas: These are great finds

neekolas (approved):

neekolas:
This seems like a very positive change that better organizes how we handle errors processing your own messages. Lets let it rip against the E2E test suite and see what comes out.

richardhuaaa:
### Merge activity

* **Jul 15, 6:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2196).
* **Jul 15, 6:18 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2196) rebased this pull request as part of a merge.
* **Jul 15, 6:26 PM UTC**: @richardhuaaa merged this pull request with [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2196).

PR #2115, opened 2025-06-16
intent unit tests

insipx:

PR #2201, opened 2025-07-11
refactor own message processing

richardhuaaa:


richardhuaaa:
* **#2201** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2196** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2212** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1285,22 +1338,29 @@ where
...
                                 return Err(err.processing_error);
                             }
                             (err.next_intent_state, None)
-                        }
-                        Ok(commit) => {
-                            self
-                            .process_own_message(&mut mls_group, commit, &intent, &message, envelope)?
-                        }
+                        },
+                        Ok(internal_message_id) => (IntentState::Committed, internal_message_id)
```
richardhuaaa: This is the core of the functional change, which is similar to the previous PR. When `process_own_message` is called:
1. If it was successful, the intent is marked as `Committed`
2. If it was a non-retriable error, we transition the intent to `next_intent_state`
3. If was a retriable error, we bubble it up so that the transaction can be rolled back and retried

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -696,7 +714,13 @@ where
...
             ) {
                 tracing::error!("error merging commit: {err}");
-                return Ok((IntentState::ToPublish, None));
+                return Err(IntentResolutionError {
+                    processing_error: err,
+                    // If the error is non-retriable, it means the commit failed to apply due to some
+                    // issue with the commit (e.g. encryption problem). We reset the intent state to
+                    // ToPublish so that we can republish it.
+                    next_intent_state: IntentState::ToPublish,
+                });
```
richardhuaaa, line 728: Important change:
- Previously, on any error, we mark the intent as `ToPublish` and re-publish the intent
- Now, we do the same, _unless it's a retriable error_. An example of a retriable error would be a DB lock when we are merging the commit into the OpenMLS state, or when we are inserting the commit log entry after. In this case we bubble up the error so it can be retried, which prevents a fork.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -714,16 +738,45 @@ where
...
-            conn.set_delivery_status_to_published(&id, envelope_timestamp_ns, envelope.id as i64)?;
-            return Ok((IntentState::Processed, Some(id)));
+            let msg = self
+                .save_transcript_message(validated_commit, envelope_timestamp_ns, *cursor)
+                .map_err(|err| IntentResolutionError {
+                    processing_error: err,
+                    // If it is a non-retriable error, the commit will be applied, but the transcript message
+                    // will be missing. We mark the intent state as errored and continue.
+                    next_intent_state: IntentState::Error,
+                })?;
```
richardhuaaa, line 753: Unimportant change:
- If a retriable error is returned here, both the previous and current code would bubble it up and retry it.
- If a non-retriable error is returned here, the previous code would bubble it up and never transition the intent state. The current code would transition the intent to `Error` and move on.

In practice I don't see any non-retriable error that is likely with this method, so it probably doesn't change much.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -714,16 +738,45 @@ where
...
+                    );
+                }
+                IntentResolutionError {
+                    processing_error: err,
+                    // If the error is non-retriable, it means that the optimistic message (which is already in
+                    // the db) will never have its delivery status updated to published. We mark the intent state
+                    // as errored and continue.
+                    next_intent_state: IntentState::Error,
+                }
+            })?;
```
richardhuaaa, line 777: Unimportant changes:
1. Previously if `calculate_message_id_for_intent` returned `None`, we would set the intent state to `Committed`, now we set it to `Error` and log an error message. This shouldn't change much because either way we are done with this intent and move on to the next message.
2. Previously if `calculate_message_id_for_intent` returned a non-retriable error, we would unwind the transaction and bubble it up, now we mark the intent state as `Error` and move on.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -714,16 +738,45 @@ where
...
+        tracing::debug!("setting message @cursor=[{}] to published", envelope.id);
+        conn.set_delivery_status_to_published(&id, envelope_timestamp_ns, envelope.id as i64)
+            .map_err(|err| IntentResolutionError {
+                processing_error: GroupMessageProcessingError::Db(err),
+                next_intent_state: IntentState::Error,
+            })?;
+        Ok(Some(id))
```
richardhuaaa, line 784: Change:
- Previously if we were able to set the delivery status for the message, we marked the intent as `Processed`. Now, we simply return `Ok()`, and the outer function marks the intent as `Committed`.

This undoes the change from https://github.com/xmtp/libxmtp/pull/1899. I compensated by adding the logic on L663, which I believe addresses the original issue.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -653,9 +659,21 @@ where
...
+            tracing::warn!(
+                "Skipping already processed intent {} of kind {} because it is in state {:?}",
+                intent.id,
+                intent.kind,
+                intent.state
+            );
+            return Err(IntentResolutionError {
+                processing_error: GroupMessageProcessingError::IntentAlreadyProcessed,
+                next_intent_state: intent.state,
+            });
```
richardhuaaa, line 681: I believe this is an alternative way to address https://github.com/xmtp/libxmtp/pull/1899, but would be good for @neekolas to double-check. If the intent is already Committed or Processed, we return the same state as a non-retriable error, and skip the rest of the function. The outer function will just skip over the payload seeing as the state didn't change. This allowed me to simplify the return types and expected behavior - if this function returns `Ok()`, we are always transitioning the intent to `Committed`.

xmtp_db/src/encrypted_store/consent_record.rs
```diff
@@ -74,7 +74,7 @@ impl StoredConsentRecord {
                 last_consent.state,
                 hex::encode(&group.id),
             );
-            cr.store(conn)?;
+            conn.insert_newer_consent_record(cr)?;
```
richardhuaaa: Important change, cc @nplasterer and @humanagent:
- This function is called when processing the Welcome for a DM. Previously, if the consent sync inserted the consent entry before the Welcome for the DM was processed, the call to `cr.store(conn)?;` would fail with a unique constraint error, because we're trying to insert a consent record that already exists. This would abort processing of the Welcome and lead to a DM sub-group that was never received, causing the appearance of a fork.
- Now, we check if the consent record already exists, and either update it or no-op without returning an error, so that the Welcome is not dropped.

bindings_ffi/src/mls.rs
```diff
@@ -6967,7 +6967,7 @@ mod tests {
         // Wait for alix_a to send out the consent on the sync group
         alix_a
             .worker()
-            .wait(SyncMetric::ConsentSent, 3)
```
richardhuaaa, line 6970: I'm suspecting these tests were relying on racey behavior or error retries. After the changes in this PR, the third `ConsentSent` now never arrives. The reason I'm comfortable making these changes:

1. When I reason through the logic of the tests, there should only be two consent records sent.
2. These lines are used to synchronize timing before the final consent state is checked. The final consent state is still correct after this change.

xmtp_mls/src/groups/mls_sync.rs
neekolas: This seems like a nice cleanup as well

xmtp_mls/src/groups/mls_sync.rs
neekolas: This new setup is definitely clearer to read and reason through

xmtp_mls/src/groups/mls_sync.rs
neekolas: It would be great if we could mock/simulate some of these error conditions to enshrine the behaviour in tests

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -654,9 +660,21 @@ where
         intent: &StoredGroupIntent,
         message: &ProtocolMessage,
         envelope: &GroupMessageV1,
-    ) -> Result<(IntentState, Option<Vec<u8>>), GroupMessageProcessingError> {
-        if intent.state == IntentState::Committed {
-            return Ok((IntentState::Committed, None));
+    ) -> Result<Option<Vec<u8>>, IntentResolutionError> {
```
neekolas, line 667: Maybe we can add some comments to this function that explains clearly how to use `IntentResolutionError` and when it is safe to return an `Ok(..)` value

xmtp_db/src/encrypted_store/consent_record.rs
```diff
@@ -59,7 +59,7 @@ impl StoredConsentRecord {
 
     /// This function will perform some logic to see if a new group should be auto-consented
     /// or auto-denied based on past consent.
-    pub fn persist_consent<C: ConnectionExt>(
+    pub fn stitch_dm_consent<C: ConnectionExt>(
```
neekolas: We might want to rip out this one change to its own PR just so we can revert separately if needed and have a clearer commit history for anyone looking back.

neekolas (approved):

xmtp_mls/src/groups/mls_sync.rs
insipx: https://github.com/xmtp/libxmtp/pull/2115 should give us very fine-grained ability to mock every piece of intent code & isolate each mls operation

xmtp_db/src/encrypted_store/consent_record.rs
richardhuaaa: Split it into https://github.com/xmtp/libxmtp/pull/2212

insipx (approved):

richardhuaaa:
### Merge activity

* **Jul 15, 6:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2201).
* **Jul 15, 6:28 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2201) rebased this pull request as part of a merge.
* **Jul 15, 6:36 PM UTC**: @richardhuaaa merged this pull request with [Graphite](https://app.graphite.dev/github/pr/xmtp/libxmtp/2201).

PR #2213, opened 2025-07-15
Clean up local commit log non-retriable errors

richardhuaaa:
Some earlier refactors make these fixes possible now:
1. Log non-retriable errors when processing your own intent - previously these were not logged
2. Log all non-retriable errors within the same transaction as the one that increments the cursor

Easiest to review with whitespace diff turned off

PR #2223, opened 2025-07-17
Only deserialize once

richardhuaaa:
We are deserializing/extracting the same envelope between **two to four times** in each iteration of a retry loop. I've refactored to:
- Only deserialize once at the beginning
- Instead of passing down the whole deserialized payload, pass only the information from the payload that is needed

### DIFF
--- a/xmtp_mls/src/groups/mls_ext/commit_log_storer.rs
+++ b/xmtp_mls/src/groups/mls_ext/commit_log_storer.rs
@@ -7,7 +7,6 @@ use openmls::prelude::CredentialWithKey;
 use openmls::prelude::GroupEpoch;
 use openmls::prelude::GroupId;
 use openmls::prelude::StagedWelcome;
-use openmls::schedule::EpochAuthenticator;
 use xmtp_db::{
     local_commit_log::{CommitType, NewLocalCommitLog},
     remote_commit_log::CommitResult,
@@ -50,13 +49,11 @@ pub trait CommitLogStorer: std::marker::Sized {
     /// Specifically, do not call this for retryable errors, or
     /// VersionTooLow/GroupPaused errors.
     fn mark_failed_commit_logged<Db: ConnectionExt>(
+        &self,
         provider: &XmtpOpenMlsProvider<Db>,
-        group_id: &[u8],
         commit_sequence_id: u64,
         commit_epoch: GroupEpoch,
         error: &GroupMessageProcessingError,
-        last_epoch_number: GroupEpoch,
-        last_epoch_authenticator: &EpochAuthenticator,
     ) -> Result<(), StorageError>;
 }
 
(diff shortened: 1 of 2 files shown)


richardhuaaa:
* **#2213** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2223** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2223?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2223?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xmtp_mls/src/groups/mls_ext/commit_log_storer.rs
```diff
@@ -192,21 +189,22 @@ impl CommitLogStorer for MlsGroup {
     }
 
     fn mark_failed_commit_logged<Db: ConnectionExt>(
+        &self,
```
richardhuaaa, line 192: Unrelated simplification that was unlocked by a previous refactor, it was a tiny self-contained change so I just included it here.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1237,18 +1236,59 @@ where
         envelope: &GroupMessageV1,
         trust_message_order: bool,
     ) -> Result<MessageIdentifier, GroupMessageProcessingError> {
+        let mls_message_in = MlsMessageIn::tls_deserialize_exact(&envelope.data)?;
+        let message_entity_kind = match mls_message_in.wire_format() {
+            WireFormat::Welcome => EntityKind::Welcome,
+            _ => EntityKind::Group,
+        };
```
richardhuaaa, line 1243: This code block was pulled up from `process_message_inner`, which was the first function being called by this method.

You can see that this same code was pulled down from `consume_message()`, which is one of the callers of this method.

This is one way we deduplicate.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1237,18 +1236,59 @@ where
...
+
+        if trust_message_order {
```
richardhuaaa, line 1245: Right now this method has two callers, one from sync and one from streams, hence the need for this check. The need to check the `trust_message_order` param can be removed when we have totally ordered streams.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1237,18 +1236,59 @@ where
...
                 result = self
                     .post_process_message(
                         &mls_group,
                         result,
-                        envelope,
-                        mls_group.epoch(),
-                        mls_group.epoch_authenticator(),
+                        envelope.id,
+                        message_type,
+                        message_epoch,
```
richardhuaaa, line 1291: Previously, lower in the call-tree of `post_process_message`, the envelope was being deserialized again in order to derive `message_type` and `message_epoch`. We pass these new params down so that we no longer have to deserialize lower in the tree.

Separately, `mls_group.epoch()` and `mls_group.epoch_authenticator()` were removed because they can be derived from the `mls_group`.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1780,73 +1794,18 @@ where
         Ok(Some(msg))
     }
 
-    fn insert_commit_entry_for_failed_commit(
```
richardhuaaa, line 1777: The caller of this method now calls `mark_failed_commit_logged()` directly, seeing as we no longer need to deserialize first.

Issue #2184, opened 2025-07-09
Key Package rotation on 3 month expiry


nplasterer:
If a person is active but not receiving any new welcomes we still want to rotate their key package every 3 months.

PR #2224, opened 2025-07-17
rotate keypackage before the lifetime ends

mchenani:
### Rotate key packages before their lifetime ends by implementing proactive rotation scheduling with 30-day intervals and 5-second queue timing
- Changes the key package rotation system from reactive to proactive by setting `next_key_package_rotation_ns` to current time plus `KEY_PACKAGE_ROTATION_INTERVAL_NS` (30 days) when creating `StoredIdentity` instances in [xmtp_mls/src/identity.rs](https://github.com/xmtp/libxmtp/pull/2224/files#diff-5275a736595dacbf9d901365ab7dc26abd388f54986946de8e4ea539dc7cb16b)
- Adds `KEY_PACKAGE_ROTATION_INTERVAL_NS` constant set to 30 days and `KEY_PACKAGE_QUEUE_INTERVAL_NS` constant set to 5 seconds in configuration files [xmtp_mls/src/configuration.rs](https://github.com/xmtp/libxmtp/pull/2224/files#diff-80e1437765f38906f066439ad9805b99205fb55f992bceb400a9d091628b6942) and [xmtp_db/src/configuration.rs](https://github.com/xmtp/libxmtp/pull/2224/files#diff-7b56ec5258ef9c988a6390a3387f34819eb83f404965daea96d59fd7074fce8d)
- Modifies `StoredIdentity` struct to allow direct setting of `next_key_package_rotation_ns` field and renames `clear_key_package_rotation_queue` to `reset_key_package_rotation_queue` in [xmtp_db/src/encrypted_store/identity.rs](https://github.com/xmtp/libxmtp/pull/2224/files#diff-0eadc16eb390b4b83daba33d86321f37822a654963113c8fd5eff1f2cd4cbd36)
- Updates `queue_key_package_rotation` method to only update rotation time when the current value is greater than the new value and changes `is_identity_needs_rotation` to return true when `next_rotation_opt` is None

#### 📍Where to Start
Start with the `TryFrom<&Identity>` implementation for `StoredIdentity` in [xmtp_mls/src/identity.rs](https://github.com/xmtp/libxmtp/pull/2224/files#diff-5275a736595dacbf9d901365ab7dc26abd388f54986946de8e4ea539dc7cb16b) where the proactive rotation scheduling is initialized.

----

_[Macroscope](https://app.macroscope.com) summarized 100e5dc._


mchenani:
closes #2184

xmtp_db/src/encrypted_store/key_package_history.rs
```diff
@@ -90,7 +94,11 @@ impl<C: ConnectionExt> DbConnection<C> {
         use crate::schema::key_package_history::dsl;
         self.raw_query_read(|conn| {
             dsl::key_package_history
-                .filter(dsl::delete_at_ns.le(now_ns()))
+                .filter(
+                    dsl::delete_at_ns
+                        .le(now_ns())
+                        .or(dsl::delete_at_ns.is_null()),
```
mchenani, line 100: this will cover old keypackages where the `delete_at_ns` is none to be rotated.

xmtp_mls/src/identity.rs
```diff
@@ -667,7 +667,11 @@ impl Identity {
         let kp_bytes = kp.tls_serialize_detached()?;
         let hash_ref = serialize_key_package_hash_ref(&kp, &provider)?;
         let history_id = conn
-            .store_key_package_history_entry(hash_ref.clone(), pq_pub_key.clone())?
+            .store_key_package_history_entry(
+                hash_ref.clone(),
+                pq_pub_key.clone(),
+                kp.life_time().not_after() as i64,
```
mchenani, line 673: marking the KeyPackage to be rotated and deleted based on their life_time. in the future if we change the keypackage life_time this will automatically inherit the value from that.

xmtp_db/src/encrypted_store/identity.rs
```diff
@@ -80,7 +86,10 @@ impl<C: ConnectionExt> DbConnection<C> {
                 .first::<Option<i64>>(conn)
         })?;
 
-        Ok(matches!(next_rotation_opt, Some(rotate_at) if now_ns() >= rotate_at))
+        Ok(match next_rotation_opt {
+            Some(rotate_at) => now_ns() >= rotate_at,
+            None => true,
```
mchenani, line 89: this will cover the cases where we didn't have the lifetime as rotation time. cover old users.

xmtp_mls/src/identity.rs
```diff
@@ -677,7 +677,7 @@ impl Identity {
                     conn.mark_key_package_before_id_to_be_deleted(history_id)?;
                     Ok::<(), StorageError>(())
                 })?;
-                conn.clear_key_package_rotation_queue()?;
+                conn.reset_key_package_rotation_queue(kp.life_time().not_after() as i64)?;
```
mchenani: To ensure that we always set the default value as the KP lifetime, even if we change the lifetime in the future, we have it covered.

xmtp_db/src/encrypted_store/identity.rs
```diff
@@ -58,12 +57,15 @@ impl<C: ConnectionExt> DbConnection<C> {
...
+    pub fn reset_key_package_rotation_queue(
+        &self,
+        new_kp_valid_not_after: i64,
+    ) -> Result<(), StorageError> {
         use crate::schema::identity::dsl;
 
         self.raw_query_write(|conn| {
             diesel::update(dsl::identity)
-                .set(dsl::next_key_package_rotation_ns.eq::<Option<i64>>(None))
+                .set(dsl::next_key_package_rotation_ns.eq(Some(new_kp_valid_not_after * NS_IN_SEC)))
```
richardhuaaa: It's a bit confusing that the units are different, and the input also doesn't specify a unit. What if we just called the input `next_key_package_rotation_ns`, and pass in timstamps in ns for max transparency?

xmtp_mls/src/configuration.rs
```diff
@@ -24,6 +24,10 @@ pub const MAX_INTENT_PUBLISH_ATTEMPTS: usize = 3;
 
 pub const GROUP_KEY_ROTATION_INTERVAL_NS: i64 = NS_IN_30_DAYS;
 
+/// Only used to seed the initial `next_key_package_rotation_ns`.
+/// This does *not* affect the actual key-package lifetime.
+pub const DEFAULT_INITIAL_KEY_PACKAGE_NEXT_ROTATION_NS: i64 = 90 * NS_IN_DAY; // 90 days
```
richardhuaaa: KEY_PACKAGE_ROTATION_INTERVAL_NS?

And I'm wondering if we should rotate it earlier than the actual expiry of 90 days. You'd have to log on exactly at the 90 day mark to avoid having an expired key package.

Rotating it more frequently has the benefits of slightly more security + better chance of avoiding expiration, and the drawback of consuming more network resources. Given the limited drawback I'm wondering if we should rotate it like every 7 days, or 30 days?

xmtp_mls/src/identity.rs
richardhuaaa: Just double-checking:

1. Outside of this change, how do we handle the case when the key package upload fails, eg due to network error, or the app is restarted? Are we guaranteed to upload the key package the next time?

2. I'm a bit worried about not setting the rotation in all cases. What if we additionally set it at the same time that we insert the key package, i.e. just set the field when the key package is created?

mchenani: 1- in here https://github.com/xmtp/libxmtp/pull/2224/files#diff-0eadc16eb390b4b83daba33d86321f37822a654963113c8fd5eff1f2cd4cbd36R83 we check if `next_key_package_rotation_ns` is none or not valid we rotate ASAP the worker start working.

2- we only rotate and upload the key_package in this function, where other places we do that?

xmtp_db/src/encrypted_store/identity.rs
mchenani: the unit we get from the key_package is in seconds, but we store everything in NS cross libxmtp, didn't want to break that.

xmtp_mls/src/configuration.rs
mchenani: Totally agree, I would say `60 days`, cuz the kp is 90 days.

xmtp_mls/src/identity.rs
mchenani: what we can do in the future prop, set a trigger on the key_package table, to update this value based on that table. then we can ensure will keep track of all the keypackages.

richardhuaaa: > 1 - we check if next_key_package_rotation_ns is none or not valid we rotate ASAP the worker start working.

Ah, that makes sense. If the upload fails, we don't clear the rotation flag, so we rotate again next time the app is started.

> 2- we only rotate and upload the key_package in this function, where other places we do that?

You're right. I misremembered this as a field on the keypackage, so was wondering why we don't set it when we write the keypackage. But it's written on the identity.

xmtp_db/src/encrypted_store/identity.rs
richardhuaaa: Put another way - if you looked at just the function interface without looking at the implementation here, would you know what unit of timestamp to pass in for `new_kp_valid_not_after`, and what it is used for? Given this is just a simple setter method, for consistency I would suggest that the input argument is named after the property being set and use the same units, i.e. `next_key_package_rotation_ns`, and any conversion should be the caller's responsibility. Especially given we use ns everywhere else in libxmtp, this prevents mistakes.

It's only a small difference but I do believe consistency adds up. Of course the best solution would be to use a strongly typed timestamp type!

xmtp_mls/src/configuration.rs
richardhuaaa: Could we actually do 30 days or less? The shorter the rotation interval, the bigger the window that people have to rotate their keys. As a concrete example - with an interval of 60 days, they could be active within the 60 days, and then not active for 30 days, and their key package would expire.

xmtp_db/src/encrypted_store/identity.rs
mchenani: Thanks man, I thought you're asking about why one is in secs and why is NS, otherwise I totally agree and we're on the same page in this topic. 🙌

xmtp_mls/src/configuration.rs
mchenani: I agree, this is only the initial value since we can't use the KP lifetime itself, after first rotation we're gonna use the 90days from the KP Lifetime, I guess if we want to rotate every 30 days, we need to ignore the KP lifetime here too: https://github.com/xmtp/libxmtp/pull/2224/files#diff-0eadc16eb390b4b83daba33d86321f37822a654963113c8fd5eff1f2cd4cbd36R64

mchenani: @richardhuaaa udpated.

xmtp_db/src/encrypted_store/identity.rs
mchenani: @richardhuaaa updated.

xmtp_db/src/encrypted_store/identity.rs
```diff
@@ -44,26 +43,25 @@ impl<C: ConnectionExt> DbConnection<C> {
...
+    pub fn reset_key_package_rotation_queue(
+        &self,
+        rotation_interval_ns: i64,
+    ) -> Result<(), StorageError> {
         use crate::schema::identity::dsl;
 
         self.raw_query_write(|conn| {
             diesel::update(dsl::identity)
-                .set(dsl::next_key_package_rotation_ns.eq::<Option<i64>>(None))
+                .set(dsl::next_key_package_rotation_ns.eq(Some(now_ns() + rotation_interval_ns)))
```
richardhuaaa, line 70: We need to be careful not to overwrite a queued rotation set by `queue_key_package_rotation`, right? How about we add something like `.filter(dsl::next_key_package_rotation_ns.lt(new_rotation_time_ns)` or equivalent, here and in `queue_key_package_rotation`?

mchenani: Good catch, thanks, will do that.

Issue #2029, opened 2025-05-22
Disappearing Messages don't clean up messages after settings changed


nplasterer:
<@U055PVDSDSR> found a nice bug with disappearing messages.

if you change the disappearing message settings and some messages before didn't have time to run through a clean up because of the previous settings, those will be not deleted.

So:
1. Set the settings to like 10 secs
2. Send a message `A`
3. Set the settings to 1 sec
4. Send a message `B`
`B` will get deleted but not `A` even if it was sent while we had the 10 secs settings

[Slack Message](https://xmtp-labs.slack.com/archives/C07NSHXK693/p1747936169948679?thread_ts=1747936169.948679&cid=C07NSHXK693)

PR #2229, opened 2025-07-19
fix disappearing messages if it gets disabled

mchenani:
Issue: if a message sent during the disappearing messages is enabled, if the settings gets disabled, the message never gets deleted.

Why: in the current design, we look at the group settings and then delete the messages accordingly.

Solution: When we want to insert a message, store the disappearing settings with the message upon receiving it.


nplasterer (approved):

nplasterer:
Fixes https://github.com/xmtp/libxmtp/issues/2029

Issue #2102, opened 2025-06-12
Automated fork recovery work items


richardhuaaa:
Just an informal list for some lightweight tracking. Most low-risk approach is to finish and ship detection first, and then continue based on the result.

Design

- Protos/DB schemas - https://github.com/xmtp/proto/pull/276 https://github.com/xmtp/libxmtp/pull/2101 #2175 
- XIP writing and review - [draft published](https://community.xmtp.org/t/xip-68-draft-automated-fork-recovery/951)

Detection

- Storing updates in the local commit log - https://github.com/xmtp/libxmtp/pull/2106 #2179 #2180 
- Publishing updates to the remote commit log
  - Backend API support - https://github.com/xmtp/xmtp-node-go/pull/460
  - Client-side publishing - https://github.com/xmtp/libxmtp/pull/2183 https://github.com/xmtp/libxmtp/pull/2186
  - Verify groups are consented to and active before commit log download/upload - https://github.com/xmtp/libxmtp/pull/2303
- Cryptography
  - Setting secret on group creation - https://github.com/xmtp/libxmtp/pull/2231
  - Creating/retrieving key and signing on publish - https://github.com/xmtp/libxmtp/pull/2265
  - Bookkeeping consensus key and verifying on receive - https://github.com/xmtp/libxmtp/pull/2280
  - Updating mutable metadata when consensus key is determined - https://github.com/xmtp/libxmtp/pull/2302
  - (Puntable) Stitch a new DM group with consensus key on mutable metadata if none exists
  - Update XIP
- Parsing and storing updates from the remote commit log - https://github.com/xmtp/libxmtp/pull/2244
- Detecting mismatched state, with SDK hooks and telemetry - https://github.com/xmtp/libxmtp/pull/2272 https://github.com/xmtp/libxmtp/pull/2305
- Add handling + tests for multiple super_admins, changing super_admins
- Start worker behind dev flag - https://github.com/xmtp/libxmtp/pull/2312
- documentation updates - how can devs / users detect if a group is forked? (maybe we dont need docs until recovery is implemented)

Recovery

- Ability to send and receive oneshot messages - done
  - (Optional) Delete oneshot group after sending
- Sending readd request - done
- Receiving readd request, with bookkeeping - done
- Ability to perform readd (intent, commit, sending welcomes) - done
- Ability to receive readd commit, with bookkeeping - done
- Performing the readd, with bookkeeping - done
- Applying/validating readd welcomes, with bookkeeping - done
[...]

PR #2244, opened 2025-07-24
functions for saving the remote commit log

cameronvoell:
See tracking issue https://github.com/xmtp/libxmtp/issues/2102

### Add functions for saving the remote commit log by implementing separate upload and download operations for commit logs in the XMTP database and MLS groups
- Splits commit log operations into separate upload and download workflows by renaming `QueryGroup::get_conversation_ids_for_remote_log()` to `get_conversation_ids_for_remote_log_publish()` and adding `get_conversation_ids_for_remote_log_download()` in [xmtp_db/src/encrypted_store/group.rs](https://github.com/xmtp/libxmtp/pull/2244/files#diff-e3d4940853a1045db514d5c1a48581d9ef53da4ba45a7cb8387c72c99c475b5a)
- Replaces `EntityKind::PublishedCommitLog` with `EntityKind::CommitLogUpload` and adds `EntityKind::CommitLogDownload` to distinguish between upload and download cursor tracking in [xmtp_db/src/encrypted_store/refresh_state.rs](https://github.com/xmtp/libxmtp/pull/2244/files#diff-3eb320843309b9973b2ceaf8240120758bd1f2524b263d9a6b869b33266804ea)
- Implements `CommitLogWorker::save_remote_commit_log()` method to query, download, and store remote commit log entries locally with cursor management in [xmtp_mls/src/groups/commit_log.rs](https://github.com/xmtp/libxmtp/pull/2244/files#diff-c25dc4d36a6249375b1001e9c4b66e50034909d2fc05025a45476e115429dc3a)
- Adds `RemoteCommitLog` store functionality and `From<ProtoCommitResult>` conversion with MAX_PAGE_SIZE constant of 100 for pagination in [xmtp_db/src/encrypted_store/remote_commit_log.rs](https://github.com/xmtp/libxmtp/pull/2244/files#diff-cfcb5baf7f2bdfbb485a1f78af11355094f6dd1c3dc120f9bfe41312c6ee731a)
- Extends `ConnectionError` enum with `DecodeError` variant for `prost::DecodeError` handling in [xmtp_db/src/encrypted_store/mod.rs](https://github.com/xmtp/libxmtp/pull/2244/files#diff-7c11a42d3a8d1e571eea6a833dfe388063c036317c4d7a6fbce207fbb3beb387)

#### 📍Where to Start
Start with the `save_remote_commit_log()` method in [xmtp_mls/src/groups/commit_log.rs](https://github.com/xmtp/libxmtp/pull/2244/files#diff-c25dc4d36a6249375b1001e9c4b66e50034909d2fc05025a45476e115429dc3a) to understand the main workflow for downloading and saving remote commit logs.

#### Changes since #2244 opened

- Removed membership state filters from database queries in QueryGroup implementation [44c832a]
----

_[Macroscope](https://app.macroscope.com) summarized 44c832a._

### DIFF
--- a/xmtp_db/src/encrypted_store/group.rs
+++ b/xmtp_db/src/encrypted_store/group.rs
@@ -595,11 +595,13 @@ impl<C: ConnectionExt> DbConnection<C> {
         self.raw_query_read(|conn| query.load::<Vec<u8>>(conn))
     }
 
+    // All dms and groups that are note sync groups or rejected
     pub fn get_conversation_ids_for_remote_log_download(
         &self,
     ) -> Result<Vec<Vec<u8>>, crate::ConnectionError> {
         let query = dsl::groups
             .filter(dsl::conversation_type.ne(ConversationType::Sync))
+            .filter(dsl::membership_state.ne(GroupMembershipState::Rejected))
             .select(dsl::id);
 
         self.raw_query_read(|conn| query.load::<Vec<u8>>(conn))
--- a/xmtp_db/src/encrypted_store/group.rs
+++ b/xmtp_db/src/encrypted_store/group.rs
@@ -575,8 +575,9 @@ impl<C: ConnectionExt> DbConnection<C> {
         })
     }
 
-    /// Get conversation IDs for all conversations that require a remote commit log publish (DMs and groups where user is super admin, excluding sync groups)
-    pub fn get_conversation_ids_for_remote_log(
+    /// Get conversation IDs for all conversations that require a remote commit log publish
+    /// (DMs and groups where user is super admin, excluding sync groups and rejected groups)
+    pub fn get_conversation_ids_for_remote_log_publish(
         &self,
     ) -> Result<Vec<Vec<u8>>, crate::ConnectionError> {
         let query = dsl::groups
@@ -587,8 +588,19 @@ impl<C: ConnectionExt> DbConnection<C> {
                         .eq(ConversationType::Group)
                         .and(dsl::should_publish_commit_log.eq(true))),
             )
+            .filter(dsl::membership_state.ne(GroupMembershipState::Rejected))
             .select(dsl::id)
             .order(dsl::created_at_ns.asc());
+
+        self.raw_query_read(|conn| query.load::<Vec<u8>>(conn))
+    }
+
+    pub fn get_conversation_ids_for_remote_log_download(
+        &self,
+    ) -> Result<Vec<Vec<u8>>, crate::ConnectionError> {
+        let query = dsl::groups
+            .filter(dsl::conversation_type.ne(ConversationType::Sync))
+            .select(dsl::id);
 
         self.raw_query_read(|conn| query.load::<Vec<u8>>(conn))
     }
--- a/xmtp_db/src/encrypted_store/group.rs
+++ b/xmtp_db/src/encrypted_store/group.rs
@@ -673,7 +673,6 @@ impl<C: ConnectionExt> QueryGroup<C> for DbConnection<C> {
                         .eq(ConversationType::Group)
                         .and(dsl::should_publish_commit_log.eq(true))),
             )
-            .filter(dsl::membership_state.ne(GroupMembershipState::Rejected))
             .select(dsl::id)
             .order(dsl::created_at_ns.asc());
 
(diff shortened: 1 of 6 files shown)


cameronvoell:
* **#2244** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2244?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2244?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xmtp_db/src/encrypted_store/group.rs
```diff
@@ -587,11 +588,24 @@ impl<C: ConnectionExt> DbConnection<C> {
                         .eq(ConversationType::Group)
                         .and(dsl::should_publish_commit_log.eq(true))),
             )
+            .filter(dsl::membership_state.ne(GroupMembershipState::Rejected))
```
richardhuaaa: I'm not sure if this is consent state, but it just occurred to me that we probably only want to publish for groups we've allowed (not pending or rejected)

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -115,61 +153,120 @@ where
...
+                    Vec<i32>,
+                ) = conn
                     .get_group_logs_for_publishing(&conversation_id, published_commit_log_cursor)?
                     .iter()
-                    .map(PlaintextCommitLogEntry::from)
-                    .collect();
+                    .map(|log| (PlaintextCommitLogEntry::from(log), log.rowid))
+                    .unzip();
+
+                let max_rowid = rowids.into_iter().max().unwrap_or_else(|| {
```
richardhuaaa: I think we can just get this from the last row, seeing as the query is sorted by row ID. And if the query is empty I think we can skip publishing here?

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -184,37 +281,134 @@ where
...
+        let remote_log_cursors =
+            conn.get_remote_log_cursors(conversation_ids_for_remote_log_download.as_slice())?;
+        let query_log_requests: Vec<QueryCommitLogRequest> = remote_log_cursors
+            .iter()
+            .map(|(conversation_id, cursor)| QueryCommitLogRequest {
+                group_id: conversation_id.clone(),
+                paging_info: Some(PagingInfo {
+                    direction: SortDirection::Ascending as i32,
+                    id_cursor: *cursor as u64,
+                    limit: remote_commit_log::MAX_PAGE_SIZE,
```
richardhuaaa, line 311: Are we relying on the next iteration of the worker (5 mins) to page through the results if there is more than one page of entries? I'd worry that we end up paging too slowly, but I suppose it would be unrealistic for there to be more than a page of new entries being published every 5 mins?

xmtp_db/src/encrypted_store/group.rs
cameronvoell: good call updated to only publish to or download from remote log if it is consent state allowed - 4efdba52414855ad15590a039c6a3f7072fb25f7

cameronvoell: looks like groups table GroupMembershipState is not really used anywhere, the update function is never used. 

Removed for now here 44c832a7106eb81020c4c5790a26c2c3d0124259, think ill add the filtering for correct consent state in a follow up.

xmtp_mls/src/groups/commit_log.rs
cameronvoell: FIxed in latest update. publish flow had a few inefficiencies that I cleaned up. Most notably it was doing an api call per group, now it batches them all into a single call - 6d2e1af503d1f9c93814099c8e52248205505311

xmtp_mls/src/groups/commit_log.rs
cameronvoell: My thought is that this should be okay for now as long as it's clear in comments, added 13210522c213bfb9e975cfcdd043a51d6d8b009e. 

100 commits per group per 5 minutes seems pretty reasonable. Feels like it follows the design philosophy of commit log not causing harm. If one of your groups is in some state where it is publishing to commit log much more than 100 commits in 5 minutes, it's possible it's stuck in some kind of loop and maybe we dont want to exhaust resources with the client trying to keep up with the remote commit log. 

Another case to consider is if you come back from not using the app for multiple months. It could be the case that there are many more than 100 commits per group since you last sync remote commit log. I think we'd rather err on the side of the remote log needing some time to catch up in this case, rather than eagerly trying to get everything, so it does not become a new source of bad performance. Relying on the worker iterations seems like the simplest way to get this throttling behavior when starting a client after not using for a while. 

Interested in your thoughts though, if this might not be a good idea.

richardhuaaa: Sounds very reasonable and well thought through!

PR #2249, opened 2025-07-24
Adds conversation_type to MlsGroup struct in favor of reading group metadata state for conversation_type

cameronvoell:
See iOS repro test PR this was tested against for performance:

https://github.com/xmtp/xmtp-ios/pull/542

got this three times in a row on my repro before this libxmtp update:
```
syncAllConversations() ▸ 65.385 s for 1201 convos

=== TIMING STATISTICS ===
conversations.list() ▸ avg: 1.465s, max: 50.721s (43 calls)
messages() ▸ avg: 0.002s, max: 0.005s (43 calls)
findConversation() ▸ avg: 0.001s, max: 0.003s (43 calls)
updateConsent() ▸ avg: 0.000s, max: 0.001s (43 calls)
=== END ===
```

and then after conversations.list() never hit the high max again:
```
conversations.list() ▸ avg: 0.017s, max: 0.039s (1698 calls)
messages() ▸ avg: 0.001s, max: 0.013s (1698 calls)
findConversation() ▸ avg: 0.000s, max: 0.006s (1698 calls)
updateConsent() ▸ avg: 0.000s, max: 0.006s (1698 calls)
```

### Store conversation type as field in MlsGroup struct instead of retrieving from group state metadata
The `MlsGroup` struct now stores `conversation_type` as a direct field rather than retrieving it asynchronously from metadata. The `conversation_type()` method changes from an async method that queries metadata to a synchronous method that returns the stored field. All `MlsGroup` creation methods across the codebase are updated to accept and pass the `conversation_type` parameter. The `FfiConversation.conversation_type()` method in [bindings_ffi/src/mls.rs](https://github.com/xmtp/libxmtp/pull/2249/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f) becomes synchronous and directly returns the stored value. A new `get_conversation_type()` method is added to the `QueryGroup` trait in [xmtp_db/src/encrypted_store/group.rs](https://github.com/xmtp/libxmtp/pull/2249/files#diff-e3d4940853a1045db514d5c1a48581d9ef53da4ba45a7cb8387c72c99c475b5a) to retrieve conversation types directly from the database.

#### 📍Where to Start
Start with the `MlsGroup` struct definition and its `conversation_type()` method in [xmtp_mls/src/groups/mod.rs](https://github.com/xmtp/libxmtp/pull/2249/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3) to understand how the conversation type field is now stored and accessed.

#### Changes since #2249 opened

- Added conversation_type parameter to MlsGroup constructor call [0c65c4a]
- Changed conversation type parameter in MlsGroup initialization during message recovery [ee2d364]
----

_[Macroscope](https://app.macroscope.com) summarized ee2d364._

### DIFF
--- a/xmtp_mls/src/subscriptions/process_message/factory.rs
+++ b/xmtp_mls/src/subscriptions/process_message/factory.rs
@@ -11,6 +11,7 @@ use crate::subscriptions::process_message::MessageIdentifierBuilder;
 use crate::subscriptions::SubscribeError;
 use tracing::Instrument;
 use xmtp_common::{retry_async, Retry};
+use xmtp_db::group::ConversationType;
 use xmtp_db::prelude::*;
 use xmtp_db::{group_message::StoredGroupMessage, refresh_state::EntityKind, StorageError};
 use xmtp_proto::mls_v1::group_message;
@@ -107,6 +108,7 @@ where
             self.0.clone(),
             msg.group_id.clone(),
             None,
+            ConversationType::Sync,
             msg.created_ns as i64,
         );
         match group.sync_with_conn().await {
--- a/xmtp_mls/src/subscriptions/process_message/factory.rs
+++ b/xmtp_mls/src/subscriptions/process_message/factory.rs
@@ -108,7 +108,7 @@ where
             self.0.clone(),
             msg.group_id.clone(),
             None,
-            ConversationType::Sync,
+            ConversationType::Group,
             msg.created_ns as i64,
         );
         match group.sync_with_conn().await {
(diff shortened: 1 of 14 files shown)


cameronvoell:
* **#2249** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2249?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2249?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xmtp_mls/src/subscriptions/process_message/factory.rs
```diff
@@ -107,6 +108,7 @@ where
             self.0.clone(),
             msg.group_id.clone(),
             None,
+            ConversationType::Group,
```
cameronvoell, line 111: This is the one place the new conversation type needs to be added where we dont know its value. I think we're also populating dummy data for dm id here and the code path looks like conversation_type is not used, so I think this is safe for now, but seems like a code smell.

PR #2250, opened 2025-07-25
Create builder for intents 

insipx:
The builder makes it easier to write unit tests, as well as makes the code when working with bulk groups clearer. 

**Reasoning:**
Q: If we are to batch intents, how does the function fit into libxmtp

`queue_intents` is implemented on the group. having a method on group accept multiple other groups where the intent would be queued seems confusing (why does one group get "priority"?). + the ergonomics of randomly choosing a group to queue an intent on many other groups. The intent is the same for all groups, so it shouldn't rely on getting any singular group.

In that case, we could put the batch queue on Client. But, now we have separated logic/places for queuing intents, which again seems like it would be confusing and might lead to bugs. There's also many other functions implemented on client, which can make encapsulating functionality tricky

Making intents a first-class citizen improves ergonomics, and also lets us write better tests for the queueing logic itself, while keeping the implementation details private

### DIFF
--- a/xmtp_mls/src/groups/intents/queue.rs
+++ b/xmtp_mls/src/groups/intents/queue.rs
@@ -42,46 +42,52 @@ impl QueueIntentBuilder {
         let intent = self.build()?;
         intent.queue_with_conn(conn, group)
     }
-
-    /// Create an intent to send a message
-    pub fn send_message(&mut self) -> &mut Self {
-        self.kind = Some(IntentKind::SendMessage);
-        self
-    }
-
-    /// Create an intent to update the keys for a group
-    pub fn key_update(&mut self) -> &mut Self {
-        self.kind = Some(IntentKind::KeyUpdate);
-        self
-    }
-
-    /// Create an intent to update the metadata of a group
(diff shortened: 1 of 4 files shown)


insipx:
* **#2250** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2250?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2251](https://github.com/xmtp/libxmtp/pull/2251) <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2251?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2258](https://github.com/xmtp/libxmtp/pull/2258) <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2258?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2250?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2252, opened 2025-07-25
fix lifetime validation failure for other node in welcomes

### LINKED ISSUE #2228: Error: Lifetime is not acceptable
Added my own local agent and its not streaming messages for this specific group as well :thinking_face:, these are the logs when i added it to the group

[Slack Message](https://xmtp-labs.slack.com/archives/C095DFN7YBG/p1752865875577019)

mchenani:
Context:
https://github.com/openmls/openmls/issues/1810

closes #2228

### DIFF
--- a/xmtp_mls/src/utils/test/mod.rs
+++ b/xmtp_mls/src/utils/test/mod.rs
@@ -5,6 +5,7 @@ pub mod tester_utils;
 
 #[cfg(any(test, feature = "test-utils"))]
 pub mod fixtures;
+pub mod test_mocks_helpers;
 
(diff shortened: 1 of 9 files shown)


xmtp_mls/src/utils/test/mod.rs
```diff
@@ -295,93 +293,3 @@ pub async fn wait_for_min_intents<C: ConnectionExt>(
     })
     .await
 }
-
```
mchenani, line 298: moved this changes to make them easier to remove when mocking mechanism lands

PR #2262, opened 2025-07-28
Initial implementation of a cursor store

mkysel:


xmtp_cursor_state/src/store.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /// Get the current vector clock for this topic
+    pub fn get_latest(&self, topic: &[u8]) -> Option<&Cursor> {
+        self.topics.get(topic)
+    }
+
+    /// Compute the lowest common cursor across a set of topics.
+    /// For each node_id, uses the **minimum** sequence ID seen across all topics.
+    pub fn lowest_common_cursor(&self, topics: &[Vec<u8>]) -> Option<Cursor> {
```
mkysel, line 32: this one is worth discussing a bit more. It is obviously not efficient to re-load a bunch of messages that we have seen already, just because a new topic was added. Handling this correctly is a large project. Doing the LCC is guaranteed to be correct. We need to discuss what happens if the client receives duplicates. Fundamentally last_seen when used with many groups has a "at-least-once" semantic. Not sure what that means for the rest of the system.

xmtp_api_d14n/src/queries/combined.rs
mkysel, line 1: what is this file? seems like dead duplicated code to me. What does it do?

xmtp_api_d14n/src/protocol/extractors.rs
```diff
@@ -188,10 +192,14 @@ pub struct GroupMessageExtractor {
 }
 
 impl Extractor for GroupMessageExtractor {
-    type Output = mls_v1::GroupMessage;
+    type Output = (mls_v1::GroupMessage, u32, u64);
```
insipx, line 195: could make this a struct to avoid opaque tuple accesses

xmtp_cursor_state/src/store.rs
```diff
@@ -0,0 +1,227 @@
...
+mod tests {
+    use super::*;
+
+    fn cursor_with(kvs: &[(u32, u64)]) -> Cursor {
+        Cursor {
+            node_id_to_sequence_id: kvs.iter().cloned().collect(),
+        }
+    }
+
+    #[test]
```
insipx: use `#[xmtp_common::test]` from the `xmtp_common` crate so that this tests runs in webassembly as well as native

xmtp_cursor_state/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+edition = "2021"
```
insipx, line 2: can use 2024 edition

xmtp_cursor_state/src/store.rs
```diff
@@ -0,0 +1,227 @@
...
+
+impl CursorStore {
+    pub fn new() -> Self {
+        Self {
+            topics: HashMap::new(),
+        }
+    }
+
+    /// Record that a message for this topic with the given clock was processed
+    pub fn processed(&mut self, topic: Vec<u8>, new_clock: &Cursor) {
```
insipx: I think we should make a distinction here between 'processed' and 'received' -- processed is maybe overused within the mls side of things and confused me at first. This isn't decrypting anything w.r.t mls, but just taking out of the outer proto envelopes for d14n

xmtp_cursor_state/Cargo.toml
```diff
@@ -0,0 +1,10 @@
...
+name = "xmtp_cursor_state"
+rust-version = "1.70"
+version.workspace = true
+license.workspace = true
+
+[dependencies]
+hex = "0.4.3"
+xmtp_proto = { path = "../xmtp_proto", features = ["convert"] }
```
insipx, line 11: for persistence, you can create migrations with the diesel crate (in `xmtp_db`) for sqlite table, then store a reference to the `XmtpDb` or just `ConnectionExt` generic trait on the store. probably OK to do in a stacked PR to keep this one small, if you want to continue with this

xmtp_api_d14n/src/queries/combined.rs
insipx: at some point there was a migration plan that was reading/writing to both centralized and decentralized backend. nothing was set in stone yet, so rather than deleting any code we decided to just create a "combined" client according to the migration spec, for experimentation/testing purposes

I think that Borja is working on that now, i'd have to sync with him to figure out if this client still applies

The d14n clients are pretty easy to mix and match based on the endpoints and traits, that's why its a separate/duplicate file, easy to delete and or revert

xmtp_cursor_state/src/store.rs
insipx: If we receive a duplicate, won't we know its a duplicate based on what's in persistent storage?

insipx (commented):
Left a general review with comments/questions, since this is a concept some of it might not apply

overall like the direction for this

xmtp_cursor_state/src/store.rs
mkysel: for sure. We should be able to de-dup fairly easily. We just need to decide which layer of the code is responsible for that.

There is a scenario in which a publish to a node succeeds, but the connection gets torn somewhere along the way and the client gets an error. In some cases, the payer will hide the error and re-publish the same thing to another node.
This means that one "logical" payload can have multiple copies with different node/sequence IDs on the network. We need to be able to de-dup these as well. I suspect that might be more complicated.

richardhuaaa: We did discuss this last year, what do you think of [this](https://www.notion.so/ephemerahq/Efficient-and-reliable-batch-subscriptions-10a30823ce928060b6a0cf4e38ee7138?source=copy_link)?

mkysel: that is the last time we talked about it. All my knowledge is based on your insight. I am re-surfacing this conversation a year later!

richardhuaaa: Ah this is my bad, I totally misunderstood what was being discussed here! You brought up a great point about deduping the 'logical' payload. I wonder if it would make sense to dedupe against [this field](https://github.com/xmtp/proto/blob/637b892e32c5409a57f18c84291fc9007b4fb160/proto/xmtpv4/envelopes/envelopes.proto#L41), or a hash of it? That's assuming that the `depends_on` in the AAD won't be different between re-publishes. We could maybe structure it as a unique constraint in the sqlite database?

mkysel: the entire `ClientEnvelope` will be the "same". The wrapping PayerEnvelope and OriginatorEnvelope will be different.

Not sure whether it is guaranteed that the serialized bytes will be exactly the same. I would not want to depend on it. The client might decide to republish something via a different payer.

richardhuaaa: It's a fair point that the serialization might be different. Also, we would need to ensure there is some kind of timestamp or nonce (if there isn't already one), so that just sending the same message twice doesn't get deduplicated.

A solution could be for the client to add a non-colliding deduplication key to the AAD, derived from a sequential ID/timestamp/random value, and dedupe based on sender+this value

mkysel: I like that!

insipx:
closing this since #2482 merged with these changes

PR #2263, opened 2025-07-28
Make Sync GRPC Channel

nplasterer:
SyncAll can take up a lot of GRPC threads and cause large queue delays for other functions like send message. Lets move sync to it's own GRPC channel so that it doesn't block the main channel actions.
Confirmed SyncAll improved from 60s to .6s for 200 groups and dms.


insipx (approved):
looks great!

PR #2230, opened 2025-07-19
Protos for commit log signing

richardhuaaa:
This is the accompanying PR for https://github.com/xmtp/proto/pull/284, including minimal code changes to ensure Rust compiles.

PR #2265, opened 2025-07-29
Generate or fetch signing key when publishing to commit log

richardhuaaa:
Implements commit log signing, including generating or fetching a consensus signing key. For newly created groups, the consensus signing key is effectively determined by the group creator, but we also need something that can extend backwards to older groups.

The consensus works roughly as follows:

Rule 1: The first entry in the commit log (the winner of the 'race') determines the consensus signing key to use for the rest of the log. We can store this result on the group row when verifying the log entries (next PR in stack).
Rule 2: Until there is a first entry in the commit log, any key may be used to sign commit log entries.
Rule 3: Once a key has been confirmed as the first entry in the commit log, the private key may be shared with the group via mutable metadata. Until it is shared, only the creator of the key is able to publish to the log. The mutable metadata update can be implemented later in the stack, although could also be skipped.

Some optimizations:
- The creator can pre-write a key to the mutable metadata when creating a group (lower PR in the stack), even though the first entry in the commit log has not been published yet. In general, when rule 2 is being followed, all installations will try to use any existing key on the mutable metadata, if there is one.
- Any time a key is used to sign a remote commit log entry, we store it in the key store so that we can find it later without needing to check the mutable metadata, which requires getting a lock on the OpenMLS group.

Some missing things:
- Unit tests are implemented higher in the stack
- There are race conditions here to think through - will address these as follow-ups:
  - We should never end up in a scenario where we think we need to generate a new signing key, and we overwrite an existing copy of the consensus signing key.
  - If two installations race to be the first to publish to the remote commit log, one of the installations' updates will be discarded. However, their publish cursor will have been incremented. We'll need to fix this eventually, perhaps by coupling the publish cursor with the parsing cursor.
  
Feedback on races, or ideas for simplification welcome!


richardhuaaa:
* **#2318** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2318?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2312** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2305** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2305?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2302** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2291** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2280** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2280?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2265** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2265?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2231** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2231?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2230** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2230?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

cameronvoell (approved):

PR #2272, opened 2025-07-30
fork detection and tests

cameronvoell:


cameronvoell:
* **#2272** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2272?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2272?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,28 +316,41 @@ where
...
-                applied_epoch_number: Some(log_entry.applied_epoch_number as i64),
-                applied_epoch_authenticator: Some(log_entry.applied_epoch_authenticator),
+                applied_epoch_number: log_entry.applied_epoch_number as i64,
+                applied_epoch_authenticator: log_entry.applied_epoch_authenticator,
             }
             .store(conn)?;
-            if entry.sequence_id > latest_download_cursor {
-                latest_download_cursor = entry.sequence_id;
+            if commit_log_entry.sequence_id > latest_download_cursor {
+                latest_download_cursor = commit_log_entry.sequence_id;
```
richardhuaaa: If the last entries are skipped, we should probably update the cursor regardless so that we don't re-process them. Maybe it makes sense to derive this from the last entry, and also to not update the cursor on L356 if the list is empty?

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -375,15 +360,153 @@ where
...
-    PublishCommitLogsToRemote,
-    SaveRemoteCommitLog,
-    All,
+    fn should_skip_remote_commit_log_entry(
+        validation_info: &RemoteLogValidationInfo,
+        entry: &PlaintextCommitLogEntry,
+    ) -> bool {
+        let is_applied = entry.commit_result == ProtoCommitResult::Applied as i32;
+        // Should skip if:
+        // 1. The entry signature is invalid - TODO(cam)
```
richardhuaaa: I have a PR for this, can rebase it on top of your changes

xmtp_mls/src/groups/commit_log.rs
cameronvoell: updated [74db657cedfd0b5ee07820b647102aad61051afa](https://github.com/xmtp/libxmtp/pull/2272/files#diff-c25dc4d36a6249375b1001e9c4b66e50034909d2fc05025a45476e115429dc3aR353-R361) 👌

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
+            let commit_log_entry: &CommitLogEntry = entry;
+            let log_entry = PlaintextCommitLogEntry::decode(
+                commit_log_entry.serialized_commit_log_entry.as_slice(),
+            )?;
+
+            // From the stored remote commit log, fetch the following info:
+            // 1. The latest applied epoch authenticator
+            // 2. The latest applied epoch number
+            // 3. The latest stored sequence id
+            let validation_info = conn.get_remote_log_validation_info(&group_id)?;
```
richardhuaaa: We are doing two DB calls on each iteration here - would we be able to just use the previous row, seeing as the `applied_epoch_authenticator` and `applied_epoch_number` fields are non-optional? Or would there be edge cases/inconsistencies with those fields? Just wondering if it would remove a few structs/traits and simplify the code a little

richardhuaaa: I suppose the thing I missed in the validation logic is verifying that the `applied_epoch_number` and `applied_epoch_authenticator` for failed commits matches the previous `applied_epoch_number` and `applied_epoch_authenticator`, but maybe that's easy to add? And I guess the verbiage in the XIP made all of this non-obvious, which is my bad!

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
+                applied_epoch_number: log_entry.applied_epoch_number as i64,
+                applied_epoch_authenticator: log_entry.applied_epoch_authenticator,
             }
             .store(conn)?;
-            if entry.sequence_id > latest_download_cursor {
-                latest_download_cursor = entry.sequence_id;
-            }
-            if log_entry.commit_sequence_id > latest_sequence_id {
-                latest_sequence_id = log_entry.commit_sequence_id;
+            if log_entry.commit_sequence_id as i64 > latest_commit_sequence_id {
```
richardhuaaa: This should always be the case based on the validation rules right?

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
+            )?;
+        }
+
         Ok(UpdateCursorsResult {
             conversation_id: group_id,
-            last_entry_saved_commit_sequence_id: latest_sequence_id as i64,
-            last_entry_saved_remote_log_sequence_id: latest_download_cursor as i64,
+            num_entries_saved,
+            last_entry_saved_commit_sequence_id: latest_commit_sequence_id,
+            last_entry_saved_remote_log_sequence_id: latest_download_cursor,
```
richardhuaaa: Are these two fields used anywhere? I'm wondering if we can get away with only returning `num_entries_saved` and nothing else, but I definitely could have misunderstood

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
             }
         }
-        conn.update_cursor(
-            &group_id,
-            xmtp_db::refresh_state::EntityKind::CommitLogDownload,
-            latest_download_cursor as i64,
-        )?;
+        if let Some(last_entry) = commit_log_response.commit_log_entries.last() {
+            latest_download_cursor = last_entry.sequence_id as i64;
+            conn.update_cursor(
```
richardhuaaa, line 411: Reasoning through whether or not we need a transaction. If a failure occurs halfway through this function, the worst case is we re-download some log entries, which we throw away when we validate them, which should be fine right?

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
+    ) -> Result<ForkedStateCheckResult, CommitLogError> {
+        // Get cursors for this conversation
+        let fork_check_local_cursor = conn.get_last_cursor_for_id(
+            conversation_id,
+            xmtp_db::refresh_state::EntityKind::CommitLogForkCheckLocal,
+        )?;
+        let fork_check_remote_cursor = conn.get_last_cursor_for_id(
+            conversation_id,
+            xmtp_db::refresh_state::EntityKind::CommitLogForkCheckRemote,
+        )?;
```
richardhuaaa, line 492: Not important for this PR, just something I thought about with the consensus key stuff too. Would it be fair to say that we expect no concurrency with any of this code? Wondering if it makes sense to prove it by acquiring a lock on each iteration of the worker (which self-documents the concurrency model and protects against any unanticipated scenarios, like the previous iteration is still running when the next iteration starts, or there are two workers or something)

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
+            if let Some(matching_remote_log) =
+                self.find_matching_remote_log(&remote_logs, local_log.commit_sequence_id)
+            {
+                // Found a matching commit_sequence_id - check if forked
+                let is_forked = local_log.applied_epoch_authenticator
+                    != matching_remote_log.applied_epoch_authenticator;
+
+                if is_forked {
+                    println!("Forked state check result: {:?}", matching_remote_log);
+                    println!("Local log: {:?}", local_log);
```
richardhuaaa: Should these be `tracing` calls rather than prints?

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
+                    forked_epoch_number: if is_forked {
+                        Some(matching_remote_log.applied_epoch_number as u64)
+                    } else {
+                        None
+                    },
+                    forked_commit_sequence_id: if is_forked {
+                        Some(matching_remote_log.commit_sequence_id as u64)
+                    } else {
+                        None
+                    },
```
richardhuaaa: I'm wondering if these fields are meaningful - we are looking at the latest commit sequence ID in common, so it's possible that it could have forked earlier than this right?

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
+                });
+            }
+        }
+
+        // No matching commit_sequence_id found for any local log = no fork detected
+        Ok(ForkedStateCheckResult {
+            is_forked: false,
+            forked_epoch_number: None,
+            forked_commit_sequence_id: None,
+        })
```
richardhuaaa: You're probably already planning to do this, but once we persist the fork state, once a group is forked we probably shouldn't run this fn again, otherwise we might accidentally return `is_forked: false` here if there are no new entries

richardhuaaa (commented):
It looks robust and well-thought through :+1:, just being extra careful with the review seeing as this is the most critical part of the feature. Mainly have some small simplifications and paranoia about edge cases on my end. Can give a speedy review whenever ready!

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,248 @@ where
...
+    }
+
+    #[cfg(test)]
+    pub(crate) fn should_skip_remote_commit_log_entry_test(
+        &self,
+        validation_info: &RemoteLogValidationInfo,
+        entry: &PlaintextCommitLogEntry,
+    ) -> bool {
+        self.should_skip_remote_commit_log_entry(validation_info, entry)
+    }
```
richardhuaaa: Is this method needed?

xmtp_mls/src/groups/commit_log.rs
cameronvoell: Updated to only do one db call per group and otherwise use the previous entries values for applied epoch auth and applied epoch number

Also added the extra check and unit tests for verifying non applied commits epoch and auth match the latest applied 79d2e07dfc645d8125b11d995cc9d15da6379da8

xmtp_mls/src/groups/commit_log.rs
```diff
@@ -344,46 +316,263 @@ where
...
+                    } else {
+                        None
+                    },
+                    forked_commit_sequence_id: if is_forked {
+                        Some(matching_remote_log.commit_sequence_id as u64)
+                    } else {
+                        None
+                    },
+                });
+            }
```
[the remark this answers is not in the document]
cameronvoell: It's possible that your local commit log is ahead of the remote log, so it may find an entry that matches your latest local log on a subsequent iteration of the worker

xmtp_mls/src/groups/commit_log.rs
cameronvoell: removed 07a28084188ecc7ac7c0f5f5614227211476b4e4

xmtp_mls/src/groups/commit_log.rs
cameronvoell: removed 07a28084188ecc7ac7c0f5f5614227211476b4e4

xmtp_mls/src/groups/commit_log.rs
cameronvoell: Yep good to reason through, but seems validation we weed out any re-downloaded entries in the case where a failure stopped the cursor update which I think should be fine.

xmtp_mls/src/groups/commit_log.rs
cameronvoell: ill make a note to revisit this for all of our workers before we turn on auto fork recovery 👍

xmtp_mls/src/groups/commit_log.rs
cameronvoell: removed 97664f22ff8903ec107254fd2ba1eac7db1e50a9

xmtp_mls/src/groups/commit_log.rs
cameronvoell: updated 97664f22ff8903ec107254fd2ba1eac7db1e50a9

xmtp_mls/src/groups/commit_log.rs
cameronvoell: added a reminder - https://github.com/xmtp/libxmtp/issues/2102

xmtp_mls/src/groups/commit_log.rs
richardhuaaa: Makes sense - it's to expose a non-visible function to tests

richardhuaaa (approved):
Looks great 👏

PR #2276, opened 2025-07-31
stitch dms grouped based on ordered group ids

### LINKED ISSUE #2267: Bug: DM stitching can break across installations
### Describe the bug

when creating a DM group between inboxes that results in 2 separate groups, future installations (of either inbox) may not stitch the DM group correctly.

### Expected behavior

the DM group should be stitched across all installations

to see how this surfaces, checkout the `rygine/dm-stitching` branch in the `xmtp/xmtp-js` repo. at the repo root, run the following:

```bash
yarn
cd sdks/node-sdk
yarn test test/stitching.test.ts
```

the first test should pass, the second one will not. the only difference is the order that the clients are synced using `syncAll`.

the tests will log the relevant values.

mchenani:
In the current implementation, DM stitching groups all conversations and sorts them by the timestamp of the last message, which can yield different “stitched” groups for different users. This change instead orders grouped DMs by their group ID—since that ID is the same for all participants, everyone will see the same stitched DM.
closes #2267

### DIFF
--- a/xmtp_db/src/encrypted_store/conversation_list.rs
+++ b/xmtp_db/src/encrypted_store/conversation_list.rs
@@ -95,7 +95,7 @@ impl<C: ConnectionExt> QueryConversationList<C> for DbConnection<C> {
                 "id IN (
                     SELECT id FROM (
                         SELECT id,
-                            ROW_NUMBER() OVER (PARTITION BY COALESCE(dm_id, id) ORDER BY last_message_ns DESC) AS row_num
+                            ROW_NUMBER() OVER (PARTITION BY dm_id ORDER BY id DESC) AS row_num
                         FROM groups
                     ) AS ranked_groups
                     WHERE row_num = 1
--- a/xmtp_db/src/test_utils.rs
+++ b/xmtp_db/src/test_utils.rs
@@ -16,6 +16,9 @@ pub trait XmtpTestDb {
     async fn create_persistent_store(
         path: Option<String>,
     ) -> EncryptedMessageStore<crate::DefaultDatabase>;
+    /// Create an empty unencrypted database
+    /// does no validation and does not run migrations.
+    async fn create_unencrypted_persistent_store(path: Option<String>) -> crate::DefaultStore;
     /// Create an empty database
     /// does no validation and does not run migrations.
     async fn create_database(path: Option<String>) -> crate::DefaultDatabase;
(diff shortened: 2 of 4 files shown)


xmtp_db/src/encrypted_store/conversation_list.rs
```diff
@@ -95,7 +95,7 @@ impl<C: ConnectionExt> QueryConversationList<C> for DbConnection<C> {
                 "id IN (
                     SELECT id FROM (
                         SELECT id,
-                            ROW_NUMBER() OVER (PARTITION BY COALESCE(dm_id, id) ORDER BY last_message_ns DESC) AS row_num
```
mchenani, line 110: last_message_ns is not a reliable variable here, order based on the group_id make it consistent for everyone to see the same stitched dm.

richardhuaaa:
Could you go over a bit more on how this fixes the issue - why is it important that the group ID is agreed upon by all participants? I would assume that even if this were not the case, participants can send a message on any of the stitched groups, and all other participants should receive it and stitch it into the same DM?

mchenani:
> In the current implementation, DM stitching groups all conversations and sorts them by the timestamp of the last message, which can yield different “stitched” groups for different users. This change instead orders grouped DMs by their group ID—since that ID is the same for all participants, everyone will see the same stitched DM.

This was my question too, maybe @rygine can answer this since he reported/found the issue via his test.

mchenani:
Although this PR doesn’t contain the original fix (that’s in https://github.com/xmtp/libxmtp/pull/2277), it’d still be a nice addition: with this change, all participants would communicate through the same group/group_id instead of different groups!
@richardhuaaa wdyt?

xmtp_db/src/encrypted_store/conversation_list.rs
richardhuaaa: My only hesitation comes from the fact that the most recently used group is the one most likely to have the correct set of members. And if the goal is to keep the group ID consistent, neither way is consistent, because the very fact that somebody sent a message on this most recently used group (with the greatest `last_message_ns`) means that at one point they believed this was the correct group ID to use (most likely because: they don't have access to the other groups). So even with this PR, the consensus group ID from that participant's POV has to evolve, so in both approaches, we start without agreement on the group ID and converge to agreement eventually.

To be fair - if everything worked perfectly, I think both ways would produce the correct result. I have a slight preference, but if you're confident we wouldn't be adding new regressions with this change,  I'm not strongly opposed!

richardhuaaa:
My only hesitation comes from the fact that the most recently used group is the one most likely to have the correct set of members. And if the goal is to keep the group ID consistent, neither way is consistent, because the very fact that somebody sent a message on this most recently used group (with the greatest `last_message_ns`) means that at one point they believed this was the correct group ID to use (most likely because: they don't have access to the other groups). So even with this PR, the consensus group ID from that participant's POV has to evolve, so in both approaches, we start without agreement on the group ID and converge to agreement eventually.

To be fair - if everything worked perfectly, I think both ways would produce the correct result. I have a slight preference, but if you're confident we wouldn't be adding new regressions with this change,  I'm not strongly opposed!

richardhuaaa:
Thinking about this more - always selecting the most recently used DM, rather than the one with the highest `id`, also leaves open the option of initiating fork recovery by creating a new DM, rather than trying to fix an existing DM, cc @cameronvoell.

Unsure if we will go down that route, but zooming out, regardless of the type of problem, it might be nice to always have the ability to 'abandon' older, broken DM's in the event of issues (including, for example, if we're simply not a member of the older DM).

PR #2278, opened 2025-07-31
Move event sends outside of transactions

nplasterer:
Defer sending of local event/device sync messages until transactions finish.


insipx:
Cherry-picked into #2286

PR #2279, opened 2025-07-31
Add a cursor to the messages stream

nplasterer:
Looks up the sequence ID from the database before streaming messages. Falls back to network.


nplasterer:
@insipx might need your help fixing these mock tests. 🤔 But this did fix my streaming syncing in parallel performance issue.

bindings_ffi/src/mls.rs
```diff
@@ -6937,10 +6938,18 @@ mod tests {
         stream.wait_for_ready().await;
 
         alix_group.send("first".as_bytes().to_vec()).await.unwrap();
+        bo.conversations()
+            .sync_all_conversations(None)
+            .await
+            .unwrap(); // ✅ Force sync
```
insipx: what does force syncing do here

xmtp_mls/src/subscriptions/stream_messages/test_case_builder.rs
```diff
@@ -298,6 +299,20 @@ impl StreamSequenceBuilder {
...
 
     fn init_session(&mut self, groups: Vec<GroupTestCase>) {
+        // Set up db mock expectation
+        let db_calls = || {
+            let mut mock_db = xmtp_db::mock::MockDbQuery::new();
+            // Set up expectations for commonly called db methods
+            mock_db
+                .expect_get_latest_sequence_id_for_group()
+                .returning(|_| Ok(None));
+            mock_db
```
insipx, line 309: Do the tests fail if we don't set this expectation? I can't find anywhere we're calling just the latest sequence id

insipx (approved):

bindings_ffi/src/mls.rs
nplasterer: @insipx the tests were failing because it refetches the messages on stream and doesn't move the cursor unless we sync.

xmtp_mls/src/subscriptions/stream_messages/test_case_builder.rs
nplasterer: Let me try to remove it

PR #2306, opened 2025-08-08
fix: conflicting ReactionV2 import in module

hexcow:
Rust doesn't allow importing the same name from different places in the same module, even if the paths are identical.
the second `ReactionV2` import was causing a conflict and has been removed.


insipx:
The second import does not cause a conflict gg

PR #2370, opened 2025-08-22
Do not log application message errors on commit log

richardhuaaa:
@cameronvoell performed some thorough validation that revealed that application message intents are throwing an `IntentAlreadyProcessed` error when read back from the network, and these errors are then being logged in the local commit log.

This fix is part 1, which prevents us from logging application message errors to the local commit log. Unit test fails before this change and passes after.

~I'll fix the root cause of why messages produce an `IntentAlreadyProcessed` error in the next PR.~ I'm not able to reproduce the `IntentAlreadyProcessed` via unit tests. I think these errors were thrown during validation on an emulator because multiple threads were processing the same payload, so the first thread would succeed and the second thread failed. If that theory is correct, there's nothing to fix - just trying to work out how to prove that this is what's happening.

### DIFF
--- a/xmtp_mls/src/groups/mls_sync.rs
+++ b/xmtp_mls/src/groups/mls_sync.rs
@@ -1415,8 +1415,9 @@ where
                     intent.kind
                 );
 
-                let message = message.into();
-                let maybe_validated_commit = self
+                let message: ProtocolMessage = message.into();
+                let message_type = message.content_type();
+                let validation_result = self
                     .stage_and_validate_intent(mls_group, &intent, &message, envelope)
                     .await;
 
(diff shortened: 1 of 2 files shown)


richardhuaaa:
* **#2370** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2370?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2370?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -1451,10 +1452,10 @@ where
                         identifier.previously_processed(true);
                         return Ok(());
                     }
-                    let result: Result<Option<Vec<u8>>, IntentResolutionError> = match maybe_validated_commit {
+                    let result: Result<Option<Vec<u8>>, IntentResolutionError> = match validation_result {
                         Err(err) => Err(err),
-                        Ok(commit) => {
-                            self.process_own_message(mls_group, commit, &intent, &message, envelope, &storage)
+                        Ok(validated_intent) => {
```
cameronvoell, line 1457: good rename 👍

cameronvoell (approved):
LGTM!

PR #2401, opened 2025-08-27
Route subscribe errors from stream setup to onClose callback

cameronvoell:


cameronvoell:
Aiming to resolve https://github.com/xmtp/libxmtp/issues/2405. May wait to hear confirmation if xmtp-android development build `4.5.0-dev.8f99e29` seems to address the issue before merging here.

PR #2485, opened 2025-09-20
Update kotlin bindings release action for 16kb page size requirement

cameronvoell:


cameronvoell:
looks like this was already done via Cargo config - https://github.com/xmtp/libxmtp/pull/2009/files

I verified that that we can run the library on a 16kb emulator and that our generated .so files are 16kb aligned via the check_elf_alignment.sh script from google here https://developer.android.com/guide/practices/page-sizes#elf-alignment

Issue #2354, opened 2025-08-20
Bug: message stream fails to process more than 1000 messages from same inbox


rygine:
### Describe the bug

the 1000th message from the same inbox fails to process on a message stream with the following message:

```
xmtp_mls::subscriptions::process_message::factory: synced 1000 messages, 1 failed 999 succeeded from cursor Some(4142050)
	┝━> [4143049] openmls process message error: Generation is too far in the future to be processed.
```

### Expected behavior

all messages should be processed

### Steps to reproduce the bug

first, ensure that a local node is running.

checkout my branch `rygine/stream-fixes` and run the new CLI tools as follows:

```bash
# checkout branch
git checkout rygine/stream-fixes
# build CLI tools
cd examples/cli
./build-release.sh
```

from the repo root, run the `stream-monitor` CLI tool:

```bash
target/releases/stream-monitor --max-messages 1002
```

check the terminal for the inbox ID. you'll see output like this:

```bash
Starting XMTP Stream Monitor
Generated wallet address: 0x2a41BE7c8E44634941bAEE4a8eF705210ee50AeF
Creating XMTP client...
Using ephemeral storage
2025-08-19T19:41:45.935324Z ERROR xmtp_mls::worker: KeyPackageCleaner worker error: Metadata(Connection(Database(NotFound)))
Client created and registered successfully!
Inbox ID: 2c0d33b57f90c60790a66e6917bf311cdc3ef9e0cee053085e4181370962646b
```

copy the inbox ID and open a new terminal so that we can send messages to this inbox. in the new terminal, run the `message-sender` tool:

```bash
target/releases/message-sender 2c0d33b57f90c60790a66e6917bf311cdc3ef9e0cee053085e4181370962646b --threads 1 --messages 1002
```

this will send 1002 messages from a single client.

now, notice that the `stream-monitor` is unable to process all messages.

to see the issue logged in the terminal, enable logging in the `stream-monitor` with the `--enable-logging` option as follows:

```bash
target/releases/stream-monitor --max-messages 1002 --enable-logging
```

### More context from Cryspen

`1000` is the default limit for future epochs.

> My hunch is that this is a storage issue. If I recall correctly, you store the group in between every operation. And if this secret here is not persisted correctly, you don’t increment the generation of the secret, while trying to generate a later generation every time you encrypt a message.

richardhuaaa:
This is possibly fixed by https://github.com/openmls/openmls/pull/1846:

[...]

PR #2498, opened 2025-09-23
Pull in fixes from OpenMLS

richardhuaaa:
Updates openMLS dep to include the latest upstream changes. Importantly, includes https://github.com/openmls/openmls/pull/1846, which possibly fixes https://github.com/xmtp/libxmtp/issues/2354

> The gist is that we didn’t persist the private tree correctly, which means that we may keep old secret trees around on clients (depending on sending behaviour), which violates the RFC and of course breaks forward secrecy within the epoch.
In really bad cases this may actually also break functionality (decryption), if a client is passive and the epochs are long (because the forward ratchet limit may be reached).

Changes to libxmtp:
- We were previously pointing directly at https://github.com/xmtp/openmls/pull/43, rather than main. The changes previously hardcoded into that PR have been merged as a config flag in upstream, so that PR no longer needs to merge into main. I believe I've set the corresponding libxmtp flags via the `StagedWelcome` builder in `decrypted_welcome.rs`, but it's worth double-checking @neekolas @mchenani.
   - What I checked: `test_sync_welcomes_when_kp_life_time_ended` fails without my change, fails with my change but without `.skip_lifetime_validation()`, and passes when `.skip_lifetime_validation()` is set.
- The OpenMLS `ProcessMessageError` now takes in a generic, so also updated libxmtp code to reflect this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3072,57 +3072,19 @@ dependencies = [
 
 [[package]]
 name = "hpke-rs"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e11bd4ee27b79fa1820e72ef8489cc729c87299ec3f7f52b8fc8dcb87cb2d485"
-dependencies = [
- "hpke-rs-crypto 0.2.0",
- "log",
- "serde",
- "tls_codec",
- "zeroize",
-]
-
-[[package]]
-name = "hpke-rs"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "36874953dfe0223fd877a77b0eefcd84f8da36161b446c6fcb47b8311fa0251a"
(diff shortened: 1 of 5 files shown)


richardhuaaa:
* **#2498** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2498?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2498?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=2498).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #2500, opened 2025-09-24
Failing lifetime validation unit test

richardhuaaa:
This is failing on the last line of the test, even before the openmls changes were pulled in. Feel free to pull this branch and verify. Is there something wrong with the test?


richardhuaaa:
Moving into a separate PR on latest main that includes the fix

PR #2504, opened 2025-09-25
Bump MLS version and stop pushing to infra

mkysel:
Our infra deploy stack is something like 26 elements deep. Most commits are unrelated to the MLS validation service so triggering a new redeploy (which takes 10 minutes) in 3 different environments is not the best use of our time.

I released 0.1.4 from main today and will pin that version in all environments. If we ever introduce a breaking change, we can do it manually or use the auto-generated from-main SHA versions.

### DIFF
--- a/.github/workflows/deploy-validation-server.yml
+++ b/.github/workflows/deploy-validation-server.yml
@@ -42,24 +42,3 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-
-  deploy:
-    name: Deploy new images to infra
-    runs-on: ubuntu-latest
-    needs: push_to_registry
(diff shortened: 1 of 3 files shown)


neekolas:
I think this is fine...but we need to circulate this information widely and make sure the protocol team is aware and regularly pushes up updates for protocol changes.

If the MLS validation service in dev/prod is running out of date code, you can run into nasty bugs that only emerge post-release.

A compromise would be to only deploy automatically on tagged releases without a `dev` or `rc` suffix. Then it only happens every few weeks.

mkysel:
Sure. The issue is also that in decentralization, our partners need to know that they need to upgrade. They wont be running main!

mkysel:
@insipx @neekolas let me know if the new version is GTG!

insipx (approved):

PR #2505, opened 2025-09-25
Adds action for auto generating xmtp-android branch and PR for updating libxmtp version

cameronvoell:


.github/workflows/update-mobile-repositories.yml
```diff
@@ -0,0 +1,358 @@
...
+    steps:
+      - name: Checkout libxmtp
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.tag || github.ref }}
+
+      - name: Get version and SHA
+        id: version_info
+        run: |
+          # Get the original version from cargo metadata
```
neekolas, line 37: This is long enough it probably deserves to go into a shell script so that it gets shellcheck and proper syntax highlighting.

neekolas:
This is going to be so good when it all works

PR #2544, opened 2025-10-02
D14N: Multi node client

fbac:
### Introduce multi-node client middleware and route API requests through `xmtp_api_d14n::middleware::MultiNodeClient` with a 100ms node selection timeout to support D14N
Add a `middleware` module in `xmtp_api_d14n` that includes a `MultiNodeClient` implementing `Client` and selecting a healthy node discovered from the gateway under a configurable timeout, expose the middleware publicly, and update proto and client builder types to support node discovery and health checks. The proto changes add `HealthCheck` and `GetNodes` endpoint identifiers and change `GetNodesResponse` to a `map<uint32, string>`, while `ClientBuilder` becomes `Clone` and `HealthCheck` gains `Clone`. Cargo manifests and lockfiles are updated for `http` and `url`, and Docker compose image tags are advanced.

- Implement a multi-node client and builder in [multi_node_client.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-c6b13327fb82bc847cf6f3c8a0f113b7b1ae6b1e19afc3c2605252e4926b2552) that discovers nodes via `GetNodes`, validates TLS against URL scheme, performs concurrent health checks, and proxies requests to the selected node
- Expose middleware via `xmtp_api_d14n::middleware` in [lib.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-681fccc9f3c92bb252163b40c3e36f4eb3042f3d2fb13c944510e7c1ea5f62ba) and [mod.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-d9f2ff298157c6ba3db30d694d5c683bfeaf2a278cee63c7cbe01f8f57338e53)
- Update proto to add `HealthCheck` and `GetNodes` endpoint variants in [error.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-1ba145e9996e1f4242914a55ad212ee52c712213b47f48c63ee49fbd23089c62) and change `GetNodesResponse` to a node ID to URL map in [xmtp.xmtpv4.payer_api.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-96f98b1b71081ef3b8543147df11b65bfcfe2c37647c7cef3a3415fdfb574b64) with corresponding serde in [xmtp.xmtpv4.payer_api.serde.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-067327d1062f6c9e16fff59850321fa1184e76e3629bc9c7195ff9dfd3b2e4d8)
- Make `xmtp_api_grpc::ClientBuilder` `Clone` in [client.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-9fb1ae52e16798d7a38d2a3fae784b2aedafb84be7be76737ffe42205d1d4121) and make `HealthCheck` `Clone` in [health_check.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-bc1eb7fe0712626b37f84761c85c9f97f6b4371846d61022adc613b39065f628)
- Update dependencies in [Cargo.toml](https://github.com/xmtp/libxmtp/pull/2544/files#diff-4f852b7aca87c43016619ef0c0ead683b676213925d9f3f78ec4ba58e955f373) and lockfile in [Cargo.lock](https://github.com/xmtp/libxmtp/pull/2544/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87e); bump docker image tags in [docker-compose-d14n.yml](https://github.com/xmtp/libxmtp/pull/2544/files#diff-f3b95b320e721deae58c0f228c225c5dc895a08336200f71f989ac0764d95776)

#### 📍Where to Start
Start with the `xmtp_api_d14n::middleware::MultiNodeClient` and its initialization path in [multi_node_client.rs](https://github.com/xmtp/libxmtp/pull/2544/files#diff-c6b13327fb82bc847cf6f3c8a0f113b7b1ae6b1e19afc3c2605252e4926b2552), focusing on `init_inner`, `get_nodes`, and `get_fastest_node`.

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 61c94b7. 4 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
No issues evaluated.

</details>

### DIFF
--- a/xmtp_mls/Cargo.toml
+++ b/xmtp_mls/Cargo.toml
@@ -214,6 +214,15 @@ harness = false
 name = "groups"
 required-features = ["bench"]
 
+[[bench]]
+harness = false
+name = "messages"
+required-features = ["bench"]
+[[bench]]
+harness = false
+name = "sync_conversations"
+required-features = ["bench"]
+
 [[test]]
 harness = true
 name = "chaos"
--- a/xmtp_mls/benches/messages.rs
+++ b/xmtp_mls/benches/messages.rs
@@ -1,0 +1,492 @@
+//! Benchmarks for message finding operations with shared setup
+//!
+//! This version shares the expensive setup between benchmark samples to reduce total runtime.
+
+use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
+use std::{hint::black_box, sync::Arc, time::Duration};
+use tokio::runtime::{Builder, Runtime};
+use tracing::{Instrument, trace_span};
+use xmtp_common::bench::{self, BENCH_ROOT_SPAN};
+use xmtp_db::encrypted_store::group_message::{
+    ContentType, DeliveryStatus, GroupMessageKind, MsgQueryArgs, SortDirection,
+};
+use xmtp_mls::utils::bench::{MessageBenchSetup, new_client, setup_group_with_messages};
+
+pub const MESSAGE_COUNTS: [usize; 5] = [10, 100, 1000, 10000, 50000];
+pub const SAMPLE_SIZE: usize = 10;
+
+fn setup_runtime() -> Runtime {
+    Builder::new_multi_thread()
+        .enable_time()
+        .enable_io()
+        .thread_name("xmtp-messages-bencher")
+        .build()
+        .unwrap()
+}
+
+/// Shared setup for all benchmarks - creates client and group with messages once per MESSAGE_COUNT
+async fn setup_benchmark(total_messages: usize) -> Arc<MessageBenchSetup> {
+    let client = new_client(false).await;
+    setup_group_with_messages(client, total_messages).await
+}
+
+fn bench_find_messages(c: &mut Criterion) {
+    bench::logger();
+    let mut benchmark_group = c.benchmark_group("find_messages_shared");
+    benchmark_group.sample_size(SAMPLE_SIZE);
+    benchmark_group.measurement_time(Duration::from_secs(30));
+    benchmark_group.warm_up_time(Duration::from_secs(3));
+
+    let runtime = Arc::new(setup_runtime());
+
+    for &total_messages in MESSAGE_COUNTS.iter() {
+        benchmark_group.throughput(Throughput::Elements(10_u64)); // Limit of 10
+
+        // Setup once per MESSAGE_COUNT - completely outside the benchmark
+        let setup = runtime.block_on(setup_benchmark(total_messages));
+        let runtime_clone = runtime.clone();
+
+        benchmark_group.bench_function(
+            BenchmarkId::new("find_messages_limit_10", total_messages),
+            move |b| {
+                let span = trace_span!(BENCH_ROOT_SPAN, total_messages);
+                let setup = setup.clone();
+                let runtime = runtime_clone.clone();
+
+                b.iter(|| {
+                    runtime.block_on(
+                        async {
+                            let messages = setup
+                                .group
+                                .find_messages(&MsgQueryArgs {
+                                    limit: Some(10),
+                                    ..Default::default()
+                                })
+                                .unwrap();
+
+                            assert_eq!(
+                                messages.len(),
+                                10,
+                                "Expected exactly 10 messages, got {}",
+                                messages.len()
+                            );
+                            black_box(messages);
+                        }
+                        .instrument(span.clone()),
+                    )
+                });
+            },
+        );
+    }
+
+    benchmark_group.finish();
+}
+
+fn bench_find_messages_v2(c: &mut Criterion) {
+    bench::logger();
+    let mut benchmark_group = c.benchmark_group("find_messages_v2_shared");
+    benchmark_group.sample_size(SAMPLE_SIZE);
+    benchmark_group.measurement_time(Duration::from_secs(30));
+    benchmark_group.warm_up_time(Duration::from_secs(3));
+
+    let runtime = Arc::new(setup_runtime());
+
+    for &total_messages in MESSAGE_COUNTS.iter() {
+        benchmark_group.throughput(Throughput::Elements(10_u64)); // Limit of 10
+
+        // Setup once per MESSAGE_COUNT - completely outside the benchmark
+        let setup = runtime.block_on(setup_benchmark(total_messages));
+        let runtime_clone = runtime.clone();
+
+        benchmark_group.bench_function(
+            BenchmarkId::new("find_messages_v2_limit_10", total_messages),
+            move |b| {
+                let span = trace_span!(BENCH_ROOT_SPAN, total_messages);
+                let setup = setup.clone();
+                let runtime = runtime_clone.clone();
+
+                b.iter(|| {
+                    runtime.block_on(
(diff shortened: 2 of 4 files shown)


xmtp_api_d14n/src/middleware/aggregate_client.rs
```diff
@@ -0,0 +1,156 @@
...
+) -> Result<GrpcClient, Box<dyn std::error::Error + Send + Sync>> {
+    let endpoint = HealthCheck::builder()
+        .build()
+        .map_err(|e| format!("get health check build failed: {e}"))?;
+
+    let futures = clients.into_iter().map(|(node_id, client)| {
+        let endpoint = endpoint.clone();
+        async move {
+            let start = Instant::now();
+            let result = xmtp_common::time::timeout(timeout, endpoint.query(&client)).await;
```
insipx: it might be nice to include a latency combinator, like `api::measured` or something that automatically runs the request w/ a timeout, then returns a struct with the query result and measured `std::time::Duration`

i have some example combinators in my branch: https://github.com/xmtp/libxmtp/tree/insipx/v3-combinators/xmtp_proto/src/traits/combinators

the trait changed a little but should apply. that way we can re use this logic for any other endpoint anywhere in the code, which might be useful for debugging in the future

insipx: can also be done in a followup

xmtp_api_d14n/src/clients/aggregate_client.rs
```diff
@@ -0,0 +1,156 @@
...
+        return Err("All node clients failed to build".into());
+    }
+
+    Ok(clients)
+}
+
+async fn get_fastest_node(
+    clients: HashMap<u32, GrpcClient>,
+    timeout: Duration,
+) -> Result<GrpcClient, Box<dyn std::error::Error + Send + Sync>> {
```
insipx: since we're indicating the `GrpcClient` anyway, we can return a `ApiClientError<GrpcError>` here -- it should make the error handling easier

xmtp_api_d14n/src/middleware/aggregate_client.rs
```diff
@@ -0,0 +1,156 @@
...
+
+            match result {
+                Ok(Ok(_)) => Some((node_id, client, start.elapsed().as_millis() as u64)),
```
insipx: it would be good to log the failures here with `tracing::` so we know what fails

further, you can use a combination of [`flatten`](https://doc.rust-lang.org/std/result/enum.Result.html#method.flatten) and [`ok`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) to make  this a bit more readable. flatten  requires  the same error type (as well as latest rust `v1.89`) so you may need to map the outer timeout error to apiclient error first.

then you can use `inspect_err` to inspect the error for a log before using `ok()` to return it

xmtp_api_d14n/src/clients/aggregate_client.rs
```diff
@@ -0,0 +1,156 @@
...
+    gateway_client: C,
+    inner: C,
+    timeout: Duration,
+}
+
+impl AggregateClient<GrpcClient> {
+    pub async fn new(
+        gateway_client: GrpcClient,
+        timeout: Duration,
+    ) -> Result<Self, Box<dyn std::error::Error + Send + Sync>> {
```
insipx: generally we want to stay away from `Box<dyn Error>`. it would be more idiomatic to create your own Error enum and `impl From<YourError> for ApiClientError`. for errors we use the [`thiserror`](https://docs.rs/thiserror) crate

when you write that conversion, `YourError` can turn into a `Box<dyn>` on `ApiClientError::Other`. however keeping an enum of errors around in the middleware implementation is useful for implementing `RetryableError` and possibly for tests. its much easier to match on an enum, then downcasting a `Box<>`. further people looking at the code will also have a good reference for what can go wrong w/o going into code details (as long as the enum remains small).

xmtp_api_d14n/src/middleware/aggregate_client.rs
```diff
@@ -0,0 +1,156 @@
...
+        let mut client_builder = GrpcClient::builder();
+        let is_tls = url.parse::<url::Url>()?.scheme() == "https";
+
+        client_builder.set_tls(is_tls);
+        client_builder.set_host(url);
+
+        let client = client_builder.build().await?;
+
+        Ok::<_, Box<dyn std::error::Error + Send + Sync>>((node_id, client))
+    });
```
insipx: since this is essentially a `Vec<Future>`, you can make the operations more clear and process as items are ready (rather than waiting for every item to be ready with join_all) by using [`stream::iter`](https://docs.rs/futures/latest/futures/stream/fn.iter.html) together with one of the [`buffer_unordered`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered) on the `StreamExt` trait. 

```rust
let values = vec![future1, future2];
let clients = stream::iter(values).buffer_unordered(some_limit);
```

then you can use the other fns on [`StreamExt`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.for_each) or just loop/collect the collection

```rust
clients.collect::<Result<_, _>>().await?;
```

or
```rust
while let Some(client) = clients.next().await? {
    client.insert(client);
}
```

etc.etc. theres a bunch of useful stuff in the StreamExt trait

if you collect into a list, you can then `into_iter()` and use the `partion_results` of itertools to easily handle result values too: https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.partition_result

xmtp_api_d14n/src/middleware/aggregate_client.rs
fbac: Great idea, I'll do this in a followup PR, so we focus here on the client and its builder.

xmtp_api_d14n/src/middleware/aggregate_client.rs
```diff
@@ -0,0 +1,331 @@
...
+        ApiClientError::<E>::Other(Box::new(value))
+    }
+}
+
+/// RetryableError for AggregateClientError is used to determine if the error is retryable.
+/// Trait needed by the From<AggregateClientError> for ApiClientError<E> implementation.
+/// All errors are not retryable.
+impl RetryableError for AggregateClientError {
+    fn is_retryable(&self) -> bool {
+        false
```
insipx: going to need to passthrough is_retryable to errors in the enum which might be retryable, like the rest of ApiClientError or BodyError, i.e

```rust
fn is_retryable(&self) -> bool {
    use AggregateClientError::*;
    match self {
        GrpcError(e) => e.is_retryable(),
        BodyError(e) => e.is_retryable(),
    }
}
```

if we don't, then upstream code which expects some api errors to be retryable that would otherwise succeed will fail. for instance, on network instability

fbac: Implemented.

PR #2568, opened 2025-10-09
welcome pointer protos

tylerhawkes:
### Introduce Welcome Pointer protocol support across generated proto types and runtime handling, adding `WelcomeMessage.Version::WelcomePointer` serialization/deserialization and integrating pointer-aware sizing in `xmtp_mls::groups::Group::sync`
This change adds Welcome Pointer support across generated proto definitions and serde, integrates pointer-aware handling in group sync, and wires new optional fields and tasks into database and API layers. It introduces new enum and message types for pointer contents and wrapper algorithms, updates serialization/deserialization for the new oneof variants, and adds explicit `todo!`/`unimplemented!` guards where runtime handling is not yet implemented.

- Add `WelcomeMessage.Version::WelcomePointer` and `WelcomeMessageInput.Version::WelcomePointer` with associated structs and serde in [xmtp.mls.api.v1.rs](https://github.com/xmtp/libxmtp/pull/2568/files#diff-7aaf219648e096ca94f4232a745675c0454d8fa76552259b7a0e655ae1c930b0) and [xmtp.mls.api.v1.serde.rs](https://github.com/xmtp/libxmtp/pull/2568/files#diff-5ba5626760dd0656b6a1290c3835f281e50a96922a3dae21800c78c25bf69994)
- Introduce message and enum types for Welcome Pointer and pointee AEAD metadata with serde in [xmtp.mls.message_contents.rs](https://github.com/xmtp/libxmtp/pull/2568/files#diff-d831f53a5ef95f66ec27ea70736b593fa5f0bea5337da312f933619f551799ba) and [xmtp.mls.message_contents.serde.rs](https://github.com/xmtp/libxmtp/pull/2568/files#diff-122b6d63d45ac71bb5d12c5c19db44cdc885d615984b73a44575252c30287341)
- Extend `xmtp_mls::groups::Group::sync` to size payloads for `WelcomeMessageInputVersion::WelcomePointer` in [mls_sync.rs](https://github.com/xmtp/libxmtp/pull/2568/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e)
[...]

PR #2569, opened 2025-10-09
welcome pointer v3 impl

tylerhawkes:
### Introduce welcome pointer v3 by adding symmetric AEAD welcome pointer support, variant-based `xmtp_proto::types::WelcomeMessage` models, and mixed send logic in `xmtp_mls::groups::mls_sync::Group<Context>::send_welcomes` with chunking clamped to 1–50 and pointer sending gated by `INSTALLATION_THRESHOLD_FOR_WELCOME_POINTER_SENDING=100000`
This adds a v3 welcome pointer implementation across message models, sending, decryption, and processing, and consolidates error handling to `ConversionError` for proto conversions.

- Add variant-based welcome message types (`V1`, `WelcomePointer`, `DecryptedWelcomePointer`) and update extractors and envelopes to dispatch accordingly in `xmtp_proto/src/types/welcome_message.rs` and `xmtp_api_d14n/src/protocol/envelopes.rs`.
- Implement mixed send logic to emit `WelcomePointer` and V1 messages with symmetric encryption when compatible installations exceed `INSTALLATION_THRESHOLD_FOR_WELCOME_POINTER_SENDING=100000` in `xmtp_mls/src/groups/mls_sync.rs`.
- Add symmetric AEAD decrypt/unwrap helpers and pointer resolution, and support constructing `DecryptedWelcome` from proto variants in `xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs` and `xmtp_mls/src/groups/welcome_pointer.rs`.
- Advertise welcome pointee AEAD capabilities via an UnknownExtension and include it in key packages in `xmtp_mls/src/groups/mls_ext/mls_ext_welcome_pointee_encryption_aead_type.rs` and `xmtp_mls/src/identity.rs`.
- Standardize prost encode/decode and enum conversion errors to `xmtp_proto::ConversionError` across groups and verification code.
- Cap send chunk size to 50 and adjust wrapper API to pass algorithms by value across wrappers and benches.

#### 📍Where to Start
Start with the mixed send path and pointer construction in `Group<Context>::send_welcomes` in [mls_sync.rs](https://github.com/xmtp/libxmtp/pull/2569/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e), then follow pointer resolution and decryption through `DecryptedWelcome::from_welcome_proto` in [decrypted_welcome.rs](https://github.com/xmtp/libxmtp/pull/2569/files#diff-9825b13d685491fae6f92c0a816a683e6cdd5e91b0d39963afd2c1247b295cb2) and `resolve_welcome_pointer` in [welcome_pointer.rs](https://github.com/xmtp/libxmtp/pull/2569/files#diff-5153b06b316345923b4f192c248f8e9645130e15b8eef678e3baf624f8f4bc27).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 92b906a. 27 files reviewed, 59 issues evaluated, 59 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>examples/cli/debug.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 66](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/examples/cli/debug.rs#L66): `unwrap()` on the result of `api_client.query_welcome_messages(&installation_id).await` will panic on any API/network/deserialization error, terminating the process instead of returning a `Result` error. This violates the function’s error-handling contract (`Result<(), String>`), can crash at runtime, and prevents graceful handling/logging. Replace `unwrap()` with proper error propagation (e.g., `.await.map_err(|e| e.to_string())?`) so that failures return `Err(String)` rather than panic. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/extractors/welcomes.rs — 0 comments posted, 8 evaluated, 8 filtered</summary>

- [line 66](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L66): On error paths before `self.welcome_message` is assigned (e.g., `installation_key` or `wrapper_algorithm` `try_into()` failures), the method returns `Err` while leaving any prior `self.welcome_message` value intact. This stale state can lead to incorrect downstream behavior if the extractor instance is reused, because the prior message remains set despite the current message failing. The extractor should either clear `self.welcome_message` at the start of the visit, clear it on error, or ensure a fresh extractor instance per message. <b>[ Low confidence ]</b>
- [line 72](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L72): `self.welcome_message` is overwritten unconditionally (`self.welcome_message = Some(builder)`) without guarding against multiple invocations or enforcing uniqueness/mutual exclusion. If the extractor can visit more than one welcome-related message in a single processing run, the earlier value is silently discarded and replaced, violating sequence constraints (uniqueness and adjacency) and causing potential data loss without an error. This applies in both `visit_welcome_message_v1` and `visit_welcome_pointer`. A guard (e.g., returning an error if `self.welcome_message` is already `Some(...)`) or explicit reset logic is required to preserve invariants. <b>[ Invalidated by documentation search ]</b>
- [line 82](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L82): On error paths in `visit_welcome_pointer` before `self.welcome_message` is assigned (e.g., `installation_key` or `wrapper_algorithm` `try_into()` failures), the method returns `Err` while leaving any prior `self.welcome_message` value intact. If the extractor instance is reused, this stale state can cause incorrect behavior. As in `visit_welcome_message_v1`, the extractor should clear state at the start or on error, or ensure single-use instances. <b>[ Low confidence ]</b>
- [line 87](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L87): `self.welcome_message` is overwritten unconditionally in `visit_welcome_pointer` (`self.welcome_message = Some(builder)`) without guarding against multiple invocations or enforcing uniqueness/mutual exclusion, mirroring the issue in `visit_welcome_message_v1`. If multiple visit methods are called on the same instance, earlier state is silently discarded, violating sequence constraints and causing potential data loss. <b>[ Invalidated by documentation search ]</b>
- [line 119](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L119): Potentially lossy and unchecked cast of `message.created_ns` to `i64` before passing to `DateTime::from_timestamp_nanos` at `created_ns(DateTime::from_timestamp_nanos(message.created_ns as i64))`. Using `as i64` performs a truncating, potentially wrapping conversion for non-`i64` numeric types (e.g., `u64`). If `message.created_ns` exceeds `i64::MAX` or is an unsigned value near `u64::MAX`, this cast can produce a negative or otherwise incorrect timestamp. Depending on the behavior of `DateTime::from_timestamp_nanos`, this may either silently record an incorrect creation time or trigger a panic for out-of-range input. The code provides no bounds check or error handling around this conversion, making the path unsafe for large or unsigned inputs. <b>[ Invalidated by documentation search ]</b>
- [line 125](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L125): Behavioral contract change: `wrapper_algorithm` is now converted via `try_into()` with `?` (`.wrapper_algorithm(message.wrapper_algorithm.try_into()?)`) instead of being passed through directly. This introduces a new fallible path where previously the value was accepted as-is. If `message.wrapper_algorithm` contains an unrecognized or out-of-range value, this code will return an error (`Self::Error`) and abort processing, whereas prior behavior would continue. This can cause previously accepted messages to be rejected, changing externally visible contract/compatibility without explicit handling or fallback. <b>[ Low confidence ]</b>
- [line 138](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L138): Partial mutation of `self.welcome_message` occurs before any fallible conversions are validated, causing inconsistent internal state on error. The calls to `cursor(...)` (lines 138–141) and `created_ns(...)` (line 142) mutate `self.welcome_message` before the fallible builder chain inside `.variant(...)` is evaluated. If `installation_key(message.installation_key.as_slice().try_into()?)`, `wrapper_algorithm(message.wrapper_algorithm.try_into()?)`, or `.build()?` fails, the function returns early with `Err` after having already updated `cursor` and `created_ns`. This leaves `self.welcome_message` in a partially-updated state derived from an invalid message. To preserve invariants, compute the `WelcomePointer` fully first (including all `?` validations) and only then apply mutations to `self.welcome_message` atomically, or buffer mutations in a local builder and swap only on success. <b>[ Low confidence ]</b>
- [line 142](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L142): Potentially invalid timestamp due to unchecked narrowing cast: `DateTime::from_timestamp_nanos(message.created_ns as i64)` blindly casts `message.created_ns` to `i64`. If `created_ns` exceeds `i64::MAX`, the cast yields a negative value due to two's-complement wrap, producing an incorrect or out-of-range `DateTime`. This can silently corrupt the timestamp or even panic depending on `DateTime` implementation. Validate the range before casting (e.g., return an error if `created_ns > i64::MAX`) or use an API that accepts `u64`/checked conversion. <b>[ Invalidated by documentation search ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/traits/visitor.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 85](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/traits/visitor.rs#L85): Misleading log message in `visit_welcome_message_version`: the method logs `"noop_visit_group_message_version"`, which contradicts the method’s purpose and name and can mislead diagnostics and telemetry. This is a runtime-observable behavior issue that can hinder debugging or log-based monitoring, especially since this method is invoked by `WelcomeMessageVersion::accept` prior to variant-specific visits. <b>[ Out of scope ]</b>
- [line 189](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_api_d14n/src/protocol/traits/visitor.rs#L189): Misleading log message in `visit_subscribe_welcome_messages_request`: the method logs `"noop_visit_subscribe_group_messages_request"`, which refers to a different method and can cause confusion during troubleshooting or log analysis. This is runtime-observable and contradicts the method name/purpose. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_db/src/xmtp_openmls_provider.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 36](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_db/src/xmtp_openmls_provider.rs#L36): The doc comment for the `savepoint` method was changed to say "Start a new transaction", which explicitly contradicts the method name `savepoint` and the adjacent comments/TODO that say it "Must only be used when already in a transaction". This contradiction can lead callers to incorrectly use `savepoint` as if it begins a new transaction, risking runtime issues (e.g., SQLite race conditions mentioned in the comments) when used outside an existing transaction. The implementation contract is unclear due to this conflict between documentation and intended semantics. <b>[ Low confidence ]</b>
- [line 42](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_db/src/xmtp_openmls_provider.rs#L42): The `savepoint` method has a critical precondition ("Must only be used when already in a transaction") that is not enforced by the API (no type-state indicating an active transaction, no runtime guard specified). The comments acknowledge that violating this can cause SQLite race conditions because it does not use BEGIN IMMEDIATE. Without an enforced invariant, callers can invoke `savepoint` outside a transaction, risking runtime errors/races. Consider enforcing this via a typed transaction context (e.g., requiring a `&mut Self::TxQuery` that is only obtainable within `transaction`), or perform a runtime check of transaction depth as suggested in the TODO. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/benches/crypto.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 45](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/benches/crypto.rs#L45): Unwraps may panic at runtime: both `crypto.random_vec(CIPHERSUITE.hash_length()).unwrap()` and `crypto.derive_hpke_keypair(CIPHERSUITE.hpke_config(), ikm.as_slice()).unwrap()` will panic if the underlying operations fail (e.g., RNG failure, misconfiguration, or any internal error). In a benchmark run, this causes abrupt termination. If panics are acceptable for unrecoverable initialization failures, consider using `.expect("...reason...")` to make intent explicit, or handle errors to ensure every input reaches a defined terminal state. <b>[ Out of scope ]</b>
- [line 49](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/benches/crypto.rs#L49): Potential out-of-memory or excessive resource usage for large payload sizes: allocating and filling `payload` with `vec![0; size]` where `size` goes up to `134_217_728` bytes per iteration can lead to OOM or swapping on constrained systems, causing process abort (Rust aborts on OOM) or severe runtime degradation during the benchmark. Consider reducing the upper bound, using `BatchSize::PerIteration` with controlled memory reuse, or gating large sizes behind a feature flag/environment check to ensure benchmarks complete reliably. <b>[ Out of scope ]</b>
- [line 53](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/benches/crypto.rs#L53): Possible silent benchmarking of failure path: the return value of `wrap_welcome(&payload, &[], &key, WrapperAlgorithm::Curve25519)` is ignored, so if `wrap_welcome` returns an `Err` for any input (e.g., due to key/ciphersuite constraints or internal failures), the benchmark would still run and record timings for the failure path rather than asserting success or reporting an error. This produces misleading benchmark results and hides runtime errors. Consider asserting/propagating success (e.g., `.expect("wrap_welcome should succeed")`), or otherwise recording failures explicitly so all inputs reach a defined and correct terminal state. <b>[ Test / Mock code ]</b>
</details>

<details>
<summary>xmtp_mls/src/client.rs — 0 comments posted, 6 evaluated, 6 filtered</summary>

- [line 1062](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/client.rs#L1062): Possible panic if the value for the key is `None`: `kp1.remove(&installation_public_key).unwrap().unwrap()` performs a second `unwrap()` on the map value (likely `Option<KeyPackage>`). If `get_key_packages_for_installation_ids` indicates "no key package" by returning `None` for that key, this will panic. Handle the `None` with an explicit assertion or branch to produce a clearer failure. <b>[ Test / Mock code ]</b>
- [line 1062](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/client.rs#L1062): Possible panic if the requested key is absent in the returned map: `kp1.remove(&installation_public_key).unwrap()` unconditionally unwraps the `Option` from `remove`. Even though the code asserts `kp1.len() == 1`, that does not guarantee the presence of `installation_public_key` in the map. If `get_key_packages_for_installation_ids` omits missing IDs or returns a different key, this will panic. Consider handling the `None` case with an explicit assertion or error to keep the failure clearer and avoid a runtime panic. <b>[ Test / Mock code ]</b>
- [line 1064](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/client.rs#L1064): Possible panic if `StoredIdentity` is missing from the database: `client.context.db().fetch(&()).unwrap().unwrap()` unwraps both a `Result` and an `Option`. If the fetch returns `Ok(None)`, this will panic. While the test client likely initializes an identity, the code here assumes its existence without an explicit guard. Consider asserting presence with a helpful message to avoid a generic panic. <b>[ Out of scope ]</b>
- [line 1072](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/client.rs#L1072): Timing-dependent test may be flaky: the test sleeps for 11 seconds after queueing key rotation and then expects a rotated key (asserting `init1 != init2`). Without a deterministic signal that rotation has completed, this can be nondeterministic across slower environments or scheduling differences (e.g., in CI or wasm). Consider replacing fixed sleep with a polling loop that times out, or a hook/event indicating rotation completion, to ensure a well-defined terminal state. <b>[ Out of scope ]</b>
- [line 1079](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/client.rs#L1079): Possible panic if the requested key is absent in the returned map during the second fetch: `kp2.remove(&installation_public_key).unwrap()` unconditionally unwraps and will panic if the key is missing. Asserting `kp2.len() == 1` does not strictly guarantee the presence of `installation_public_key`. Consider explicit presence checks for clearer failures. <b>[ Test / Mock code ]</b>
- [line 1079](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/client.rs#L1079): Possible panic if the value for the key is `None` during the second fetch: `kp2.remove(&installation_public_key).unwrap().unwrap()` will panic if the map contains the key with a `None` value. Handle this case explicitly to produce a clearer failure rather than a panic. <b>[ Test / Mock code ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 109](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs#L109): `welcome_from_decrypted_welcome_pointer` always attempts to decrypt `v1.welcome_metadata` even when the ciphertext is empty. On valid inputs where metadata is intentionally omitted (empty), AEAD decryption on an empty ciphertext can fail and erroneously return a `CryptoError`, causing the whole path to fail instead of cleanly treating metadata as absent. The decryption at lines `109–114` should be conditional on `!v1.welcome_metadata.is_empty()` to avoid spurious failures for welcomes without metadata. <b>[ Low confidence ]</b>
- [line 116](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs#L116): `welcome_from_decrypted_welcome_pointer` silently converts an empty decrypted metadata payload into `None` by filtering with `Some(...).filter(|data| !data.is_empty())`. This loses information about whether metadata was present but empty versus truly absent, which can break consumers that rely on that distinction. If the contract differentiates between empty metadata and no metadata, this is silent data loss. Consider preserving `Some(Vec::new())` or explicitly documenting/validating that empty is treated as absent. <b>[ Low confidence ]</b>
- [line 171](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs#L171): `find_key_package_hash_ref` serializes `hpke_public_key: &[u8]` using `tls_serialize_detached()` before performing the storage lookup. If the storage key under `KEY_PACKAGE_REFERENCES` was originally stored using raw HPKE public key bytes (not TLS-serialized form), double-serializing here will produce a different key and cause a spurious `NotFound` error, breaking valid welcome pointer decryption. Ensure the serialization format used here exactly matches the format used when writing the references, or remove the extra serialization if the bytes are already in their canonical persisted form. <b>[ Low confidence ]</b>
- [line 195](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs#L195): In `find_private_key`, a failure to serialize `hash_ref` using `bincode::serialize` is mapped to `GroupError::NotFound(NotFound::PostQuantumPrivateKey)`. This conflates a serialization error with a missing key and produces misleading error semantics. If serialization fails (e.g., due to incompatible type changes or serialization settings), the caller will incorrectly be told the private key was not found rather than that the input could not be serialized. This violates clear error reporting and can misdirect recovery logic that might retry or attempt re-population of keys.
Problematic code: `let serialized_hash_ref = bincode::serialize(hash_ref).map_err(|_| GroupError::NotFound(NotFound::PostQuantumPrivateKey))?;` <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_ext/mls_ext_wrapper_encryption.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 97](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/mls_ext_wrapper_encryption.rs#L97): `try_from` only maps `WelcomePointerWrapperAlgorithmProto::XwingMlkem768Draft6` and treats all other variants (including the valid `Curve25519`) as errors. This will cause deserialization or conversion from proto to `WrapperAlgorithm` to fail for traditional algorithms, breaking round-trip and parity with the `Into` conversions used elsewhere. The conversion should handle all supported algorithms explicitly, at minimum mapping `WelcomePointerWrapperAlgorithmProto::Curve25519` to `WrapperAlgorithm::Curve25519`, rather than rejecting it via the catch-all `_` case. <b>[ Low confidence ]</b>
- [line 117](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/mls_ext_wrapper_encryption.rs#L117): The `try_from` implementation for converting `WrapperAlgorithm` to the proto enum (`WelcomePointerWrapperAlgorithmProto`) only accepts `WrapperAlgorithm::XWingMLKEM768Draft6` and returns a `ConversionError::InvalidValue` for all other variants (e.g., `WrapperAlgorithm::Curve25519`). This makes valid wrapper algorithms unsupported in this conversion path and will cause runtime failures (conversion errors) when attempting to encode or map Curve25519 wrapper algorithms into proto messages. Given `WrapperAlgorithm` has a `Curve25519` variant and is used elsewhere (including decryption logic that supports Curve25519), rejecting it here breaks contract parity and introduces a runtime error for valid inputs.
Problematic code: `match value { WrapperAlgorithm::XWingMLKEM768Draft6 => Ok(...), _ => Err(xmtp_proto::ConversionError::InvalidValue { ... }) }` <b>[ Low confidence ]</b>
- [line 172](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/mls_ext_wrapper_encryption.rs#L172): Behavioral contract change: `WrapperEncryptionExtension::try_from` now uses `proto.algorithm.try_into()?` instead of the previous infallible conversion (e.g., `into()`). This introduces a new error path where decoding fails if `proto.algorithm` is an unknown/unsupported value. Previously, with an infallible conversion, unknown enum values must have been mapped to some default or `Unknown` variant by `From`, allowing the decode to succeed and the extension to be constructed. Callers that previously received `Ok(WrapperEncryptionExtension { algorithm: Unknown, ... })` for unknown algorithm values will now receive an `Err(Self::Error)`. This is an externally visible behavior change that can cause runtime failures (rejecting inputs that were previously accepted), violating contract parity unless explicitly documented and intended. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs — 0 comments posted, 8 evaluated, 8 filtered</summary>

- [line 13](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs#L13): The static `LIBCRUX_CRYPTO_PROVIDER` is initialized with `openmls_libcrux_crypto::CryptoProvider::new().expect("Failed to create CryptoProvider")`, which will panic at runtime if the provider cannot be created. This is a hard crash during module initialization, and any call to functions using `LIBCRUX_CRYPTO_PROVIDER` (e.g., `wrap_welcome_symmetric`) will never be reached if initialization fails. Consider returning a fallible handle or deferring creation and propagating an error instead of panicking to avoid non-graceful failure. <b>[ Low confidence ]</b>
- [line 160](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs#L160): Potential runtime panic on first use of `LIBCRUX_CRYPTO_PROVIDER` due to `expect` during lazy initialization. At line 160, calling `(*LIBCRUX_CRYPTO_PROVIDER)` may trigger the `LazyLock` initialization which uses `openmls_libcrux_crypto::CryptoProvider::new().expect("Failed to create CryptoProvider")`. If `CryptoProvider::new()` returns an `Err` (e.g., due to unavailable randomness or environment issues), the process will panic instead of returning an error. This violates the function’s error-handling contract (`Result<_, UnwrapWelcomeError>`) and bypasses the newly added `Crypto` error propagation path. The safer behavior is to avoid `expect` and propagate the initialization failure as `UnwrapWelcomeError::Crypto` (or another appropriate variant), ensuring callers receive a structured error rather than a crash. <b>[ Low confidence ]</b>
- [line 193](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs#L193): `new_key_package(...).unwrap()` will panic if `new_key_package` returns `Err`. In this test, there is no guard ensuring the MLS provider and identity setup can never fail at runtime (e.g., misconfigured provider, unsupported algorithm, key store issues). Panic-based termination is not constrained to a pre-effect initialization invariant; it occurs after external effects (client construction) and will abort the test without a graceful failure path. <b>[ Out of scope ]</b>
- [line 213](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs#L213): `wrap_welcome(...).unwrap()` will panic if `wrap_welcome` returns `Err` (e.g., due to invalid public key material, algorithm mismatch, or cryptographic failure). The test provides no explicit guard to prove the error is impossible, so a runtime panic is reachable. <b>[ Out of scope ]</b>
- [line 224](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs#L224): `unwrap_welcome(...).unwrap()` will panic if `unwrap_welcome` returns `Err` (e.g., bad private key lookup, corrupt ciphertext, or mismatched algorithm). There is no visible guard ensuring these cannot occur, so a runtime panic path exists. <b>[ Out of scope ]</b>
- [line 238](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs#L238): The test unconditionally unwraps `pq_pub_key` with `let pq_pub_key = maybe_pq_pub_key.unwrap();`. If the environment/provider does not produce a post-quantum public key (e.g., feature disabled or not supported), this will panic, making the test environment-dependent and brittle. A guard that skips or conditionally runs the test when `pq_pub_key` is `None` would avoid runtime panics. <b>[ Test / Mock code ]</b>
- [line 264](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs#L264): The test asserts that the ciphertext is not equal to the plaintext values using `assert_ne!` (e.g., `assert_ne!(&to_encrypt, &wrapped.0);` and `assert_ne!(&to_encrypt_metadata, &wrapped.1);`). While extremely unlikely, HPKE ciphertexts are byte strings and can, with negligible probability, match the plaintext exactly, making the test probabilistically flaky. Tests should avoid probabilistic failure conditions and instead validate structural properties (e.g., successful decrypt round-trip, non-empty encapsulated key, different lengths if guaranteed by format, or metadata like AAD impact). <b>[ Test / Mock code ]</b>
- [line 420](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs#L420): The test asserts `assert_ne!(to_encrypt_metadata, wrapped.1);` relying on ciphertext inequality to plaintext. As with the other test, this is probabilistic and can theoretically fail, leading to flakiness. Prefer deterministic validation (e.g., successful decryption round-trip) instead of asserting byte-wise inequality. <b>[ Test / Mock code ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_sync.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 2405](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_sync.rs#L2405): Negative `message_cursor` values are cast from `i64` to `u64` via `as` without validation: `message_cursor.unwrap_or(0) as u64`. For negative inputs this silently wraps to a very large `u64`, producing incorrect `WelcomeMetadata.message_cursor` and potentially breaking downstream logic. Either reject negative cursors with an error, or clamp at zero before casting. <b>[ Invalidated by documentation search ]</b>
- [line 2464](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/mls_sync.rs#L2464): Potential algorithm mismatch in `WelcomePointerProto`: the `aead_type` embedded in the welcome pointer is hard-coded to `Chacha20Poly1305`, while the actual symmetric wrapping earlier in the function uses the value from `WelcomePointersExtension::preferred_type()`. If `preferred_type()` returns a value other than ChaCha20Poly1305 (now or in the future), recipients will interpret the pointer using the wrong AEAD and fail to decrypt, causing runtime failures in welcome processing. Update the `aead_type` field in the proto to match the `aead_type` actually used for symmetric wrapping. <b>[ Invalidated by documentation search ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/welcome_sync.rs — 0 comments posted, 5 evaluated, 5 filtered</summary>

- [line 297](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcome_sync.rs#L297): `welcome.tls_serialize_detached().unwrap()` will panic on serialization failure. `MlsMessageOut::tls_serialize_detached()` returns a `Result`, and this path does not guard against errors. Any malformed or otherwise non-serializable `welcome` will cause a runtime crash rather than returning a defined fallback or error. <b>[ Out of scope ]</b>
- [line 299](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcome_sync.rs#L299): `message_cursor.unwrap_or(0)` silently maps `None` to `0` when creating `WelcomeMetadata`. If `0` is a valid cursor value or if `None` is expected to encode absence distinctly from `0`, this loses semantic information and can cause downstream misinterpretation. There is no visible guarantee that `0` is an appropriate sentinel. <b>[ Out of scope ]</b>
- [line 302](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcome_sync.rs#L302): `public_key` is accepted and used without validation. Passing an empty or malformed `public_key` is reachable (no guards), which can lead to `wrap_welcome` failure (panic due to `unwrap()`), or worse, construction of a `WelcomeMessage` with an invalid `hpke_public_key` field (`hpke_public_key: public_key` at line 316). This can violate external contract expectations and produce a malformed artifact without a clear error. <b>[ Out of scope ]</b>
- [line 305](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcome_sync.rs#L305): `wrap_welcome(...).unwrap()` will panic if `wrap_welcome` returns an error (e.g., due to an invalid `public_key`, HPKE failure, or metadata encoding issues). This function performs external crypto processing and can fail at runtime; panicking here yields undefined behavior for callers rather than a controlled error. <b>[ Out of scope ]</b>
- [line 308](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcome_sync.rs#L308): `random.as_v1().unwrap()` will panic if `WelcomeMessage::generate()` returns a non-V1 variant. There is no visible invariant guaranteeing that `generate()` yields `V1`. If future changes introduce additional variants or conditional generation, this path will crash unexpectedly. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/welcomes/xmtp_welcome.rs — 0 comments posted, 9 evaluated, 9 filtered</summary>

- [line 194](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L194): validate_membership no longer isolates potential side effects behind a rollback-only transaction. The prior implementation decrypted/parsed the welcome inside an `mls_storage` transaction that intentionally rolled back (`IntentionalRollback`), guaranteeing that any accidental or library-internal writes to the key store/persistent state were discarded. The new implementation calls `DecryptedWelcome::from_welcome_proto(welcome, &self.context).await?` directly, outside any rollback guard, while the surrounding comment still states: "Process the welcome without affecting persistent state." If `from_welcome_proto` (or any of its callees) performs writes (e.g., precomputations, provisional key/credential caching, or any side-effect in the MLS provider), these will now persist on success paths that later return an error during validation. That violates the function’s contract, can leave the store in a partially initialized state, and breaks the invariant that validation is read-only and reversible. This regression risks subtle corruption and double-application bugs on reprocessing. <b>[ Low confidence ]</b>
- [line 271](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L271): Potential emission of deferred events for a failed (rolled back) welcome commit. In `commit_or_fail_forever`, the commit logic runs inside a savepoint: `storage.savepoint(|conn| self.commit(conn, events, decrypted_welcome))`. If `commit` returns a non-retryable error, the savepoint rolls back any DB changes, and the outer `transaction` updates the cursor and returns `Ok(CommitResult::FailedForever(err))`. After the transaction returns `Ok(...)` (including the `FailedForever` case), the code unconditionally calls `events.send_all(&self.context)`. If `self.commit` enqueued any events into `events` before returning the error, those events will now be emitted even though the intended state changes were rolled back. This violates the invariant that effects are dispatched only after successful persistence and risks emitting notifications corresponding to uncommitted work. To be correct, events should be queued only after the point-of-no-return (or cleared on error), or the savepoint should isolate and discard any enqueued events on failure. <b>[ Low confidence ]</b>
- [line 305](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L305): Potential incorrect comparison due to casting `current_cursor` to `u64` at `welcome.sequence_id() > current_cursor as u64`. If `current_cursor` can be negative (e.g., sentinel values), casting a negative `i64` to `u64` yields a very large number, causing `requires_processing` to be `false` and erroneously skipping valid welcomes. This results in returning `Err(ProcessIntentError::WelcomeAlreadyProcessed(welcome.cursor).into())` and permanently preventing processing. <b>[ Invalidated by documentation search ]</b>
- [line 315](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L315): The welcome cursor is updated (`db.update_cursor(context.installation_id(), EntityKind::Welcome, welcome.cursor)?;`) before the rest of the processing completes. If subsequent steps fail with a retriable error, the cursor advancement may cause permanent skipping of earlier welcomes. The code even notes this: `// TODO: ... this will permanently skip them.` This violates the effect-invariant: advancing state before ensuring processing success without guaranteed rollback. <b>[ Low confidence ]</b>
- [line 444](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L444): Error handling/contract change for payload encoding: previously errors from `encoded_added_payload.encode(...)` were mapped to `GroupError::EncodeError` via `map_err(GroupError::EncodeError)`, now the code uses `?` directly (`encoded_added_payload.encode(&mut encoded_added_payload_bytes)?;`). This alters the externally visible error variant emitted by this path (e.g., it may now surface `CodecError` or another error type via `From`). This is a contract parity change: callers expecting `GroupError::EncodeError` will not receive it anymore. <b>[ Low confidence ]</b>
- [line 452](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L452): Fallback to an `unknown` content type when `encoded_added_payload.r#type` is `None` constructs a `StoredGroupMessage` with `authority_id: "unknown"`, `type_id: "unknown"`, and version `0.0`. Emitting a published message with an invalid/malformed content type may violate downstream consumers’ grammar/contract and cause failures or silent misinterpretation. <b>[ Low confidence ]</b>
- [line 476](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L476): Potential integer overflow/truncation: `sequence_id` and `originator_id` are cast to `i64` (`sequence_id: welcome.sequence_id() as i64`, `originator_id: welcome.originator_id() as i64`). If these values exceed `i64::MAX`, the cast will wrap/truncate, producing incorrect values persisted into `StoredGroupMessage`. This can break ordering, deduplication, and downstream logic. <b>[ Invalidated by documentation search ]</b>
- [line 502](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L502): Defaulting the application message cursor to `0` when `welcome_metadata` is absent (`let cursor = welcome_metadata.map(|m| m.message_cursor as i64).unwrap_or_default();`) can regress progress by resetting the cursor to a lower value than previously stored. The subsequent `db.update_cursor` may move the cursor backward, causing reprocessing/duplicates and violating monotonicity. <b>[ Low confidence ]</b>
- [line 511](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L511): Under the `d14n` feature, the cursor update uses a hardcoded `originator_id: 100` instead of the actual originator (`Cursor { originator_id: 100, sequence_id: cursor as u64 }`). This violates the cursor’s semantic contract and can break synchronization, ordering, and duplicate detection. The TODO comment acknowledges the issue: Originator must be included in welcome. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/identity.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 714](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/identity.rs#L714): When `include_post_quantum` is true, the code pushes a post-quantum public key extension into the `key_package_extensions` (`build_post_quantum_public_key_extension(&keypair.public)` at line 705). However, the `Capabilities` built at lines 714–729 do not declare support for that post-quantum extension type. In MLS, a LeafNode’s capabilities must announce support for any extensions it uses (in the leaf or key package). Omitting the post-quantum extension type can cause `KeyPackage::build` to fail at runtime (e.g., `KeyPackageNewError`) or cause downstream verification/rejection by peers. The fix is to add the corresponding `ExtensionType` for the post-quantum public key extension to the `extensions` list in `Capabilities` whenever `include_post_quantum` is true. <b>[ Out of scope ]</b>
- [line 1051](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/identity.rs#L1051): Multiple `unwrap()` calls in the `post_quantum_interop` test can panic at runtime if upstream operations fail (e.g., network fetches, message version mismatches via `as_v1().unwrap()`). While acceptable in tests, these introduce brittle behavior and make failures less diagnosable. Prefer explicit `expect()` with messages or handling the `Result/Option` to surface clearer errors. Affected lines include unwrapping results of `rotate_and_upload_key_package`, `find_or_create_dm_by_inbox_id`, `sync_welcomes`, `list_conversations`, `wrapper_encryption`, and `as_v1()`. <b>[ Code style ]</b>
</details>

<details>
<summary>xmtp_mls/src/subscriptions/process_welcome.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 311](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_mls/src/subscriptions/process_welcome.rs#L311): Documentation explicitly states the function "Returns an error if synchronization fails or the group cannot be found," but the implementation returns `Ok(None)` when no group is found (e.g., when `welcomes.process_new_welcome` returns `Ok(None)` or `self.load_from_store(id)` returns `Ok(None)`). This creates a contract mismatch and can cause callers relying on the documented behavior to mis-handle the "group not found" case. To fix, either change the implementation to return `Err` when the group is not found, or update the documentation to reflect that `None` is a successful outcome indicating no group was found. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_proto/src/types/welcome_message.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 72](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_proto/src/types/welcome_message.rs#L72): Asymmetric conversion API for `WelcomeMessageType`: a free function named `from` is provided only for `DecryptedWelcomePointer` that wraps it into `WelcomeMessageType::DecryptedWelcomePointer`, but there are no equivalent conversions for `WelcomeMessageV1` or `WelcomePointer`. For a wrapper/adapter type aggregating related variants, callers expect uniform conversions or a standard trait-based `From`/`TryFrom` implementation. The partial helper creates an inconsistent external contract and can lead to misuse (e.g., developers assuming similar helpers exist for other variants or choosing the wrong path), increasing the chance of runtime errors elsewhere. At minimum, provide symmetric helpers or trait impls for all variants, or remove the standalone `from` function to avoid misleading API surface. <b>[ Code style ]</b>
- [line 152](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_proto/src/types/welcome_message.rs#L152): `DecryptedWelcomePointer::decode` returns a `DecryptedWelcomePointer` without enforcing the documented invariant that `encryption_key`, `data_nonce`, and `welcome_metadata_nonce` lengths MUST match the selected `aead_type`. The function assigns `wp.encryption_key`, `wp.data_nonce`, and `wp.welcome_metadata_nonce` directly without validating their lengths against `aead_type`. This allows malformed inputs (wrong key/nonce sizes) to propagate as "success" and can cause later cryptographic operations to fail unpredictably or panic. The error type already supports `InvalidLength`, so the function should validate sizes and return an error when they don't match. <b>[ Low confidence ]</b>
- [line 165](https://github.com/xmtp/libxmtp/blob/92b906a86c05a2607192bd87537496274c8becb1/xmtp_proto/src/types/welcome_message.rs#L165): `created_ns` is constructed via `chrono::DateTime::from_timestamp_nanos(xmtp_common::rand_i64())`. Using an unconstrained random `i64` for the timestamp means the input can be out of the valid range for `DateTime`. Depending on the Chrono API version and implementation, `from_timestamp_nanos` may panic or otherwise fail at runtime for out-of-range values. Since there are no guards restricting `rand_i64()` to a safe range, this path is reachable and can cause a runtime termination during `generate()`. This violates the requirement that every input reaches a defined terminal state and is particularly unsafe in a generator helper that should not crash on random inputs. <b>[ Invalidated by documentation search ]</b>
</details>

</details>


tylerhawkes:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2569?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2573** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2570** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2569** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2569?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2568** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2568?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=2569).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

xmtp_mls/src/groups/welcome_pointer.rs
```diff
@@ -0,0 +1,74 @@
...
+        }
+        retries += 1;
+        tracing::info!(
+            "Welcome pointer not found, backing off... (attempt {})",
+            retries
+        );
+        if let Some(d) = retry.backoff(retries, time_spent) {
+            xmtp_common::time::sleep(d).await;
+        } else {
+            return Err(super::GroupError::MissingField(
```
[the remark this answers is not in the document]
tylerhawkes: This is fixed in a downstream pr

xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs
```diff
@@ -1,12 +1,18 @@
...
 use openmls::ciphersuite::hpke::Error as OpenmlsHpkeError;
 use openmls::prelude::tls_codec::Error as TlsCodecError;
+use openmls_traits::crypto::OpenMlsCrypto;
 use openmls_traits::types::HpkeCiphertext;
 use thiserror::Error;
 use tls_codec::{Deserialize, Serialize};
 use xmtp_common::RetryableError;
 use xmtp_configuration::WELCOME_HPKE_LABEL;
 
+static LIBCRUX_CRYPTO_PROVIDER: std::sync::LazyLock<openmls_libcrux_crypto::CryptoProvider> =
```
tylerhawkes, line 11: Need to ensure that this is ok with the underlying rng.

xmtp_mls/src/groups/mls_ext/welcome_wrapper.rs
```diff
@@ -132,10 +140,56 @@ pub fn unwrap_welcome(
...
+        aead_type,
+        symmetric_key,
+        data[1],
+        nonces[1],
+        &[],
+    )?;
+    Ok([data_0, data_1])
+}
+
 #[cfg(test)]
```
tylerhawkes, line 165: These can be simplified now that we're not passing nonces as a Vec.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -2422,23 +2423,105 @@ where
...
+                {
+                    Ok(WelcomeMessageInput {
+                        version: Some(WelcomeMessageInputVersion::WelcomePointer(
+                            WelcomePointerInput {
+                                destination: installation.installation_key,
+                                welcome_pointer: wrap_welcome(
+                                    welcome_pointer,
+                                    &[],
+                                    &installation.hpke_public_key,
+                                    WrapperAlgorithm::XWingMLKEM768Draft6,
```
tylerhawkes: change this to use the installations wrapper algorithm because it will always be XWing

xmtp_mls/src/groups/welcome_pointer.rs
```diff
@@ -0,0 +1,93 @@
...
+            .map_err(|e| xmtp_api::ApiError::Api(Box::new(e)))?
+            .messages
+            .into_iter()
+            .next()
+        {
+            break first;
+        }
+        retries += 1;
+        tracing::info!(
+            "Welcome pointer not found, backing off... (attempt {})",
```
tylerhawkes: Change this to welcome pointee

xmtp_mls/src/groups/welcome_pointer.rs
```diff
@@ -0,0 +1,93 @@
...
+            xmtp_common::time::sleep(d).await;
+        } else {
+            return Err(xmtp_proto::ConversionError::InvalidValue {
+                item: "WelcomeMessage",
+                expected: "WelcomeMessage from Node",
+                got: "None".into(),
+            }
+            .into());
+        }
+        tracing::debug!("welcome pointer not found, retrying...");
```
tylerhawkes: Also change to welcome pointee

xmtp_mls/src/groups/welcome_pointer.rs
```diff
@@ -0,0 +1,93 @@
...
+    };
+    let welcome = match welcome.version {
+        Some(Version::V1(welcome)) => welcome,
+        Some(Version::WelcomePointer(_)) => {
```
tylerhawkes: Add comment about these being non-retryable and incrementing the cursor

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
```diff
@@ -35,23 +37,25 @@ impl DecryptedWelcome {
...
         let hash_ref = find_key_package_hash_ref(provider, hpke_public_key)?;
         let private_key = find_private_key(provider, &hash_ref, &wrapper_ciphersuite)?;
 
-        let (welcome_bytes, welcome_metadata_bytes) = unwrap_welcome(
-            encrypted_welcome_bytes,
-            encrypted_welcome_metadata_bytes,
-            &private_key,
-            wrapper_ciphersuite,
-        )?;
+        let (welcome_bytes, welcome_metadata_bytes) =
```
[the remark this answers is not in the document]
tylerhawkes: I don't know what you're on, but this is rust with the tracing macros and that's not how they work. The code as is works as expected.

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
```diff
@@ -64,10 +74,72 @@ impl DecryptedWelcome {
...
+        context: &impl crate::context::XmtpSharedContext,
+    ) -> Result<(openmls::messages::Welcome, Option<WelcomeMetadata>), GroupError> {
+        let v1 = super::super::welcome_pointer::resolve_welcome_pointer(
+            decrypted_welcome_pointer,
+            context,
+        )
+        .await?;
+
+        let aead_type = match decrypted_welcome_pointer.aead_type {
+            xmtp_proto::xmtp::mls::message_contents::WelcomePointeeEncryptionAeadType::Chacha20Poly1305 => {
```
mchenani, line 90: Nit: I know why the original aead_type enum is named `WelcomePointeeEncryptionAeadType`, but I feel it's a little bit confusing since we use it in both places, for pointer and pointee, can you think of a more general name? something like `WelcomeEncryptionAeadType`?

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
```diff
@@ -64,10 +74,72 @@ impl DecryptedWelcome {
...
+            }
+            WelcomeMessageType::DecryptedWelcomePointer(w) => {
+                Self::welcome_from_decrypted_welcome_pointer(w, context).await?
+            }
+        };
 
         let join_config = build_group_join_config();
 
-        let builder = StagedWelcome::build_from_welcome(provider, &join_config, welcome.clone())?;
+        let provider = XmtpOpenMlsProviderRef::new(provider);
```
mchenani: nit: I think having another name for this filed would be less confusing, or just move the initialization in line in the next line when we create the builder. wdyt?

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
tylerhawkes: It's only defined on the pointer and while we use it to decrypt the pointee, it still makes sense in that context too.

xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs
tylerhawkes: The builder ends up borrowing from the provider. The parameter is named provider, so I think it's fine. It's only needed to get the builder working on a few lines.

xmtp_api_d14n/src/protocol/extractors/welcomes.rs
```diff
@@ -92,17 +102,45 @@ impl EnvelopeVisitor<'_> for V3WelcomeMessageExtractor {
...
+        &mut self,
+        message: &V3ProtoWelcomePointer,
+    ) -> Result<(), Self::Error> {
+        let originator_node_id = xmtp_configuration::Originators::WELCOME_MESSAGES;
+        self.welcome_message
+            .cursor(Cursor {
+                originator_id: originator_node_id,
+                sequence_id: message.id,
+            })
+            .created_ns(DateTime::from_timestamp_nanos(message.created_ns as i64))
```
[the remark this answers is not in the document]
tylerhawkes: we can worry about that in 200 years and still have at least 30 years to fix the issue.

xmtp_api_d14n/src/protocol/traits/visitor.rs
```diff
@@ -103,6 +116,11 @@ pub trait EnvelopeVisitor<'env> {
         Ok(())
     }
 
+    fn visit_v3_welcome_pointer(&mut self, _m: &V3WelcomePointer) -> Result<(), Self::Error> {
+        tracing::trace!("noop_visit_v3_welcome_pointer");
+        panic!()
```
insipx: i'm guessing this was for testing -- should remove/error

tylerhawkes: Yeah, this was for testing, however, I'm kind of wondering if this should be the default so that we know that we're never calling methods that aren't implemented.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -2400,23 +2399,115 @@ where
...
+            .iter()
+            .filter(|installation| {
+                installation
+                    .welcome_pointee_encryption_aead_types
+                    .compatible()
+            })
+            .count();
+
+        let (welcome_pointer_bytes, welcome_pointee) = if wp_capable
+            > xmtp_configuration::INSTALLATION_THRESHOLD_FOR_WELCOME_POINTER_SENDING
```
insipx, line 2420: now that the extractors are included -- do welcome pointers work with d14n or do we need to think about a switch on Client here?

xmtp_api_d14n/src/protocol/traits/visitor.rs
insipx: it would be nice to have something to indicate if we missed an implementation, but it would force all implementers to implement each function in a visitor, most of which might be empty. maybe a macro or something similar could make that easier for us

xmtp_mls/src/groups/mls_sync.rs
tylerhawkes: They should work with d14n. They pass all the tests when the `d14n` feature is enabled.

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -2400,23 +2399,115 @@ where
...
+        {
+            let destination = xmtp_common::rand_array::<32>();
+            tracing::debug!(
+                wp_capable,
+                destination = %hex::encode(destination),
+                "Using welcome pointers"
+            );
+            let symmetric_key = xmtp_common::rand_array::<32>();
+            let data_nonce = xmtp_common::rand_array::<12>();
+            let mut welcome_metadata_nonce = xmtp_common::rand_array::<12>();
```
tylerhawkes: Should use zeroize for these and the welcome_pointer_bytes

PR #2600, opened 2025-10-15
Make code cov informational and not blocking PRs for now

cameronvoell:
Code coverage setup is relatively new, and may not be configured to ignore all benign cases and therefore we do not want to block PRs in many cases for now. Will set the check to informational, which can aid reviewers in identifying important missed coverage.  

See config option here: https://docs.codecov.com/docs/commit-status#informational

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,6 +3,9 @@ coverage:
     project:
       default:
         informational: true
+    patch:
+      default:
+        informational: true
 ignore:
   - "xmtp_proto/src/gen"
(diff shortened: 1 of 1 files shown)


tylerhawkes (approved):

neekolas:
Was this actually blocking before? Graphite shows it as a non-blocking status check. Anyways, I think that's right.

cameronvoell:
Guess GH rules make the check not blocking already, but this codecov setting seems correct for making it easier to see when a PR IS blocked by a status check.

PR #2605, opened 2025-10-16
Self Removal: add leave group

### LINKED ISSUE #2335: Self Removal work items
Decision:
Draft Published internally
Publish Draft for public - WIP

Tasks:
* #2336: the client send the leave request by adding their inbox id in the group pending_remove / pending_leave metadata. 
* #2337: the main client mark the group as pending_leave and filter-out the group from conversations list, by adding a new filter to the group_query -- group table migration needed
* #2338 : on other installations of the user, they must mark the group as pending_remove as the main client.
* #2339 : upon receiving the main client request, the admin client must remove the client from the group via a worker
* #2340 after the user successfully removed from the group, the admin needs to remove that user's inbox_id from the pending_removal metadata.
* #2341  after the admin removed a member that was a response to an inbox_id that in pending_removal metadata, we need to put a new transcript message as `X left the group`

mchenani:
closes #2335 

### Add group self-removal by introducing `Group::leave_group` and periodic `PendingSelfRemove` worker, and expose `PendingRemove` membership state across bindings and storage
This change adds a leave request flow and background processing for self-removal, classifies left vs removed members in updates, and persists pending-remove state across storage and bindings.

- Add `Group::leave_group` with validation and leave-request message handling in [mod.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3)
- Run periodic `PendingSelfRemove` background worker via `ClientBuilder.build` to process pending leaves in [builder.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-2df8df9acb5405ca5c2dde2487e92ffc9a7daf6be38dd075a2a2d4878cc35ce6) and [pending_self_remove_worker.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-43bcc9597a35d2a8e7e3c031e1407266941aff569a722c56ca2713f1318eab49)
- Distinguish `left_inboxes` from `removed_inboxes` using `ValidatedCommit::into_with` in [validated_commit.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-779227110741fa6ce78048adfaac1895d306d8fd502dab800774b62aada72365)
- Persist and query pending-remove entries and group `has_pending_leave_request` in DB schemas and traits in [pending_remove.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-04cee635dfe69106f967f5a06e09d0a55a15d8c31d0c2860ab86a29cfe86f1b7) and [group.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-e3d4940853a1045db514d5c1a48581d9ef53da4ba45a7cb8387c72c99c475b5a)
- Process leave-request content in sync and external paths and update pending-remove bookkeeping in [mls_sync.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e)
- Expose `PendingRemove` membership state and `LeaveRequest` content type in bindings and content types

#### 📍Where to Start
Start with the leave flow entry points: review `Group::leave_group` and related helpers in [mod.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3), then the periodic processing in `PendingSelfRemoveWorker.run` in [pending_self_remove_worker.rs](https://github.com/xmtp/libxmtp/pull/2605/files#diff-43bcc9597a35d2a8e7e3c031e1407266941aff569a722c56ca2713f1318eab49).

#### Changes since #2605 opened

- Reformatted worker registration method chains in `ClientBuilder.build` async method [43ec37b]
- Cleaned up duplicate imports in `xmtp_db` encrypted store wasm database module [43ec37b]
- Reduced timeout values for WebAssembly test environment variables [60dd5be]
- Removed logger initialization from test method [f9ade53]
- Modified super-admin status detection in admin pending remove actions to use OpenMlsGroup metadata directly [7f66ae6]
- Renamed method and updated call site in groups module [77d83e8]
- Updated documentation and removed unused import in encrypted store database module [77d83e8]
----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 77d83e8. 19 files reviewed, 37 issues evaluated, 35 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>bindings_node/src/content_types/mod.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 20](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/bindings_node/src/content_types/mod.rs#L20): Adding the `LeaveRequest` variant to the `#[napi]`-exposed Rust enum `ContentType` without explicit discriminant values changes the numeric discriminants of all subsequent variants at runtime. With `#[napi]` (without `string_enum`), this enum is exposed to Node.js as a numeric enum, so clients or persisted data relying on specific numeric values for variants like `GroupMembershipChange`, `GroupUpdated`, etc., will now see different numbers. This breaks ABI/contract parity and can cause misinterpretation of values on the JavaScript side or anywhere the numeric representation is used. To maintain stability, either assign explicit discriminants (e.g., `Unknown = 0, Text = 1, ...`) or use `#[napi(string_enum)]` to expose stable string variants. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_content_types/src/leave_request.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 45](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_content_types/src/leave_request.rs#L45): `decode` silently ignores `content.parameters` without validation. If upstream or future versions include parameters that materially affect semantics, dropping them without an explicit check or error leads to silent data loss or misinterpretation. At minimum, the codec should validate that parameters are empty (or known and expected) for this content type and return `InvalidContentType` or a specific error if unexpected parameters are present. <b>[ Low confidence ]</b>
- [line 46](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_content_types/src/leave_request.rs#L46): `decode` ignores `content.compression` and attempts to decode `content.content` bytes directly. If the `EncodedContent` is compressed (e.g., gzip), this will fail with a `Decode` error or, worse, misinterpret the payload. The codec should either reject compressed input with a clear error or handle decompression before `prost::Message::decode`. <b>[ Low confidence ]</b>
- [line 46](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_content_types/src/leave_request.rs#L46): `decode` does not validate `content.r#type` for presence or that it matches `LeaveRequestCodec::content_type()` (authority, type_id, and version). This allows decoding arbitrary bytes even when the content type is missing or mismatched, violating contract parity and failing to use the existing `CodecError::InvalidContentType`. If the bytes happen to decode, the codec would silently accept the wrong type; if they don't, it returns a generic `Decode` error rather than a clear invalid type error. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_content_types/src/lib.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 75](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_content_types/src/lib.rs#L75): The `ContentType::DeviceSyncMessage` variant (line 48) cannot be produced via `TryFrom<&str>` because `fn try_from` (lines 55–77) has no match arm mapping any device sync message type ID to `ContentType::DeviceSyncMessage`. Any device-sync type ID string will fall through to the default case and return `Err("Unknown content type ID: ...")` (line 75), causing valid inputs to be rejected and making the `DeviceSyncMessage` variant effectively unreachable through this conversion path. This creates an interface/contract asymmetry: `ContentType` advertises the `DeviceSyncMessage` variant but the canonical string-to-enum conversion omits it. If device-sync content is possible upstream, this is a runtime failure path. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/group.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 110](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_db/src/encrypted_store/group.rs#L110): Semantic contradiction between the field name `has_pending_leave_request` and its documentation. The docstring says "Whether the pending-remove list is empty", which implies `true` means the list is empty. However, the field name `has_pending_leave_request` implies `true` means the list is non-empty (i.e., there exists a pending leave request). With `Option<bool>` this inversion can lead to silently inverted logic across the codebase, producing incorrect runtime behavior (e.g., treating a non-empty list as empty, or vice versa). The second doc line ("NULL if the pending-remove didn't receive an update yet") matches the `Option<bool>` encoding but does not resolve the inversion in the primary meaning. This is an implementation-vs-docs contradiction that creates uncertainty about which interpretation is correct and is likely to cause real runtime misbehavior wherever the value is read or written. <b>[ Low confidence ]</b>
- [line 1191](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_db/src/encrypted_store/group.rs#L1191): Updating `dsl::groups.find(group_id)` does not check the number of affected rows and always returns `Ok(())`, silently succeeding even when no row exists for the provided `group_id`. `diesel::update(...).execute(conn)` returns `Ok(0)` if no row matches, but the code neither maps this case to `StorageError::NotFound` nor reports any error. This creates a runtime bug where a missing group is indistinguishable from a successful update. Given the presence of `StorageError::NotFound`, callers likely expect a `NotFound` error when the target group does not exist. You should check the returned `usize` from `execute` and return `StorageError::NotFound` (or a suitable error) when it is zero. <b>[ Low confidence ]</b>
- [line 1225](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_db/src/encrypted_store/group.rs#L1225): New enum variant `GroupMembershipState::PendingRemove = 5` introduces a new externally visible state without any accompanying guarantees that all consumers handle it. Given `#[repr(i32)]` and Diesel integration, records may persist with value `5`. If any reading/deserializing or business logic elsewhere assumes only the previous set of variants or performs range checks/conversions, it may fail at runtime (e.g., deserialization errors) or mis-handle the state (e.g., fall back to default behavior). This is a contract-parity issue: adding a new variant requires auditing all match statements, persistence/serialization paths, and state-machine logic to ensure the new path is handled consistently. Without that, reachable inputs containing `PendingRemove` can cause runtime errors or incorrect behavior. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/builder.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 302](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/builder.rs#L302): `WorkerRunner` is cloned into the context (`context.workers = workers.clone()`) before workers are registered. Per the inline comment on `WorkerRunner`, this leads to the context and client having different views. Specifically, the `factories` inside the clone stored in the context remain empty while the main `workers` instance later registers multiple workers and then spawns them. Any code that uses `context.workers` to inspect registered workers or perform interactions that depend on the registered factories may observe an empty set, leading to runtime inconsistency and potential functional failures. To preserve a single coherent view, either clone after registration/spawn or refactor `WorkerRunner` to share its registry state (e.g., wrap `factories` in an `Arc<...>`). <b>[ Out of scope ]</b>
- [line 328](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/builder.rs#L328): PendingSelfRemoveWorker is registered twice, causing duplicate background processing and potential double-application of its effects. Inside `build`, the worker is registered once at lines 328–330. The diff also adds an identical `register_new_worker::<PendingSelfRemoveWorker<...>>` call, which would result in two registrations if both are in effect. Duplicate registration can produce duplicated task handling, race conditions, and violates the at-most-once and no-double-application guarantees for background jobs. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/error.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 246](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/error.rs#L246): `is_retryable` returns `false` unconditionally and does not inspect `self` (the error instance). If the enclosing error type has variants that represent transient or recoverable conditions (e.g., network timeouts, temporary unavailability, concurrent modification races), this implementation will suppress all retries, leading to avoidable failures or degraded behavior. The presence of a method named `is_retryable` strongly suggests that different errors should have different retryability semantics. Returning a constant `false` violates that expectation and may break callers that rely on this method to decide on retry logic. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_sync.rs — 0 comments posted, 19 evaluated, 18 filtered</summary>

- [line 801](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L801): Potentially unsafe casts of timestamps from signed to unsigned can produce incorrect values. In `process_own_message`, `envelope_timestamp_ns` (likely signed) is cast to `u64` when passed to `save_transcript_message` and `set_delivery_status_to_published`: `envelope_timestamp_ns as u64`. If `envelope_timestamp_ns` is negative (e.g., from a malformed or legacy envelope), this will wrap to a very large `u64` value, causing misordered timestamps, incorrect expiration times, or other downstream inconsistencies. There are no guards ensuring non-negative values before casting. <b>[ Previously rejected ]</b>
- [line 975](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L975): When `allow_cursor_increment` is false, the code determines whether processing is required by fetching the current cursor via `db.get_last_cursor_for_originator(...)` and comparing sequence IDs (lines 975-978). If there is no existing cursor for the given group/originator (e.g., first message), depending on the DB implementation, `get_last_cursor_for_originator` may return an error (e.g., NotFound) rather than a default cursor. The `?` will propagate that error, aborting the transaction and the whole message processing, even though the correct behavior should be to treat the message as requiring processing. This turns a normal “first message” condition into an error path. To fix, handle the ‘missing cursor’ case explicitly by interpreting it as no prior cursor (thus `requires_processing = true`) instead of propagating an error. <b>[ Low confidence ]</b>
- [line 1038](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1038): Risky integer cast: `envelope_timestamp_ns as u64` is used in multiple places (e.g., passing to `extract_message_sender` and `save_transcript_message`). If `message_envelope.timestamp()` can return a negative value (e.g., bad input or clock skew), the `as u64` cast will wrap to a huge positive number, leading to incorrect timing-related behavior without an error. Add validation to ensure timestamp is non-negative (or use a checked conversion) before casting, and handle invalid timestamps with a defined error path. <b>[ Out of scope ]</b>
- [line 1108](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1108): Non-atomic effects on error path: In the `Content::V1` branch, the message is stored (`message.store_or_ignore(&storage.db())?`) and `identifier.internal_id(...)` is set before calling `self.process_leave_request_message(mls_group, storage, &message)?` when `message.content_type == ContentType::LeaveRequest` (lines 1108–1110). If `process_leave_request_message` returns an error, the function exits with `Err(...)` after having already persisted the message and possibly queued events. This yields partially applied state and may cause inconsistent behavior (e.g., the message appears stored even though handling failed, leading to mismatch with retry logic or downstream invariants). Consider performing leave-request handling in the same transaction as the message store, or deferring persistence until after successful processing, or adding explicit rollback/compensation on failure. <b>[ Previously rejected ]</b>
- [line 1229](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1229): Potential runtime panic: `validated_commit.expect("Needs to be present when this is a staged commit")` will terminate the process if `validated_commit` is `None` when handling `ProcessedMessageContent::StagedCommitMessage`. There is no visible guard ensuring `validated_commit` is `Some` for this branch, and this function is not in an initialization path. A `None` value is reachable via the function signature `validated_commit: Option<ValidatedCommit>`. Replace the `expect` with an explicit error return (e.g., `return Err(GroupMessageProcessingError::IntentMissingStagedCommit)` or a new error variant) to avoid crashing. <b>[ Low confidence ]</b>
- [line 1319](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1319): Missing escalation or retry when the storage read for the message fails or returns `None` in `process_own_leave_request_message`. The code only processes the leave request if `get_group_message(message_id)` returns `Ok(Some(message))` and the content type matches. If the storage returns `Err(_)` or `Ok(None)`, the function silently does nothing. Given this function is invoked immediately after setting the delivery status to published for the message, a `None` result could indicate a race or storage inconsistency. Without a retry or error propagation, the leave request may never be processed. <b>[ Low confidence ]</b>
- [line 1322](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1322): Errors from `process_leave_request_message` are swallowed in `process_own_leave_request_message`, potentially leaving state inconsistent without any escalation or retry. The function logs at `debug` level on failure and returns `()`, and the caller (`process_own_message`) proceeds to return `Ok(Some(id))` after marking the message as published. If processing a `LeaveRequest` is required to maintain correctness (e.g., removing local membership state, triggering cleanup), silently ignoring failures means the intent will be considered successfully processed while the leave operation may not have been applied locally. This can result in divergence between MLS group state and local storage. <b>[ Already posted ]</b>
- [line 1344](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1344): In `process_leave_request_message`, `update_group_membership` is invoked with `&self.group_id` while the `PendingRemove` record uses `message.group_id.clone()`. If this function can be called with a message whose `group_id` does not match `self.group_id` (e.g., due to upstream routing bugs or cross-group messages), the code would update membership for one group while recording a pending removal for another, leading to inconsistent state across storage. Consider asserting or checking that `message.group_id == self.group_id` before performing effects. <b>[ Low confidence ]</b>
- [line 1377](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1377): In `process_admin_pending_remove_actions`, errors from `is_super_admin_without_lock` are swallowed and the function returns `Ok(())`. This means transient or resolvable errors in checking super-admin status lead to silently skipping admin pending remove actions and potentially leaving the group in a stale state (e.g., not marking the group pending when it should). Consider propagating the error or recording a failure state so that the condition can be retried rather than silently ignored. <b>[ Low confidence ]</b>
- [line 1398](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1398): Calls to `update_group_pending_status` in both `process_admin_pending_remove_actions` and `handle_super_admin_status_change` do not handle or propagate a potential error. If `update_group_pending_status` returns a `Result` (which is plausible for a storage-updating method), those errors would be silently ignored, causing the group’s pending status to not reflect intended changes. Ensure the return value is checked and errors are propagated or logged and retried as appropriate. <b>[ Low confidence ]</b>
- [line 1465](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1465): In `handle_super_admin_status_change`, simultaneous promotion and demotion of the current user are not handled correctly. The code uses `if was_promoted { ... } else if was_demoted { ... }`, which means if both `was_promoted` and `was_demoted` are `true` for the same metadata update (e.g., due to contradictory or duplicate operations), only the promotion branch will execute and the demotion branch will be skipped. This can leave the group pending status set to `true` even though the user was also demoted from super-admin in the same change, resulting in incorrect state. Consider explicitly handling the case where both conditions are `true` (e.g., prefer demotion to clear pending status, or resolve by validation). <b>[ Low confidence ]</b>
- [line 1506](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1506): Errors returned by `storage.db().set_group_has_pending_leave_request_status(...)` are only logged and not propagated or surfaced to the caller in any way. Given the function’s side-effecting contract (update persistent group status), silent failure can leave the system in an inconsistent state where the in-memory notion and persistent store diverge, with no way for the caller to detect and react (e.g., retry, escalate). Consider returning a `Result<(), E>` or otherwise signaling failure to preserve invariants and enable callers to handle errors. <b>[ Low confidence ]</b>
- [line 1510](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1510): Inconsistent error logging across branches causes asymmetric structured fields in telemetry. In the `has_pending_removes == true` branch, the error is captured as a structured field (`error = %e`) and includes `inbox_id`, while in the `else` branch the error is passed as a positional format argument (`"... {}", e`) and omits the `error` field and `inbox_id`. This inconsistency can break log parsing/alerting pipelines that rely on a consistent schema for error events (e.g., expecting an `error` field), and loses valuable context for troubleshooting. Align both branches to use the same structured fields (e.g., `error = %e`, include `inbox_id`) and avoid mixing positional formatting with structured fields. <b>[ Low confidence ]</b>
- [line 1870](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L1870): `track_err!` macro usage at `track_err!("Process message", &process_result, group: &self.group_id);` contradicts the macro's documented contract. The macro documentation states it "returns the original result unchanged" to be transparent to control flow, but the implementation does not return the original `$result` and instead returns `()` (unit). In this function it's used as a standalone statement, so it does not break control flow here; however, the contradiction creates uncertainty and can lead to subtle runtime bugs if future maintainers (or other call sites) rely on the documented return value for chaining. This is a contract mismatch between documentation and implementation that can affect runtime behavior at call sites. <b>[ Low confidence ]</b>
- [line 2073](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L2073): Message ID collision risk due to using `timestamp_ns.to_string()` as the idempotency key in `calculate_message_id`. The function’s contract says it "Relies on a client-created idempotency_key (which could be a timestamp)", but here the server-side save uses the provided `timestamp_ns`. If multiple distinct transcript messages have identical payload bytes and identical `timestamp_ns` (e.g., same ns granularity or retries), their computed `message_id` will collide, causing `store_or_ignore` to ignore a legitimate new message. Consider using a stronger idempotency key (e.g., a UUID or cursor-based unique value) or including additional unique entropy. <b>[ Code style ]</b>
- [line 2089](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L2089): Non-atomic side effects: `handle_metadata_update_from_commit` mutates group metadata before the transcript message is persisted. If `msg.store_or_ignore(&storage.db())?` fails, the function returns an error with metadata changes already applied, leaving storage in an inconsistent state (updated metadata without a matching stored transcript message). This violates atomicity and can produce divergent state between metadata and message history. Consider wrapping both operations in a single transaction or applying metadata changes only after the message is stored, or provide explicit rollback on failure. <b>[ Previously rejected ]</b>
- [line 2094](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L2094): Potential integer overflow/wrap when casting `u64` to `i64` for `sent_at_ns`, `sequence_id`, and `originator_id`. The code uses `timestamp_ns as i64`, `cursor.sequence_id as i64`, and `cursor.originator_id as i64`. If these values exceed `i64::MAX`, the cast will wrap and produce negative values, corrupting stored data and violating invariants for timestamps and identifiers. Use checked conversion (e.g., `try_into()` with error handling) or store as `i128`/`u64` in the database schema. <b>[ Previously rejected ]</b>
- [line 2214](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mls_sync.rs#L2214): Non-atomic state transition: the code marks the intent as `published` in the local DB before the external side effect of actually sending the message succeeds. Specifically, `set_group_intent_published(...)` is executed inside a transaction (lines 2214–2224), and only after that `api().send_group_messages(messages).await?` is called (lines 2239–2241). If the API call fails, the intent remains in a `published` state with its hash recorded, but the message was never sent. There is no compensating action or rollback and the code will not retry this intent because it is already marked as published, creating a reachable path to message loss (missed publish) and violating at-most-once/at-least-once delivery expectations and invariants for intent lifecycle. This breaks the required invariant that after any effect, all exit paths preserve consistency: DB state indicates published, but the external world has not observed it. The log comments recognize missed messages if the transaction is removed, but do not address this atomicity gap between DB and network publish. A fix would be to either (a) defer marking as published until after a successful send; (b) mark an intermediate state (e.g., `publishing`) and only transition to `published` after the send succeeds; or (c) make the send idempotent and retryable and retry on API errors while ensuring the intent is not moved to a terminal state prematurely. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mod.rs — 0 comments posted, 6 evaluated, 6 filtered</summary>

- [line 1017](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mod.rs#L1017): Permission check contradicts the documented behavior: the docstring states “Only admins and super admins can call this function,” but the implementation only permits super admins. Specifically, `is_super_admin` is checked and non-super admins are allowed to continue with `Ok(())` while the log message claims the caller is not in admin or super admin. This mismatch between documentation/logging and code can lead to silently skipping legitimate admin requests and inconsistent behavior for callers who are admins but not super admins. <b>[ Low confidence ]</b>
- [line 1033](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mod.rs#L1033): Duplicate inbox IDs in the pending removal list are not deduplicated before removal. `valid_removals` is built directly from `pending_removal_list` and may include duplicates, leading to multiple attempts to remove the same member in a single batch. This can violate at-most-once behavior expectations and could cause errors in downstream membership update logic if it expects unique removals. <b>[ Low confidence ]</b>
- [line 1033](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mod.rs#L1033): Potential self-removal: the function does not exclude the current caller’s inbox ID from `valid_removals`. If the pending removal list includes the current admin/super admin, the code will attempt to remove them as well, which can lead to loss of required permissions mid-operation or unexpected state changes. A guard should prevent removing the actor performing the administrative action unless explicitly allowed and safe. <b>[ Low confidence ]</b>
- [line 1069](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mod.rs#L1069): Pending removal list is not cleared or updated after successful removals, which may leave stale entries and cause repeated attempts to remove already-processed members. After a successful call to `remove_members_by_inbox_id`, the function simply returns `Ok(())` without purging processed inbox IDs from the pending list or marking completion. If other parts of the system rely on this list as a work queue, it can lead to repeated work, confusing logs, or future errors. <b>[ Low confidence ]</b>
- [line 1131](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mod.rs#L1131): Possible duplicate deletions due to not enforcing uniqueness in `removed_members`. The code constructs `removed_members` by filtering `pending_removal_list` without deduplicating entries: `let removed_members: Vec<String> = pending_removal_list.iter().filter(|pending_user| !current_member_ids.contains(pending_user)).cloned().collect();`. If `pending_removal_list` contains duplicate user IDs, this will pass duplicates to `delete_pending_remove_users`, potentially causing repeated deletion attempts, errors from the backing store, or double-application of side effects. To ensure correctness under all inputs, `removed_members` should be de-duplicated (e.g., via a `HashSet`) before calling `delete_pending_remove_users`. <b>[ Low confidence ]</b>
- [line 1199](https://github.com/xmtp/libxmtp/blob/77d83e87017eae6671c4359cf7941c596975f6b1/xmtp_mls/src/groups/mod.rs#L1199): Race condition leading to multiple leave requests being sent. In `leave_group`, the code checks `!self.is_in_pending_remove(self.context.inbox_id())?` and, if true, sends a leave request via `self.send_message(...).await?`. There is no atomic guard or state update before sending. If `leave_group` is invoked concurrently (e.g., by multiple UI interactions or retries), both calls can observe that the user is not in the pending-remove list and both proceed to send the leave request, violating at-most-once semantics and causing duplicate requests. This should be guarded by an atomic update (e.g., setting a pending flag transactionally before sending, or holding a lock) or making the operation idempotent at the storage level. <b>[ Previously rejected ]</b>
</details>

</details>

### DIFF
--- a/xmtp_db/src/encrypted_store/database/wasm.rs
+++ b/xmtp_db/src/encrypted_store/database/wasm.rs
@@ -1,11 +1,10 @@
 //! WebAssembly specific connection for a SQLite Database
-//! Stores a single connection behind a mutex that's used for every libxmtp operation
+//! Stores a single connection behind a RefCell that's used for every libxmtp operation
 use crate::DbConnection;
 use crate::PersistentOrMem;
 use crate::{ConnectionExt, StorageOption, XmtpDb};
 use diesel::prelude::SqliteConnection;
 use diesel::{connection::SimpleConnection, prelude::*};
-use parking_lot::Mutex;
 use sqlite_wasm_rs::sahpool_vfs::OpfsSAHPoolCfg;
 use std::cell::RefCell;
 use std::rc::Rc;
--- a/xmtp_mls/src/groups/mls_sync.rs
+++ b/xmtp_mls/src/groups/mls_sync.rs
@@ -1371,7 +1371,7 @@ where
         // Process admin actions based on current group state
         // If the current user is super-admin and there are pending remove requests, mark the group accordingly
         let is_super_admin = match self
-            .is_super_admin_from_metadata(mls_group, self.context.inbox_id().to_string())
+            .is_super_admin_without_lock(mls_group, self.context.inbox_id().to_string())
         {
             Ok(is_admin) => is_admin,
             Err(e) => {
--- a/xmtp_mls/src/groups/mod.rs
+++ b/xmtp_mls/src/groups/mod.rs
@@ -1620,7 +1620,7 @@ where
     }
 
     /// Checks if the given inbox ID is a super admin of the group at the most recently synced epoch
-    pub fn is_super_admin_from_metadata(
+    pub fn is_super_admin_without_lock(
         &self,
         mls_group: &OpenMlsGroup,
         inbox_id: String,


mchenani:
* **#2611** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2610** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2608** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2608?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2606** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2606?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2605** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2605?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2605?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=2605).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

cameronvoell:
adding a reminder here to revisit the worker interval duration

https://github.com/xmtp/libxmtp/pull/2610#discussion_r2445230129

cameronvoell (approved):
looking good now, thanks @mchenani !

PR #2631, opened 2025-10-17
Streaming benchmark

codabrink:
### Add streaming benchmark CLI and gate worker spawning via `xmtp_mls::ClientBuilder::with_disable_workers` to support streaming benchmark
Introduce a `benches` workspace member with a `streaming` CLI for message send/receive benchmarking in [streaming.rs](https://github.com/xmtp/libxmtp/pull/2631/files#diff-3b1a2906ffa276ff948acbb8d6ed8139f583ae54e895aa1666d184da96c84f6b), and add a `disable_workers` flag to `xmtp_mls::ClientBuilder` to skip background workers. Switch Diesel crates to a git branch and add serde for `StoredIdentity`.

- Add `benches` crate and `streaming` binary with RLIMIT NOFILE set to `4096` in [streaming.rs](https://github.com/xmtp/libxmtp/pull/2631/files#diff-3b1a2906ffa276ff948acbb8d6ed8139f583ae54e895aa1666d184da96c84f6b)
- Add `disable_workers` to `xmtp_mls::ClientBuilder` and propagate through builder methods in [builder.rs](https://github.com/xmtp/libxmtp/pull/2631/files#diff-2df8df9acb5405ca5c2dde2487e92ffc9a7daf6be38dd075a2a2d4878cc35ce6)
- Update test utilities to `TesterBuilder::new()` and expand configuration, including snapshots and dev API, in [tester_utils.rs](https://github.com/xmtp/libxmtp/pull/2631/files#diff-727ea71bc8afdfdf6f0b1d9813b2e6e46f6fdd647abbc9b56be80bd873034c7d)
- Derive `serde::Serialize` and `serde::Deserialize` for `xmtp_db::StoredIdentity` in [identity.rs](https://github.com/xmtp/libxmtp/pull/2631/files#diff-0eadc16eb390b4b83daba33d86321f37822a654963113c8fd5eff1f2cd4cbd36)
- Fetch Diesel crates from `ephemerahq/diesel` git branch `codabrink/writable-deserialize` in [Cargo.toml](https://github.com/xmtp/libxmtp/pull/2631/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542)

#### 📍Where to Start
Start with the CLI flow in `main` within [streaming.rs](https://github.com/xmtp/libxmtp/pull/2631/files#diff-3b1a2906ffa276ff948acbb8d6ed8139f583ae54e895aa1666d184da96c84f6b), then review worker gating in `xmtp_mls::ClientBuilder::build` in [builder.rs](https://github.com/xmtp/libxmtp/pull/2631/files#diff-2df8df9acb5405ca5c2dde2487e92ffc9a7daf6be38dd075a2a2d4878cc35ce6).

#### Changes since #2631 opened

- Removed benchmark identity file containing credential data [cd1b9b9]
- Refactored test database creation infrastructure in tester utilities [a870afb]
- Standardized indicatif dependency management across workspace [a870afb]
- Updated diesel dependency configuration in root manifest [a870afb]
----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 347909d. 8 files reviewed, 24 issues evaluated, 23 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>benches/src/streaming.rs — 0 comments posted, 6 evaluated, 5 filtered</summary>

- [line 109](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/benches/src/streaming.rs#L109): Potential integer overflow and silent data loss when reporting: `let sent = app.ctx.msg_tx.load(Ordering::SeqCst) as i64;` and `let received = app.ctx.msg_rx.load(Ordering::SeqCst) as i64;` cast from `u64` to `i64`. If counts exceed `i64::MAX`, the cast will wrap to a negative value, corrupting the report (including `dropped`). Use `u128` or keep `u64` for arithmetic and formatting. <b>[ Already posted ]</b>
- [line 159](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/benches/src/streaming.rs#L159): Panic risk: `tx.send(andre.inbox_id().to_string()).expect("Failed to share inbox_id")` will panic if the `oneshot` receiver (`rx`) is dropped before the send (e.g., if `setup_monitor` returns early or the awaiting task is cancelled). This should return an error rather than panic inside the spawned task. <b>[ Low confidence ]</b>
- [line 172](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/benches/src/streaming.rs#L172): Potential hang: `main` awaits `monitor_ready.notified().await` even if the monitor task fails before calling `ready.notify_one()`. In `monitor_messages`, `ready.notify_one()` is invoked only after several fallible operations (`sync_welcomes`, `stream_all_messages`). If any of these calls error, the spawned task returns early without notifying, causing `main` to wait indefinitely. <b>[ Already posted ]</b>
- [line 243](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/benches/src/streaming.rs#L243): Errors from individual senders are silently ignored. In `setup_send_messages`, `join_all(futs).await;` collects a `Vec<Result<()>>` from `send_messages`, but the results are discarded. This means message send failures are not surfaced and the benchmark continues reporting success, corrupting metrics and masking failures. <b>[ Previously rejected ]</b>
- [line 278](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/benches/src/streaming.rs#L278): Incorrect accounting: `ctx.msg_tx.fetch_add(1, Ordering::SeqCst);` and `ctx.tx_progress.inc(1);` are performed before `dm.send_message(...).await?`. If the send fails, the counters and progress are still incremented, inflating the "sent" count and corrupting the dropped calculation. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/identity.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 74](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_db/src/encrypted_store/identity.rs#L74): `queue_key_package_rotation` computes `rotate_at_ns = now_ns() + KEY_PACKAGE_QUEUE_INTERVAL_NS` without validating that `KEY_PACKAGE_QUEUE_INTERVAL_NS` is non-negative and within safe bounds. If the interval is negative (or so large it overflows `i64`), the method can schedule rotation in the past or trigger undefined behavior due to integer overflow. Similarly, `reset_key_package_rotation_queue` blindly uses `rotation_interval_ns` provided by the caller to set `next_key_package_rotation_ns`, allowing negative intervals to keep the rotation perpetually in the past and repeatedly eligible for resetting. Both methods should validate the interval (e.g., `>= 0` and no overflow on addition) to ensure a well-defined scheduling contract. <b>[ Already posted ]</b>
- [line 76](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_db/src/encrypted_store/identity.rs#L76): `queue_key_package_rotation` only updates rows where `next_key_package_rotation_ns > rotate_at_ns` and does not include the `NULL` case. When `next_key_package_rotation_ns` is `NULL`, the method will perform no update and return `Ok(())`, leaving the identity perpetually in a state where `is_identity_needs_rotation()` returns `true` (because `None` maps to needs-rotation). If `queue_key_package_rotation` is meant to schedule rotation for identities not yet queued, it should include a `IS NULL` condition or unconditionally set a value to ensure progress. As written, a `NULL` field will never be scheduled by `queue_key_package_rotation`. <b>[ Low confidence ]</b>
- [line 113](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_db/src/encrypted_store/identity.rs#L113): `is_identity_needs_rotation` uses `first::<Option<i64>>` to fetch `next_key_package_rotation_ns` and will return an error if the `identity` table has no rows. This creates an inconsistent contract relative to `queue_key_package_rotation` and `reset_key_package_rotation_queue`, which both succeed even when zero rows are affected (i.e., when no identity exists). With no explicit guard ensuring the identity row exists prior to these queries, a caller can observe `Ok(())` from queue/reset but then get an error from `is_identity_needs_rotation` in the same state. To provide a defined terminal state for all inputs, `is_identity_needs_rotation` should gracefully handle the "no row" case (e.g., treat it as needing rotation or return `Ok(false)` with a documented reason) rather than propagating `NotFound`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/builder.rs — 0 comments posted, 7 evaluated, 7 filtered</summary>

- [line 173](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/builder.rs#L173): `from_client` hardcodes `identity_strategy: IdentityStrategy::CachedOnly` rather than preserving the client’s original identity strategy. While the builder also sets `identity: Some(...)`, forcing `CachedOnly` may alter startup semantics (e.g., erroring if the identity is not found in the store) instead of matching the original client’s behavior. This is a contract parity issue that can lead to unexpected failures depending on downstream builder logic. <b>[ Low confidence ]</b>
- [line 179](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/builder.rs#L179): `from_client` does not preserve the `allow_offline` configuration from the source `client`. It hardcodes `allow_offline: false`, which can change runtime behavior when rebuilding a `ClientBuilder` from an existing `Client` that was configured to allow offline operation. <b>[ Low confidence ]</b>
- [line 180](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/builder.rs#L180): `from_client` does not preserve the `disable_events` setting from the source `client`. It hardcodes `disable_events` based on `cfg(test)` rather than copying from the existing client, causing behavior changes (events disabled in tests, enabled otherwise) regardless of the original client configuration. <b>[ Already posted ]</b>
- [line 184](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/builder.rs#L184): `from_client` resets `disable_commit_log_worker` to `false` instead of preserving the existing client’s configuration. This can re-enable commit log workers when reconstructing a builder from a client that had them disabled, changing runtime side effects and worker behavior. <b>[ Already posted ]</b>
- [line 187](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/builder.rs#L187): `from_client` assigns `cursor_store: None` unconditionally. If the source `client` had a configured `cursor_store`, this silently drops that configuration, potentially leading to missing cursors or reprocessing and can cause later startup/build paths to fail or behave incorrectly if a non-`None` `cursor_store` is expected. <b>[ Already posted ]</b>
- [line 188](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/builder.rs#L188): `from_client` hardcodes `disable_workers: false` instead of reflecting the source `client`’s configuration. Reconstructed builders may re-enable background workers unexpectedly, changing concurrency and side-effect behavior. <b>[ Already posted ]</b>
- [line 642](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/builder.rs#L642): In `enable_api_stats`, the error returned when either `api_client` or `sync_api_client` is missing is always `ClientBuilderError::MissingParameter { parameter: "api_client" }`, which is misleading when `sync_api_client` is the missing parameter. The guard `if self.api_client.is_none() || self.sync_api_client.is_none()` conflates the two cases but the error message specifically names only `api_client`. This can lead to incorrect diagnostics and user confusion at runtime. Consider returning a distinct error for each missing parameter or adjusting the error message to reflect which parameter is actually missing. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/device_sync/archive.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 208](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/groups/device_sync/archive.rs#L208): The test relies on exact entity counts and specific IDs that may be environment- or timing-dependent, leading to flakiness. Examples: asserting `groups.len() == 2` after loading from `alix`'s DB (line 208) and `old_messages.len() == 6` (line 215), asserting exactly one message synced (line 271). Background processes, schema changes, or additional system groups/messages can make these counts differ without indicating a functional regression. Prefer asserting minimums or properties that are invariant (e.g., presence of expected records) rather than exact totals unless the test fully isolates the DB state and guarantees no other records exist. <b>[ Test / Mock code ]</b>
- [line 228](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/groups/device_sync/archive.rs#L228): The test uses a fixed filesystem path `Path::new("archive.xmtp")` for creating and deleting the backup file. In a parallel test environment (Rust tests can run concurrently by default), multiple test instances or other tests using the same path can race: one test may delete the file while another is writing or reading it, leading to nondeterministic failures, partial reads/writes, or permission errors. This also risks leaving residue between runs. Use a unique, per-test temporary path (e.g., via `tempfile`, a random filename in a temp dir, or a per-test directory) to prevent collisions. <b>[ Test / Mock code ]</b>
- [line 243](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/groups/device_sync/archive.rs#L243): The test calls `insert_importer(&mut importer, &alix2.context).await.unwrap()`, implying that import failures would surface via the `Result`. However, the referenced `insert_importer` implementation logs per-element insert errors via `tracing::warn!` and continues without propagating them, returning `Ok(())` even on partial failures. This creates a contract mismatch and false confidence: the `unwrap` cannot catch insertion failures, only stream or importer-level errors. If insertions fail, the test only detects it later via assertions, making debugging harder and potentially masking partial failure cases. Consider making `insert_importer` propagate errors or have the test explicitly validate insertion outcomes without relying on `unwrap`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_sync.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 2864](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/groups/mls_sync.rs#L2864): Incorrect error handling for spawned tasks: the code collects only `JoinError`s from the `tokio::task::JoinSet::join_all()` results and silently ignores inner errors returned by `g.publish_intents().await`. Each spawned task returns `Result<_, _>`, so the outer type is `Result<Result<_, _>, JoinError>`. Filtering with `r.is_err()` only catches join failures (panics/cancellations), while `Ok(Err(e))` (where the task completed but returned an error) is treated as success and never logged nor triggers the final panic. This can cause the test to pass the error collection phase even when `publish_intents` fails in multiple tasks, relying solely on the later `assert_eq!` checks to detect issues. To fix, inspect both layers: for each `r` in the results, handle `Err(JoinError)` and also `Ok(inner)` where `inner.is_err()`, logging and failing accordingly. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/identity_updates.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 882](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/identity_updates.rs#L882): Background worker lifecycle is not managed: `DeviceSyncClient::new` followed by `wait_for_sync_worker_init` starts/initializes a sync worker, but the test never tears it down or shuts it off. This can leak tasks/resources and cause interference with other tests (e.g., additional logs or state changes), violating single paired cleanup and isolation. In long-running test processes, such leaked tasks can produce nondeterministic effects after the test completes. <b>[ Out of scope ]</b>
- [line 889](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/identity_updates.rs#L889): Global log count assumptions at the start of the test (`assert_logged!("Loaded association", 0)`) can fail if logs from prior tests are not isolated/reset. Without explicit log isolation per test, the baseline count may already be non-zero, producing false negatives unrelated to the code under test. The test should ensure log isolation or query scoped/log IDs associated only with this client/test. <b>[ Out of scope ]</b>
- [line 889](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/identity_updates.rs#L889): Flaky test due to asserting exact log occurrence counts with `assert_logged!` in a concurrent environment. The test initializes `DeviceSyncClient` and waits for the sync worker (`wait_for_sync_worker_init`), which may emit logs concurrently and unpredictably. Exact count assertions such as `assert_logged!("Loaded association", 0)` and subsequent increments (lines with `assert_logged!` at 889, 891, 906, 907, 925, 930, 931) can intermittently fail depending on timing and other background operations that may also log those messages. This violates determinism and can produce nondeterministic runtime failures. <b>[ Out of scope ]</b>
- [line 923](https://github.com/xmtp/libxmtp/blob/347909d4ac10ad8dc68d63e60a8822f647706915/xmtp_mls/src/identity_updates.rs#L923): Potential race condition after applying the identity association: the test immediately calls `get_association_state` after `apply_signature_request(...).await.unwrap()` (lines 917–923). If `apply_signature_request` triggers asynchronous persistence or propagation, `get_association_state` may read stale state, causing intermittent failures in subsequent assertions (`assert_logged!` counts and members length). The test should either verify the operation is synchronous and fully durable, or await a condition/event indicating persistence before reading. <b>[ Out of scope ]</b>
</details>

</details>

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -12,7 +12,11 @@ rustflags = ["--cfg", "tracing_unstable"]
 
 [target.wasm32-unknown-unknown]
 runner = 'wasm-bindgen-test-runner'
-rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
+rustflags = [
+  '--cfg',
+  'getrandom_backend="wasm_js"',
+  '-Ctarget-feature=+bulk-memory,+mutable-globals',
+]
 
 [target.aarch64-linux-android]
 rustflags = ["-Clink-arg=-Wl,-z,max-page-size=16384"]
--- a/.github/workflows/lint-wasm-bindings.yaml
+++ b/.github/workflows/lint-wasm-bindings.yaml
@@ -12,7 +12,6 @@ on:
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
-  RUSTFLAGS: --cfg tracing_unstable -Ctarget-feature=+bulk-memory,+mutable-globals --cfg getrandom_backend="wasm_js"
   CARGO_PROFILE_TEST_DEBUG: 0
 jobs:
   lint:
--- a/.github/workflows/test-webassembly.yml
+++ b/.github/workflows/test-webassembly.yml
@@ -53,31 +53,23 @@ jobs:
           ./emsdk install latest
           ./emsdk activate latest
       - name: Build WebAssembly Packages
-        env:
-          RUSTFLAGS: --cfg tracing_unstable -Ctarget-feature=+bulk-memory,+mutable-globals --cfg getrandom_backend="wasm_js"
         run: |
           source ./emsdk/emsdk_env.sh
           cargo build --locked --tests --release --target wasm32-unknown-unknown -p xmtp_id -p xmtp_mls -p xmtp_cryptography -p xmtp_common -p bindings_wasm -p xmtp_api_d14n -p xmtp_db
       - name: test libraries
         env:
-          RUSTFLAGS: --cfg tracing_unstable -Ctarget-feature=+bulk-memory,+mutable-globals --cfg getrandom_backend="wasm_js"
           CHROMEDRIVER: "chromedriver"
         run: |
-          cargo test --locked --release --target wasm32-unknown-unknown -p xmtp_id -p xmtp_cryptography -p xmtp_api -p bindings_wasm -p xmtp_api_d14n -p xmtp_db -- \
-            --skip encrypted_store::group_message::tests::it_cannot_insert_message_without_group \
+          cargo test --locked --release --target wasm32-unknown-unknown -p xmtp_id -p xmtp_cryptography -p xmtp_api -p bindings_wasm -p xmtp_api_d14n -p xmtp_db
         working-directory: ./
       - name: test xmtp_mls client
         env:
-          RUSTFLAGS: --cfg tracing_unstable -Ctarget-feature=+bulk-memory,+mutable-globals --cfg getrandom_backend="wasm_js"
           CHROMEDRIVER: "chromedriver"
         run: |
-          cargo test --locked --release --target wasm32-unknown-unknown -p xmtp_mls -- \
-            --skip encrypted_store::group_message::tests::it_cannot_insert_message_without_group \
-            --skip subscriptions::
+          cargo test --locked --release --target wasm32-unknown-unknown -p xmtp_mls -- --skip subscriptions::
         working-directory: ./
       - name: test xmtp_mls subscriptions
         env:
-          RUSTFLAGS: --cfg tracing_unstable -Ctarget-feature=+bulk-memory,+mutable-globals --cfg getrandom_backend="wasm_js"
           CHROMEDRIVER: "chromedriver"
         run: |
           cargo test --locked --release --target wasm32-unknown-unknown -p xmtp_mls -- subscriptions::
--- a/common/src/retry.rs
+++ b/common/src/retry.rs
@@ -310,7 +310,7 @@ impl Retry {
 ///     Err(MyError::Retryable)
 /// }
 ///
-/// #[tokio::main]
+/// #[tokio::main(flavor = "current_thread")]
 /// async fn main() -> Result<(), MyError> {
 ///
 ///     let (tx, mut rx) = mpsc::channel(3);
--- a/dev/test/wasm
+++ b/dev/test/wasm
@@ -1,17 +1,20 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -eou pipefail
 
-export RUSTFLAGS="-Ctarget-feature=+bulk-memory,+mutable-globals --cfg getrandom_backend=\"wasm_js\"${RUSTFLAGS:+ $RUSTFLAGS}"
+PACKAGE="${1:-}"
+if [[ -n "${PACKAGE}" ]]; then
+  PACKAGE=" -p ${PACKAGE}"
+else
+  PACKAGE=" --workspace --exclude xdbg --exclude xmtpv3 --exclude bindings_node --exclude mls_validation_service --exclude xmtp_cli --exclude xmtp_api_grpc --exclude xmtp_proto"
+fi
 
-PACKAGE=${1:-}
-TESTS=${2:-}
+TESTS="${2:-}"
 
-RUST_LOG=off \
-WASM_BINDGEN_SPLIT_LINKED_MODULES=1 \
-WASM_BINDGEN_TEST_ONLY_WEB=1 \
-WASM_BINDGEN_TEST_TIMEOUT=180 \
-RSTEST_TIMEOUT=60 \
-cargo test --locked --release --target wasm32-unknown-unknown \
-  -p $PACKAGE -- \
-  --skip encrypted_store::group_message::tests::it_cannot_insert_message_without_group \
+export RUST_LOG=off
+export WASM_BINDGEN_TEST_ONLY_WEB=1
+export WASM_BINDGEN_SPLIT_LINKED_MODULES=1
+export WASM_BINDGEN_TEST_TIMEOUT=180
+export RSTEST_TIMEOUT=60
+cargo test --locked --release --target wasm32-unknown-unknown --timings \
+  $PACKAGE -- \
   $TESTS
--- a/dev/test/wasm-interactive
+++ b/dev/test/wasm-interactive
@@ -1,23 +1,42 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -eou pipefail
 
-RED='\033[0;31m'
-NC='\033[0m' # No Color
-
-if [ -z "${1:-}" ]; then
-  echo -e "${RED}First argument must be package to test -- EX: \`test-wasm-interactive xmtp_mls\`${NC}"
-  exit
+PACKAGE="${1:-}"
+if [[ -n "${PACKAGE}" ]]; then
+  PACKAGE=" -p ${PACKAGE}"
+else
+  PACKAGE=" --workspace --exclude xdbg --exclude xmtpv3 --exclude bindings_node --exclude mls_validation_service --exclude xmtp_cli --exclude xmtp_api_grpc"
 fi
 
-PACKAGE=$1
-TESTS=${2:-}
+TESTS="${2:-}"
 
-export RUSTFLAGS="-Ctarget-feature=+bulk-memory,+mutable-globals --cfg getrandom_backend=\"wasm_js\"${RUSTFLAGS:=}"
+# Setup required tools
+if [[ "$(uname -s)" == "Darwin" ]]; then
+  LLVM="$(brew list | grep -E "^llvm$" || true)"
+  if [[ -z "${LLVM}" ]]; then
+    echo "Installing llvm"
+    brew install llvm
+    echo "Add llvm environment variables to your shell"
+    exit 1
+  fi
+fi
+
+WASM_BINDGEN_CLI="$(which wasm-bindgen-test-runner 2>/dev/null || true)"
+if [[ -z "${WASM_BINDGEN_CLI}" ]]; then
+  echo "Installing wasm-bindgen-cli"
+  cargo install wasm-bindgen-cli
+fi
+
+WASM_TARGET="$(rustup target list --installed | grep wasm32-unknown-unknown 2>/dev/null || true)"
+if [[ -z "${WASM_TARGET}" ]]; then
+  echo "Adding wasm32-unknown-unknown target"
+  rustup target add wasm32-unknown-unknown
+fi
+
 export WASM_BINDGEN_TEST_ONLY_WEB=1
 # export SPLIT_LINKED_MODULES=1
 export NO_HEADLESS=1
 
 cargo test --target wasm32-unknown-unknown --release \
   -p $PACKAGE -- \
-  --skip xmtp_mls::storage::encrypted_store::group_message::tests::it_cannot_insert_message_without_group \
   $TESTS
--- a/dev/test/wasm-nextest
+++ b/dev/test/wasm-nextest
@@ -1,13 +1,21 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -eou pipefail
 
-export RUSTFLAGS="-Ctarget-feature=+bulk-memory,+mutable-globals --cfg getrandom_backend=\"wasm_js\"${RUSTFLAGS:=}"
+# Setup required tools
+if [[ "$(uname -s)" == "Darwin" ]]; then
+  CHROMEDRIVER="$(which chromedriver 2>/dev/null || true)"
+  if [[ -z "${CHROMEDRIVER}" ]]; then
+    echo "Installing chromedriver"
+    brew install chromedriver
+  fi
+fi
 
-WASM_BINDGEN_TEST_ONLY_WEB=1 \
-  WASM_BINDGEN_TEST_TIMEOUT=180 \
-  CHROMEDRIVER="chromedriver" \
-  cargo nextest run \
+export WASM_BINDGEN_TEST_ONLY_WEB=1
+export WASM_BINDGEN_TEST_TIMEOUT=180
+export CHROMEDRIVER="chromedriver"
+
+cargo nextest run \
   --profile ci \
-  --workspace -E 'platform(target) and not test(it_cannot_insert_message_without_group)' \
+  --workspace -E 'platform(target)' \
   --target wasm32-unknown-unknown --release \
   --exclude xdbg --exclude xmtpv3 --exclude bindings_node --exclude mls_validation_service --exclude xmtp_cli --exclude xmtp_api_grpc
--- a/nix/package/bindings_wasm.nix
+++ b/nix/package/bindings_wasm.nix
@@ -44,7 +44,6 @@ let
     buildInputs = [ zstd sqlite ];
     doCheck = false;
     cargoExtraArgs = "--workspace --exclude xmtpv3 --exclude bindings_node --exclude xmtp_cli --exclude xdbg --exclude mls_validation_service --exclude xmtp_api_grpc";
-    RUSTFLAGS = [ "--cfg" "tracing_unstable" "--cfg" "getrandom_backend=\"wasm_js\"" "-C" "target-feature=+bulk-memory,+mutable-globals" ];
     CARGO_BUILD_TARGET = "wasm32-unknown-unknown";
     hardeningDisable = [ "zerocallusedregs" "stackprotector" ];
     NIX_DEBUG = 1;
--- a/xmtp_db/src/encrypted_store/group_message/tests.rs
+++ b/xmtp_db/src/encrypted_store/group_message/tests.rs
@@ -2,10 +2,7 @@
 wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_dedicated_worker);
 
 use super::*;
-use crate::{
-    EncryptedMessageStore, Store, group::tests::generate_group, prelude::*,
-    test_utils::with_connection,
-};
+use crate::{Store, group::tests::generate_group, prelude::*, test_utils::with_connection};
 use xmtp_common::{assert_err, assert_ok, rand_time, rand_vec};
 
 pub(crate) fn generate_message(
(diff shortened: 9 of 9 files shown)


benches/src/streaming.rs
```diff
@@ -0,0 +1,252 @@
...
+    tracing_subscriber::fmt()
+        // filter spans/events with level TRACE or higher.
+        // .with_max_level(Level::TRACE)
+        .with_env_filter(EnvFilter::new("streaming=trace"))
+        // build but do not install the subscriber.
+        .init();
+
+    let args = Args::parse();
+
+    let num_msgs = args.senders * args.count;
```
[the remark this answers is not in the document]
codabrink: It will never be that large.

benches/src/streaming.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    info!("Sending messages...");
+    let start = Instant::now();
+    fut.await;
+    let elapsed = start.elapsed();
+    app.ctx.tx_progress.finish();
+
+    let _ = app.ctx.barrier.lock().await;
+
+    let sent = app.ctx.msg_tx.load(Ordering::SeqCst) as i64;
```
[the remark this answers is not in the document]
codabrink: It will never be that large.

benches/src/streaming.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+
+    Ok(())
+}
+
+async fn setup_send_messages(
+    inbox_id: String,
+    ctx: &Arc<Context>,
+) -> Result<impl Future<Output = ()>> {
```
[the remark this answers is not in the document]
codabrink: There is a portion of the work I want done immediately, and a portion I want done later.

neekolas (commented):
Why a new crate rather than the existing `benches`  folder in `xmtp_mls`? Do we want to move all the benches over to here eventually?

codabrink:
Good callout. Something I meant to address before making the PR ready to review. I'll move it over to benches.

neekolas:
Maybe we do want all benches in their own crate, idk. The longer I spend working in Rust the more I think: "more crates good". @insipx do you have any thoughts?

insipx:
the more crates the better -- we would be able to simplify xmtp_mls Cargo.toml, since a bunch of those optional dependencies are for benchmarks, + it would be easier to have a place to put benchmark-specific code for Clients etc in the future. crates also act as individual compilation targets which lets them be compiled in parallel as well

codabrink:
Sounds good. I'll rename `performance` to `benches` and we can move the other benches to the new crate in a later PR.

codabrink:
I'll integrate this script into criterion when I move the benches over.

Cargo.toml
```diff
@@ -73,8 +74,8 @@ const_format = "0.2"
 ctor = "0.6"
 criterion = { version = "0.7", features = ["html_reports", "async_tokio"] }
 derive_builder = "0.20"
-diesel = { version = "2.3", default-features = false }
-diesel_migrations = { version = "2.2", default-features = false }
+diesel = { git = "https://github.com/ephemerahq/diesel", branch = "codabrink/writable-deserialize", default-features = false }
+diesel_migrations = { git = "https://github.com/ephemerahq/diesel", branch = "codabrink/writable-deserialize", default-features = false }
```
codabrink: We forked diesel to allow for loading in writable database snapshots into ephemeral db's, which makes it easy to load in thousands of clients without making network requests. I'll work on getting the function up-streamed, but in the meantime...

benches/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+name = "streaming"
+path = "src/streaming.rs"
+
+[dependencies]
+anyhow.workspace = true
+clap = { version = "4.5.49", features = ["derive"] }
+futures.workspace = true
+indicatif = "0.18.0"
+parking_lot.workspace = true
+rlimit = "0.10.2"
```
insipx, line 17: didn't know about this crate, we should replace `fdlimit` with it if its more up-to-date/maintained. do you mind adding an issue to the dev board for that if not in this pr?

benches/src/streaming.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+const RLIMIT: u64 = 4096;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    tracing_subscriber::fmt()
+        // filter spans/events with level TRACE or higher.
+        // .with_max_level(Level::TRACE)
+        .with_env_filter(EnvFilter::new("streaming=trace"))
```
tylerhawkes, line 65: Probably want to use `RUST_LOG` env var and default to this if not set.

benches/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[[bin]]
+name = "streaming"
+path = "src/streaming.rs"
+
+[dependencies]
+anyhow.workspace = true
+clap = { version = "4.5.49", features = ["derive"] }
+futures.workspace = true
+indicatif = "0.18.0"
```
insipx: since we're using in multiple places, we should extract any common deps into the workspace folder, to avoid any possible dups

benches/src/streaming.rs
```diff
@@ -0,0 +1,286 @@
...
+        .map(|rx| rx.to_string())
+        .unwrap_or("Unknown".to_string());
+
+    info!(
+        "\nREPORT:\n\
+        {sent} messages sent across {senders} senders,\n\
+        {received} messages received ({dropped} dropped)\n\
+        rx time: {rx_elapsed} seconds ({rx_rate} msgs/s)\n\
+        tx time: {tx_elapsed} seconds ({tx_rate} msgs/s)",
+    );
```
codabrink: I plan to switch thi to a criterion report when the rest of the benches are moved over.

xmtp_mls/src/utils/test/tester_utils.rs
```diff
@@ -52,42 +63,67 @@ use xmtp_id::{
...
-pub struct Tester<Owner, Client>
+pub struct Tester<Owner = PrivateKeySigner, Client = FullXmtpClient>
 where
     Owner: InboxOwner,
 {
     pub builder: TesterBuilder<Owner>,
     pub client: Client,
     pub worker: Option<Arc<WorkerMetrics<SyncMetric>>>,
+    #[cfg(target_arch = "wasm32")]
     pub stream_handle: Option<Box<dyn StreamHandle<StreamOutput = Result<(), SubscribeError>>>>,
```
insipx, line 82: if you use tokio_with_wasm, you can avoid this `StreamHandle` stuff, since the spawn apis are the same. at some point we need to remove `StreamHandle` entirely, since originally it was created to deal with differences between tokio & wasm_bindgen_futures

the [`xmtp_common::task::`](https://xmtp.github.io/libxmtp/xmtp_common/wasm/task/index.html) path should have the spawn re-exported. it has the advantage of using the nice tokio handle types as well. this will avoid the cfg trick here

benches/Cargo.toml
codabrink: Yeah I saw fdlimit in the benches. I can make one.

xmtp_mls/src/utils/test/tester_utils.rs
```diff
@@ -138,37 +185,73 @@ where
...
-            .with_disable_events(Some(!self.events))
             .with_commit_log_worker(self.commit_log_worker)
             .fork_recovery_opts(self.fork_recovery_opts.clone().unwrap_or_default());
 
+        // Setup the database. Snapshots are always ephemeral.
+        if self.ephemeral_db || self.snapshot.is_some() {
+            #[cfg(not(target_arch = "wasm32"))]
+            let db = NativeDb::new_unencrypted(&StorageOption::Ephemeral).unwrap();
+            #[cfg(target_arch = "wasm32")]
+            let db = WasmDb::new(&StorageOption::Ephemeral).await.unwrap();
```
insipx: you can use the [`TestDb`](https://xmtp.github.io/libxmtp/xmtp_db/test_utils/type.TestDb.html) alias here along with the [`create_database`](https://xmtp.github.io/libxmtp/xmtp_db/test_utils/type.TestDb.html#impl-XmtpTestDb-for-EncryptedMessageStore%3CNativeDb%3E) function to create databases & avoid cfg

Cargo.toml
```diff
@@ -73,8 +74,8 @@ const_format = "0.2"
 ctor = "0.6"
 criterion = { version = "0.7", features = ["html_reports", "async_tokio"] }
 derive_builder = "0.20"
-diesel = { version = "2.3", default-features = false }
-diesel_migrations = { version = "2.2", default-features = false }
+diesel = { git = "https://github.com/ephemerahq/diesel", branch = "codabrink/writable-deserialize", default-features = false }
```
insipx: we should put this in the [patch] section

tylerhawkes (approved):

Cargo.toml
insipx: Also, is there a plan for upstreaming? I think Weiznich would be open to this change, it seems harmless & as long as it is well-documented, I don't see it causing any issues. it would be nice to not worry about maintaining our own fork in the future, esp if we eventually aim to release libxmtp on crates.io

xmtp_mls/src/utils/test/tester_utils.rs
codabrink: Oh nice

Cargo.toml
codabrink: Yeah I think it should be pretty easy to upstream it too. I'll make an effort to get it up.

PR #2684, opened 2025-10-24
Sort and paginate by inserted_at

neekolas:
### TL;DR

Added `inserted_at_ns` field to messages to track when they were added to the database, enabling insertion-time based sorting and pagination.

Default all existing rows to use their `sent_at_ns`​ as their `inserted_at_ns`​ in the absence of anything more accurate.

### Notes

We initially talked about using the built in `rowid`​ for this sort. There are a few problems:  
1\. Takes a ton of diesel gymnastics to expose the rowid and query by it. Not pretty.

1. `rowid`​ can change in a vacuum, mangling the sort order. We don't have an integer primary key and don't want one here

The migration for this was a bit of a journey. Turns out SQLite does not allow you to create a default value with a function on an existing table, only on new table creation. I didn't want to have to pass in `now_ns`​ every time you save a message. Especially since you might actually insert in a slightly different order from when you create the timestamps. Feels more sensible to let SQLite handle this for us.

I added a benchmark on doing this migration for 10k 500 byte messages and it was <100ms, which I think is in the range of acceptable. One time pain to get a better schema.

### What changed?

- Added `inserted_at_ns` field to `StoredGroupMessage` and related message structures
- Created a database migration to add the column to the `group_messages` table
- Added new `SortBy` enum with `SentAt` and `InsertedAt` options
- Extended `MsgQueryArgs` with `sort_by`, `inserted_after`, and `inserted_before` parameters
- Updated FFI bindings to expose the new field and query parameters
- Modified message storage to let the database set the `inserted_at_ns` value automatically
- Added comprehensive tests for the new sorting and filtering capabilities

### How to test?

1. Use the new query parameters to sort messages by insertion time:
2. Implement cursor-based pagination using the `inserted_at_ns` field:

### Why make this change?

This change addresses several important use cases:

1. **Reliable pagination**: Using `inserted_at_ns` for pagination provides a stable ordering that won't change when messages are added or removed, unlike offset-based pagination.
2. **Chronological insertion order**: Some applications need to display messages in the order they were received by the client, not by their sent timestamp.
3. **Sync efficiency**: The new fields enable more efficient syncing by allowing clients to request only messages inserted after their last sync time.
4. **Conflict resolution**: When multiple messages have the same sent timestamp, insertion time provides a deterministic ordering.

### DIFF
--- a/bindings_ffi/src/message.rs
+++ b/bindings_ffi/src/message.rs
@@ -347,6 +347,7 @@ pub struct FfiDecodedMessageMetadata {
     pub sender_inbox_id: String,
     pub content_type: FfiContentTypeId,
     pub conversation_id: Vec<u8>,
+    pub inserted_at_ns: i64,
 }
 
 #[derive(uniffi::Enum, Clone, Debug)]
@@ -896,6 +897,7 @@ impl From<DecodedMessageMetadata> for FfiDecodedMessageMetadata {
             conversation_id: metadata.group_id,
             sender_inbox_id: metadata.sender_inbox_id,
             content_type: metadata.content_type.into(),
+            inserted_at_ns: metadata.inserted_at_ns,
         }
     }
 }
@@ -1051,6 +1053,7 @@ pub struct FfiDecodedMessage {
     reactions: Vec<Arc<FfiDecodedMessage>>,
     delivery_status: FfiDeliveryStatus,
     num_replies: u64,
+    inserted_at_ns: i64,
 }
 
 #[uniffi::export]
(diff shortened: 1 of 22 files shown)


neekolas:
* **#2684** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2684?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2684?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=2684).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

insipx (approved):

PR #2704, opened 2025-10-31
repro xmtp-node-go segfault

insipx:
this is the diff with my branch that causes the segfault. If i revert this change i no longer get segfaults. this is part of some tester code that is setting up Api Clients. The segfault occurs when an api client is setup with the default cursor store, which sets all id_cursors to `0`

```
Modified regular file src/utils/test/tester_utils.rs:
    ...
 211  211:             client = client.temp_store().await;
 212  212:         }
 213  213: 
 214  214:         let f(mut local_client, mut sync_api_client) = match self.api_endpoint {
 215  215:             ApiEndpoint::Local => (TestClient::create_local(), TestClient::create_local()),
 216  216:             ApiEndpoint::Dev => (TestClient::create_dev(), TestClient::create_dev()),
 217  217:         };
 218     :         let store = Arc::new(SqliteCursorStore::new(client.store.as_ref().unwrap().db()));
 219     :         let mut local_client = TestClient::with_cursor_store(f, store.clone());
 220     :         let mut sync_api_client = TestClient::with_cursor_store(f, store);
 221     : 
 222  218:         let mut proxy = None;
 223  219:         if self.proxy {
 224  220:             proxy = Some(local_client.with_toxiproxy().await);
    ...
```


insipx:
fix is in https://github.com/xmtp/xmtp-node-go/pull/533

PR #2715, opened 2025-11-03
Streaming Stresstest Dashboard

codabrink:
### Add an interactive TUI streaming stresstest dashboard to the bench binary and raise the default timeout to 4
Introduce a ratatui-based dashboard in the bench target that renders logs, charts, and gauges, wire UI updates from the streaming monitor with per-message timing and reconnection telemetry behind `feature "test-utils"`, and switch progress reporting from `ProgressBar` to in-app gauges while updating imports to `xmtp_mls::mls_common::*`.

#### 📍Where to Start
Start with the synchronous `main` in [streaming.rs](https://github.com/xmtp/libxmtp/pull/2715/files#diff-3b1a2906ffa276ff948acbb8d6ed8139f583ae54e895aa1666d184da96c84f6b), then review the `App` run loop and render methods, followed by `benchmark` and `monitor_messages` for the data flow and telemetry integration.

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 1023893. 10 files reviewed, 32 issues evaluated, 26 issues filtered, 3 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>benches/src/streaming.rs — 3 comments posted, 20 evaluated, 15 filtered</summary>

- [line 72](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L72): `ExpAvg::new` accepts any `alpha: f64` without validation. Values outside the [0.0, 1.0] range (including `NaN` or infinities) cause nonsensical or unstable behavior in `add`: negative or >1 weights, potential overflow in `(1.0 - self.alpha)`, and loss of normalization. For example, `alpha < 0.0` yields negative weighting, `alpha > 1.0` can make `(1.0 - alpha)` negative, and `alpha = NaN` propagates `NaN` to `avg`, permanently corrupting the state. The typical EMA smoothing factor is in [0.0, 1.0]; the constructor should validate and reject invalid `alpha` or explicitly document and handle them (e.g., clamp, return `Result`, or panic during initialization before external effects). <b>[ Low confidence ]</b>
- [line 77](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L77): Initialization of `ExpAvg` is incorrectly inferred from the value of `avg` being exactly `0.0`. In `add`, the condition `if self.avg == 0.0` treats an actual, legitimate average of zero as "uninitialized" and reseeds the average with the next `value`, bypassing smoothing for that call. This breaks the intended exponential moving average behavior whenever the average legitimately becomes `0.0`, causing a discontinuity and overweighting the next sample (setting `avg = value` instead of `alpha * value + (1 - alpha) * avg`). This should be implemented using an explicit initialization flag (e.g., `Option<f64>` or a separate `initialized` boolean), not by comparing `avg` to `0.0`. <b>[ Low confidence ]</b>
- [line 190](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L190): The background Tokio runtime thread spawned in `main` is never shut down, even when the TUI exits. The code spawns a thread that calls `runtime.block_on(async { tokio::spawn(benchmark(ctx, stats_tx)).await })` and then `unwrap()` twice. If `benchmark` runs indefinitely (typical for streaming/benchmark loops), the thread will block forever. Pressing 'q' in the UI causes `App::run` to return and `main` to exit `Ok(())`, but the spawned thread keeps the process alive, preventing program termination. Additionally, using `unwrap()` on the join results will panic if the task or runtime fails, causing inconsistent shutdown behavior. A coordinated shutdown signal with joining the thread and aborting the spawned tasks is needed. <b>[ Low confidence ]</b>
- [line 242](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L242): On quit ('q'), `run` returns early without ensuring background tasks are torn down. If the welcome sender task was enabled (`self.ctx.welcome_sender` is `Some`), it is not aborted and may continue running, potentially leaking resources or keeping the runtime alive after UI exit. <b>[ Low confidence ]</b>
- [line 255](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L255): In `keypress`, repeatedly pressing '-' grows `self.graph_window` via `Duration + Duration`. `std::time::Duration` addition panics on overflow. Although reaching overflow is unlikely in practice, there is no guard against it, leaving a latent runtime panic. <b>[ Low confidence ]</b>
- [line 271](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L271): `keypress` calls `self.ctx.tokio_handle.block_on(...)` inside the UI event loop. If `run()` is executed on a Tokio runtime thread (no guard prevents this), `Handle::block_on` will panic at runtime or deadlock. This makes the 'w' toggle path unsafe in runtime contexts where the UI runs within Tokio. <b>[ Low confidence ]</b>
- [line 273](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L273): `keypress` uses `unwrap()` on the result of `continuous_new_welcomes(...)` inside the `KeyCode::Char('w')` toggle path. If `continuous_new_welcomes` returns an error, this will panic at runtime, crashing the app instead of returning an error or logging it. <b>[ Low confidence ]</b>
- [line 288](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L288): The `Promise::Receiver` handling in `App::tick` only transitions to `Promise::Item` on `Ok(stats)` from `oneshot::Receiver::try_recv()`. If the sender side is dropped or the oneshot is closed without sending (which is reachable in error paths), `try_recv()` returns a `Closed` error. The current code leaves `self.stats_rx` as `Promise::Receiver` forever, causing the stats never to be processed and a latent leak of the receiver state. The code should handle `Closed` by transitioning to a terminal state (e.g., a `None` equivalent or a default stats handle) to ensure every input reaches a defined terminal state. <b>[ Low confidence ]</b>
- [line 316](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L316): Unbounded growth of UI data vectors leads to memory usage increasing without limit during long runs. In `tick`, `self.data.rx_process_time` and `self.data.rx_process_time_rolling_average` are continuously appended to on every message and every ~20ms respectively, and `self.data.reconnection_time` grows when the welcome sender is active. There is no pruning of old entries to the current `graph_window`. Over time, these vectors can become very large, causing increased memory usage and degraded performance. Implement periodic pruning of entries that fall outside `start_ns..end_ns` or cap their size. <b>[ Low confidence ]</b>
- [line 351](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L351): Casting `self.data.gauges.len()` (a `usize`) to `u16` for a layout constraint can silently truncate if the number of gauges exceeds `u16::MAX`, causing the computed layout height to wrap-around to a small value and rendering a malformed UI. There is no guard or clamp to prevent truncation. <b>[ Low confidence ]</b>
- [line 458](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L458): `render_rolling_avg` also risks producing a degenerate Y-axis range when `min == max` (e.g., only one constant datapoint), which some charting libs reject or render poorly. Providing a small padding or default range avoids potential runtime errors or unreadable output. <b>[ Code style ]</b>
- [line 529](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L529): The result of `fut.await` in `benchmark` is ignored, which silently discards any error from the message-sending future. This can cause the benchmark to report misleading stats (e.g., fewer sent messages) without any indication of failure, and may leave the system in an inconsistent state. The call should propagate or at least log the error, e.g. `fut.await?` or `if let Err(e) = fut.await { error!(...); }`. <b>[ Low confidence ]</b>
- [line 601](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L601): Potential overflow when computing the jittered frequency: `let freq = freq + Duration::from_nanos(jitter as u64);` where `jitter` is `welcome_sender_jitter_ms * 1_000_000`. If `welcome_sender_jitter_ms` is large enough, `Duration::from_nanos(jitter as u64)` can approach `u64::MAX`, and the addition with `freq` can overflow and panic. Use `checked_add` or `saturating_add` on `Duration` to avoid panics. <b>[ Previously rejected ]</b>
- [line 674](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L674): UI update sends are performed without handling failures, causing silent loss of updates if the receiver is dropped. Examples: `ctx.ui.send(UiUpdate::MessageReceiveNs(...))` and `let _ = ctx.ui.send(UiUpdate::Gauge(...))`. At minimum, log failures to preserve observability. <b>[ Code style ]</b>
- [line 743](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/benches/src/streaming.rs#L743): Filesystem write may fail due to missing parent directory: `fs::write(&snapshot_path, snapshot).await?;` is executed without ensuring the directory containing `snapshot_path` exists. If the provided `snapshot_path` points to a non-existent directory, this will error at runtime. Create the directory (e.g., with `fs::create_dir_all`) before writing. <b>[ Low confidence ]</b>
</details>

<details>
<summary>bindings_node/src/permissions.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 69](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/bindings_node/src/permissions.rs#L69): Loss of policy specificity and non-round-trippable membership and metadata policies: `From<&MembershipPolicies> for PermissionPolicy` maps `BasePolicies::AllowSameMember` to `PermissionPolicy::Other`, and any non-`Standard` variants to `PermissionPolicy::Other`. Similarly, `From<&MetadataPolicies>` and `From<&PermissionsPolicies>` return `PermissionPolicy::Other` for non-`Standard` variants. However, the corresponding `TryInto` implementations for `MembershipPolicies`, `MetadataPolicies`, and `PermissionsPolicies` only accept `Allow`, `Deny`, `Admin`, or `SuperAdmin` and return `Err(InvalidPermissionPolicyOption)` for `Other`. This makes valid non-standard policies impossible to round-trip: reading yields `Other`, writing the same value back fails. This is a runtime contract mismatch likely to surface when encountering such variants in existing groups. <b>[ Low confidence ]</b>
- [line 201](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/bindings_node/src/permissions.rs#L201): Non-round-trip behavior and reachable runtime error when metadata policies are absent: `policy_set()` returns `PermissionPolicy::DoesNotExist` for missing metadata keys (via `get_policy`), but `TryFrom<PermissionPolicySet> for PolicySet` unconditionally calls `try_into()` on each corresponding field, which returns `Err(GroupMutablePermissionsError::InvalidPermissionPolicyOption)` for `DoesNotExist`. Any consumer that attempts to convert a `PermissionPolicySet` produced by `policy_set()` back into a `PolicySet` without explicitly supplying all metadata policies will hit a runtime error. This breaks contract parity and makes a valid read-only state (missing metadata policy) impossible to round-trip back into a writable `PolicySet` without error. <b>[ Low confidence ]</b>
- [line 254](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/bindings_node/src/permissions.rs#L254): Silent data loss and behavior change for `update_permissions_policy`: `policy_set()` does not expose any field for `update_permissions_policy`, and `TryFrom<PermissionPolicySet> for PolicySet` unconditionally sets `update_permissions_policy` to `PermissionsPolicies::allow_if_actor_super_admin()` (hardcoded), discarding the original value. Any conversion from a `PermissionPolicySet` will overwrite whatever the underlying group had, potentially weakening or strengthening permissions without visibility. This breaks contract parity and can yield incorrect runtime behavior. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/subscriptions/mod.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 332](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/xmtp_mls/src/subscriptions/mod.rs#L332): `on_close()` is not guaranteed to be invoked on task cancellation or abrupt termination outside of the controlled paths. In all four streaming-with-callback functions, `on_close()` is invoked only on explicit error during stream creation or after the loop ends. If the spawned task is cancelled (e.g., runtime shutdown or external cancellation) while inside the `while let Some(...)` loop or before reaching the `on_close()` calls, `on_close()` will not run, violating the at-most-once/at-least-once close notification expectation and potentially leaking higher-level state that depends on `on_close()` for cleanup. <b>[ Low confidence ]</b>
- [line 339](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/xmtp_mls/src/subscriptions/mod.rs#L339): Errors during stream creation are swallowed and never propagated via the returned `StreamHandle`'s `StreamOutput`, which is `Result<()>`. In `stream_conversations_with_callback`, when `client.stream_conversations(...)` returns `Err`, the code logs a warning, calls `on_close()`, and then returns `Ok(())` (`return Ok::<_, SubscribeError>(())`). This makes it impossible for the caller to distinguish between a normal completion and a failure to create the stream. The same pattern exists in `stream_all_messages_with_callback`. If the external contract expects `Err` propagation through the `StreamHandle` for failures, this is a runtime semantics bug — the failure path is indistinguishable from successful termination. <b>[ Out of scope ]</b>
- [line 423](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/xmtp_mls/src/subscriptions/mod.rs#L423): Same swallowed error propagation in `stream_all_messages_with_callback`: when `StreamAllMessages::new(&context, ...)` returns `Err`, the code logs a warning, calls `on_close()`, and returns `Ok(())`. This prevents the caller from observing the failure via the returned `StreamHandle` result, conflating failure to start with normal termination. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/subscriptions/stream_messages.rs — 0 comments posted, 5 evaluated, 4 filtered</summary>

- [line 145](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/xmtp_mls/src/subscriptions/stream_messages.rs#L145): `new_with_factory` forwards the input `groups` directly to `subscribe_group_messages` without any validation or normalization. If `groups` contains duplicate `GroupId`s, the subscription will include duplicates. Depending on the underlying implementation of `XmtpMlsStreams::subscribe_group_messages`, this can result in duplicated delivery and processing of the same group's messages or undefined behavior in the subscription. The code should either de-duplicate `groups` before subscribing or explicitly document and reject duplicates. <b>[ Low confidence ]</b>
- [line 145](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/xmtp_mls/src/subscriptions/stream_messages.rs#L145): `new_with_factory` allows subscribing with an empty `groups` list and forwards it to `subscribe_group_messages`. If the underlying API rejects empty subscriptions, this produces an error that is propagated, but the method provides no early validation or clear error context. At minimum, returning a defined error when `groups.is_empty()` would make the outcome explicit and avoid making an external call with a known-invalid input. <b>[ Low confidence ]</b>
- [line 317](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/xmtp_mls/src/subscriptions/stream_messages.rs#L317): Contract/documentation mismatch in the subscription path: the `subscribe` helper (invoked by `resolve_group_additions`, reached via the Adding state in `poll_next`) is documented to "get the last synced cursor" and "determine the appropriate cursor position", but it unconditionally returns `Some(Cursor { sequence_id: 1, originator_id: 0 })`. This contradicts the stated behavior and can lead to starting at an incorrect position, potentially missing messages or causing replay inconsistencies. While the cursor logic is outside the immediate code-object lines, `poll_next` relies on the returned cursor to set state (`this.set_cursor(...)`) making the mismatch observable here. <b>[ Low confidence ]</b>
- [line 428](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/xmtp_mls/src/subscriptions/stream_messages.rs#L428): `resolve_group_additions` preemptively marks the newly added group as having "seen" cursor `Cursor::new(1, 0u32)` via `this.groups.add(&group.group_id, MessagePosition::new(Cursor::new(1, 0u32)));` before subscription re-establishment and without consulting persisted state. Downstream, `on_waiting` checks `self.groups.has_seen(next_msg.cursor)` and will skip messages whose cursor is considered seen. If the first message delivered for the newly added group has cursor `1`, it will be skipped even if it is new, resulting in potential message loss at the start of the added group. The correct behavior is to initialize the group's position from a persisted last cursor (if any) or leave it unset until `subscribe` returns a determined starting cursor; otherwise, do not mark any cursor as seen prematurely. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/subscriptions/stream_messages/types.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 122](https://github.com/xmtp/libxmtp/blob/10238939f54e31ba8589752f7a71f9f7c3c9a0d0/xmtp_mls/src/subscriptions/stream_messages/types.rs#L122): Unbounded growth of `GroupList.seen` (`HashSet<Cursor>`) without eviction or compaction can lead to memory exhaustion over time. The `set` method unconditionally inserts every `cursor` into `self.seen`, and `has_seen` only checks for presence; there is no cleanup strategy, cap, or time/window-based eviction. This violates the "no leaks" invariant for long-running services and is explicitly hinted at in the comment above `seen` that suggests a Bloom filter or GC strategy. Under sustained message traffic with unique `Cursor` values (e.g., monotonically increasing `sequence_id`), the set will grow indefinitely, potentially causing runtime failures due to memory pressure. Consider bounding the size, adding eviction, or using a probabilistic dedup structure as noted in the comment. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/benches/src/streaming.rs
+++ b/benches/src/streaming.rs
@@ -16,7 +16,7 @@ use rlimit::{Resource, setrlimit};
 use std::{
     collections::HashMap,
     f64,
-    path::PathBuf,
+    path::{Path, PathBuf},
     sync::{
         Arc,
         atomic::{AtomicU64, Ordering},
@@ -41,7 +41,6 @@ use tui_logger::{
 };
 use xmtp_mls::{
     common::time::now_ns,
-    db::group_message::ContentType,
     subscriptions::stream_messages::{StreamStat, StreamState, StreamStats},
     tester,
     utils::Tester,
@@ -53,7 +52,7 @@ struct Args {
     count: u64,
     #[arg(short, long, default_value = "10")]
     senders: u64,
-    #[arg(short, long, default_value = "2")]
+    #[arg(short, long, default_value = "4")]
     timeout: u64,
     #[arg(short, long, default_value = "false")]
     dev: bool,
@@ -304,7 +303,9 @@ impl App {
                 }
             }
         }
-        if let Some(duration) = self.data.reconnection_duration {
+        if let Some(duration) = self.data.reconnection_duration
+            && self.ctx.welcome_sender.lock().is_some()
+        {
             let duration = duration.as_nanos() as f64;
             self.data.reconnection_time.push((now_ns, duration));
         }
@@ -646,7 +647,7 @@ async fn monitor_messages(
     let (mut stream, stats) = andre
         .stream_all_messages_owned_with_stats_handle(None, None)
         .await?;
-    stats_tx.send(stats);
+    let _ = stats_tx.send(stats);
 
     let mut monitoring_start: Option<Instant> = None;
     let grace_period = Duration::from_secs(ctx.args.timeout);
@@ -697,11 +698,17 @@ async fn setup_send_messages(
     ctx: &Arc<Context>,
 ) -> Result<impl Future<Output = ()>> {
     info!("Registering {} senders...", ctx.args.senders);
-    let _ = tokio::fs::create_dir_all("snapshots").await;
+    let mut snapshot_path = PathBuf::from("snapshots");
+    if ctx.args.dev {
+        snapshot_path = snapshot_path.join("dev")
+    } else {
+        snapshot_path = snapshot_path.join("local")
+    };
+    let _ = tokio::fs::create_dir_all(&snapshot_path).await;
 
     let mut futs = vec![];
     for i in 0..ctx.args.senders {
-        futs.push(create_client(i, ctx));
+        futs.push(create_client(i, ctx, &snapshot_path));
     }
     let testers: Vec<Tester> = futures::stream::iter(futs)
         .buffer_unordered(100)
@@ -725,8 +732,8 @@ async fn setup_send_messages(
     })
 }
 
-async fn create_client(i: u64, ctx: &Context) -> Result<Tester> {
-    let snapshot_path = PathBuf::from(format!("snapshots/{i}.db3"));
+async fn create_client(i: u64, ctx: &Context, snapshot_path: &Path) -> Result<Tester> {
+    let snapshot_path = snapshot_path.join(format!("{i}.db3"));
     let snapshot = fs::read(&snapshot_path).await.ok().map(Arc::new);
 
     tester!(bo, with_dev: ctx.args.dev, ephemeral_db, with_snapshot: snapshot.clone(), disable_workers);


xmtp_mls/src/subscriptions/stream_messages/test_utils.rs
```diff
@@ -1,4 +1,125 @@
...
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
+};
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel};
+use xmtp_common::time::now_ns;
+use xmtp_proto::prelude::XmtpMlsStreams;
+
+pin_project! {
```
insipx, line 19: rather than using #[cfg] test, this would be much cleaner if you used composition to compose The messages stream with a new stream type. the state enum is `pub(super)`, so it can be accessed in sibling modules.

This would make the new stream type highly coupled to stream messages, but that's better than maintaining a test-only mirror of the type.

I.E the impl would look roughly something like this:

```rust
pin_project!
    pub struct StreamMessagesStats<'a, Context: Clone, Subscription, Factory =ProcessMessageFuture<Context>>> {
        inner: StreamGroupMessages<'a, Context, Subscription, Factory>,
        stats: StreamStats
    }
}

impl<'a, Context, Subscription, Factory> Stream for StreamMessagesStats<'a, Context, Subcription, Factory> {
    type Item = StreamGroupMessages<'a, Context, Subscription, Factory>::Item;
    fn poll_next(mut self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Option<Self::Item>> {
         match self.inner.state {
              Waiting => {
                  // do stats stuff
                  self.inner.poll_next(cx)
              },
              // set all the stats
              // any new states that stats are tracking can be new fields on "state"
             // for instance, "FinishReconnect" can become a state like "State::WaitingAfterReconnect" that follows the same codepath as Waiting
         }
    }
}
```

then you don't even need to change the "with_handle" function, since it returns an `impl Stream` anyway, you can just compose the Stream groups with this new messages stats type.

i.e something like

```rust
let s = StreamMessages::new();
let with_stats = StreamStats::new(s);
```


To be even more generic you could define a trait like 

```rust
trait StreamState {
    fn get_state(&self) -> State;
}
```

then require that this trait is implemented for your Stats stream type, i.e

```rust
pin_project! {
    pub struct StreamStats<S: StreamState> {
        inner: S,
        stats: ApiStats
    }
}
```


This has the chief benefit of making it very easy to grok what `StreamMessageStats` does, since it removes all unnecessary type arguments
```rust
impl<'a, S> Stream for StreamMessagesStats<'a, S> 
where
    S: Stream + StreamState
{
    type Item = S::Item;
    fn poll_next(mut self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Option<Self::Item>> {
         match self.inner.get_state() {
              Waiting => {
                  // do stats stuff
                  self.inner.poll_next(cx)
              }
              // set all the stats
              // any new states that stats are tracking can be new fields on "state"
             // for instance, "FinishReconnect" can become a state like "State::WaitingAfterReconnect" that follows the same codepath as Waiting
         }
    }
}
```


then we don't really have to worry about cfg(test) anymore, since the stats can be safely compiled w/o feature flags and not effect the core functionality of StreamGroupMessages

Generally, composition is the most rust-idiomatic approach to shared functionality so we also get a nice type signature of the stream to tell if stats are attached or not

PR #2716, opened 2025-11-03
Rework Send Sync bounds to be applied on traits instead of impls to avoid unsafe impls in wasm

tylerhawkes:
### Apply Send/Sync constraints on traits and remove Send/Sync bounds from impls across `xmtp_*` crates to avoid unsafe impls on wasm
Relax Send/Sync bounds from impls and propagate `MaybeSend`/`MaybeSync` onto core traits and error types, unifying worker spawning, stream handles, and API client traits for wasm by eliminating unsafe Send/Sync impls and cfg splits; add `BoxDynError` where needed and introduce a non-retryable `ApiClientError::OtherUnretryable`.

#### 📍Where to Start
Start with the trait bound changes in `common` and `xmtp_proto` to see the new `MaybeSend`/`MaybeSync` foundation: review `RetryableError` and stream handle changes in [common/src/retry.rs](https://github.com/xmtp/libxmtp/pull/2716/files#diff-1fc729bbb038af084238293889e90354fb5eca3dfd6641183f6afcacfd945c24) and [common/src/stream_handles.rs](https://github.com/xmtp/libxmtp/pull/2716/files#diff-afc279a63ab19e2fc8affd22da405474e5d582ccdcdf0496b31125162ee74b02), then the API trait updates in [xmtp_proto/src/api_client.rs](https://github.com/xmtp/libxmtp/pull/2716/files#diff-7e862515880042ad7a89178c4850f4edd3b1a449ebd83515598088bd845fe67b) and [xmtp_proto/src/traits.rs](https://github.com/xmtp/libxmtp/pull/2716/files#diff-a040511e61519da940d7c0956bd7917a6a8d2ce1a929d3b4b8978cc7084befb5). After that, examine worker unification in [xmtp_mls/src/worker.rs](https://github.com/xmtp/libxmtp/pull/2716/files#diff-ea560e63ca633c03d1189957b9efca26de6aa33294f42c116a574e5bf259ce23).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 1d16a56. 52 files reviewed, 48 issues evaluated, 45 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>common/src/stream_handles.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 142](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/common/src/stream_handles.rs#L142): `AbortHandle::is_finished` for `CloseHandle` returns `self.0.is_closed()`, which reports whether the `mpsc::Receiver` was dropped, not whether the task/future has completed or a close signal was processed. During normal completion, the result is sent on the `oneshot` and the task ends, which drops the receiver and makes `is_closed()` return true. However, if a close signal was sent (`end()` called) but the spawned task has not yet processed it (buffer full or not polled yet), `is_finished()` will still return false despite a termination being in progress. Conversely, if the handle is dropped without closing (see previous issue), `is_finished()` may never become true even though the handle is gone and the task may be stuck forever. This can mislead callers relying on `is_finished()` to reflect completion state and cause incorrect control flow or resource management. <b>[ Low confidence ]</b>
- [line 205](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/common/src/stream_handles.rs#L205): Dropping a `WasmStreamHandle` without calling `end()` leaks the spawned task: when the `mpsc::Sender` (`closer`) is dropped, `CloserHandle::poll` returns `Poll::Pending` on `Poll::Ready(None)` and the `select` never resolves via the closer branch. If the `future` also never completes, the spawned task will stay alive for the duration of the program, never sending on the `oneshot` result channel, and cannot be cancelled anymore because the handle was dropped. This is a lifecycle/cancellation bug: dropping the handle commonly implies cancellation, but here it guarantees the opposite (task retention). The code path is reachable whenever a caller drops the handle without explicitly calling `end()`. <b>[ Low confidence ]</b>
- [line 278](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/common/src/stream_handles.rs#L278): Infinite recursion in `impl AbortHandle for tokio::task::AbortHandle` for method `is_finished`: the implementation `fn is_finished(&self) -> bool { self.is_finished() }` calls itself, causing unbounded recursion and a stack overflow at runtime whenever `AbortHandle::is_finished` is invoked. If the intent was to forward to an inherent method on `tokio::task::AbortHandle`, it must use fully-qualified syntax (e.g., `<tokio::task::AbortHandle>::is_finished(self)`), and such an inherent method must actually exist. If there is no inherent `is_finished` on Tokio's type, this method should be implemented using a correct runtime check or removed. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_api/src/debug_wrapper.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 70](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api/src/debug_wrapper.rs#L70): Double-wrapping errors that already include stats: when `req().await` returns `Err(ApiClientError::ErrorWithStats { .. })`, the catch-all arm `e => Err(ApiClientError::ErrorWithStats { e: Box::new(e), stats: stats(), })` will wrap the existing `ErrorWithStats` inside a new `ErrorWithStats`, causing nested error structures and duplicated stats. This can lead to confusing and ever-growing error nesting if upstream callers re-apply `wrap_err`, and violates at-most-once addition of stats. A dedicated match arm should return `Err(err)` unchanged for `ApiClientError::ErrorWithStats { .. }`, similar to the handling for `ClientWithEndpointAndStats`. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/middleware/multi_node_client/gateway_api.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 187](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/middleware/multi_node_client/gateway_api.rs#L187): TLS detection in `validate_tls_guard` treats only the `https` scheme as TLS by checking `url.scheme() == "https"`. This misclassifies other valid TLS URL schemes (e.g., `grpcs`, potentially `wss` if supported by the system) as non-TLS. As a result, node URLs that are legitimately secured but use a different TLS-denoting scheme will be rejected with `MultiNodeClientError::TlsChannelMismatch`, even though the template's `tls_channel` is correctly set to require TLS. This can cause all node clients to fail to build if the gateway returns such URLs. Consider recognizing a broader set of TLS schemes or explicitly mapping supported schemes to TLS/non-TLS. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/traits/cursor_store.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 64](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/protocol/traits/cursor_store.rs#L64): The default implementation of `latest_for_originator` silently assumes that `latest_per_originator(topic, &[originator])` will always include the requested `originator`. It immediately reads `sid` via `GlobalCursor::get(originator)` and constructs a `Cursor` without validating presence. If the implementor returns a `GlobalCursor` that does not contain the given `originator` (which is plausible under this trait’s contract, e.g., missing data or partial results), this path can either: (a) return a default/zero sequence ID (if `get` returns a default), silently misrepresenting state; or (b) panic (if `get` expects presence). Either outcome is a runtime bug: missing originators must be explicitly handled (e.g., returning `CursorStoreError` or a documented fallback) rather than silently defaulting or panicking. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/queries/d14n/identity.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 96](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/identity.rs#L96): In `get_identity_updates_v2`, the method only returns responses for inbox IDs that have at least one update. Requested inbox IDs that have zero updates are omitted entirely. If the external API contract expects a response entry for every requested inbox ID (with an empty `updates` list when there are none), this omission creates ambiguity and forces callers to infer absence vs zero updates from missing entries. <b>[ Out of scope ]</b>
- [line 134](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/identity.rs#L134): In `get_inbox_ids`, every mapped response sets `identifier_kind` to `IdentifierKind::Ethereum` regardless of the actual identifier kind. This silently mislabels passkey-derived inbox IDs (and any future kinds) as Ethereum. The mapping at `responses: res.responses.iter().map(...)` should preserve the kind returned by the V4 response or be derived consistently from the original request, rather than hardcoding `IdentifierKind::Ethereum` for all entries. <b>[ Out of scope ]</b>
- [line 162](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/identity.rs#L162): In `verify_smart_contract_wallet_signatures`, a single verifier error (`self.scw_verifier.is_valid_signature(...).await`) aborts the entire batch with a top-level error, discarding any other signature validations. The response type supports per-signature `error` in `ValidationResponse`, suggesting the API may expect partial results even when some validations fail. Current behavior prevents returning per-item errors for the failing signature(s) and any successful validations for others, which can break batch semantics. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/queries/d14n/mls.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 90](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/mls.rs#L90): `send_group_messages` performs per-envelope publishing in a loop without transactional guarantees or compensating actions (lines 90-97). If any publish fails mid-loop, earlier messages have already been published, and the method returns an error with a partially applied effect. This can leave the system in an inconsistent state relative to the caller's expectation of "send all or error" and provides no visibility about which messages succeeded. A similar issue exists in `send_welcome_messages` (lines 109-116). Until batch publishing is implemented, this path should either document and surface partial successes (e.g., per-message results) or add best-effort retries/compensation to reduce inconsistency. <b>[ Low confidence ]</b>
- [line 109](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/mls.rs#L109): `send_welcome_messages` also performs per-envelope publishing in a loop (lines 109-116) with the same partial side-effect behavior as `send_group_messages`. A failure mid-loop returns an error after some messages have already been published, with no per-message result reporting. This duplicates the consistency risk described above. <b>[ Low confidence ]</b>
- [line 157](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/mls.rs#L157): `query_latest_group_message` returns `Err` when the newest envelope exists as a response object but contains no inner message, instead of returning `Ok(None)`. The function only checks `response.results.is_empty()` (line 157) and then unconditionally calls `accept` and `get` on `GroupMessageExtractor` for the first response (lines 160-166). If the first `Response` has `originator_envelope: None` (a reachable case, as evidenced by tests for similar extractors), `GroupMessageExtractor` is likely to surface an `EnvelopeError` rather than a `None` value. This violates the external contract implied by the method's `Result<Option<GroupMessage>, _>` return type, which should use `None` to represent absence of a message. The method needs to detect and return `Ok(None)` when the inner envelope is missing, not just when `results` is empty. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/queries/d14n/streams.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 39](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/streams.rs#L39): All three subscription methods (`subscribe_group_messages`, `subscribe_group_messages_with_cursors`, and `subscribe_welcome_messages`) accept slices (`group_ids`, `groups_with_cursors`, `installations`) without validating that they are non-empty. If an empty slice is passed, the code constructs `topics` as an empty `Vec` and proceeds to build a `SubscribeEnvelopes` request with empty topics and, in the with-cursors path, an empty `GlobalCursor`. Depending on the contract of `SubscribeEnvelopes::builder().topics(...)` and the downstream service, this can yield a malformed subscription (rejected by the builder, by the client, or by the server) or undefined behavior (e.g., an error or a stream that never yields). The current implementation does not provide a defined terminal state or explicit error for empty inputs, making this a reachable runtime path with ambiguous outcomes. <b>[ Out of scope ]</b>
- [line 66](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/streams.rs#L66): In `subscribe_group_messages_with_cursors`, the code comment says "Compute the lowest common cursor from the provided cursors", but the implementation builds `min_clock` over the union of all node IDs across cursors, taking the per-node minimum sequence ID. A 'common' cursor would typically be defined over the intersection of node IDs present in all provided cursors. Using the union contradicts the comment and could produce a `GlobalCursor` containing nodes that are not common to all topics, which may be semantically incorrect for consumers expecting a truly 'common' cursor and may be rejected or mishandled downstream. This contradiction creates uncertainty about intended behavior and risks incorrect or malformed subscriptions. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/queries/d14n/xmtp_query.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 33](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/xmtp_query.rs#L33): The handling of `at: Option<GlobalCursor>` uses `unwrap_or_default()` when `None` is passed, forcing a non-absent `GlobalCursor` into the query via `.last_seen(at.unwrap_or_default())`. This silently converts "no cursor" into a specific default cursor value. If `GlobalCursor::default()` is not semantically equivalent to "no cursor" (e.g., it represents the beginning of history or some sentinel that changes query semantics), this can cause incorrect query ranges, excessive data retrieval, or missed results compared to omitting `last_seen`. This violates the principle of preserving the caller's intent for `None` as absent and can lead to subtle runtime misbehavior rather than a clear, handled fallback. <b>[ Low confidence ]</b>
- [line 34](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/d14n/xmtp_query.rs#L34): The call `.limit(MAX_PAGE_SIZE)` blindly applies a global constant without validating bounds or ensuring it is within the server/API’s acceptable range. If `MAX_PAGE_SIZE` is zero, negative (if type permits), or exceeds the maximum allowed by the API, the query may fail at runtime (server-side rejection) or behave unexpectedly (e.g., zero results). There is no explicit guard or fallback to a safe default, making this path prone to runtime errors dependent on configuration. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/queries/v3/identity.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 44](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/v3/identity.rs#L44): In `get_inbox_ids`, requests with `identifier_kind` values other than `IdentifierKind::Ethereum` or `IdentifierKind::Passkey` are silently dropped. The code filters to addresses and passkeys only and ignores all other kinds without error or visible outcome: `filter(|r| r.identifier_kind == IdentifierKind::Ethereum as i32)` and `filter(|r| r.identifier_kind == IdentifierKind::Passkey as i32)`. If the input includes other supported or future `IdentifierKind` variants, those entries will be lost, producing incomplete results without any explicit failure or warning. This violates the requirement to ensure every input reaches a defined terminal state and to avoid silently discarding cases that are not proven unreachable by guards. The method should either explicitly reject unknown kinds with a clear error or handle them appropriately, rather than dropping them. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/queries/v3/xmtp_query.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 87](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_api_d14n/src/queries/v3/xmtp_query.rs#L87): `unreachable!()` in the default match arm for `topic.kind()` will cause a runtime panic if a `TopicKind` other than the explicitly handled ones (`GroupMessagesV1`, `WelcomeMessagesV1`, `IdentityUpdatesV1`, `KeyPackagesV1`) is passed. There is no guard in this function ensuring that only those four kinds are ever supplied, and per the contract you must treat inputs as reachable unless a concrete guard exists. This makes the function prone to crashing at runtime when new or unexpected topic kinds appear. Replace `unreachable!()` with a defined error outcome or a graceful fallback so that every input reaches a defined terminal state without runtime termination. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_archive/src/export_stream.rs — 0 comments posted, 3 evaluated, 2 filtered</summary>

- [line 66](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_archive/src/export_stream.rs#L66): `BatchExportStream::next` reverses the order of elements within each batch because it stores the entire batch in `self.buffer` and then yields elements using `self.buffer.pop()`. If `R::backup_records` returns elements in a semantically ordered sequence (e.g., chronological), popping from the end reverses that order, potentially violating ordering constraints (the code already comments that order matters for certain elements). Consider using `remove(0)` or pushing into the buffer in reverse so that `pop()` preserves the intended order. <b>[ Low confidence ]</b>
- [line 150](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_archive/src/export_stream.rs#L150): `self.cursor` uses `i64` and is increased by `R::BATCH_SIZE` each iteration without bounds checks. Over extremely long runs, this can overflow, causing wrap-around to negative values in release builds and leading to undefined paging behavior (potential infinite loops, missed or duplicated ranges). Consider using checked arithmetic (`checked_add`) with a clear error or termination when nearing limits. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_archive/src/exporter.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 54](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_archive/src/exporter.rs#L54): Unchecked `key` length can cause a runtime panic when constructing the cipher in `Self::new(options, db, key)`. The `new` method calls `Aes256Gcm::new(GenericArray::from_slice(key))`, which requires a 32-byte key; if `key.len() != 32`, `GenericArray::from_slice` panics at runtime. `export_to_file` does not validate `key` before calling `new`, so callers can trigger a panic rather than a recoverable error. To fix, enforce the key length (e.g., return an error if `key.len() != 32`) before calling `new`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/mod.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 249](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_db/src/encrypted_store/mod.rs#L249): Potential panic due to unchecked indexing into `sqlite_version` vector. In `init`, after executing `sql_query("SELECT sqlite_version() AS version").load::<SqliteVersion>(conn)?;`, the code immediately logs `sqlite_version[0].version` without verifying that the result set is non-empty. If the query returns zero rows (e.g., due to an unexpected driver behavior, misconfiguration, or a different SQLite build), this will result in an out-of-bounds index and panic at runtime. Add an explicit check for `sqlite_version.is_empty()` and handle it gracefully (e.g., return a meaningful error or fallback log). <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_db_test/src/chaos.rs — 0 comments posted, 2 evaluated, 1 filtered</summary>

- [line 167](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_db_test/src/chaos.rs#L167): `static_start_transaction_hook` and `start_transaction_hook` registrations are never executed. While the code defines `STATIC_TRANSACTION_START_HOOK` and `TRANSACTION_START_HOOK` and provides methods to register these hooks, there is no call site that actually runs them (no `run_static_hooks(STATIC_TRANSACTION_START_HOOK)` or `run_hook(TRANSACTION_START_HOOK)` anywhere). The only executed hooks in this implementation are read/write pre/post hooks. This leads to a functional mismatch: users registering transaction start hooks will observe that they never fire. To fix, ensure that the transaction start hooks are invoked at the appropriate time (e.g., in a transaction-starting method) or explicitly remove/deprecate the unused APIs. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_db_test/src/lib.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 90](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_db_test/src/lib.rs#L90): Unconditional `unwrap()` on `self.db.init()` can panic at runtime if initialization fails. This occurs during builder construction and there is no error propagation nor cleanup for any partially-initialized resources. If `init()` opens connections or allocates resources, a panic here can leave them in an indeterminate state without a paired `disconnect`/cleanup. Consider returning a `Result` and handling failure paths explicitly. <b>[ Out of scope ]</b>
- [line 96](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_db_test/src/lib.rs#L96): Unconditional `unwrap()` on `chaos::ChaosConnection::builder().db(self.db.conn()).error_frequency(self.error_frequency).build()` can panic at runtime if the chaos builder rejects the provided configuration (e.g., invalid `error_frequency` range or a failing underlying `conn()`). There is no input validation for `error_frequency` (e.g., negative, >1, or NaN), and no graceful error propagation. <b>[ Out of scope ]</b>
- [line 105](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_db_test/src/lib.rs#L105): Unconditional `unwrap()` on `EncryptedMessageStore::<ChaosDb<Db>>::builder().db(chaos_db).build()` can panic if the store builder fails (e.g., due to invalid DB configuration or missing dependencies). This builder returns `Result` and the code discards error handling, risking a runtime crash and lack of cleanup for allocated resources. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_id/src/scw_verifier/mod.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 33](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_id/src/scw_verifier/mod.rs#L33): The `MalformedEipUrl` error message (`#[error("URLs must be preceded with eip144:")]`) is misleading and contradicts the implementation. The code treats `id` as an EIP-155 chain ID string like `"eip155:1"`, not a URL, and it uses the `eip155` namespace rather than `eip144`. This creates confusion about the expected input format and may lead to misdiagnosis when the error is emitted. <b>[ Out of scope ]</b>
- [line 183](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_id/src/scw_verifier/mod.rs#L183): Fragile parsing of chain ID in `MultiSmartContractSignatureVerifier::upgrade`: `let eip_id = id.split(":").nth(1).ok_or(VerifierError::MalformedEipUrl)?;` assumes exactly one `:` in `id` (e.g., `"eip155:1"`). If `id` contains zero or multiple colons (e.g., user-provided IDs via `add_verifier` like `"namespace:eip155:1"`), `nth(1)` will produce the wrong segment or error. This can lead to constructing the wrong environment variable name (`CHAIN_RPC_{eip_id}`) or spuriously returning `MalformedEipUrl`. There is no validation that the prefix is `eip155`, and no robust split. A safer approach is to validate the namespace and split only once at the first colon. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/builder.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 271](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/builder.rs#L271): Inconsistent use of normalized vs. raw identity values. The code logs `identity.inbox_id()` (method) and later passes `identity.inbox_id.as_str()` (field) to `load_identity_updates`. If the method `inbox_id()` applies normalization/transformation (e.g., canonicalization, trimming, case-folding) while the raw field `inbox_id` does not, this creates a subtle inconsistency where different forms of the value are used for logging and for API/DB operations. At data-conversion boundaries, the canonicalized form should be used consistently. Using the raw field directly can lead to lookup failures or divergence if the method is intended to supply the canonical value. <b>[ Out of scope ]</b>
- [line 297](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/builder.rs#L297): Potential strong reference cycle causing resource leaks: `context` stores a clone of `workers` (`context.workers: workers.clone()`), and each `register_new_worker::<...>(context.clone())` likely causes `WorkerRunner` to retain a strong `Arc` to `context` for executing work. If `WorkerRunner` holds strong references to the passed `context` (common in worker registries), then `context -> workers` and `workers -> context` creates a cycle that prevents both from being dropped when the `Client` is dropped, leaking threads/resources. Without an explicit teardown or use of `Weak` references inside `WorkerRunner`, this is a runtime leak risk. <b>[ Out of scope ]</b>
- [line 309](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/builder.rs#L309): Global `EVENTS_ENABLED` is set to `true` when `!disable_events`, but there is no corresponding reset or teardown anywhere in `build` for later clients that might specify `disable_events = true`. This persistent mutation of a global flag can cause subsequent `Client` instances in the same process to observe events being enabled even when they requested events to be disabled, violating per-client configuration and leading to inconsistent behavior across instances. Since the flag is set only when `!disable_workers` as well, a later client constructed with `disable_workers = true` cannot unset the flag, further compounding the inconsistency. This is a runtime configuration leak across instances, not a compile-time issue. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/subscriptions.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 110](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/groups/subscriptions.rs#L110): The readiness oneshot (`tx`) is only sent on successful stream creation, and never sent in the failure path. If the `StreamHandle` or surrounding infrastructure expects to receive a readiness signal (either a successful `send(())` or an explicit error) to transition state, omitting `tx.send(())` on failure can lead to a mismatch or race conditions in the handle's lifecycle. Specifically, on error at lines `104..107`, the code calls `on_close()` and returns without sending a readiness signal. Depending on `StreamHandle` semantics, callers awaiting readiness may not receive the intended signal, and could be left in an inconsistent state. <b>[ Low confidence ]</b>
- [line 116](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/groups/subscriptions.rs#L116): Errors during stream setup and termination are silently swallowed and never propagated via the returned `StreamHandle`'s `StreamOutput`. In the failure path when `StreamGroupMessages::new` returns `Err`, the code logs a warning, calls `on_close()`, and then returns `Ok(())` instead of an error. In the normal termination path after the stream ends, it also returns `Ok(())`. As a result, callers cannot distinguish between a normal stream completion and an error-induced closure, which can lead to silent data loss and prevent appropriate retries or error handling by the caller. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/subscriptions/mod.rs — 0 comments posted, 12 evaluated, 11 filtered</summary>

- [line 330](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L330): Readiness signal is never sent when `client.stream_conversations(...)` fails, which can leave the associated `rx` in `xmtp_common::spawn(Some(rx), ...)` permanently pending. The code only calls `tx.send(())` after successfully creating and pinning the stream (line 338). In the early error path (lines 330–335), it logs, calls `on_close()`, and returns without sending on `tx`. If `xmtp_common::spawn` relies on this readiness signal to mark the task as started or to unblock the handle, the caller may observe a hung or unclosable stream handle, or other startup synchronization issues. <b>[ Low confidence ]</b>
- [line 332](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L332): On the error path when creating the stream fails (e.g., `client.stream_conversations(...).await` or `StreamAllMessages::new(...).await` returns `Err`), `on_close` is invoked but the 'ready' oneshot (`tx`) is never signaled. Although `tx` will be dropped when the future returns, depending on how `xmtp_common::spawn` and the associated `StreamHandle` use this channel, the handle may temporarily await readiness and only observe a cancellation after the task exits. If the handle expects a positive ready signal to transition, this can lead to a brief hang or racey behavior. To make readiness signaling unambiguous, consider sending an explicit terminal signal on the ready channel (e.g., using a separate error path) or document that a dropped `tx` indicates initialization failure. <b>[ Out of scope ]</b>
- [line 340](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L340): User-supplied callbacks `convo_callback`, `callback` (for messages), `callback` (for consent), and `callback` (for preferences) are invoked without panic containment. If a callback panics, the async task will unwind and terminate immediately, and `on_close` will not be invoked. This violates the implied cleanup invariant that `on_close` is called exactly once when the stream terminates, and can leave external state in an inconsistent or hanging state (e.g., user expecting a close notification). Consider wrapping callback invocations in `std::panic::catch_unwind` (with `AssertUnwindSafe` as needed) to ensure `on_close` is invoked even on callback panic, or clearly document that panic in callbacks forfeits `on_close` execution. <b>[ Out of scope ]</b>
- [line 343](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L343): `on_close()` is not guaranteed to be invoked on all termination paths. Specifically, if `convo_callback` panics (line 340) or the task is cancelled externally (e.g., via the returned `StreamHandle`), the async task will unwind or be aborted and the code path that calls `on_close()` (line 343) will not execute, violating at-most-once and terminal notification guarantees. There is no `catch_unwind` or scoped guard to ensure `on_close` runs on panic/cancellation. This can leave FFI callers (see example usage) without a closure notification, breaking lifecycle contracts. <b>[ Low confidence ]</b>
- [line 343](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L343): `on_close` is called after the stream ends, but there is no guard against multiple terminal events or duplicated `on_close` calls if upstream stream mistakenly yields `None` followed by further items due to a bug or cancellation/resumption. While typical `Stream` contracts prevent this, defensive coding could ensure `on_close` is at-most-once by tracking a closed flag to avoid double invocation if future changes introduce re-entrancy. As written, current paths appear single-invocation, but lack explicit at-most-once guard. <b>[ Out of scope ]</b>
- [line 343](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L343): Lifecycle callback `on_close` is not guaranteed to run on task cancellation. The async task is spawned via `xmtp_common::spawn` with a `oneshot::Receiver` for a start signal, but there is no explicit cancellation token checked inside the loop. If the `StreamHandle` cancels the task (e.g., by dropping/aborting the spawned future), the future may be terminated without reaching the end-of-stream path that calls `on_close`. This can leave clients without a close notification on cancellation. To preserve lifecycle contract parity, ensure cancellation invokes `on_close` exactly once, either by using a cancellation-aware stream that yields a terminal event, adding an explicit cancellation listener in the task, or arranging for the spawning mechanism to invoke `on_close` on cancellation. <b>[ Out of scope ]</b>
- [line 344](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L344): The returned stream output is always `Ok(())` regardless of whether the stream fails to be created (lines 330–335) or ends due to an error later. This function logs and delegates per-conversation errors via `convo_callback`, but it does not propagate any error through the `StreamHandle<StreamOutput = Result<()>>`. If downstream code relies on the `StreamHandle` emitting `Err(...)` on failure to create or premature termination, this suppresses error signaling and can lead to silent failures from the perspective of the handle consumer. <b>[ Code style ]</b>
- [line 402](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L402): The readiness `oneshot` signal (`tx.send(())`) is only sent on the success path after the stream is created and pinned. On the error path when `StreamAllMessages::new` fails, `tx` is dropped without sending any explicit error or success signal, leaving the receiver with a closed channel. If upstream waits specifically for a positive readiness signal to proceed, this asymmetric signaling can produce ambiguous outcomes and potentially hang or mis-handle startup (depending on how `xmtp_common::spawn` and the caller interpret a closed `rx`). To preserve a defined terminal state and explicit outcome for the readiness handshake, send an explicit error or a distinct signal on failure rather than dropping the channel silently. <b>[ Low confidence ]</b>
- [line 408](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L408): The `on_close` callback is not guaranteed to be invoked on all exit paths: if the `callback` passed to `stream_all_messages_with_callback` panics during message handling, the async task will unwind and terminate without calling `on_close`. This violates the invariant that cleanup callbacks fire exactly once on completion or failure, and can leave upstream state waiting for a close signal indefinitely. A scoped guard (e.g., a `Drop`-based guard) or `catch_unwind` wrapper around the callback dispatch is needed to ensure `on_close` is invoked even if the `callback` panics. <b>[ Low confidence ]</b>
- [line 425](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L425): The result of `tx.send(())` from the `oneshot::channel` is ignored. If the receiver side `rx` is dropped early (e.g., due to an error or a miscoordination in `xmtp_common::spawn`), `send` will return `Err(())`. Ignoring this can lead to a readiness coordination mismatch: the returned `StreamHandle` might be waiting for the readiness signal or might observe a cancellation while the task continues processing. At best it's harmless, but at worst it can cause the handle to never become "ready" or to proceed with incorrect assumptions, depending on how `spawn(Some(rx), ...)` uses `rx`. This is a runtime coordination bug risk. <b>[ Previously rejected ]</b>
- [line 430](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/subscriptions/mod.rs#L430): `on_close` is only invoked after the `while let Some(message)` loop completes. If the task is cancelled (e.g., the returned `StreamHandle`/closer is dropped or closed) or if the user-provided `callback` panics, the future is dropped and the code after the loop does not run. In those cases, `on_close` will never be called, breaking the expected cleanup/notification contract and causing the JS-side `on_close` to be skipped (as shown in example usage). This violates the requirement that callbacks be consumed exactly once and that cleanup happen on all exit paths, including cancellation or panic. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/utils/test/mod.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 76](https://github.com/xmtp/libxmtp/blob/1d16a56cb5a134b85fde2dbb35b06df8e194c4e1/xmtp_mls/src/utils/test/mod.rs#L76): The method `build_unchecked` calls `self.build().await.unwrap()` at runtime. If `self.build()` returns an `Err`, `unwrap()` will panic, causing abrupt termination. Because this is executed after awaiting an async operation, there is a risk that partial resources created by `build()` (e.g., connections or handles) are left without cleanup on the panic path. This helper is not guarded by `#[cfg(test)]`, so it is available in non-test builds as well. Prefer returning the `Result` or using an explicit error handling path to avoid panicking, or clearly constrain the function to test-only usage where panic is acceptable pre-effect. If panicking is intended, ensure `build()` performs no effects prior to returning `Err` to avoid leaks. <b>[ Code style ]</b>
</details>

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -206,3 +206,4 @@ diesel_migrations = { git = "https://github.com/ephemerahq/diesel", branch = "co
 
 [workspace.lints.clippy]
 uninlined_format_args = "allow"
+arc_with_non_send_sync = "allow"
--- a/bindings_node/src/streams.rs
+++ b/bindings_node/src/streams.rs
@@ -18,7 +18,7 @@ pub struct StreamCloser {
 
 impl StreamCloser {
   pub fn new(
-    handle: impl XmtpStreamHandle<StreamOutput = Result<(), SubscribeError>> + Send + Sync + 'static,
+    handle: impl XmtpStreamHandle<StreamOutput = Result<(), SubscribeError>> + 'static,
   ) -> Self {
     let abort = handle.abort_handle();
     Self {
--- a/bindings_wasm/src/client.rs
+++ b/bindings_wasm/src/client.rs
@@ -252,7 +252,6 @@ pub async fn create_client(
 
   Ok(Client {
     account_identifier,
-    #[allow(clippy::arc_with_non_send_sync)]
     inner_client: Arc::new(xmtp_client),
     app_version,
   })
--- a/common/src/retry.rs
+++ b/common/src/retry.rs
@@ -17,6 +17,7 @@
 //! ```
 
 use crate::time::Duration;
+use crate::{MaybeSend, MaybeSync};
 use rand::Rng;
 use std::error::Error;
 use std::sync::Arc;
@@ -36,17 +37,13 @@ pub fn arc_retryable_to_error(retryable: Arc<dyn RetryableError>) -> Arc<dyn Err
     retryable
 }
 
-#[cfg(not(target_arch = "wasm32"))]
-pub type BoxedRetry = Retry<Box<dyn Strategy + Send + Sync>>;
-
-#[cfg(target_arch = "wasm32")]
 pub type BoxedRetry = Retry<Box<dyn Strategy>>;
 
 pub struct NotSpecialized;
 
 /// Specifies which errors are retryable.
 /// All Errors are not retryable by-default.
-pub trait RetryableError<SP = NotSpecialized>: std::error::Error {
+pub trait RetryableError<SP = NotSpecialized>: std::error::Error + MaybeSend + MaybeSync {
     fn is_retryable(&self) -> bool;
 }
 
@@ -108,7 +105,7 @@ impl<S: Strategy + 'static> Retry<S> {
 }
 
 /// The strategy interface
-pub trait Strategy {
+pub trait Strategy: MaybeSend + MaybeSync {
     /// A time that this retry should backoff
     /// Returns None when we should no longer backoff,
     /// despite possibly being below attempts
--- a/common/src/stream_handles.rs
+++ b/common/src/stream_handles.rs
@@ -1,11 +1,14 @@
 //! Consistent Stream behavior between WebAssembly and Native utilizing `tokio::task::spawn` in native and
 //! `wasm_bindgen_futures::spawn` for web.
 
-#[cfg(target_arch = "wasm32")]
-pub type GenericStreamHandle<O> = dyn StreamHandle<StreamOutput = O>;
+use crate::{MaybeSend, MaybeSync, if_native, if_wasm};
 
-#[cfg(not(target_arch = "wasm32"))]
-pub type GenericStreamHandle<O> = dyn StreamHandle<StreamOutput = O> + Send + Sync;
+if_wasm! {
+    pub type GenericStreamHandle<O> = dyn StreamHandle<StreamOutput = O>;
+}
+if_native! {
+    pub type GenericStreamHandle<O> = dyn StreamHandle<StreamOutput = O> + Send + Sync;
+}
 
 #[derive(thiserror::Error, Debug)]
 pub enum StreamHandleError {
@@ -27,7 +30,7 @@ pub enum StreamHandleError {
 #[allow(async_fn_in_trait)]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
-pub trait StreamHandle {
+pub trait StreamHandle: MaybeSend + MaybeSync {
     /// The Output type for the stream
     type StreamOutput;
 
@@ -57,20 +60,14 @@ pub trait StreamHandle {
 }
 
 /// A handle that can be moved/cloned/sent, but can only close the stream.
-pub trait AbortHandle: Send + Sync {
+pub trait AbortHandle: crate::MaybeSend + crate::MaybeSync {
     /// Send a signal to end the stream, without waiting for a result.
     fn end(&self);
     fn is_finished(&self) -> bool;
 }
 
-#[cfg(not(target_arch = "wasm32"))]
-pub use native::*;
-
-#[cfg(target_arch = "wasm32")]
-#[allow(unused)]
+if_wasm! {
 pub use wasm::*;
-
-#[cfg(target_arch = "wasm32")]
 mod wasm {
     use std::{
         future::Future,
@@ -209,9 +206,10 @@ mod wasm {
             }
         }
     }
-}
+}}
 
-#[cfg(not(target_arch = "wasm32"))]
+if_native! {
+pub use native::*;
 mod native {
     use super::*;
     use std::future::Future;
@@ -309,4 +307,4 @@ mod native {
             ready,
         }
     }
-}
+}}
--- a/common/src/wasm.rs
+++ b/common/src/wasm.rs
@@ -11,11 +11,11 @@ pub use tokio::*;
 crate::if_wasm! {
     /// Marker trait to determine whether a type implements `Send` or not.
     pub trait MaybeSend {}
-    impl<T> MaybeSend for T {}
+    impl<T: ?Sized> MaybeSend for T {}
 
     /// Marker trait to determine whether a type implements `Send` or not.
     pub trait MaybeSync {}
-    impl<T> MaybeSync for T {}
+    impl<T: ?Sized> MaybeSync for T {}
 
     /// Global Marker trait for WebAssembly
     pub trait Wasm {}
@@ -24,20 +24,26 @@ crate::if_wasm! {
     pub struct StreamWrapper<'a, I> {
         inner: Pin<Box<dyn Stream<Item = I> + 'a>>,
     }
+
+    pub type BoxDynError = Box<dyn std::error::Error>;
+
 }
 
 crate::if_native! {
     /// Marker trait to determine whether a type implements `Send` or not.
     pub trait MaybeSend: Send {}
-    impl<T: Send> MaybeSend for T {}
+    impl<T: Send + ?Sized> MaybeSend for T {}
 
     /// Marker trait to determine whether a type implements `Sync` or not.
-    pub trait MaybeSync {}
-    impl<T: Sync> MaybeSync for T {}
+    pub trait MaybeSync: Sync {}
+    impl<T: Sync + ?Sized> MaybeSync for T {}
 
     pub struct StreamWrapper<'a, I> {
         inner: Pin<Box<dyn Stream<Item = I> + Send + 'a>>,
     }
+
+    pub type BoxDynError = Box<dyn std::error::Error + Send + Sync>;
+
 }
 
 impl<I> Stream for StreamWrapper<'_, I> {
--- a/xmtp_api/src/debug_wrapper.rs
+++ b/xmtp_api/src/debug_wrapper.rs
@@ -54,7 +54,7 @@ async fn wrap_err<T, R, F, E>(
 where
     R: FnOnce() -> F,
     F: Future<Output = Result<T, ApiClientError<E>>>,
-    E: std::error::Error + RetryableError + Send + Sync + 'static,
+    E: RetryableError + 'static,
 {
     let res = req().await;
     if let Err(e) = res {
@@ -98,8 +98,8 @@ where
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 impl<A, E> XmtpMlsClient for ApiDebugWrapper<A>
 where
-    A: XmtpMlsClient<Error = ApiClientError<E>> + Send + Sync,
-    E: std::error::Error + RetryableError + Send + Sync + 'static,
+    A: XmtpMlsClient<Error = ApiClientError<E>>,
+    E: RetryableError + 'static,
     A: HasStats,
 {
     type Error = ApiClientError<E>;
@@ -219,8 +219,8 @@ where
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 impl<A, E> XmtpMlsStreams for ApiDebugWrapper<A>
 where
-    A: XmtpMlsStreams<Error = ApiClientError<E>> + Send + Sync + 'static,
-    E: std::error::Error + RetryableError + Send + Sync + 'static,
+    A: XmtpMlsStreams<Error = ApiClientError<E>>,
+    E: RetryableError + 'static,
     A: HasStats,
 {
     type GroupMessageStream = <A as XmtpMlsStreams>::GroupMessageStream;
@@ -270,8 +270,8 @@ where
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 impl<A, E> XmtpIdentityClient for ApiDebugWrapper<A>
 where
-    A: XmtpIdentityClient<Error = ApiClientError<E>> + Send + Sync,
-    E: std::error::Error + RetryableError + Send + Sync + 'static,
+    A: XmtpIdentityClient<Error = ApiClientError<E>>,
+    E: RetryableError + 'static,
     A: HasStats,
 {
     type Error = ApiClientError<E>;
--- a/xmtp_api/src/identity.rs
+++ b/xmtp_api/src/identity.rs
@@ -63,7 +63,7 @@ where
     ) -> Result<impl Iterator<Item = (String, Vec<T>)>>
     where
         T: TryFrom<IdentityUpdateLog>,
-        <T as TryFrom<IdentityUpdateLog>>::Error: RetryableError + Send + Sync + 'static,
+        <T as TryFrom<IdentityUpdateLog>>::Error: RetryableError + 'static,
     {
         if filters.is_empty() {
             return Ok(vec![].into_iter());
--- a/xmtp_api/src/lib.rs
+++ b/xmtp_api/src/lib.rs
@@ -29,14 +29,14 @@ pub mod strategies {
 }
 
 // Erases Api Error type (which may be Http or Grpc)
-pub fn dyn_err(e: impl RetryableError + Send + Sync + 'static) -> ApiError {
+pub fn dyn_err(e: impl RetryableError + 'static) -> ApiError {
     ApiError::Api(Box::new(e))
 }
 
 #[derive(Debug, thiserror::Error)]
 pub enum ApiError {
     #[error("api client error {0}")]
-    Api(Box<dyn RetryableError + Send + Sync>),
+    Api(Box<dyn RetryableError>),
     #[error(
         "mismatched number of results, key packages {} != installation_keys {}",
         .key_packages,
--- a/xmtp_api/src/scw_verifier.rs
+++ b/xmtp_api/src/scw_verifier.rs
@@ -11,7 +11,7 @@ use xmtp_proto::xmtp::identity::api::v1::VerifySmartContractWalletSignaturesResp
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 impl<C> SmartContractSignatureVerifier for ApiClientWrapper<C>
 where
-    C: Send + Sync + XmtpIdentityClient,
+    C: XmtpIdentityClient,
 {
     /// Verifies an ERC-6492<https://eips.ethereum.org/EIPS/eip-6492> signature.
     ///
--- a/xmtp_api_d14n/src/middleware/multi_node_client/errors.rs
+++ b/xmtp_api_d14n/src/middleware/multi_node_client/errors.rs
@@ -1,6 +1,6 @@
 use thiserror::Error;
 use xmtp_api_grpc::error::GrpcError;
-use xmtp_common::RetryableError;
+use xmtp_common::{MaybeSend, MaybeSync, RetryableError};
 use xmtp_proto::api::{ApiClientError, BodyError};
 
 /// Errors that can occur during multi-node client operations.
@@ -31,7 +31,7 @@ pub enum MultiNodeClientError {
 /// Required by the Client trait implementation, as request and stream can return MultiNodeClientError.
 impl<E> From<MultiNodeClientError> for ApiClientError<E>
 where
-    E: std::error::Error + Send + Sync + 'static,
+    E: std::error::Error + MaybeSend + MaybeSync + 'static,
 {
     fn from(value: MultiNodeClientError) -> ApiClientError<E> {
         ApiClientError::<E>::Other(Box::new(value))
--- a/xmtp_api_d14n/src/middleware/multi_node_client/gateway_api.rs
+++ b/xmtp_api_d14n/src/middleware/multi_node_client/gateway_api.rs
@@ -5,7 +5,10 @@ use crate::{
 use futures::StreamExt;
 use std::collections::HashMap;
 use xmtp_api_grpc::{ClientBuilder, GrpcClient};
-use xmtp_common::time::{Duration, Instant};
+use xmtp_common::{
+    BoxDynError,
+    time::{Duration, Instant},
+};
 use xmtp_proto::api::Query;
 use xmtp_proto::api_client::ApiBuilder;
 use xmtp_proto::{ApiEndpoint, api::ApiClientError};
@@ -52,7 +55,7 @@ pub async fn get_nodes(
 
             let client = client_builder.build().map_err(|e| (node_id, e.into()))?;
 
-            Ok::<_, (u32, Box<dyn std::error::Error + Send + Sync>)>((node_id, client, url))
+            Ok::<_, (u32, BoxDynError)>((node_id, client, url))
         }))
         .buffer_unordered(max_concurrency);
 
@@ -175,19 +178,16 @@ pub async fn get_fastest_node(
 }
 
 /// Validate that the template's TLS configuration matches the URL scheme.
-pub fn validate_tls_guard(
-    template: &ClientBuilder,
-    url: &str,
-) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+pub fn validate_tls_guard(template: &ClientBuilder, url: &str) -> Result<(), BoxDynError> {
     let url_is_tls = url
         .parse::<url::Url>()
-        .map_err(|e| -> Box<dyn std::error::Error + Send + Sync> { Box::new(e) })?
+        .map_err(|e| -> BoxDynError { Box::new(e) })?
         .scheme()
         == "https";
 
     (template.tls_channel == url_is_tls)
         .then_some(())
-        .ok_or_else(|| -> Box<dyn std::error::Error + Send + Sync> {
+        .ok_or_else(|| -> BoxDynError {
             Box::new(MultiNodeClientError::TlsChannelMismatch {
                 url_is_tls,
                 client_builder_tls_channel: template.tls_channel,
--- a/xmtp_api_d14n/src/protocol/traits.rs
+++ b/xmtp_api_d14n/src/protocol/traits.rs
@@ -53,7 +53,7 @@ pub enum EnvelopeError {
     // for extractors defined outside of this crate or
     // generic implementations like Tuples
     #[error("{0}")]
-    DynError(Box<dyn RetryableError + Send + Sync>),
+    DynError(Box<dyn RetryableError>),
 }
 
 impl RetryableError for EnvelopeError {
--- a/xmtp_api_d14n/src/protocol/traits/cursor_store.rs
+++ b/xmtp_api_d14n/src/protocol/traits/cursor_store.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use xmtp_common::RetryableError;
+use xmtp_common::{MaybeSend, MaybeSync, RetryableError};
 use xmtp_proto::{
     api::ApiClientError,
     types::{ClockOrdering, Cursor, GlobalCursor, OriginatorId, Topic, TopicKind},
@@ -14,7 +14,7 @@ pub enum CursorStoreError {
     #[error("the store cannot handle topic of kind {0}")]
     UnhandledTopicKind(TopicKind),
     #[error("{0}")]
-    Other(Box<dyn RetryableError + Send + Sync>),
+    Other(Box<dyn RetryableError>),
 }
 
 impl RetryableError for CursorStoreError {
@@ -36,7 +36,7 @@ impl<E: std::error::Error> From<CursorStoreError> for ApiClientError<E> {
 /// Trait defining how cursors should be stored, updated, and fetched
 /// _NOTE:_, implementations decide retry strategy. the exact implementation of persistence (or lack)
 /// is up to implementors. functions are assumed to be idempotent & atomic.
-pub trait CursorStore: Send + Sync {
+pub trait CursorStore: MaybeSend + MaybeSync {
     // /// Get the last seen cursor per originator
     // fn last_seen(&self, topic: &Topic) -> Result<GlobalCursor, Self::Error>;
 
--- a/xmtp_api_d14n/src/protocol/traits/envelope_collection.rs
+++ b/xmtp_api_d14n/src/protocol/traits/envelope_collection.rs
@@ -1,10 +1,12 @@
 //! Traits and blanket implementations representing a collection of [`Envelope`]'s
+use xmtp_common::{MaybeSend, MaybeSync};
+
 use super::*;
 
 // TODO: Maybe we can do something with Iterators and Extractors
 // to avoid the Combinators like `CollectionExtractor`?
 /// A Generic Higher-Level Collection of Envelopes
-pub trait EnvelopeCollection<'env> {
+pub trait EnvelopeCollection<'env>: MaybeSend + MaybeSync {
     /// Get the topic for an envelope
     fn topics(&self) -> Result<Vec<Topic>, EnvelopeError>;
     /// Get the payload for an envelope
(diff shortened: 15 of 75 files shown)


tylerhawkes:
* **#2718** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2718?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2717** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2716** <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2716?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/libxmtp/2716?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=2716).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #2729, opened 2025-11-04
remove suspected race condition in `StreamGroupMessages::add()`

insipx:
since `add()` is external to `StreamGroupMessages`, it could be possible that we access a codepath with an incorrect state transition if the add() is lined up while processing a message

this PR removes the call to `resolve_group_additions` in `add()`. instead, all group additions are acted on at only one place, in `fn poll_next(`

I removed `skip()` because we get the same effect by just queueing the future immediately again, so that we can call next_message() in `waiting`

I don't know if using the `waker` like this is "correct". If you google it, the sentiment seems to be against using a `waker` like this. However, if we poll the underlying stream then we have to immediately act on the Poll::Ready, which becomes pretty complicated in a stream impl.

### DIFF
--- a/xmtp_mls/src/subscriptions/stream_messages.rs
+++ b/xmtp_mls/src/subscriptions/stream_messages.rs
@@ -186,20 +186,16 @@ where
     /// This is an asynchronous operation that transitions the stream to the `Adding` state.
     /// The actual subscription update happens when the stream is polled.
     pub(super) fn add(mut self: Pin<&mut Self>, group: MlsGroup<C>) {
+        // we unconditionally notify, otherwise
+        // test failures if we hit a group that is already in the stream.
         if self.groups.contains(&group.group_id) {
             tracing::debug!("group {} already in stream", hex::encode(&group.group_id));
             return;
         }
-
-        // if we're waiting, resolve it right away
-        if let State::Waiting = self.state {
-            self.resolve_group_additions(group);
-        } else {
-            tracing::debug!("stream busy, queuing group add");
-            // any other state and the group must be added to queue
-            let this = self.as_mut().project();
-            this.add_queue.push_back(group);
-        }
+        tracing::debug!("queuing group add");
+        // any other state and the group must be added to queue
+        let this = self.as_mut().project();
+        this.add_queue.push_back(group);
     }
 
     /// Internal API to re-subscribe to a message stream.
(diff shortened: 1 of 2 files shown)


xmtp_mls/src/subscriptions/stream_messages.rs
```diff
@@ -434,31 +434,6 @@ where
         });
     }
 
-    // iterative skip to avoid overflowing the stack
```
neekolas, line 437: I do like removing the `skip` function. That was a lot of surface area for things to be subtly different in how we process messages.

xmtp_mls/src/subscriptions/stream_messages.rs
```diff
@@ -388,7 +386,8 @@ where
                 next_msg.cursor,
                 next_msg.group_id
             );
-            next_msg = ready!(self.as_mut().skip(cx, next_msg))?;
+            cx.waker().wake_by_ref();
```
neekolas, line 389: So, the concern on the internet is that we are telling it to wake up before we've told it that the last poll is pending?

The worry then is that the executor might ignore the `wake_by_ref` in this state and do nothing (since it has free reign to combine multiple polls into one)? Or at least that the behavior is undefined and might change between executors?

xmtp_mls/src/subscriptions/stream_messages.rs
```diff
@@ -379,7 +377,7 @@ where
         if next_msg.is_none() {
             return Poll::Ready(None);
         }
-        let mut next_msg = next_msg.expect("checked for none")?;
+        let next_msg = next_msg.expect("checked for none")?;
```
tylerhawkes: This should be 

```
let Some(next_msg) = next_msg else {
    return Poll::Ready(None);
}
```

neekolas (approved):

xmtp_mls/src/subscriptions/stream_messages.rs
insipx: there's the risk that calling `cx.waker().wake_by_ref()` we starve other tasks of progressing. in a worst-case scenario that leads to an infinite busy loop of just a single task

tokios smart enough to handle it for us in most cases. I guess its just not optimal, but works for us

xmtp_mls/src/subscriptions/stream_messages.rs
```diff
@@ -376,10 +374,9 @@ where
         cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<<Self as Stream>::Item>> {
         let next_msg = ready!(self.as_mut().next_message(cx));
-        if next_msg.is_none() {
+        if let Some(next_msg) = next_msg else {
```
tylerhawkes: No `if`, just `let Some(..) = _ else {`

PR #2730, opened 2025-11-04
Move workers into a futures unordered

codabrink:
### Move workers into a FuturesUnordered by centralizing `WorkerRunner.spawn` orchestration and storing `Client.workers` as `Arc<WorkerRunner>`
Refactor worker lifecycle by making `WorkerRunner.spawn` build and manage a `FuturesUnordered` of worker futures under a single abortable handle, change `Client` to hold `Arc<WorkerRunner>`, and replace context exposure of `WorkerRunner` with direct `task_channels` and `sync_metrics` accessors.

#### 📍Where to Start
Start with `WorkerRunner.spawn` in [xmtp_mls/src/worker.rs](https://github.com/xmtp/libxmtp/pull/2730/files#diff-ea560e63ca633c03d1189957b9efca26de6aa33294f42c116a574e5bf259ce23), then follow how `ClientBuilder<ApiClient,S,Db>.build` wires the `Arc<WorkerRunner>` and context fields in [xmtp_mls/src/builder.rs](https://github.com/xmtp/libxmtp/pull/2730/files#diff-2df8df9acb5405ca5c2dde2487e92ffc9a7daf6be38dd075a2a2d4878cc35ce6).

#### Changes since #2730 opened

- Modified worker spawn system to return `WorkerKind` instead of unit type [24150ad]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 24150ad. 7 files reviewed, 5 issues evaluated, 5 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_mls/src/client.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 291](https://github.com/xmtp/libxmtp/blob/24150ada76634ee47e2729409bd092cb9ed0ae29/xmtp_mls/src/client.rs#L291): When workers are disabled via the builder (`disable_workers = true`), the `Client` still exposes `device_sync_client()` without guarding for disabled worker mode. The returned `DeviceSyncClient` contains metrics obtained via `self.context.sync_metrics()` (likely `None` when no workers are registered), causing the fallback path to create an isolated `WorkerMetrics` instance that is never driven by any worker. Subsequent calls like `wait_for_sync_worker_init()` will wait indefinitely and operations that increment metrics will be recorded on a handle that no worker observes. The method should either: (a) reject creation of a device sync client when the device sync worker is disabled, (b) attach a no-op metrics handle that short-circuits waits and increments, or (c) expose a clear error/fallback indicating that device sync functionality is disabled. <b>[ Already posted ]</b>
- [line 291](https://github.com/xmtp/libxmtp/blob/24150ada76634ee47e2729409bd092cb9ed0ae29/xmtp_mls/src/client.rs#L291): The `device_sync_client` method constructs a `DeviceSyncClient` using `self.context.sync_metrics()`, and if that returns `None` it creates a fresh `Arc<WorkerMetrics<SyncMetric>>` via `WorkerMetrics::new(...)`. This newly created metrics instance is not registered in the `WorkerRunner`'s metrics map and is not used by any spawned workers. As a result, any code that relies on these metrics for coordination (e.g., `wait_for_sync_worker_init` or metric increment visibility) will observe a different, uninitialized metrics handle and may hang indefinitely or produce inconsistent telemetry. In particular: (a) if `device_sync_client()` is called before `WorkerRunner::spawn()` has populated the metrics map, or (b) if workers are disabled, `wait_for_sync_worker_init()` will wait on a metrics handle that will never be set to initialized. This violates the intended invariant that metrics are shared and reflect worker state. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 155](https://github.com/xmtp/libxmtp/blob/24150ada76634ee47e2729409bd092cb9ed0ae29/xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs#L155): Potential panic when building `NewTask`: the call to `welcome.timestamp()` uses `chrono::DateTime::timestamp_nanos_opt().expect(...)` in `WelcomeMessage::timestamp()`. If `welcome.created_ns` is outside the `i64` range (e.g., malformed or extreme future timestamp), `expect` will panic, causing runtime termination during normal message processing. This function (`from_welcome_proto`) does not guard against out-of-range timestamps and will crash the process rather than returning an error. <b>[ Low confidence ]</b>
- [line 166](https://github.com/xmtp/libxmtp/blob/24150ada76634ee47e2729409bd092cb9ed0ae29/xmtp_mls/src/groups/mls_ext/decrypted_welcome.rs#L166): Possible silent failure when scheduling retry: `context.task_channels().send(task);` ignores any potential error return or failure from the `send` operation. If the underlying channel is disconnected or full (depending on implementation), the task may not be enqueued, but the function still returns `Err(GroupError::WelcomeDataNotFound(destination))`, losing the opportunity to retry. This violates the at-most-once/cleanup guarantees for effectful operations because a failure to enqueue is neither handled nor reported distinctly. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_mls/src/tasks.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 70](https://github.com/xmtp/libxmtp/blob/24150ada76634ee47e2729409bd092cb9ed0ae29/xmtp_mls/src/tasks.rs#L70): `TaskWorkerChannels::send` panics if the channel's receiver has been dropped. It calls `self.task_sender.send(new_task).expect("Task receiver is owned by same struct")`. `tokio::sync::mpsc::UnboundedSender::send` returns `Err` when all receivers are dropped. This is reachable if a caller retains a cloned `TaskWorkerChannels` after the worker subsystem has been shut down or reinitialized and invokes `send` later, causing a runtime panic. A safer approach is to return a `Result` from `send` or to log and ignore when the receiver is gone instead of panicking. <b>[ Out of scope ]</b>
</details>

</details>

### DIFF
--- a/xmtp_mls/src/worker.rs
+++ b/xmtp_mls/src/worker.rs
@@ -6,7 +6,7 @@ use parking_lot::Mutex;
 use std::fmt::Debug;
 use std::pin::Pin;
 use std::{any::Any, collections::HashMap, hash::Hash, sync::Arc};
-use xmtp_common::{MaybeSend, MaybeSync, StreamHandle};
+use xmtp_common::{MaybeSend, MaybeSync, StreamHandle, if_native, if_wasm};
 use xmtp_configuration::WORKER_RESTART_DELAY;
 
 pub mod metrics;
@@ -96,8 +96,8 @@ impl WorkerRunner {
                 futs.push(worker.spawn());
             }
 
-            while futs.next().await.is_some() {
-                tracing::warn!("Worker completed unexpectedly")
+            while let Some(kind) = futs.next().await {
+                tracing::warn!("Worker {kind:?} completed unexpectedly")
             }
         });
 
@@ -118,11 +118,12 @@ impl WorkerRunner {
 }
 
 pub type WorkerResult<T> = Result<T, Box<dyn NeedsDbReconnect>>;
-
-#[cfg(not(target_arch = "wasm32"))]
-type SpawnWorkerFut = dyn Future<Output = ()> + Send;
-#[cfg(target_arch = "wasm32")]
-type SpawnWorkerFut = dyn Future<Output = ()>;
+if_native! {
+    type SpawnWorkerFut = dyn Future<Output = WorkerKind> + Send;
+}
+if_wasm! {
+    type SpawnWorkerFut = dyn Future<Output = WorkerKind>;
+}
 
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
@@ -163,6 +164,8 @@ pub trait Worker: MaybeSend + MaybeSync + 'static {
                     }
                 }
             }
+
+            self.kind()
         };
 
         Box::pin(fut)


xmtp_mls/src/worker.rs
```diff
@@ -63,21 +71,37 @@ impl WorkerRunner {
...
+                    m.insert(worker.kind(), metrics);
+                }
+                futs.push(worker.spawn());
             }
 
-            if let Some(metrics) = worker.metrics() {
-                let mut m = self.metrics.lock();
-                m.insert(worker.kind(), metrics);
+            while futs.next().await.is_some() {
+                tracing::warn!("Worker completed unexpectedly")
```
tylerhawkes: Can we log the worker that quit?

xmtp_mls/src/worker.rs
```diff
@@ -95,6 +119,11 @@ impl WorkerRunner {
 
 pub type WorkerResult<T> = Result<T, Box<dyn NeedsDbReconnect>>;
 
+#[cfg(not(target_arch = "wasm32"))]
+type SpawnWorkerFut = dyn Future<Output = ()> + Send;
+#[cfg(target_arch = "wasm32")]
+type SpawnWorkerFut = dyn Future<Output = ()>;
+
```
tylerhawkes: use `if_native!` and `if_wasm!` macros for these

tylerhawkes: use `if_native!` and `if_wasm!` macros for these

tylerhawkes (approved):

xmtp_mls/src/worker.rs
```diff
@@ -63,21 +71,37 @@ impl WorkerRunner {
...
+                if let Some(metrics) = metrics {
+                    this.metrics.lock().insert(factory.kind(), metrics);
+                }
 
-            if let Some(metrics) = metrics {
-                self.metrics.lock().insert(factory.kind(), metrics);
+                if let Some(metrics) = worker.metrics() {
+                    let mut m = this.metrics.lock();
+                    m.insert(worker.kind(), metrics);
+                }
```
tylerhawkes, line 95: There's probably a better way to handle metrics than getting them passed back and forth and potentially overriding the metrics from the factory with the worker.

PR #2743, opened 2025-11-05
add experimental ryow for identities in xdbg

insipx:
### Add experimental RYOW to identity generation and gate it behind `--ryow` with `BackendOpts.ryow_timeout=5s`, using `main::XDBG_ID_NONCE=1` in `xdbg`
Introduce optional RYOW flow in identity generation, refactor client creation to return unregistered clients, and add a decentralized read/write API builder in `BackendOpts`.

#### 📍Where to Start
Start with `GenerateIdentity::create_identities` in [xmtp_debug/src/app/generate/identity.rs](https://github.com/xmtp/libxmtp/pull/2743/files#diff-5810cc3f49b5afa472f60492ce4999aadf64464660c04f8251d27b1e6abfcf1c).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized c468102. 8 files reviewed, 7 issues evaluated, 7 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_debug/src/app/clients.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 68](https://github.com/xmtp/libxmtp/blob/c468102940a781f44de355d32d06e14e7a0fe011/xmtp_debug/src/app/clients.rs#L68): `new_client_inner` is documented as "Create a new client + Identity & register it" but the implementation no longer registers the client (the call to `register_client(&client, wallet).await?` was removed). This breaks contract parity: callers expecting a registered/ready identity will now get an unregistered client, causing operations gated by `ensure_identity_ready()` to fail at runtime (e.g., group creation, DM creation, sync). Either restore registration or update all callers and documentation to reflect the unregistered behavior. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_debug/src/app/generate.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 42](https://github.com/xmtp/libxmtp/blob/c468102940a781f44de355d32d06e14e7a0fe011/xmtp_debug/src/app/generate.rs#L42): `run` forwards `concurrency` without validation. If the CLI-supplied `concurrency` is 0, downstream code (e.g., identity generation) will panic in `buffer_unordered(0)` or hang on `Semaphore::new(0)`. Add a guard in `run` (e.g., clamp to at least 1) or validate in argument parsing. <b>[ Already posted ]</b>
</details>

<details>
[...]

PR #2736, opened 2025-11-05
fix concurrency when generating messages with identities

insipx:
- adds a query "all-key-packages" to get all key packages for identities in xdbg database
- adds `load_all_identities` fn which creates a client for every identity in xdbg database for that network. this allows optimizing concurrency. Returns as a `HashMap<Inbox, Mutex<Client>>`

fixes #2708

PR #2747, opened 2025-11-06
key package debug

insipx:
### Add debug-focused offline client construction and concurrency-safe preloaded clients across `xmtp_debug` to support key package workflows
Shift client creation to synchronous/offline builders, preload clients behind mutexes for concurrent group/message ops, add key package listing/printing via direct network connections, and remove identity file locks. Introduce a new CLI subcommand for all key packages, log stream item receipt, and update docker image tags.

#### 📍Where to Start
Start with the client construction changes in `xmtp_debug` at `app::clients` entry points in [xmtp_debug/src/app/clients.rs](https://github.com/xmtp/libxmtp/pull/2747/files#diff-f251302a34aaefa8d847fbe261211c21aa97857edc8225ba0fc8042f87c16b70), then follow usage in generators and queries such as [xmtp_debug/src/app/generate/identity.rs](https://github.com/xmtp/libxmtp/pull/2747/files#diff-5810cc3f49b5afa472f60492ce4999aadf64464660c04f8251d27b1e6abfcf1c) and [xmtp_debug/src/app/query.rs](https://github.com/xmtp/libxmtp/pull/2747/files#diff-02228818084d108ca6d2633e4e2238c55e84687068962939d99e4276ebcf11e3).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized c9aeed2. 18 files reviewed, 30 issues evaluated, 26 issues filtered, 2 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_api_grpc/src/streams/non_blocking_stream.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 143](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_api_grpc/src/streams/non_blocking_stream.rs#L143): `poll_next` logs `"got an item!"` on any `Poll::Ready(Some(_))`, including when the stream yields `Err(Status)`. This produces misleading success logs on error items. Change the match to only log on `Ok(_)`, e.g., `if let Poll::Ready(Some(Ok(_))) = next { ... }`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_debug/src/app/clients.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 144](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/clients.rs#L144): `existing_client_inner` converts `db_path` to `String` via `db_path.clone().into_os_string().into_string().unwrap()` which will panic at runtime if the path is not valid UTF-8. On Unix, file system paths are byte sequences and may be non-UTF-8. Use a fallible conversion and return an error (as done in `new_client_inner`) instead of `unwrap()`. <b>[ Out of scope ]</b>
- [line 174](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/clients.rs#L174): `load_all_identities` calls `client_from_identity(&i.value(), network)` which internally uses `existing_client_inner` to build a client. In `existing_client_inner`, the database path is converted with `db_path.clone().into_os_string().into_string().unwrap()`. If `db_path` contains non-UTF-8 bytes (valid on Unix), this will panic at runtime. This makes `load_all_identities` susceptible to a panic on any identity whose path is not valid UTF-8. <b>[ Low confidence ]</b>
- [line 197](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/clients.rs#L197): `load_n_identities` builds a `HashMap<InboxId, Arc<Mutex<crate::DbgClient>>>` from randomly selected identities. If two or more identities share the same `inbox_id`, later entries will silently overwrite earlier ones, reducing the number of clients below `n` without any warning. This violates uniqueness guarantees and yields fewer clients than requested with no error or log. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_debug/src/app/generate/groups.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 41](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/groups.rs#L41): `create_groups` uses `ProgressStyle::with_template(...).unwrap()` which can panic if the template is invalid. While the template string is constant, any future change or upstream behavior change would crash group generation. Prefer handling the `Result` and returning a user-facing error. <b>[ Low confidence ]</b>
- [line 86](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/groups.rs#L86): `create_groups` panics on invalid group id length via `expect("Group id expected to be 32 bytes")`. If the backend returns a non-32-byte `group_id`, the process crashes. Replace with an explicit error conversion and propagate a user-facing error. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_debug/src/app/generate/identity.rs — 0 comments posted, 7 evaluated, 7 filtered</summary>

- [line 46](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/identity.rs#L46): `ProgressStyle::with_template(...).unwrap()` can panic at runtime if the provided template is invalid. This occurs in the creation of the progress bar. Replace the `unwrap()` with handling that reports an error (e.g., `?` converting to your `Result`) or use a safe default when the style fails to construct. <b>[ Code style ]</b>
- [line 48](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/identity.rs#L48): Background progress ticker task spawned at lines 48–59 never terminates or is signaled to stop. It continues calling `b.tick()` indefinitely even after the function returns and after the progress bar is finished/reset, leaking a task and wasting CPU. Provide a shutdown signal and join the task on all exit paths. <b>[ Low confidence ]</b>
- [line 84](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/identity.rs#L84): `buffer_unordered(concurrency)` and `try_buffer_unordered(concurrency)` are called with an unchecked `concurrency` parameter (lines 84, 160, 186). Passing `0` causes a runtime panic in `futures` utilities. Add a guard to ensure `concurrency >= 1` or default to `1`. <b>[ Out of scope ]</b>
- [line 90](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/identity.rs#L90): Progress bar is finished (`bar.finish()`) and then reset (`bar.reset()`) at lines 90–91 while the background tick task continues. Concurrent operations on a finished/reset bar from another task can produce misleading UI and racey updates. Either stop the ticker before finishing/resetting, or keep the same lifecycle. <b>[ Low confidence ]</b>
- [line 93](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/identity.rs#L93): `Arc::into_inner(s).expect("only one reference exists after tasks finish")` can panic at runtime if any remaining `Arc` clones still exist. While all spawned tasks appear completed, any retained clone (e.g., captured and kept alive) will cause a panic. Use `Arc::try_unwrap(s)` and handle the `Err` case gracefully, or refactor to avoid the need to unwrap the `Arc`. <b>[ Code style ]</b>
- [line 139](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/identity.rs#L139): RyOW subscription loop (lines 114–138) exits with `Ok(())` when the stream ends even if `needed_installations` is not empty, silently failing read-your-own-writes guarantees. After the loop, check `needed_installations.is_empty()` and return an error if not, so incomplete observation is not treated as success. <b>[ Out of scope ]</b>
- [line 197](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/identity.rs#L197): If errors occur during the association state fetch (lines 190–203), the function `bail!`s and returns without awaiting or cancelling the `read_writes` task created earlier, leaving the subscription task running in the background. Ensure the `read_writes` task is awaited or cancelled on all exit paths (including error) to avoid leaks. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_debug/src/app/generate/messages.rs — 0 comments posted, 5 evaluated, 5 filtered</summary>

- [line 83](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/messages.rs#L83): `send_many_messages` uses `ProgressStyle::with_template(...).unwrap()` which can panic if the template is invalid. Prefer handling the error and surfacing it to the user. <b>[ Low confidence ]</b>
- [line 110](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/messages.rs#L110): `send_many_messages` misclassifies task outcomes and suppresses inner errors. The code only collects and logs `JoinError`s and counts any successfully spawned task as a sent message even if the task returned `Err(eyre::Error)`. Specifically: building `errors` via `res.iter().filter(|r| r.is_err())` only captures `JoinError`s, not inner `Result<(), eyre::Error>` failures; and `msgs_sent` counts all `Ok` outer results (which include `Ok(Err(...))`). This yields incorrect success metrics and hides actual send failures. Fix by inspecting the inner result of each `Ok` and counting/logging only `Ok(())`, while recording inner `Err`s. <b>[ Out of scope ]</b>
- [line 151](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/messages.rs#L151): `rng.gen_range(0..*max_message_size)` will panic if `max_message_size` is `0`, because `gen_range` requires a non-empty range. Validate `max_message_size > 0` before calling, or use an inclusive range, or handle the `0` case explicitly (e.g., force sending an empty message when zero). <b>[ Low confidence ]</b>
- [line 160](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/messages.rs#L160): `send_message` unconditionally calls `.build().unwrap()` on `SendMessageOptsBuilder`. If the builder validation fails (e.g., incompatible options in future changes), this will panic. Replace with proper error handling and propagate an error. <b>[ Previously rejected ]</b>
- [line 160](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/generate/messages.rs#L160): `SendMessageOptsBuilder::default().should_push(true).build().unwrap()` can panic if `build()` returns an error. Replace the `unwrap()` with proper error handling (e.g., `?`) to propagate a meaningful error via `MessageSendError`. <b>[ Code style ]</b>
</details>

<details>
<summary>xmtp_debug/src/app/modify.rs — 1 comment posted, 3 evaluated, 1 filtered</summary>

- [line 90](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/modify.rs#L90): `AddExternal` branch updates the remote group via `add_members_by_inbox_id` and `update_admin_list` but does not update the locally stored `local_group` or persist it (no `group_store.set(...)`). This leaves local state inconsistent with the remote state, violating the stated invariant seen in `Remove` where local storage is updated after the remote change. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_debug/src/app/query.rs — 1 comment posted, 7 evaluated, 6 filtered</summary>

- [line 62](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/query.rs#L62): Potential truncation/overflow and invalid timestamp passed to `chrono::DateTime::from_timestamp_nanos` when converting `update.server_timestamp_ns` (and similarly `new_update.client_timestamp_ns`) via `as i64`. The code does an unchecked cast: `chrono::DateTime::from_timestamp_nanos(update.server_timestamp_ns as i64)`. If `update.server_timestamp_ns` (or `new_update.client_timestamp_ns`) is wider than `i64` or is an unsigned value larger than `i64::MAX`, the cast will truncate data and may produce an incorrect or negative timestamp. If the resulting value is out-of-range for `from_timestamp_nanos` this may produce incorrect times or runtime errors depending on the `chrono` implementation. Handle timestamps using a safe conversion (preserve range or validate) or use the timestamp type expected by `chrono` without unchecked narrowing. <b>[ Low confidence ]</b>
- [line 62](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/query.rs#L62): `identity` constructs timestamps with `chrono::DateTime::from_timestamp_nanos(update.server_timestamp_ns as i64)` and similarly for `client_timestamp_ns`. Casting from (likely) `u64` to `i64` can wrap for large values, and `from_timestamp_nanos` may panic on out-of-range inputs. Use the fallible APIs (e.g., `DateTime::from_timestamp_nanos` if it returns `Option` in your chrono version, or convert via `NaiveDateTime::from_timestamp_opt`/`from_timestamp_nanos` fallible equivalents) and handle the `None` case gracefully. Avoid lossy `as i64` casts by using checked conversions. <b>[ Previously rejected ]</b>
- [line 87](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/query.rs#L87): Ambiguous handling of `add_association.new_member_identifier` and `revoke_association.member_to_revoke`: the code maps `None` to `String::new()` silently (`None => String::new()`). This discards the fact that an identifier was absent versus present-but-empty and may hide missing data. A missing identifier is potentially significant and should be logged or reported explicitly rather than turned into an empty string without context. <b>[ Code style ]</b>
- [line 106](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/query.rs#L106): Possible misuse of an Option when printing `new_recovery_address`: the code assigns `let new_recovery_address = change_recovery_address.new_recovery_identifier;` and then prints it with `println!("    change_recovery_address: new_recovery_address: {new_recovery_address}");` If `new_recovery_identifier` is an `Option<T>` (common in proto-generated structs), this will print `Some(...)` or `None` rather than the intended inner value. The code does not explicitly handle `None` or extract the inner value; this produces noisy/ambiguous output and may hide the absence of a value. Explicitly match/unpack the Option and handle the `None` case (e.g., print a clear message). <b>[ Code style ]</b>
- [line 206](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/query.rs#L206): `all_key_packages` uses `unwrap()` on `XmtpInstallationCredential::from_bytes(&i.value().installation_key)` inside an iterator. If any stored `installation_key` is invalid/corrupt, this will panic and crash the command. Replace `unwrap()` with error handling (e.g., `ok_or_else` or `and_then` with `?`) and provide a graceful error path. <b>[ Low confidence ]</b>
- [line 246](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/app/query.rs#L246): `print_kps` uses multiple `unwrap()` calls that can panic at runtime on valid but unexpected input: `verified.life_time().unwrap()` may be `None`, and `chrono::DateTime::from_timestamp(...).unwrap()` can return `None` for out-of-range timestamps. Replace `unwrap()` with explicit handling (e.g., `ok_or_else`/`context` or fallback printing) to avoid crashing. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_debug/src/args.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 286](https://github.com/xmtp/libxmtp/blob/c9aeed2ad34fb8c786225ccc433de1db565a73f7/xmtp_debug/src/args.rs#L286): `default_ryow_timeout` uses `"5s".parse::<humantime::Duration>().unwrap()` which can panic at runtime if parsing fails. While the literal `"5s"` is currently valid for `humantime`, relying on `unwrap()` ties correctness to third-party parser behavior and configuration. Prefer constructing the duration without parsing to avoid any panic path, e.g., `std::time::Duration::from_secs(5).into()` to produce a `humantime::Duration`. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/dev/docker/docker-compose-d14n.yml
+++ b/dev/docker/docker-compose-d14n.yml
@@ -1,6 +1,6 @@
-x-xmtpd-server-image: &x-xmtpd-server-image ghcr.io/xmtp/xmtpd:sha-5d27ea4
-x-xmtpd-gateway-image: &x-xmtpd-gateway-image ghcr.io/xmtp/xmtpd-gateway:sha-5d27ea4
-x-xmtpd-cli-image: &x-xmtpd-cli-image ghcr.io/xmtp/xmtpd-cli:sha-5d27ea4
+x-xmtpd-server-image: &x-xmtpd-server-image ghcr.io/xmtp/xmtpd:sha-d3bb0f8
+x-xmtpd-gateway-image: &x-xmtpd-gateway-image ghcr.io/xmtp/xmtpd-gateway:sha-d3bb0f8
+x-xmtpd-cli-image: &x-xmtpd-cli-image ghcr.io/xmtp/xmtpd-cli:sha-d3bb0f8
 x-xmtpd-contracts-image: &x-xmtpd-contracts-image ghcr.io/xmtp/contracts:v0.5.5
 x-postgres-image: &x-postgres-image postgres:16
 x-redis-image: &x-redis-image redis:7-alpine
--- a/xmtp_api_grpc/src/grpc_client/client.rs
+++ b/xmtp_api_grpc/src/grpc_client/client.rs
@@ -146,7 +146,6 @@ impl Client for GrpcClient {
         path: http::uri::PathAndQuery,
         body: Bytes,
     ) -> Result<http::Response<Bytes>, ApiClientError<Self::Error>> {
-        tracing::debug!(?path, "request");
         let client = &mut self.inner.clone();
         self.wait_for_ready(client).await.map_err(GrpcError::from)?;
         let request = self
--- a/xmtp_api_grpc/src/streams/non_blocking_stream.rs
+++ b/xmtp_api_grpc/src/streams/non_blocking_stream.rs
@@ -140,6 +140,9 @@ where
                 if let Poll::Ready(None) = next {
                     this.state.set(StreamState::Terminated);
                 }
+                if let Poll::Ready(Some(_)) = next {
+                    tracing::info!("got an item!");
+                }
                 next
             }
             Terminated => Poll::Ready(None),
--- a/xmtp_debug/src/app/generate/identity.rs
+++ b/xmtp_debug/src/app/generate/identity.rs
@@ -110,7 +110,10 @@ impl GenerateIdentity {
                 let _ = tx.send(());
                 bar_ref.set_message("waiting for identities to be written");
                 futures::pin_mut!(s);
-                while let Some(kp) = timeout(n.ryow_timeout.into(), s.try_next()).await.wrap_err("timeout reached for reading writes on key package published")?? {
+                let duration: std::time::Duration = n.ryow_timeout.into();
+                while let Some(kp) = timeout(duration, s.try_next()).await.wrap_err(
+                    format!("timeout reached for reading writes on key package published. still need {} installations", needed_installations.len())
+                )?? {
                     // TODO: we can deserialize key packages in extractors possibly
                     let extractor =
                         CollectionExtractor::new(kp.envelopes, KeyPackagesExtractor::new());


insipx:
This PR is part of a stack of 3 bookmarks:

1. `trunk()`
1. [push-znqrouuxvqrl](https://github.com/xmtp/libxmtp/pull/2736)
1. [push-sxyuuqzkswrx](https://github.com/xmtp/libxmtp/pull/2743)
1. **push-rrpuqokryllw ← this PR**

---

dev/docker/docker-compose-d14n.yml
```diff
@@ -1,6 +1,6 @@
-x-xmtpd-server-image: &x-xmtpd-server-image ghcr.io/xmtp/xmtpd:sha-5d27ea4
-x-xmtpd-gateway-image: &x-xmtpd-gateway-image ghcr.io/xmtp/xmtpd-gateway:sha-5d27ea4
-x-xmtpd-cli-image: &x-xmtpd-cli-image ghcr.io/xmtp/xmtpd-cli:sha-5d27ea4
+x-xmtpd-server-image: &x-xmtpd-server-image ghcr.io/xmtp/xmtpd:sha-d3bb0f8
+x-xmtpd-gateway-image: &x-xmtpd-gateway-image ghcr.io/xmtp/xmtpd-gateway:sha-d3bb0f8
+x-xmtpd-cli-image: &x-xmtpd-cli-image ghcr.io/xmtp/xmtpd-cli:sha-d3bb0f8
```
fbac, line 3: Careful with this as it might break the recent Connect-RPC changes.

insipx: ah good callout, i'll need to make sure to fix it if this merges

PR #2758, opened 2025-11-10
Fix and Test Snapshots

codabrink:
- Fix the snapshot feature in Tester and create a test.
- Switch the mutex in stream stats over to a tokio mutex to allow locking across awaits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1725,7 +1725,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fde0e0ec90c9dfb3b4b1a0891a7dcd0e2bffde2f7efed5fe7c9bb00e5bfb915e"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


xmtp_mls/src/utils/test/tester_utils.rs
```diff
@@ -99,14 +100,23 @@ impl<Owner> Tester<Owner, FullXmtpClient>
...
+
         self.db()
             .raw_query_write(|conn| {
                 let buff = conn.serialize_database_to_buffer();
                 Ok(buff.to_vec())
             })
             .unwrap()
     }
+
+    pub fn save_db_snapshot_to_file(&self, path: impl Into<PathBuf>) {
```
tylerhawkes: path should probably be `impl AsRef<Path>` as that's what std::fs::write takes and it's more flexible and can avoid allocating.

xmtp_mls/src/subscriptions/stream_messages/stream_stats.rs
```diff
@@ -103,9 +101,9 @@ pub struct StreamStats {
 }
 
 impl StreamStats {
-    pub fn new_stats(&self) -> Vec<StreamStat> {
+    pub async fn new_stats(&self) -> Vec<StreamStat> {
         let mut stats = vec![];
-        let mut stats_rx = self.rx.lock();
+        let mut stats_rx = self.rx.lock().await;
```
tylerhawkes, line 106: I'm not sure that changing these to async is really necessary. If the lock isn't available immediately then this will likely return nothing as the last reader exhausted the receiver. This could still use a parking_lot Mutex and use try_lock() and if something else is using it return an empty vec.

There are no async operations happening while the lock is held.

tylerhawkes (approved):

xmtp_mls/src/subscriptions/stream_messages/stream_stats.rs
codabrink: I'm pulling stats from this channel on a task outside of this fn in another project.

PR #2761, opened 2025-11-12
Convert the archive iterators to streams

codabrink:
Convert the archive iterators to streams. This will allow archiving of things that may need network requests.

### DIFF
--- a/xmtp_archive/src/export_stream/mod.rs
+++ b/xmtp_archive/src/export_stream/mod.rs
@@ -6,7 +6,7 @@ use std::{
     sync::Arc,
     task::{Context, Poll},
 };
-use xmtp_common::{MaybeSend, if_native, if_wasm};
+use xmtp_common::{MaybeSend, MaybeSendFuture, if_native, if_wasm};
 use xmtp_db::{StorageError, prelude::*};
 use xmtp_proto::xmtp::device_sync::{
     BackupElement, BackupElementSelection, BackupOptions, consent_backup::ConsentSave,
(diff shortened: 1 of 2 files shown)


xmtp_archive/src/exporter.rs
```diff
@@ -133,94 +136,93 @@ impl AsyncRead for ArchiveExporter {
...
+            if !*this.encoder_finished {
+                let fut = this.zstd_encoder.flush();
+                pin_mut!(fut);
+                let _ = fut.poll(cx)?;
+            }
 
-        if this.zstd_encoder.get_ref().is_empty() {
-            Poll::Ready(Ok(0))
-        } else {
-            Pin::new(&mut *this).poll_read(cx, buf)
```
codabrink, line 223: Rather than recursing, I moved this into a loop to avoid potential stack overflows.

xmtp_archive/src/export_stream/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+
+if_native! {
+    pin_project! {
+        pub(crate) struct BackupRecordStreamer<R, D> {
+            cursor: i64,
+            db: Arc<D>,
+            start_ns: Option<i64>,
+            end_ns: Option<i64>,
+            #[pin] current_future: Option<Pin<Box<dyn Future<Output = Result<Vec<BackupElement>, StorageError>> + Send>>>,
```
insipx: There's a [FutureWrapper](https://xmtp.github.io/libxmtp/xmtp_common/wasm/struct.FutureWrapper.html) type that holds an inner `Box<dyn Future>` switches on this send bounds for futures you can use instead of using the `if_wasm/native` macros.

You could also experiment with a `MaybeSendFuture` super trait we could use, like

```rust
pub trait MaybeSendFuture: MaybeSend + Future {}
```

then change the current future to look like
```rust
current_future: Option<Pin<Box<dyn MaybeSendFuture<Output = Result<Vec<BackupElement>, StorageError>>>>>,
```

I have no idea if that works, though. But I don't see why it wouldn't work

insipx (approved):
looks good!


just left a small nit w.r.t removing duplication of one of the pin_project structs storing a `Box<dyn Future>`

PR #2772, opened 2025-11-13
add message deletion

mchenani:
### Add message deletion across groups by introducing `xmtp_mls::MlsGroup::delete_message` and handling `deleteMessage` content in sync and enrichment
Implement soft-delete via a new `deleteMessage` content type, persistence in `message_deletions` with schema migration, sync-time processing that validates and records deletions, enrichment that replaces content with `DeletedMessage`, and bindings (FFI/Node/WASM) exposing deletion metadata and APIs.

#### 📍Where to Start
Start with deletion processing in `xmtp_mls::MlsGroup<Context>::process_delete_message` in [file:xmtp_mls/src/groups/mls_sync.rs], then review the public API `xmtp_mls::MlsGroup<Context>::delete_message` in [file:xmtp_mls/src/groups/mod.rs].

#### Changes since #2772 opened

- Changed `ContentType::Unknown` to be non-deletable in the `is_deletable` trait implementation [0d16925]
- Refactored deletion authorization validation logic in `groups.mls_sync.sync` method [0d16925]
- Added content type metadata to deleted messages [800285f]
- Added `should_push` field initialization to `StoredGroupMessage` struct literals across test suites [9466e96]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 9466e96. 18 files reviewed, 17 issues evaluated, 15 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>db_tools/src/tasks/migrations.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 63](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/db_tools/src/tasks/migrations.rs#L63): In `extract_version`, the docstring states the format is `YYYY-MM-DD-HHMMSS[-NNNN]_name`, but the function uses `split('_').next()` which only splits on the first underscore. If a migration name contains an underscore before the timestamp portion (unusual but possible), the version extraction would be incorrect. The old approach in `migration_status` extracted ALL numeric characters from the entire name, which would behave differently for migration names with numbers in the suffix after the underscore. <b>[ Low confidence ]</b>
- [line 73](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/db_tools/src/tasks/migrations.rs#L73): In `migration_status`, the change assumes that applied versions from `db.applied_migrations()` are already numeric-only strings (e.g., `"20251115232503"`). The old code filtered numeric characters from both the migration name AND each applied version, ensuring consistent comparison. The new code only filters the name via `extract_version(name)` but compares directly with `a == &name_version`. If `applied_migrations()` returns versions with dashes (e.g., `"2025-11-15-232503"`), the comparison will always fail, causing all migrations to appear as `[pending]` even when they are applied. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/group_message.rs — 0 comments posted, 6 evaluated, 4 filtered</summary>

- [line 229](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_db/src/encrypted_store/group_message.rs#L229): The `From<ContentType> for XmtpContentType` implementation in `bindings_wasm/src/content_types/mod.rs` does not handle the new `ContentType::DeleteMessage` variant. This will cause the same issue as the node bindings - either a compile error or incorrect conversion behavior. <b>[ Out of scope ]</b>
- [line 229](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_db/src/encrypted_store/group_message.rs#L229): The `TryFrom<ContentTypeSave> for ContentType` implementation in `xmtp_db/src/encrypted_store/group_message/convert.rs` (shown in references) does not handle `DeleteMessage`. If `ContentTypeSave` enum also lacks this variant, database operations involving `DeleteMessage` content type will fail during conversion. <b>[ Already posted ]</b>
- [line 266](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_db/src/encrypted_store/group_message.rs#L266): The `From<ContentType> for XmtpContentType` implementations in `bindings_node/src/content_types/mod.rs` and `bindings_wasm/src/content_types/mod.rs` do not handle `ContentType::DeleteMessage`. Since those match statements appear to not have a wildcard pattern and the `ContentType` enum now includes `DeleteMessage`, this could cause a compile error or runtime panic if `DeleteMessage` is encountered in those bindings. <b>[ Out of scope ]</b>
- [line 266](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_db/src/encrypted_store/group_message.rs#L266): The `ContentType::DeleteMessage` variant was added to the enum but the FFI bindings in `bindings_ffi/src/mls.rs` (the `From<FfiContentType> for ContentType` implementation) does not have a corresponding `FfiContentType::DeleteMessage` variant. If FFI code attempts to work with `DeleteMessage` content type, there is no way to convert it properly, which may cause match exhaustiveness failures or missing functionality. <b>[ Already posted ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/group_message/convert.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 148](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_db/src/encrypted_store/group_message/convert.rs#L148): The `From<ContentType> for ContentTypeSave` implementation maps `WalletSendCalls`, `LeaveRequest`, `Markdown`, `Actions`, `Intent`, `MultiRemoteAttachment`, and `DeleteMessage` to `Self::Unknown`. However, the reverse conversion `TryFrom<ContentTypeSave> for ContentType` (lines 69-89) maps `ContentTypeSave::Unknown` back to `ContentType::Unknown`. This creates a lossy roundtrip: messages with these content types will lose their original type information during device sync backup/restore, permanently degrading to `Unknown` content type. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_sync.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 1385](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_mls/src/groups/mls_sync.rs#L1385): In the out-of-order case (lines 1385-1390), when a deletion message arrives before the target message exists, the deletion is stored without any authorization validation. The code stores the deletion record unconditionally, relying on a comment that says authorization will be validated at enrichment time via `is_deletion_valid()`. However, this means any user can submit delete messages for arbitrary message IDs, and these unauthorized deletions will be persisted to the database. If the enrichment-time validation is missing or flawed, this could allow unauthorized message deletions. <b>[ Already posted ]</b>
- [line 1388](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_mls/src/groups/mls_sync.rs#L1388): The `is_super_admin_deletion` flag at lines 1388-1389 is computed against the *current* MLS group state when the deletion arrives out-of-order, not the state at the time the target message was sent. If the deleter is a super admin when the deletion arrives, but is demoted before the target message arrives, the stored `is_super_admin_deletion=true` would incorrectly authorize deletion of a message sent after demotion. This is a time-of-check-time-of-use (TOCTOU) issue for out-of-order deletions. <b>[ Low confidence ]</b>
- [line 3356](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_mls/src/groups/mls_sync.rs#L3356): The error handling inside `load_mls_group_with_lock` closure swallows the actual error from `process_delete_message` and replaces it with `GroupError::InvalidGroupMembership`. This loses the original error information and could mask the actual failure reason. If `process_delete_message` returns an error (as malformed data might cause), the test converts it to a different error type but then asserts `result.is_ok()`, which would fail. The test logic appears contradictory - it expects `Ok(())` but the error conversion suggests errors are anticipated. <b>[ Already posted ]</b>
- [line 3422](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_mls/src/groups/mls_sync.rs#L3422): Same error handling issue as test 0 - the closure converts any error from `process_delete_message` to `GroupError::InvalidGroupMembership`, losing the original error context. The subsequent assertion `result.is_ok()` will fail if an error occurs, making the test's claim that "Malformed inner DeleteMessage proto should not cause error" potentially incorrect if the actual implementation does return errors for malformed data. <b>[ Already posted ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mod.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 801](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_mls/src/groups/mod.rs#L801): TOCTOU race condition: The check `is_message_deleted(&message_id)` on line 801 and the subsequent `deletion.store(&conn)` on line 838 are not atomic. Two concurrent `delete_message` calls for the same message could both pass the deletion check and attempt to store duplicate deletion records. <b>[ Previously rejected ]</b>
- [line 825](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_mls/src/groups/mod.rs#L825): Atomicity violation: If `send_message_optimistic` on line 825 succeeds but `deletion.store(&conn)` on line 838 fails, a delete message is sent over the network but not recorded locally. This leaves the system in an inconsistent state where remote peers see the deletion but the local database doesn't track it. <b>[ Previously rejected ]</b>
- [line 947](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_mls/src/groups/mod.rs#L947): The new `find_enriched_messages` method passes `args` directly to `get_group_messages` without calling `filter_out_hidden_message_types_from_query(args)`, unlike the similar `find_messages_v2_with_conn` method which applies this filter. This could expose internal/hidden message types that should not be visible to callers. <b>[ Already posted ]</b>
</details>

<details>
<summary>xmtp_mls/src/messages/enrichment.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 257](https://github.com/xmtp/libxmtp/blob/9466e9605344f3352b7e94ae24f375ee137969bd/xmtp_mls/src/messages/enrichment.rs#L257): In `get_deletions`, if multiple `StoredMessageDeletion` records exist for the same `deleted_message_id`, only one will be kept in the resulting `DeletionMap` (the last one encountered during iteration). If deletion records can have different validity states or properties, this could cause incorrect deletion validation by using the wrong deletion record. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/xmtp_db/src/encrypted_store/message_deletion.rs
+++ b/xmtp_db/src/encrypted_store/message_deletion.rs
@@ -224,6 +224,7 @@ mod tests {
             sequence_id: 1,
             originator_id: 1,
             inserted_at_ns: 0,
+            should_push: false,
         }
         .store(conn)
         .unwrap();
--- a/xmtp_mls/src/groups/mls_sync.rs
+++ b/xmtp_mls/src/groups/mls_sync.rs
@@ -3344,6 +3344,7 @@ pub(crate) mod tests {
             sequence_id: 1,
             originator_id: 1,
             inserted_at_ns: 0,
+            should_push: false,
         };
 
         // Use load_mls_group_with_lock to get access to the MLS group and call process_delete_message
@@ -3410,6 +3411,7 @@ pub(crate) mod tests {
             sequence_id: 2,
             originator_id: 1,
             inserted_at_ns: 0,
+            should_push: false,
         };
 
         let storage = alix.context.mls_storage();
@@ -3483,6 +3485,7 @@ pub(crate) mod tests {
             sequence_id: 3,
             originator_id: 1,
             inserted_at_ns: 0,
+            should_push: false,
         };
 
         let storage = alix.context.mls_storage();
--- a/xmtp_mls/src/groups/tests/test_delete_message.rs
+++ b/xmtp_mls/src/groups/tests/test_delete_message.rs
@@ -279,6 +279,7 @@ async fn test_true_out_of_order_deletion_by_sender() {
         sequence_id: 2,
         originator_id: 1,
         inserted_at_ns: 0,
+        should_push: false,
     };
     delete_message.store(&alix_conn)?;
 
@@ -324,6 +325,7 @@ async fn test_true_out_of_order_deletion_by_sender() {
         sequence_id: 1,
         originator_id: 1,
         inserted_at_ns: 0,
+        should_push: false,
     };
     message.store(&alix_conn)?;
 
@@ -397,6 +399,7 @@ async fn test_out_of_order_unauthorized_deletion_rejected() {
         sequence_id: 3,
         originator_id: 1,
         inserted_at_ns: 0,
+        should_push: false,
     };
     malicious_delete_message.store(&bo_conn)?;
 
@@ -434,6 +437,7 @@ async fn test_out_of_order_unauthorized_deletion_rejected() {
         sequence_id: 2,
         originator_id: 1,
         inserted_at_ns: 0,
+        should_push: false,
     };
     message.store(&bo_conn)?;
 
@@ -1001,6 +1005,7 @@ async fn test_out_of_order_sender_deletion_shows_correct_deleted_by() {
         sequence_id: 1,
         originator_id: 1,
         inserted_at_ns: 0,
+        should_push: false,
     };
     delete_message.store(&alix_conn)?;
 
@@ -1037,6 +1042,7 @@ async fn test_out_of_order_sender_deletion_shows_correct_deleted_by() {
         sequence_id: 2,
         originator_id: 1,
         inserted_at_ns: 0,
+        should_push: false,
     };
     original_message.store(&alix_conn)?;
 
--- a/xmtp_mls/src/messages/tests/test_deletion_validation.rs
+++ b/xmtp_mls/src/messages/tests/test_deletion_validation.rs
@@ -30,6 +30,7 @@ fn create_test_message(
         sequence_id: 1,
         originator_id: 1,
         inserted_at_ns: 0,
+        should_push: false,
     }
 }
 


cameronvoell:
This PR addresses https://github.com/xmtp/libxmtp/issues/2912. See XIP 76 for more detail: https://github.com/xmtp/XIPs/blob/main/XIPs/xip-76-delete-messages.md

xmtp_content_types/src/delete_message.rs
```diff
@@ -0,0 +1,77 @@
...
+pub struct DeleteMessageCodec;
+
+impl DeleteMessageCodec {
+    const AUTHORITY_ID: &'static str = "xmtp.org";
+    pub const TYPE_ID: &'static str = "deleteMessage";
+    pub const MAJOR_VERSION: u32 = 1;
+    pub const MINOR_VERSION: u32 = 0;
+
+    fn fallback() -> Option<String> {
+        Some("Deleted a message".to_string())
```
rygine: this would appear as just `Deleted a message` in the message feed, which i don't think is very useful. i actually don't think we need to have a fallback for this message type as i'm not sure it would provide any value. WDYT?

mchenani: I was thinking that would be a good solution for backward compatibility. Wdyt?

rygine: i'm just looking at it from a user perspective. if i saw "Deleted a message" in the message feed, i would be confused.

rygine: it's OK to have no fallback. the default behavior in that case is to display nothing.

mchenani: Fair enough. Will remove it.

bindings_node/src/content_types/deleted_message.rs
rygine, line 1: please rename this file to `deleted_message.rs`

bindings_wasm/src/content_types/decoded_message_content.rs
```diff
@@ -30,6 +30,8 @@ pub enum DecodedMessageContent {
   WalletSendCalls { content: WalletSendCalls },
   Intent { content: Option<Intent> },
   Actions { content: Option<Actions> },
+  DeleteMessage,
```
rygine: this is not necessary, right?

xmtp_db/src/encrypted_store/group_message/convert.rs
```diff
@@ -141,8 +142,16 @@ impl From<ContentType> for ContentTypeSave {
             ContentType::Reply => Self::Reply,
             ContentType::Text => Self::Text,
             ContentType::TransactionReference => Self::TransactionReference,
-            ContentType::Unknown => Self::Unknown,
-            _ => Self::Unknown,
+
+            // question: do we need to include these in the backup?
```
mchenani, line 147: @codabrink what's you opinion here?

codabrink: This field should be switched to a string. I'll create a follow-up PR.

PR #2780, opened 2025-11-13
Deprecate ReactionV2

rygine:
### Replace proto-based ReactionV2 with JSON `Reaction` v1.0 across FFI, Node, WASM, content types, and MLS decoding paths
Switch reaction content to JSON `Reaction` with string `action`/`schema` and optional `reference_inbox_id`, update codecs to `ReactionCodec` v1.0, and refactor bindings and decoded message types to use `Reaction` instead of `ReactionV2`. Key entry points include encoding/decoding functions and enum payloads updated to map to strings.

#### 📍Where to Start
Start with `ContentCodec<Reaction>` in [reaction.rs](https://github.com/xmtp/libxmtp/pull/2780/files#diff-24e4e78e8578981034245acc25394ec3f75bb7e0edf8dede5259d1ab82f96354), then follow `TryFrom<EncodedContent>` in [decoded_message.rs](https://github.com/xmtp/libxmtp/pull/2780/files#diff-0f2f8821d2e9aa1d55c1968fcc5258870086f982d88c44b4126c4af508b82763) and FFI/node/wasm reaction encode/decode functions.

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 1ac424d.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -95,7 +95,7 @@ use xmtp_proto::types::ApiIdentifier;
 use xmtp_proto::types::Cursor;
 use xmtp_proto::xmtp::device_sync::{BackupElementSelection, BackupOptions};
 use xmtp_proto::xmtp::mls::message_contents::EncodedContent;
-use xmtp_proto::xmtp::mls::message_contents::content_types::{MultiRemoteAttachment, ReactionV2};
+use xmtp_proto::xmtp::mls::message_contents::content_types::MultiRemoteAttachment;
 
 // Re-export types from message module that are used in public APIs
 pub use crate::message::{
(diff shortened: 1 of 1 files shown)


neekolas:
I think there are some apps that are now switching _to_ ReactionV2? Curious about the backstory of the deprecation

rygine:
> Curious about the backstory of the deprecation

@neekolas i'm working on ensuring that all content types are in `libxmtp` and available in the bindings. to that end, i was looking to cut down on complexity where possible. ReactionV2 was initially created as a concept when we decided to starting bringing content types into `libxmtp`. since it uses protos and is not backwards-compatible, i figured we could simplify the code to support reactions across the board by removing it.

that being said, i'm certainly not opposed to using protos for all current content types, but since most of them don't, i'm leaning towards consistency.

> I think there are some apps that are now switching _to_ ReactionV2?

if that's the case, then this PR is a no-go, which i'm OK with.

rygine:
closing this for now

PR #2794, opened 2025-11-14
Consent find_by_dm_id Performance Improvement

codabrink:
- Add a consent bench over consent `find_by_dm_id`.
- Add an index that provided a 100x performance improvement to `find_by_dm_id`.
- Update the query to provide another 20% improvement on-top of that.

### DIFF
--- a/xmtp_db/migrations/2025-11-14-185054-0000_add_dm_id_index/up.sql
+++ b/xmtp_db/migrations/2025-11-14-185054-0000_add_dm_id_index/up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX idx_groups_dm_id ON groups(dm_id) WHERE dm_id IS NOT NULL;


xmtp_db/migrations/2025-11-14-185054-0000_performance-improvements/up.sql
```diff
@@ -0,0 +1,24 @@
...
+CREATE INDEX idx_group_messages_originator_sequence
+ON group_messages(originator_id, sequence_id);
+
+-- For content_type filtering
+CREATE INDEX idx_group_messages_content_type
+ON group_messages(content_type);
+
+-- For delivery_status filtering
+CREATE INDEX idx_group_messages_delivery_status
+ON group_messages(delivery_status);
```
neekolas: Do we have benches or EXPLAINs showing these indexes improve query performance? 

Adding 6 new indexes to the table will have some negative impact on write performance and will lead to a bunch of extra storage...so its not free. We should be sure the query planner will ever use these in their current form.

xmtp_db/src/encrypted_store/association_state.rs
```diff
@@ -124,15 +124,18 @@ impl<C: ConnectionExt> QueryAssociationStateCache for DbConnection<C> {
...
-
-        let query = dsl::association_state
+        let mut query = dsl::association_state
             .select((dsl::inbox_id, dsl::sequence_id, dsl::state))
-            .filter(
+            .into_boxed();
+
+        // Build OR conditions for each pair
+        for (inbox_id, sequence_id) in identifiers {
+            query = query.or_filter(
```
neekolas: This is more intuitive query. I'd profile how this performs across hundreds of inboxes (which is a common case when you join a large group). It's a much more complex query.

The old implementation relies on the fact that `sequence_id`s are globally unique, so they should get the same results every time. We should do whichever one is faster.

xmtp_db/migrations/2025-11-14-185054-0000_performance-improvements/up.sql
codabrink: I'll be coming back through to verify / prune. These are just output from llm queries that I put in this PR as a rough draft. It originally provided much more.

PR #2802, opened 2025-11-17
ordering proptests

insipx:
#2789

PR #2801, opened 2025-11-17
dependency resolution tests

insipx:
#2789

PR #2788, opened 2025-11-14
set dependencies in send message

insipx:
### Annotate group message send with dependency cursors and add SHA-256 hashing across d14n envelopes in `xmtp_api_d14n` and `xmtp_mls`
Implement dependency resolution for group message publish by computing payload SHA-256s and setting `AAD.depends_on` in `XmtpMlsClient::send_group_messages`, add `CursorStore::find_message_dependencies`, propagate cursors and hashes through `Envelope`, `EnvelopeCollection`, and extractors, and extend SQLite cursor store to return dependency cursors.

#### 📍Where to Start
Start with dependency injection and AAD setting in `XmtpMlsClient::send_group_messages` in [mls.rs](https://github.com/xmtp/libxmtp/pull/2788/files#diff-b41f084fe8fd9295035f4b6d73b1318093b6475f1f923e681a880e6ba94a2f20), then review the `CursorStore` extensions in [cursor_store.rs](https://github.com/xmtp/libxmtp/pull/2788/files#diff-028daad07b57a99602b70e297432f143b84633ff46fa3f2b9851f6b476260eae) and the SQLite implementation in [cursor_store.rs](https://github.com/xmtp/libxmtp/pull/2788/files#diff-92939812065b708096075a08e83053a96bbb0d48d2a13d665eba3336c85740ca).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized d2ad8ec.

PR #2782, opened 2025-11-14
add ordered query combinator

insipx:
### Add ordered query combinator and modify group message querying to return ordered envelopes via `queries::combinators::ordered_query::OrderedQuery` with network backoff dependency resolution
Introduce ordered querying that resolves dependencies with exponential backoff, switch topic handling from raw bytes to `Topic`, add least-merge vector clock operations, and update group message queries to build a `TopicCursor` via LCC and return ordered envelopes.

#### 📍Where to Start
Start with `ordered` behavior in `queries` by reviewing `MlsApiImpl::query_group_messages` in [xmtp_api_d14n/src/queries/d14n/mls.rs](https://github.com/xmtp/libxmtp/pull/2782/files#diff-b41f084fe8fd9295035f4b6d73b1318093b6475f1f923e681a880e6ba94a2f20), then follow into `queries::combinators::ordered_query::OrderedQuery` in [xmtp_api_d14n/src/queries/combinators/ordered_query.rs](https://github.com/xmtp/libxmtp/pull/2782/files#diff-a7f3188e32afea719721f5098f8f9cbde6c110942050e65a7abf44fd2218d0fe) and the resolver in [xmtp_api_d14n/src/protocol/resolve/network_backoff.rs](https://github.com/xmtp/libxmtp/pull/2782/files#diff-de7799794fbc3300554d0474e251ec4117a5b0773a085ba0f94dd6f863c2f51e).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 43ff1b3.

PR #2805, opened 2025-11-18
add published_in_epoch to group message

insipx:
- add published_in_epoch column to group message
- requires companion pr in [xmtp/proto ](https://github.com/xmtp/proto/pull/310)
- sets the epoch of a group_message when storing in the database. 

This will help ensure message dependencies can be programmed defensively by ensuring the epoch of a commit is 1 less then the message we are setting a dependency on.


insipx:
This PR is part of a stack of 11 bookmarks:

1. `trunk()`
1. [insipx/dependency-resolution](https://github.com/xmtp/libxmtp/pull/2676)
1. [insipx/causal-sort](https://github.com/xmtp/libxmtp/pull/2770)
1. [insipx/resolution](https://github.com/xmtp/libxmtp/pull/2771)
1. [insipx/network-backoff](https://github.com/xmtp/libxmtp/pull/2775)
1. [insipx/ordered-combinator](https://github.com/xmtp/libxmtp/pull/2782)
1. [insipx/test-resolution](https://github.com/xmtp/libxmtp/pull/2801)
1. [insipx/test-ordering](https://github.com/xmtp/libxmtp/pull/2802)
1. **insipx/add-published-in-epoch ← this PR**
1. [insipx/sql-queries](https://github.com/xmtp/libxmtp/pull/2819)
1. [insipx/sync-conn](https://github.com/xmtp/libxmtp/pull/2820)
1. [insipx/set-dependencies](https://github.com/xmtp/libxmtp/pull/2788)

---

xmtp_proto/src/gen/xmtp.message_api.v1.rs
```diff
@@ -1,4 +1,54 @@
...
+        super::super::message_contents::Signature,
+    >,
+}
+impl ::prost::Name for Token {
+    const NAME: &'static str = "Token";
+    const PACKAGE: &'static str = "xmtp.message_api.v1";
+    fn full_name() -> ::prost::alloc::string::String {
+        "xmtp.message_api.v1.Token".into()
+    }
+    fn type_url() -> ::prost::alloc::string::String {
```
tylerhawkes, line 24: What can be done so that generation of these are stable and don't move around when we add more code?

insipx: we could try using https://docs.rs/tonic/latest/tonic/macro.include_proto.html

This would generate the protos into the rust `OUT_DIR`. the downside is first compile will be a bit longer as it needs to clone `xmtp/proto` and `googleapis` to generate the protos. but it will delete the entire `gen/` directory, which will create smaller PRs.

at that point, i would argue for moving xmtp_proto, or at least just the "proto" part into xmtp/proto, then renaming `xmtp_proto` to `xmtp_types` or `xmtp_primitives` or something. Then we could get rid of the companion PR pattern and proto generation could just be a cargo update which makes versioning clearer (rather than ad-hoc commit refs we store in `txt`, cargo would take care of it in the lockfile)

tylerhawkes: I think we open an issue with prost. My guess is they're using some hash that is dependent on the contents of the protos and that's why things change around when a single field is added.

PR #2807, opened 2025-11-18
Expose more content types in bindings

rygine:
### Expose text, read receipt, wallet send calls, transaction reference, attachment, and group-updated content codecs across FFI, Node, and WASM bindings and enforce `u32` remote attachment `content_length` with fallible mappers
Add encode/decode APIs for multiple content types across bindings, extend `GroupUpdated` fields to include admin and super admin inbox changes, and change remote attachment `content_length` to `u32` with `TryFrom` validation and standardized error mapping.

#### 📍Where to Start
Start with the FFI codecs and struct changes in [message.rs](https://github.com/xmtp/libxmtp/pull/2807/files#diff-ebd102a369643d46ce058998aa216b3d38724c782b29b2d20f6140ae7acdd691) and the new encode/decode functions in [mls.rs](https://github.com/xmtp/libxmtp/pull/2807/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f), then follow the Node bindings in [remote_attachment.rs](https://github.com/xmtp/libxmtp/pull/2807/files#diff-6b9d793d8c67322d0d2557dded41bef74983cf826275f36a7d1f6b94245269e2) and WASM bindings in [remote_attachment.rs](https://github.com/xmtp/libxmtp/pull/2807/files#diff-00e8ba5ba639bc99f49e101d55f520f0c638e2445a70c6ae5b0bc6a0f7e99c3c).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 626b394.

### DIFF
--- a/bindings_node/src/content_types/group_updated.rs
+++ b/bindings_node/src/content_types/group_updated.rs
@@ -170,7 +170,9 @@ pub fn decode_group_updated(bytes: Uint8Array) -> Result<GroupUpdated> {
   let encoded = EncodedContent::decode(bytes.as_ref()).map_err(ErrorWrapper::from)?;
 
   // Use GroupUpdatedCodec to decode into GroupUpdated and convert to GroupUpdated
-  GroupUpdatedCodec::decode(encoded)
-    .map(Into::into)
-    .map_err(|e| napi::Error::from_reason(e.to_string()))
+  Ok(
+    GroupUpdatedCodec::decode(encoded)
+      .map(Into::into)
+      .map_err(ErrorWrapper::from)?,
+  )
 }
--- a/bindings_node/src/content_types/remote_attachment.rs
+++ b/bindings_node/src/content_types/remote_attachment.rs
@@ -81,8 +81,7 @@ pub fn decode_remote_attachment(bytes: Uint8Array) -> Result<RemoteAttachment> {
   let encoded_content = EncodedContent::decode(bytes.as_ref()).map_err(ErrorWrapper::from)?;
 
   // Use RemoteAttachmentCodec to decode into RemoteAttachment
-  let attachment = RemoteAttachmentCodec::decode(encoded_content)
-    .map_err(|e| napi::Error::from_reason(e.to_string()))?;
+  let attachment = RemoteAttachmentCodec::decode(encoded_content).map_err(ErrorWrapper::from)?;
 
   // Convert to bindings type with error handling
   RemoteAttachment::try_from(attachment)
--- a/bindings_node/src/client.rs
+++ b/bindings_node/src/client.rs
@@ -205,15 +205,12 @@ pub async fn create_client(
       let key: EncryptionKey = key
         .try_into()
         .map_err(|_| Error::from_reason("Malformed 32 byte encryption key"))?;
-      let db = NativeDb::new(&storage_option, key)
-        .map_err(|e| Error::from_reason(format!("Error creating native database {}", e)))?;
-      EncryptedMessageStore::new(db)
-        .map_err(|e| Error::from_reason(format!("Error Creating Encrypted Message store {}", e)))?
+      let db = NativeDb::new(&storage_option, key).map_err(ErrorWrapper::from)?;
+      EncryptedMessageStore::new(db).map_err(ErrorWrapper::from)?
     }
     None => {
-      let db = NativeDb::new_unencrypted(&storage_option)
-        .map_err(|e| Error::from_reason(e.to_string()))?;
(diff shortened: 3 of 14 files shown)


bindings_ffi/src/mls.rs
```diff
@@ -3017,6 +3019,62 @@ pub fn decode_actions(bytes: Vec<u8>) -> Result<FfiActions, GenericError> {
...
+    encoded
+        .encode(&mut buf)
+        .map_err(|e| GenericError::Generic { err: e.to_string() })?;
+
+    Ok(buf)
+}
+
+#[uniffi::export]
+pub fn decode_text(bytes: Vec<u8>) -> Result<String, GenericError> {
+    let encoded_content = EncodedContent::decode(bytes.as_slice())
```
codabrink, line 3092: No need to change now, but for the future you can just borrow the bytes and the compiler will auto-deref to the slice. `::decode(&bytes)`

bindings_node/src/content_types/group_updated.rs
```diff
@@ -28,6 +39,71 @@ impl From<xmtp_proto::xmtp::mls::message_contents::GroupUpdated> for GroupUpdate
         .into_iter()
         .map(|c| c.into())
         .collect(),
+      left_inboxes: updated.left_inboxes.into_iter().map(|i| i.into()).collect(),
+      added_admin_inboxes: updated
```
tylerhawkes, line 43: This is repeated so often, it would probably benefit from a macro. I can help you write it. It would be nice to have something like `into!(updated.added_admin_inboxes)` and that would do the transformation.

bindings_node/src/content_types/remote_attachment.rs
```diff
@@ -9,21 +16,73 @@ pub struct RemoteAttachment {
...
+  fn try_from(
+    remote: xmtp_content_types::remote_attachment::RemoteAttachment,
+  ) -> std::result::Result<Self, Self::Error> {
+    let content_length = u32::try_from(remote.content_length).map_err(|_| {
+      Error::from_reason(format!(
+        "content_length {} exceeds maximum value of {} bytes",
+        remote.content_length,
+        u32::MAX
+      ))
+    })?;
```
tylerhawkes, line 35: Can we unify these types, so that a `RemoteAttachment.content_length` is a u32. I know everything is basically `usize == u64`, but not having to have these be fallible would be nice.

rygine: i don't see why not. i'm not sure why `usize` was originally used. i don't think we need to support attachments > 4GB. @neekolas WDYT?

bindings_wasm/src/content_types/read_receipt.rs
```diff
@@ -9,3 +14,38 @@ impl From<xmtp_content_types::read_receipt::ReadReceipt> for ReadReceipt {
...
+  let encoded =
+    ReadReceiptCodec::encode(read_receipt.into()).map_err(|e| JsError::new(&format!("{}", e)))?;
+
+  // Encode the EncodedContent to bytes
+  let mut buf = Vec::new();
+  encoded
+    .encode(&mut buf)
+    .map_err(|e| JsError::new(&format!("{}", e)))?;
+
+  Ok(Uint8Array::from(buf.as_slice()))
```
tylerhawkes, line 38: Vec's can be directly turned into Unit8Arrays, so removing the `.as_slice()` here will be more efficient.

bindings_wasm/src/content_types/transaction_reference.rs
```diff
@@ -49,3 +67,45 @@ impl From<xmtp_content_types::transaction_reference::TransactionMetadata> for Tr
...
+  let encoded = TransactionReferenceCodec::encode(transaction_reference.into())
+    .map_err(|e| JsError::new(&format!("{}", e)))?;
+
+  // Encode the EncodedContent to bytes
+  let mut buf = Vec::new();
+  encoded
+    .encode(&mut buf)
+    .map_err(|e| JsError::new(&format!("{}", e)))?;
+
+  Ok(Uint8Array::from(buf.as_slice()))
```
tylerhawkes, line 98: Here as well. Should probably grep the diff to make sure we get them all.

bindings_wasm/src/content_types/read_receipt.rs
rygine: i tried that as well as passing `&buf`. the compiler complains in both instances.

tylerhawkes: I was mistaken. It must have been the node bindings that allows this.

PR #2814, opened 2025-11-20
Fix bug with load_mls_group_with_lock

neekolas:
## tl;dr

- Fixes a bug where `load_mls_group_with_lock` does not return an error or wait if the group is already locked (synchronously or asynchronously)
- Replaces a lot of usage of `load_mls_group_with_lock` with `load_mls_group`...which does basically the same thing as the old version but more honestly.

## The issue

Previously `load_mls_group_with_lock` would call `get_lock_sync`, which returns an error. But it ignored the result and wouldn't return if it failed. That means that if anyone had already acquired the lock, it would do nothing.

If you were the only lock holder you _could_ block someone from calling `load_mls_group_with_lock_async`, which does successfully wait until the lock was released. But even that would only respect the first caller of `load_mls_group_with_lock`, so it isn't a functional semaphore or `RwLock`​ either.

## Why not just fix the bug?

I decided to change to the new `load_mls_group` in most cases for a few reasons.

1. The current implementation would have had a ton of lock errors if it actually worked. We call methods that call `load_mls_group_with_lock` from inside the callbacks to `load_mls_group_with_lock_async`
2. Most of the calls to `load_mls_group_with_lock` are for read-only data. If an app wants to call `group.members()` they don't want (and can't) coordinate their calls around whether or not we are syncing or receiving messages from a stream. If this lock actually worked it would make our SDKs extremely flaky since we would be returning lock errors everywhere. 

## So, this is safe?

I'm not so sure. We have to thoroughly audit all usage of any method that previously called any of the methods that relied on `load_mls_group_with_lock` and make sure that they shouldn't actually be taking out a lock (and erroring if one isn't available) or using the async method and waiting until the lock becomes available. We generally use `load_mls_group_with_lock_async` for heavy lifting...but maybe not exclusively. So we shouldn't merge this without some more investigation tomorrow.

## Other notes

- I kept the callback format for `load_mls_group`​...but I probably should just make it a regular function.

### DIFF
--- a/xmtp_mls/src/groups/intents.rs
+++ b/xmtp_mls/src/groups/intents.rs
@@ -1006,7 +1006,7 @@ pub(crate) mod tests {
 
         let storage = group.context.mls_storage();
         let decrypted_message = group
-            .load_mls_group_with_lock(storage, |mut mls_group| {
+            .load_mls_group(storage, |mut mls_group| {
                 Ok(mls_group
                     .process_message(&XmtpOpenMlsProviderRef::new(storage), mls_message.clone())
                     .unwrap())
--- a/xmtp_mls/src/groups/members.rs
+++ b/xmtp_mls/src/groups/members.rs
@@ -35,7 +35,7 @@ where
     pub async fn members(&self) -> Result<Vec<GroupMember>, GroupError> {
         let db = self.context.db();
         let storage = self.context.mls_storage();
-        let group_membership = self.load_mls_group_with_lock(storage, |mls_group| {
+        let group_membership = self.load_mls_group(storage, |mls_group| {
             Ok(extract_group_membership(mls_group.extensions())?)
         })?;
         let requests = group_membership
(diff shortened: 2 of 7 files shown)


neekolas:
* **#2821** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2821?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2814** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2814?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2814?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=2814).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #2815, opened 2025-11-20
Snapshot improvements

codabrink:
Some snapshot files are journaling in WAL mode, which isn't compatible with in-memory databases.

This updates the journal mode to allow loading for a more resilient snapshot feature, and then runs pending migrations on snapshots.

### DIFF
--- a/xmtp_db/src/test_utils.rs
+++ b/xmtp_db/src/test_utils.rs
@@ -192,7 +192,7 @@ mod native {
 
                 if result.is_ok() {
                     break store;
-                } else if i >= 1 {
+                } else if i >= 1 || path.is_none() {
                     #[allow(clippy::panicking_unwrap)]
                     result.unwrap();
                 }
--- a/xmtp_mls/src/utils/test/tester_utils.rs
+++ b/xmtp_mls/src/utils/test/tester_utils.rs
@@ -115,7 +115,7 @@ where
 
     pub fn save_db_snapshot_to_file(&self, path: impl AsRef<Path>) {
         let snapshot = self.db_snapshot();
-        std::fs::write(path, &snapshot);
+        std::fs::write(path, &snapshot).unwrap();
     }
 }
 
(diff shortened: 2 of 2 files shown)


xmtp_db/src/test_utils.rs
```diff
@@ -154,25 +158,67 @@ mod wasm {
...
-            let store = EncryptedMessageStore::new_uninit(db).unwrap();
+        async fn create_ephemeral_store_from_snapshot(
+            mut snapshot: &[u8],
+            path: Option<impl AsRef<Path>>,
+        ) -> crate::DefaultStore {
+            let path = path.as_ref();
+            let mut buffer;
+
+            let mut i = 0;
+            let store = loop {
```
tylerhawkes, line 183: This could be `for i in 0..2`​ and then the manual tracking of `i`​ could be removed

xmtp_mls/src/utils/test/tester_utils.rs
```diff
@@ -496,7 +500,9 @@ where
     }
 
     pub fn snapshot_file(mut self, snapshot_path: impl Into<PathBuf>) -> Self {
-        let snapshot = std::fs::read(snapshot_path.into()).unwrap();
+        let snapshot_path = snapshot_path.into();
+        let snapshot = std::fs::read(snapshot_path.clone()).unwrap();
```
tylerhawkes:
```suggestion
        let snapshot = std::fs::read(&snapshot_path).unwrap();
```

tylerhawkes (changes requested):
Looks good other than the suggested changes, and comments 1 & 2 from claude

xmtp_db/src/test_utils.rs
codabrink: Can't break with values in for loops

codabrink:
Weird lint didn't catch #2

PR #2817, opened 2025-11-20
Add contacts API

rygine:
### Add a contacts API with database tables, FTS triggers, and device sync import/export across `xmtp_mls` and `xmtp_db`
Introduce `contacts` CRUD in `xmtp_mls`, add ORM models and migrations for contact data with FTS5 and JSON view in `xmtp_db`, and integrate contacts into backup/export/import and FFI/device sync.

#### 📍Where to Start
Start with the `contacts` high-level API in [xmtp_mls/src/contacts/mod.rs](https://github.com/xmtp/libxmtp/pull/2817/files#diff-764ca8f4160f35880047d256689095f7fe7c682f937baa737f06a52620c1501e), then follow database interactions in [xmtp_db/src/encrypted_store/contacts/mod.rs](https://github.com/xmtp/libxmtp/pull/2817/files#diff-91e6719f5c6003156547da248996a516c0affa897b6a477aeb56ee5a4c291754) and the migration in [xmtp_db/migrations/2025-12-28-000000_create_contacts/up.sql](https://github.com/xmtp/libxmtp/pull/2817/files#diff-69be67309932136544221f3a9b350f8a42d65b2a8ee8f59c407106057aac2ef7).

#### Changes since #2817 opened

- Modified contact and companion table CRUD operations in `contacts::DbConnection<C>` `QueryContacts` implementation to return inserted rows directly using `get_result`, check `rows_affected` and return `diesel::result::Error::NotFound` for updates and deletes when no rows match, and improved `get_contacts` query logic to ignore empty search strings, properly build FTS search patterns, and clamp limit/offset to non-negative i64 bounds [324e0b6]
- Updated `contacts::FullContact` `From<RawFullContact>` implementation to deserialize companion field JSON with error handling that logs warnings including `inbox_id` context and defaults to empty vectors on parse failure [324e0b6]
- Moved contact-related methods from `client::Client` implementation to new dedicated `contacts::client` module [324e0b6]
- Enhanced contact tests with increased timing tolerance and expanded cascade delete verification using raw SQL count queries [324e0b6]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 324e0b6. 20 files reviewed, 25 issues evaluated, 24 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_archive/src/export_stream/contact_save.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 14](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_archive/src/export_stream/contact_save.rs#L14): The `backup_records` function never updates `state.cursor`, causing an infinite loop when called repeatedly. Each call loads the same cursor value (line 14), fetches the same batch of records, but never increments the cursor. The caller likely expects the cursor to advance after each batch. <b>[ Already posted ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/contacts/addresses.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 71](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_db/src/encrypted_store/contacts/addresses.rs#L71): The `From<AddressData> for NewContactAddress` sets `contact_id: 0` with a comment saying it will be set later. If the caller forgets to set the correct `contact_id` before insertion, the address will be associated with contact ID 0, which may not exist or may be the wrong contact. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/contacts/mod.rs — 1 comment posted, 5 evaluated, 4 filtered</summary>

- [line 566](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_db/src/encrypted_store/contacts/mod.rs#L566): OFFSET without LIMIT causes a SQL syntax error in SQLite. At lines 566-568, when `query.offset` is `Some` but `query.limit` is `None`, the generated SQL will contain `OFFSET X` without a preceding `LIMIT` clause. SQLite requires LIMIT when OFFSET is used, causing a runtime database error. <b>[ Already posted ]</b>
- [line 596](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_db/src/encrypted_store/contacts/mod.rs#L596): In `contacts_paged`, the `limit` and `offset` parameters are directly formatted into the SQL string without validation. If negative values are passed, SQLite will fail with an error like "LIMIT clause must not be negative". This is inconsistent with `get_contacts` which clamps these values to non-negative using `.clamp(0, i64::MAX)`. <b>[ Already posted ]</b>
- [line 1224](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_db/src/encrypted_store/contacts/mod.rs#L1224): The test relies on a time-based assertion (`updated_at_ns > original_created_at`) with a 20ms sleep, but this could still be flaky if the system clock has low resolution or if the `updated_at_ns` is set before the `update_contact` call rather than during it. If `update_contact` doesn't actually update the `updated_at_ns` field, the assertion on line 1224 will fail. <b>[ Already posted ]</b>
- [line 1720](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_db/src/encrypted_store/contacts/mod.rs#L1720): The test accesses `results[0]`, `results[1]`, and `results[2]` without first asserting that `results.len() == 3`. If `get_contacts` returns fewer than 3 items due to a bug in the underlying implementation, the test will panic with an index out of bounds error rather than providing a clear assertion failure message indicating the actual vs expected count. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/contacts/mod.rs — 0 comments posted, 10 evaluated, 10 filtered</summary>

- [line 82](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L82): The `update` method modifies the contact in the database but does not update the `Contact` struct's fields (e.g., `display_name`, `first_name`, etc.). After calling `update()`, the `Contact` instance holds stale data that no longer matches the database, which can lead to inconsistent state if the caller continues using the same `Contact` object. <b>[ Low confidence ]</b>
- [line 158](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L158): The `addresses()` method uses `filter_map` with `ContactAddress::new`, which silently drops addresses when the constructor returns `None`. If valid address data in the database fails to construct a `ContactAddress` for any reason, it will be silently omitted from the results with no warning or error. <b>[ Already posted ]</b>
- [line 304](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L304): The test assumes `phones[0]` will be the first phone number added ("555-1234"), but unless `phone_numbers()` guarantees insertion order, this test could be flaky. Many storage backends do not guarantee retrieval order matches insertion order. The assertion on line 304-305 may intermittently fail if the underlying implementation returns phone numbers in a different order (e.g., by ID, alphabetically, or non-deterministically). <b>[ Already posted ]</b>
- [line 439](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L439): Accessing `updated_phones[0]` without first verifying that `updated_phones` is non-empty could cause a panic with an unclear index-out-of-bounds error if the update operation unexpectedly removes the phone number entry. Adding `assert_eq!(updated_phones.len(), 1);` before accessing the element would make test failures more diagnostic. <b>[ Low confidence ]</b>
- [line 489](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L489): Direct indexing with `emails[0]`, `updated[0].email`, and `updated[0].delete()` at lines 489, 494, and 496 will panic with an unclear "index out of bounds" error if the `emails()` call returns an empty vector. While this is test code and panicking indicates a bug, the error message would be unhelpful. Consider using `assert!(!emails.is_empty())` before accessing or using `emails.first().expect("email should exist")` for clearer failure messages. <b>[ Already posted ]</b>
- [line 519](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L519): Accessing `urls[0]` without first asserting that the vector is non-empty. If `add_url` fails to add the URL (despite returning `Ok`), this will panic with an index out of bounds error rather than a meaningful test failure. Consider using `assert_eq!(urls.len(), 1)` before indexing or using `urls.first().expect("expected one url")`. <b>[ Low confidence ]</b>
- [line 526](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L526): Accessing `updated[0]` without verifying the vector is non-empty. If the update operation unexpectedly removes the URL or the refresh logic has a bug, this will panic with an unclear index out of bounds error. Adding an explicit length assertion before indexing would provide clearer test diagnostics. <b>[ Already posted ]</b>
- [line 549](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L549): Index access `wallets[0]` on line 549 assumes `add_wallet_address` succeeded and returned a non-empty collection. If the wallet wasn't added (e.g., due to a silent failure or race condition), this would panic rather than providing a meaningful test failure message. Consider using `wallets.first().expect("wallet should exist after add")` or an explicit assertion on `wallets.len()` before accessing. <b>[ Low confidence ]</b>
- [line 583](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L583): Unchecked index access at `addrs[0]` on line 583 could panic if `contact.addresses()` returns an empty vector. While the test expects one address after `add_address`, if the underlying implementation has a bug or the add fails silently, this would cause an unclear panic rather than a meaningful test failure. Consider using `addrs.first().expect("expected one address after add")` or adding an assertion like `assert_eq!(addrs.len(), 1)` before indexing. <b>[ Low confidence ]</b>
- [line 592](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/mod.rs#L592): Unchecked index access at `updated[0]` on lines 592, 593, and 595 could panic if the vector is empty. Adding an explicit length assertion before indexing would make test failures more informative. <b>[ Already posted ]</b>
</details>

<details>
<summary>xmtp_mls/src/contacts/url.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 25](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/contacts/url.rs#L25): The `update` method modifies the database but does not update the struct's `url` and `label` fields. After calling `update()`, the `ContactUrl` instance will contain stale values that don't match the database state. Callers accessing `self.url` or `self.label` after an update will see the old values instead of the newly updated ones. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/device_sync/contact_sync.rs — 0 comments posted, 6 evaluated, 6 filtered</summary>

- [line 51](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/groups/device_sync/contact_sync.rs#L51): The `import_contacts` function logs warnings for failed contact imports but continues processing. If many contacts fail to import due to a systemic issue (e.g., database corruption), the function will return a partial success count without indicating the failure rate, potentially masking serious problems. <b>[ Low confidence ]</b>
- [line 77](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/groups/device_sync/contact_sync.rs#L77): The `import_single_contact` function compares `contact_save.updated_at_ns` with `existing_contact.updated_at_ns` to determine if an update should occur, but if the source device's clock is behind, valid updates may be silently ignored. Additionally, if `updated_at_ns` is 0 or unset in the proto, this comparison may incorrectly skip updates. <b>[ Low confidence ]</b>
- [line 116](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/groups/device_sync/contact_sync.rs#L116): The `delete_all_child_data` function at line 116-119 only deletes addresses that have `Some(id)`, but `AddressData::id` could be `None` for addresses returned from `get_addresses`. If the database returns addresses with `id: None`, those addresses will not be deleted, leaving orphaned data. <b>[ Low confidence ]</b>
- [line 273](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/groups/device_sync/contact_sync.rs#L273): The 10ms sleep at line 273 is a fragile timing assumption. On heavily loaded systems or slow CI environments, system clock granularity or scheduling delays could result in both timestamps being identical or even reversed, causing the test to fail intermittently. A more robust approach would be to explicitly control the timestamps in the test data or use a larger sleep duration. <b>[ Already posted ]</b>
- [line 311](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/groups/device_sync/contact_sync.rs#L311): The 10ms sleep on line 311 relies on timestamp granularity to differentiate "older" from "newer" contacts. On systems with coarse-grained timestamps (e.g., some filesystems or databases using second-level precision), 10ms may not be sufficient to create a distinguishable timestamp difference, causing the test to be flaky. <b>[ Already posted ]</b>
- [line 324](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/groups/device_sync/contact_sync.rs#L324): The assertion `assert_eq!(imported_count, 1)` on line 324 may be incorrect. If `import_contacts` returns the count of contacts that were actually updated/imported, and the test verifies that the contact was NOT updated (because the existing one is newer), then `imported_count` should be 0, not 1. This assertion either tests the wrong expected value or indicates a misunderstanding of what `import_contacts` returns, potentially causing the test to pass with incorrect behavior. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/mls_store.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 169](https://github.com/xmtp/libxmtp/blob/324e0b658fcbc3b219995407cfbd2b503c5c6306/xmtp_mls/src/mls_store.rs#L169): In `find_groups_with_inbox_ids`, the call to `load_mls_group_with_lock(...).ok()` on line 169 silently converts all errors to `None`, causing groups that fail to load (due to transient errors, database locks, or corruption) to be silently excluded from results. This means a temporary failure could cause matching groups to be incorrectly omitted without any error indication to the caller. <b>[ Already posted ]</b>
</details>

</details>

### DIFF
--- a/xmtp_mls/src/client.rs
+++ b/xmtp_mls/src/client.rs
@@ -11,7 +11,6 @@ use xmtp_macro::log_event;
 
 use crate::{
     builder::SyncWorkerMode,
-    contacts::Contact,
     context::XmtpSharedContext,
     groups::{
         ConversationListItem, GroupError, MlsGroup,
@@ -38,10 +37,7 @@ use xmtp_db::{
     ConnectionExt, NotFound, StorageError, XmtpDb,
     consent_record::{ConsentState, ConsentType, StoredConsentRecord},
     db_connection::DbConnection,
-    encrypted_store::{
-        contacts::{ContactData, ContactsQuery, QueryContacts},
-        conversation_list::ConversationListItem as DbConversationListItem,
-    },
+    encrypted_store::conversation_list::ConversationListItem as DbConversationListItem,
     group::{ConversationType, GroupMembershipState, GroupQueryArgs},
     group_message::StoredGroupMessage,
     identity::StoredIdentity,
@@ -1035,40 +1031,6 @@ where
         }
 
         Ok(can_message)
-    }
-
-    pub fn get_contact(&self, inbox_id: &str) -> Result<Option<Contact<Context>>, ClientError>
-    where
-        Context: Clone,
-    {
-        let contact = self.context.db().get_contact(inbox_id)?;
-        Ok(contact.map(|c| Contact::new(self.context.clone(), c)))
-    }
-
-    pub fn list_contacts(
-        &self,
-        query: Option<ContactsQuery>,
-    ) -> Result<Vec<Contact<Context>>, ClientError>
-    where
-        Context: Clone,
-    {
-        let contacts = self.context.db().get_contacts(query)?;
-        Ok(contacts
-            .into_iter()
-            .map(|c| Contact::new(self.context.clone(), c))
-            .collect())
-    }
-
-    pub fn create_contact(
-        &self,
-        inbox_id: &str,
-        data: ContactData,
-    ) -> Result<Contact<Context>, ClientError>
-    where
-        Context: Clone,
-    {
-        let stored = self.context.db().add_contact(inbox_id, data)?;
-        Ok(Contact::from_stored(self.context.clone(), stored))
     }
 }
 
--- a/xmtp_mls/src/contacts/client.rs
+++ b/xmtp_mls/src/contacts/client.rs
@@ -1,0 +1,36 @@
+use super::Contact;
+use crate::{Client, client::ClientError, context::XmtpSharedContext};
+use xmtp_db::encrypted_store::contacts::{ContactData, ContactsQuery, QueryContacts};
+
+impl<Context> Client<Context>
+where
+    Context: XmtpSharedContext + Clone,
+{
+    /// Get a contact by inbox ID.
+    pub fn get_contact(&self, inbox_id: &str) -> Result<Option<Contact<Context>>, ClientError> {
+        let contact = self.context.db().get_contact(inbox_id)?;
+        Ok(contact.map(|c| Contact::new(self.context.clone(), c)))
+    }
+
+    /// List contacts with optional filtering, search, and pagination.
+    pub fn list_contacts(
+        &self,
+        query: Option<ContactsQuery>,
+    ) -> Result<Vec<Contact<Context>>, ClientError> {
+        let contacts = self.context.db().get_contacts(query)?;
+        Ok(contacts
+            .into_iter()
+            .map(|c| Contact::new(self.context.clone(), c))
+            .collect())
+    }
+
+    /// Create a new contact.
+    pub fn create_contact(
+        &self,
+        inbox_id: &str,
+        data: ContactData,
+    ) -> Result<Contact<Context>, ClientError> {
+        let stored = self.context.db().add_contact(inbox_id, data)?;
+        Ok(Contact::from_stored(self.context.clone(), stored))
+    }
+}
--- a/xmtp_mls/src/contacts/mod.rs
+++ b/xmtp_mls/src/contacts/mod.rs
@@ -8,6 +8,7 @@ use xmtp_db::{
 use crate::context::XmtpSharedContext;
 
 mod address;
+mod client;
 mod email;
 mod phone_number;
 mod url;


xmtp_mls/src/groups/members.rs
```diff
@@ -20,6 +21,43 @@ pub struct GroupMember {
     pub consent_state: ConsentState,
 }
 
+/// Represents a contact aggregated across all conversations
+#[derive(Debug, Clone)]
+pub struct Contact {
+    pub inbox_id: InboxId,
+    pub account_identifiers: Vec<Identifier>,
+    pub installation_ids: Vec<Vec<u8>>,
```
tylerhawkes: There is a type for this now from xmtp_proto `InstallationId`

xmtp_mls/src/client.rs
```diff
@@ -2035,4 +2265,386 @@ pub(crate) mod tests {
             "Deleting non-existent message should return 0"
         );
     }
+
+    #[xmtp_common::test]
+    async fn test_list_contacts() {
```
tylerhawkes: This should probably go into it's own file to avoid making this file any bigger.

xmtp_mls/src/client.rs
```diff
@@ -819,6 +819,236 @@ where
...
+    ///
+    /// Filters:
+    /// - allowed_group_ids: only include members from these specific group IDs
+    /// - denied_group_ids: exclude members from these specific group IDs
+    /// - consent_states: only return contacts with the given consent states
+    /// - conversation_type: filter by conversation type (Group or Dm)
+    /// - allowed_states: only include contacts from groups with the given membership states
+    /// - created_after_ns/created_before_ns: filter by group creation time
+    /// - limit: limit the number of contacts returned
+    pub async fn list_contacts(
```
tylerhawkes: this function is 214 lines long. Can it be broken up in to smaller functions based on the `// Step ... ` comments?

xmtp_mls/src/groups/members.rs
rygine: so i tried this, but the `installation_ids` method of `AssociationState` still returns `Vec<Vec<u8>>` so i end up having to do an `.into()`. it's the same thing for `GroupId`. the underlying APIs haven't been updated with these new types.

PR #2820, opened 2025-11-21
switch with_connection to be sync with now_or_never

insipx:
this is mostly a QoL pr since i noticed the wasm ephemeral db builds completely synchronously. this makes using some test utils more flexible, like making it possible to use proptest for db queries (not used yet)


insipx:

1. `trunk()`
1. **insipx/sync-conn ← this PR**

---

PR #2821, opened 2025-11-21
[WIP] Simplify lock usage

neekolas:
### Replace internal group mutexes with explicit `MlsGroup.lock()` guards and update `MlsGroup.sync` to require a mutable guard across commit, publish, and receive paths
Introduce an explicit per-group commit lock via `MlsGroup.lock()` and remove the internal mutex and `MutexRegistry`. Convert `sync_with_conn` into `MlsGroup.sync_inner(&mut MlsGroupGuard)` and thread a mutable guard through commit-related APIs (`sync`, `publish_intents`, `receive`, `process_message`, membership updates, and intent resolution). Replace callback-based MLS group loaders with direct `load_mls_group(...)` reads across code and tests, and update context types to drop the mutex registry.

#### 📍Where to Start
Start with the lock and sync flow in `MlsGroup`: review `MlsGroup.lock()` and `MlsGroup.sync_inner(&mut MlsGroupGuard)` in [mod.rs](https://github.com/xmtp/libxmtp/pull/2821/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3) and the refactored callsites and helpers in [mls_sync.rs](https://github.com/xmtp/libxmtp/pull/2821/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 4542761. 16 files reviewed, 60 issues evaluated, 57 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_debug/src/app/generate/messages.rs — 0 comments posted, 4 evaluated, 3 filtered</summary>

- [line 212](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_debug/src/app/generate/messages.rs#L212): `mls_commit_lock` is acquired into `guard` (lines `204-206`) and then `group.update_group_description(words).await?` is called (line `212`) without passing the same `guard`. If `update_group_description` internally acquires `mls_commit_lock`, this will deadlock because the lock is already held and is unlikely to be re-entrant. Either pass the existing `guard` to `update_group_description` (if supported) or drop the `guard` before calling it, ensuring consistent lock usage like the new guarded `sync_inner`/`maybe_update_installations` calls. <b>[ Low confidence ]</b>
- [line 287](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_debug/src/app/generate/messages.rs#L287): The `client` mutex guard (`let client = client.lock().await;`) is held across multiple `await`s (`client.sync_welcomes().await?` and subsequent operations). Holding an async mutex guard across `await` can lead to deadlocks or starvation if awaited code needs the same lock or depends on progress from tasks that do. Consider narrowing the lock’s scope or restructuring to avoid holding the guard across `await`. <b>[ Low confidence ]</b>
- [line 294](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_debug/src/app/generate/messages.rs#L294): `rng.gen_range(0..*max_message_size)` will panic at runtime when `*max_message_size == 0` because the range `0..0` is empty. This makes the function crashable under a reachable input (zero max size). Add a guard or use a non-empty range (e.g., `1..=*max_message_size` with appropriate handling) or short-circuit when zero. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/builder.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 326](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/builder.rs#L326): Global `EVENTS_ENABLED` flag is only set to `true` when `!disable_events` (and implicitly when `!disable_workers` is true enough to register workers) but is never set back to `false` when events are disabled. On subsequent client builds in the same process, a prior `true` value can persist, causing components that read `EVENTS_ENABLED` to think events are enabled even when `disable_events` or `disable_workers` is set for the new client. This stale global state creates inconsistent behavior and violates configuration contract parity across builds. Consider explicitly setting `EVENTS_ENABLED.store(false, Ordering::SeqCst)` when `disable_events` is `true` or when workers are disabled, or eliminate the global by making event enablement instance-scoped. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/device_sync.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 254](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/device_sync.rs#L254): If `MlsGroup::create_and_insert(...)` succeeds but `add_missing_installations(&mut guard).await?` or `sync_group.sync().await?` fails, the newly created group has already been inserted and appears to be the primary sync group in the database. On the next call, `db.primary_sync_group()?` will likely return `Some`, causing the `Some` branch (`self.mls_store.group(&sync_group.id)?`) to be taken and skipping the missing installation addition and sync steps. This leaves the system in a partially initialized state without an automatic recovery path. Consider using a two-phase creation (mark as pending until post-creation steps succeed, then promote to primary), rolling back on failure, or ensuring the `Some` branch detects incomplete initialization and completes it. <b>[ Out of scope ]</b>
- [line 269](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/device_sync.rs#L269): Lock guard `guard` acquired via `mls_commit_lock().get_lock_async(...)` is kept alive beyond `add_missing_installations(&mut guard).await?` and until the end of the `None` arm block, which includes the awaited call `sync_group.sync().await?`. This holds the commit lock while performing `sync()`. If `sync()` attempts to acquire the same commit lock (directly or indirectly), this can deadlock. Even if it does not, it unnecessarily extends the critical section, increasing contention and risking re-entrancy issues. The guard should be dropped before calling `sync()` (e.g., by limiting its scope or calling `drop(guard)` before `sync()`). <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/device_sync/worker.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 342](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/device_sync/worker.rs#L342): The function holds the `sync_group` lock (`guard`) beyond the `sync_inner(&mut guard).await` call and then invokes `sync_group.find_messages(...)` while the guard remains in scope. If `find_messages` (or any later method in this path) attempts to acquire the same internal lock, this will deadlock. Even if it doesn’t lock, keeping the guard longer than necessary can block other tasks and cause contention. Drop `guard` immediately after `sync_inner` before calling other `sync_group` methods. <b>[ Low confidence ]</b>
- [line 343](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/device_sync/worker.rs#L343): Awaiting `sync_inner(&mut guard).await` while holding the `MutexGuard` may cause priority inversion or deadlock if `sync_inner` performs I/O and awaits operations that need the same lock from other tasks. Prefer minimizing the critical section and avoiding `.await` inside it unless proven safe. If required, refactor `sync_inner` to operate without holding the lock or split into locked/unlocked phases. <b>[ Low confidence ]</b>
- [line 365](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/device_sync/worker.rs#L365): Non-atomic check-then-send race can produce duplicate acknowledgments. After scanning messages to see if the `request_id` was already acknowledged, the code sends a new `Acknowledge` without any atomicity or CAS. Two installations can simultaneously observe no prior ack and both send, violating the documented contract: “Returns an error if request is already acknowledged by another installation.” The second sender will return `Ok(())` instead of `AlreadyAcknowledged`. To fix, enforce at-most-once semantics (e.g., server-side CAS, unique key constraint, or re-sync and re-check with a commit token). <b>[ Previously rejected ]</b>
- [line 466](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/device_sync/worker.rs#L466): The code acquires a mutex guard `let mut guard = sync_group.lock().await;` and then awaits inside `sync_group.sync_inner(&mut guard).await` while still holding the guard, and continues to hold the guard across subsequent operations including building `DeviceSyncRequestProto` and awaiting `self.send_device_sync_message(...)`. Holding an async mutex guard across `await`s and unrelated I/O increases contention and can deadlock if any awaited path (or other concurrent code) needs the same lock. The guard's lifetime extends to the end of the function, so the lock is held longer than necessary. To avoid deadlocks and reduce contention, drop the guard immediately after `sync_inner` completes (e.g., limit its scope to a block or call `drop(guard)` before constructing/sending the request) so that later awaits do not occur while the lock is held. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/intents.rs — 0 comments posted, 5 evaluated, 5 filtered</summary>

- [line 1004](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/intents.rs#L1004): The match on `mls_message` panics for any variant other than `ProtocolMessage::PrivateMessage`, with no prior guard ensuring that `message.message` is a private message. Passing a non-private `GroupMessage` will cause a runtime panic via `panic!("error mls_message")`. <b>[ Out of scope ]</b>
- [line 1008](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/intents.rs#L1008): Unwrapping the result of `group.load_mls_group(storage)` will panic if loading the group fails (e.g., missing/corrupt storage). This is a runtime termination on a recoverable error without prior guards. <b>[ Test / Mock code ]</b>
- [line 1011](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/intents.rs#L1011): Unwrapping the result of `openmls_group.process_message(...)` will panic on any processing error (e.g., invalid/corrupted message, wrong epoch). This is a runtime termination on a recoverable error. <b>[ Out of scope ]</b>
- [line 1015](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/intents.rs#L1015): The match on `decrypted_message.into_content()` panics if the content is not `ProcessedMessageContent::StagedCommitMessage`. If a non-commit or different processed message is provided, it will panic. <b>[ Out of scope ]</b>
- [line 1020](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/intents.rs#L1020): The assertion `assert!(path_update_leaf_node.is_some())` will fail if the staged commit does not contain an update path leaf node (e.g., certain commit types may omit it). Without guards ensuring the commit is specifically a key update, this can cause a test failure. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/intents/queue.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 227](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/intents/queue.rs#L227): Using `xmtp_common::rand_vec::<2>()` for `id` risks collisions across multiple calls, especially in tests creating many groups. A collision would lead to duplicate `id` insertion and the subsequent `store(...).unwrap()` panic at line 237, making tests flaky. <b>[ Test / Mock code ]</b>
- [line 235](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/intents/queue.rs#L235): `build().unwrap()` on `StoredGroup::builder()` can panic if builder validation fails (e.g., if additional required fields are introduced or defaults change). This leads to brittle tests that crash instead of reporting a controlled failure. <b>[ Test / Mock code ]</b>
- [line 237](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/intents/queue.rs#L237): `store(&context.mls_storage().db()).unwrap()` can panic on duplicate inserts when `group()` is called multiple times with the same `id`. This is exercised by `only_queues_for_unique`, which calls `group(&context, Some(vec![0]))` 10 times. If `StoredGroup::store` enforces unique `id` and returns an error on duplicates, the `unwrap()` at line 237 will cause a runtime panic, failing the test unexpectedly. <b>[ Test / Mock code ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/members.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 43](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/members.rs#L43): Potential integer truncation/overflow when converting `sequence_id` to `i64`. The code casts `sequence_id` with `as i64` (`.map(|(inbox_id, sequence_id)| (inbox_id, sequence_id as i64))`). If the source `sequence_id` can exceed `i64::MAX`, the cast will wrap to a negative value, which will then be used in `batch_read_from_cache` and `batch_get_association_state`, risking incorrect cache lookups or erroneous identity fetches. This can yield missing or incorrect association states and cause spurious `InvalidGroupMembership` errors or silently wrong member data. <b>[ Out of scope ]</b>
- [line 53](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/members.rs#L53): Coverage check for fetched association states relies on vector length equality rather than validating that every requested `InboxId` is present exactly once. If `batch_read_from_cache` returns entries that do not correspond 1:1 to the requested set (e.g., duplicates for one `InboxId` and missing another, or unexpected extras), `association_states.len() == requests.len()` could erroneously pass, producing an incomplete or duplicated member list; conversely, `len() !=` could trigger an unnecessary error even when all requested IDs are present (e.g., duplicates inflate length). The code should verify set equality by `InboxId` with no duplicates, not just length. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_sync.rs — 1 comment posted, 10 evaluated, 9 filtered</summary>

- [line 374](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L374): Removed internal mutex locking in `sync_inner` can allow concurrent execution and races on shared group state. Previously, `sync_with_conn` acquired `self.mutex` via `let _mutex = self.mutex.lock().await;`. The new `sync_inner(&mut MlsGroupGuard)` no longer acquires any lock itself and relies on the caller-provided `guard`. If callers invoke `sync_inner` without holding an exclusive lock equivalent to the removed mutex, concurrent calls could interleave `publish_intents(guard)`, `receive(guard)`, and `post_commit()` causing state corruption, double-application, or inconsistent invariants. This is a contract/behavior change: exclusivity moved from callee to caller. Ensure that `MlsGroupGuard` actually serializes access to the same critical section and is always held when calling `sync_inner`, and that `post_commit()` (which does not take `guard`) also participates in the same exclusion to avoid races between different code paths. <b>[ Low confidence ]</b>
- [line 1535](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L1535): Concurrency/atomicity regression: the previous implementation executed the load, process, and post-process steps inside `self.load_mls_group_with_lock_async(...)`, which presumably held a group-scoped lock for the entire critical section. The new code path removes that wrapper and instead accepts an unused `_guard: &mut MlsGroupGuard` parameter without actually invoking any locking logic in this function. This makes the exclusivity of the critical section dependent on the caller correctly holding the right `MlsGroupGuard` for the whole call. If a caller passes a guard that does not actually protect this group or calls without holding the intended lock, concurrent invocations can interleave: (1) two tasks can `OpenMlsGroup::load(...)` the same prior state; (2) both run `process_message_inner(...)` based on stale state; and (3) `post_process_message(...)` (e.g., cursor/epoch updates) can be applied out of order. This can lead to state forks, last-write-wins corruption, or out-of-order cursor updates, violating the at-most-once/ordering invariants the original lock enforced. Because `_guard` is not used by this function, there is no local enforcement ensuring that the lock is actually held for this group or spans the entire critical section. <b>[ Low confidence ]</b>
- [line 2159](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L2159): Intent is marked `published` in storage before attempting the network send, with no rollback on send failure. Specifically, `db.set_group_intent_published(...)` is executed inside a transaction and committed, then `self.context.api().send_group_messages(messages).await?` is performed. If the network call fails (or the process crashes) after the DB state is committed, the intent remains in `Published` state and will not be retried by `find_group_intents(..., Some(vec![IntentState::ToPublish]), ...)`, resulting in a permanently missed message. There is also no change of state back to `ToPublish` or `Processed` on this error path. <b>[ Low confidence ]</b>
- [line 2448](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L2448): `update_interval_ns` is accepted as `Option<i64>` and used directly with `unwrap_or(...)` to set `interval_ns` without validating non-negativity. A negative `interval_ns` makes the condition `elapsed_ns > interval_ns` almost always true, defeating throttling and causing unnecessary updates. Validate that `interval_ns >= 0` (or clamp to zero/minimum) before comparing. <b>[ Out of scope ]</b>
- [line 2452](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L2452): `elapsed_ns` is computed via `now_ns - last_ns` without guarding for clock skew or a forward-moving monotonicity guarantee. If `last_ns > now_ns` (e.g., due to clock adjustments or an incorrect stored value), subtraction can underflow or produce a negative value depending on integer type. For unsigned integers, this wraps to a very large value in release builds, making `elapsed_ns > interval_ns` spuriously true and causing premature updates; for signed integers, it yields a negative `elapsed_ns` that might mask real elapsed time. Add an explicit monotonic check or saturating/checked subtraction with a fallback (e.g., treat negative/underflow as zero or clamp). <b>[ Out of scope ]</b>
- [line 2509](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L2509): The function accepts a mutable guard `_guard: &mut MlsGroupGuard` (holding it across awaits to enforce exclusivity) but then ignores the guarded in-memory state and instead reloads the group from storage via `OpenMlsGroup::load(...)`. This can compute intents against a stale snapshot if in-memory updates protected by the guard have not yet been persisted, leading to incorrect membership deltas or spurious errors. Use the guarded `mls_group` (as in the previous `load_mls_group_with_lock_async` usage) to ensure consistency with the guarded state. <b>[ Low confidence ]</b>
- [line 2541](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L2541): Potential lossy cast: `let latest_sequence_id_u64 = *latest_sequence_id as u64;` and `updates.insert(inbox_id.to_string(), *latest_sequence_id as u64);` perform unchecked casts to `u64`. If `latest_sequence_id` is a signed integer type and can be negative (e.g., default sentinel values), the cast will wrap, producing a huge `u64` and causing incorrect comparisons and updates. Validate non-negativity or use a fallible conversion before casting. <b>[ Previously rejected ]</b>
- [line 2551](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L2551): The fold over `inbox_ids` returns `Err(GroupError::MissingSequenceId)` for any `(None, Some)` or `(None, None)` case in `latest_sequence_id_map.get(...)` vs `existing_group_membership.get(...)`. This causes the entire intent computation to fail if an inbox has no latest sequence entry (e.g., a member to be removed or a member with no updates present in storage). Additionally, the code queries updates for members slated for removal (`inbox_ids_to_remove`) due to constructing `inbox_ids` from all existing members, which increases the chance of hitting the MissingSequenceId error for removed members. The TODO acknowledges this. A robust path should skip removal candidates and treat missing latest entries for unchanged members as 'no update' rather than a hard error. <b>[ Low confidence ]</b>
- [line 3072](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mls_sync.rs#L3072): Incorrect error handling of spawned tasks: the code only treats `JoinError` as a failure by filtering `res.iter().filter(|r| r.is_err())`, which captures panics/cancellation, but it ignores inner `Err` values returned by `g.publish_intents(&mut guard)`. Since `JoinSet::join_all()` returns a `Vec<Result<Result<_, _>, JoinError>>`, any `Ok(Err(e))` is considered success by this filter and will not be logged or cause the test to fail. This silently accepts publish failures. Fix by flattening the results or inspecting both layers, e.g., `for r in res { match r { Ok(Ok(_)) => {}, Ok(Err(e)) => log/panic, Err(e) => log/panic } }`, or `let errs: Vec<_> = res.into_iter().map(|r| r.and_then(|inner| inner)).filter(|r| r.is_err()).collect();`. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mod.rs — 0 comments posted, 16 evaluated, 16 filtered</summary>

- [line 347](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L347): The new `load_mls_group` function omits the prior group-level locking (`get_lock_sync`/`get_lock_async`) and the operation-closure execution within the critical section. This removes mutual-exclusion guarantees for group operations, enabling concurrent access/updates without at-most-once or swap-and-clear protections. If callers use this loader during commit/apply paths, races can cause double-application, inconsistent state, or re-entrancy issues. Provide locking at this layer or document and enforce that callers must acquire the `mls_commit_lock` before using the group. <b>[ Low confidence ]</b>
- [line 352](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L352): `OpenMlsGroup::load(storage, &GroupId::from_slice(&self.group_id))` errors are mapped with `map_err(|_| NotFound::MlsGroup)?` and a missing group (`None`) is converted with `.ok_or(NotFound::MlsGroup)?`. This conflates all storage/provider failures (I/O, deserialization, keystore errors) with an actual "group not found" condition. Upstream logic cannot distinguish transient/storage errors from a legitimate missing group, violating error-contract integrity and leading to incorrect handling (e.g., treating a storage outage as a missing group, wrong retries/recovery). Preserve the original error or wrap distinctly (e.g., `StorageError` vs `NotFound`). <b>[ Low confidence ]</b>
- [line 547](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L547): The function acquires a lock (`let mut guard = self.lock().await;`) and then awaits on `ensure_not_paused().await?` without passing the guard. If `ensure_not_paused` attempts to acquire the same lock internally, this can deadlock because the current task is holding the lock across the await. Similarly, later `update_consent_state(ConsentState::Allowed)?` is called while the lock is still held and, if it internally tries to lock `self`, it can also deadlock. To avoid this, either release the lock before awaiting these calls or pass/consume the existing guard consistently in all functions that need access to the locked state. <b>[ Low confidence ]</b>
- [line 556](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L556): The function holds the lock guard across multiple `.await` points (`maybe_update_installations(...).await?` and `sync_until_last_intent_resolved(...).await?`). Even though the guard is passed into these functions, holding an async mutex across awaits can cause deadlocks or starvation if any awaited path internally re-enters the same lock (including callbacks or nested operations) or if long I/O holds the lock, blocking other tasks that need it. Prefer scoping the lock to minimal critical sections or ensuring all awaited code paths using the shared state exclusively consume the passed guard and never attempt to re-lock. <b>[ Low confidence ]</b>
- [line 579](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L579): Lock held while calling `ensure_not_paused` without passing the guard: `let mut guard = self.lock().await;` is acquired and then `self.ensure_not_paused().await?;` is awaited. If `ensure_not_paused` attempts to lock the same mutex, this will deadlock. Either ensure `ensure_not_paused` does not lock, call it before acquiring `guard`, or provide a variant that operates with the existing guard. <b>[ Low confidence ]</b>
- [line 581](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L581): Lock held across async awaits may cause deadlocks or liveness issues: `guard` is held while awaiting `self.maybe_update_installations(..., &mut guard).await?;` and `self.sync_until_last_intent_resolved(&mut guard).await?;`. If either path performs I/O or invokes code that attempts to acquire the same lock (directly or indirectly), this can deadlock or starve other tasks needing the lock. Consider limiting the critical section: capture needed state under the lock, release it, perform async work, then re-acquire to commit results; or ensure these callees never attempt to acquire the same lock and do not call out to code that does. <b>[ Low confidence ]</b>
- [line 586](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L586): Potential self-deadlock: The function acquires a lock via `let mut guard = self.lock().await;` and keeps it held through the end of the function, then calls `self.update_consent_state(ConsentState::Allowed)?;` without passing the held `guard`. If `update_consent_state` attempts to acquire the same internal lock (consistent with the pattern that other methods requiring the lock accept `&mut guard`), this will deadlock because the lock is already held by the current task. A safer pattern would be to either pass the guard to `update_consent_state` (if supported) or explicitly drop the guard before calling `update_consent_state`. <b>[ Low confidence ]</b>
- [line 597](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L597): `update_installations` passes `&mut self.lock().await` into `maybe_update_installations(...)` and then awaits the async function. This holds the lock guard across an `await`, which can lead to runtime deadlocks or starvation if `maybe_update_installations` (or any callee) awaits or tries to re-acquire the same lock. Holding an async/sync mutex guard across suspension points is a known anti-pattern and can block other tasks indefinitely or self-deadlock if re-entrant locking occurs. The temporary guard is also created inline, increasing the risk of subtle lifetime/usage mistakes. Refactor to avoid holding the lock across `await`—e.g., acquire the lock into a local variable, perform non-awaiting work under the lock, release it before any awaits, or clone/capture the state needed before awaiting. <b>[ Low confidence ]</b>
- [line 832](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L832): `sync_until_intent_resolved` is now invoked while holding the `mls_commit_lock` guard (e.g., in `groups::mls_sync::sync` lines 320–329 for the current group and lines 310–318 for stitched DMs; and in `groups::mod::add_members_by_inbox_id` lines 832–834; and `groups::mod::remove_members_by_inbox_id` lines 904–906). If the resolution path for the queued intent requires another task/component to acquire the same `mls_commit_lock` (for example, a background processor that applies MLS commits), the caller will be awaiting resolution while monopolizing the lock, potentially causing a self-deadlock or indefinite stall. This is a runtime concurrency bug if any part of intent resolution needs that lock independently of the provided guard. The change introduces a lock being held across multiple `await` points, increasing the risk of deadlock/starvation if the resolution process is external or re-entrant. <b>[ Low confidence ]</b>
- [line 1212](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L1212): The check `if group_name.len() > MAX_GROUP_NAME_LENGTH` measures UTF-8 byte length, not character count. This can erroneously reject names with multibyte characters (e.g., emojis or non-ASCII) even when they have at most `MAX_GROUP_NAME_LENGTH` characters, contradicting the error variant `TooManyCharacters`. Use a character-aware count (e.g., `group_name.chars().count()`) if the limit is intended to be in characters, or rename the error and constant to reflect a byte limit. <b>[ Low confidence ]</b>
- [line 1394](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L1394): `group_description.len()` counts UTF-8 bytes, not characters. If `MAX_GROUP_DESCRIPTION_LENGTH` and `GroupError::TooManyCharacters` are intended to enforce a character limit (as the error name suggests), this check will incorrectly reject or accept inputs containing multi-byte Unicode characters. For example, emojis or accented characters could cause a valid description (by character count) to exceed the byte limit, or permit more characters than intended if the limit was meant in code points. Use a character-aware count (e.g., `group_description.chars().count()`) if the contract is in characters, or rename the error and constant to reflect a byte limit and document it to avoid confusion. <b>[ Low confidence ]</b>
- [line 1519](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L1519): Potential unit/overflow bug: `expire_from_ms` is documented by its name as milliseconds, but it is passed to `UpdateMetadataIntentData::new_update_conversation_message_disappear_from_ns(...)` which (by name) appears to expect nanoseconds. If the callee converts ms→ns (e.g., multiplies by 1_000_000), large millisecond values (such as epoch-based timestamps ~1.7e12 ms) will overflow `i64` when converted to ns (~1.7e21), causing incorrect behavior or panics in downstream code. Negative values would also produce negative ns. Validate input bounds, perform explicit checked conversion, and/or align naming/units to avoid misinterpretation. <b>[ Out of scope ]</b>
- [line 1524](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L1524): Locking order/race window: The intent is queued via `QueueIntent::metadata_update().data(intent_data).queue(self)?` before acquiring the MLS commit lock (`self.mls_commit_lock.get_lock_async(self.group_id.clone()).await`). If queuing and subsequent resolution-sync are meant to be serialized under the same lock to enforce ordering, this introduces a window where other operations may interleave, potentially causing out-of-order state changes or race conditions. Consider acquiring the lock before queuing if queuing must be ordered with commit synchronization, or document why queueing outside the lock is safe. <b>[ Low confidence ]</b>
- [line 1544](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L1544): No validation for negative expiration values: ``expire_in_ms: i64`` is accepted without checks and forwarded into ``UpdateMetadataIntentData::new_update_conversation_message_disappear_in_ns``. If negative durations are invalid for this metadata intent, this can cause unexpected behavior or downstream errors. Add an explicit guard (e.g., reject negatives or map to a defined behavior) before enqueuing the intent. <b>[ Low confidence ]</b>
- [line 1549](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L1549): Unit mismatch: the function and constructor names indicate nanoseconds ("`update_conversation_message_disappear_in_ns`" and ``UpdateMetadataIntentData::new_update_conversation_message_disappear_in_ns``) but the parameter is named ``expire_in_ms`` and is passed through without conversion. This strongly suggests milliseconds are being supplied where nanoseconds are expected, leading to a 1e6 scaling error in the expiration duration. If 1000 (ms) is passed, it would be interpreted as 1000 ns (1 µs), causing messages to disappear almost immediately. Either rename the parameter to reflect nanoseconds or convert ms→ns before constructing the intent. <b>[ Low confidence ]</b>
- [line 1692](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/mod.rs#L1692): Potential deadlock/indefinite wait: the function acquires `mls_commit_lock` for `self.group_id` and then awaits `sync_until_intent_resolved(&mut guard, intent.id)`. If the intent resolution path (the worker that resolves the queued intent or any code it calls) also needs to acquire the same `mls_commit_lock` for the group, it will be blocked by this held guard, while this function is awaiting that resolution, producing a deadlock or indefinite hang. This change moved from waiting without the lock to waiting while holding the lock, increasing the risk. The bug would manifest as the method never returning under certain inputs/workflows where the resolver uses the same lock. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/oneshot.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 49](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/oneshot.rs#L49): Locking asymmetry: for the empty `inbox_ids` path, a commit lock is acquired via `context.mls_commit_lock().get_lock_async(group.group_id.clone()).await` and passed to `group.add_missing_installations(&mut guard).await`. For the non-empty `inbox_ids` path, `group.add_members_by_inbox_id(&inbox_ids).await` is called without holding this lock. If `add_members_by_inbox_id` expects the caller to hold the same commit lock for atomicity/serialization, this can cause races, inconsistent state, or deadlocks when concurrent operations occur on the same group. <b>[ Low confidence ]</b>
- [line 49](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/oneshot.rs#L49): The documentation states that the sender's other installations are implicitly included and will also receive the oneshot message (`"the sender's other installations are implicitly included"`). In code, `add_missing_installations` is only called when `inbox_ids` is empty. When `inbox_ids` is non-empty, only `group.add_members_by_inbox_id(&inbox_ids).await?` is invoked and no explicit call ensures the sender's other installations are added. If `add_members_by_inbox_id` does not itself add the sender's other installations, this path violates the documented behavior. <b>[ Low confidence ]</b>
- [line 60](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/oneshot.rs#L60): Group creation via `MlsGroup::<C>::create_and_insert(...)` persists side effects before member addition. If `add_members_by_inbox_id(...).await` or `add_missing_installations(&mut guard).await` fails, the function returns an error without rolling back or deleting the newly created group, leaving an orphaned/partially initialized group in storage. This violates single paired cleanup and no-leak guarantees after effects. The inline comment `// Optional: delete group from DB here` highlights the missing cleanup logic. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/welcome_sync.rs — 0 comments posted, 5 evaluated, 5 filtered</summary>

- [line 139](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/welcome_sync.rs#L139): `num_eligible_groups` is set to `groups.len()` before filtering via `filter_groups_needing_sync(groups)`. The summary passed to `GroupSyncSummary::new` will report the count of input groups, not the count of groups actually eligible/selected for sync. This can misreport the "eligible" count and violate the function's external contract of summarizing the sync operation. Move the count to after filtering or count the filtered set. <b>[ Out of scope ]</b>
- [line 153](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/welcome_sync.rs#L153): `is_active` is now checked via `group.is_active()?` without the MLS group lock previously ensured by `load_mls_group_with_lock_async`. This introduces a race with concurrent state changes to the group's active status, potentially causing `sync()` to run on an inactive group or to skip syncing an active one. Restore the locked read or otherwise ensure atomicity around the active check and subsequent sync. <b>[ Low confidence ]</b>
- [line 154](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/welcome_sync.rs#L154): Behavioral regression: the previous implementation called `group.sync_with_conn().await?` followed by `group.maybe_update_installations(None).await?`. The new code calls only `group.sync().await?` and omits the installations update. This can break post-sync invariants (e.g., installation metadata not refreshed) and alter external side effects expected by callers. Restore the missing update or document and version the contract change. <b>[ Low confidence ]</b>
- [line 283](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/welcome_sync.rs#L283): Behavioral regression: on an active group, the previous code executed `group.sync_with_conn().await` followed by `group.maybe_update_installations(None).await`, ensuring both the sync and an installations update occurred. The new code only calls `group.sync().await` and omits `maybe_update_installations(None)`. If `sync()` does not internally perform the installations update, this silently drops an externally visible side effect and can leave installations stale after a successful sync. <b>[ Low confidence ]</b>
- [line 283](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/groups/welcome_sync.rs#L283): Contract change: `group.sync_with_conn().await` was replaced by `group.sync().await`. If `sync()` does not use the same underlying connection or equivalent resources, this can lead to incomplete synchronization or different side effects compared to the previous implementation. This is an externally visible behavior change on the success path. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/subscriptions/process_message/factory.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 112](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/subscriptions/process_message/factory.rs#L112): `group.lock().await` acquires a mutex guard and that `guard` is then held across the awaited call `group.process_message(&mut guard, msg, false).await`. Holding an async mutex guard across awaits can deadlock if any awaited sub-steps try to re-lock the same mutex (directly or indirectly), or can cause prolonged lock contention/starvation for other tasks needing the same lock. This change makes the lock duration depend on the entire processing pipeline inside `process_message`, which is opaque here and could include further awaits and callbacks. If `process_message` (or anything it awaits) attempts `group.lock()` again, the task will deadlock because the lock is already held. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/test/client_test_utils.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 90](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/test/client_test_utils.rs#L90): Using `assert_eq!(sync_group.group_id, other_sync_group.group_id)` at the end of the retry loop can panic at runtime if the IDs never converge within the loop. The function’s signature returns `Result<(), GroupError>`, but this path will terminate the task with a panic instead of returning an error, violating the Result-based contract and causing uncontrolled test failure. Prefer returning a structured error after the bounded retries instead of panicking. <b>[ Low confidence ]</b>
- [line 94](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/test/client_test_utils.rs#L94): `assert_eq!(epoch, other_epoch)` can panic at runtime if epochs differ after synchronization. The function returns `Result<(), GroupError>`, so panicking bypasses error propagation and violates the Result-based contract. Return a descriptive error instead of panicking. <b>[ Low confidence ]</b>
- [line 102](https://github.com/xmtp/libxmtp/blob/4542761cfd28e906cd0c1f49d0c04d954c0144be/xmtp_mls/src/test/client_test_utils.rs#L102): `assert_eq!(ratchet_tree, other_ratchet_tree)` can panic at runtime if ratchet trees differ. This bypasses the `Result<(), GroupError>` error path and causes uncontrolled task termination. Prefer returning a structured error with details about divergence. <b>[ Low confidence ]</b>
</details>

</details>


neekolas:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2821?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2821** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2821?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2821?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2814** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/2814?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=2821).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #2832, opened 2025-11-21
fixup wasm ci script

insipx:
this fixes ./dev/test/wasm and I am able to use it locally again. I'm hoping it fixes CI for #2807 as well

- updates android platform tools to 35


1
```diff
@@ -0,0 +1 @@
+./dev/test/wasm: line 21: [: syntax error: `-v' unexpected
```
tylerhawkes, line 1:
```suggestion
```

tylerhawkes, line 1: This file needs to be deleted.

xmtp_cryptography/Cargo.toml
```diff
@@ -20,7 +20,7 @@ alloy = { workspace = true, features = [
 ] }
 ed25519-dalek = { workspace = true, features = ["digest"] }
 hex.workspace = true
-libcrux-ed25519 = "0.0.3"
+libcrux-ed25519 = "0.0.4"
```
tylerhawkes, line 23: Can we get the other dependencies updated so we're not building these crypto libraries twice?

insipx: downgraded b/c openmls, should be done separately https://github.com/xmtp/libxmtp/pull/2841

1
insipx: https://github.com/xmtp/libxmtp/pull/2841

insipx: https://github.com/xmtp/libxmtp/pull/2841

PR #2835, opened 2025-11-22
Readonly Mode

codabrink:
### Add readonly client mode and route it through builders, context, and sync paths to make registration and rotations no-ops in xmtp_mls
Introduce `ClientMode::Readonly`, thread it via `ClientBuilder.with_client_mode(...)` into `XmtpMlsLocalContext`, and gate registration, key rotation, intents, and post-commit syncing behind `readonly_mode()`. Update FFI/Node/WASM `create_client` to accept optional mode and map to core `ClientMode`. Replace test API selection with `TesterApiMode` and add targeted readonly tests.

#### 📍Where to Start
Start with `ClientMode` and context wiring in [context.rs](https://github.com/xmtp/libxmtp/pull/2835/files#diff-dec1fc3de5dd13f41fcf44c4d5b4d1dbb7f4fb2652ead1eda88879ec391f454d), then follow builder propagation in [builder.rs](https://github.com/xmtp/libxmtp/pull/2835/files#diff-2df8df9acb5405ca5c2dde2487e92ffc9a7daf6be38dd075a2a2d4878cc35ce6) and readonly guards in client and group sync paths in [client.rs](https://github.com/xmtp/libxmtp/pull/2835/files#diff-3c0189371525b6aae3be7e5a7b5b5682204ef341fb2744fc876148c5ecab8642) and [groups/mls_sync.rs](https://github.com/xmtp/libxmtp/pull/2835/files#diff-10f818f7918c8c0b2ed32c4f63e5060050527ed58df00ec081bbecb66065144e).

#### Changes since #2835 opened

- Renamed client mode terminology from 'Notification' to 'Readonly' across FFI bindings and error handling [b31d94f]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized b31d94f. 15 files reviewed, 26 issues evaluated, 26 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>bindings_ffi/benches/create_client.rs — 0 comments posted, 7 evaluated, 7 filtered</summary>

- [line 47](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_ffi/benches/create_client.rs#L47): `unwrap()` on `ident.inbox_id(nonce)` can panic at runtime if `Identifier::inbox_id` returns an `Err` (e.g., if the generated Ethereum address is deemed invalid by `is_valid_address`). Replace with proper error handling and surfacing of the failure. <b>[ Out of scope ]</b>
- [line 48](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_ffi/benches/create_client.rs#L48): Calling `tmp_path()` (used here) can panic due to `std::env::temp_dir().to_str().unwrap()` if the temp directory path is not valid UTF-8 on the host system. Use a fallible conversion or lossless path handling (e.g., `PathBuf` or `to_string_lossy`). <b>[ Out of scope ]</b>
- [line 82](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_ffi/benches/create_client.rs#L82): `unwrap()` on the result of `xmtpv3::mls::create_client(...).await` can panic if client creation fails (e.g., network/backend issues). Prefer handling the error and reporting it within the benchmark. <b>[ Out of scope ]</b>
- [line 101](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_ffi/benches/create_client.rs#L101): `unwrap()` on `ident.inbox_id(nonce)` can panic at runtime if identifier validation fails. Replace with fallible handling to keep the benchmark from aborting. <b>[ Out of scope ]</b>
- [line 104](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_ffi/benches/create_client.rs#L104): Calling `tmp_path()` here can panic because it uses `to_str().unwrap()` on the system temp directory path, which may not be UTF-8. Use a fallible or lossy conversion, or accept a `PathBuf`. <b>[ Out of scope ]</b>
- [line 124](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_ffi/benches/create_client.rs#L124): `unwrap()` on `xmtpv3::mls::create_client(...).await` inside the warm-up `block_on` can panic if client creation fails (network/backends). Handle the error instead of aborting the benchmark process. <b>[ Out of scope ]</b>
- [line 163](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_ffi/benches/create_client.rs#L163): `unwrap()` on `xmtpv3::mls::create_client(...).await` in the benchmark iteration can panic on backend/client creation failures. Prefer propagating or recording the error instead of panicking. <b>[ Out of scope ]</b>
</details>

<details>
<summary>bindings_wasm/src/client.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 80](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_wasm/src/client.rs#L80): Removing the `Notification` variant from the `#[wasm_bindgen]`-exposed enum `ClientMode` is a breaking change for existing JavaScript callers. Any previously valid JS code passing `ClientMode.Notification` (or the corresponding discriminant value) will now be rejected at the wasm boundary, causing a runtime error. `wasm_bindgen` enforces enum value validity and will throw if an unknown discriminant is provided, so the removal can lead to immediate failures when `createClient` is invoked with `client_mode` set to the former `Notification` value. This violates contract parity for the public API and can produce runtime exceptions in production. <b>[ Low confidence ]</b>
- [line 226](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/bindings_wasm/src/client.rs#L226): In `create_client`, when an `encryption_key` is provided, the code validates it (`let _key: EncryptionKey = key.try_into()?`) but then discards it and constructs the database with `WasmDb::new(&storage_option)` without passing or applying the encryption key. As a result, the encryption key is effectively ignored and the store may be created unencrypted even though the caller supplied a key. The error messages suggest an encrypted store is intended, but no encryption configuration is applied to the DB. The key should be used to initialize the database or the `EncryptedMessageStore` so that data is actually encrypted. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/builder.rs — 0 comments posted, 8 evaluated, 8 filtered</summary>

- [line 176](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/builder.rs#L176): Contract parity break: `identity_strategy` is forced to `IdentityStrategy::CachedOnly` rather than preserving any prior strategy. If the original setup expected `CreateIfNotFound`, rebuilding may fail to create missing identities. <b>[ Out of scope ]</b>
- [line 182](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/builder.rs#L182): Contract parity break: `allow_offline` is hardcoded to `false` rather than reflecting the existing `Client` configuration. Rebuilt clients may incorrectly disallow offline mode. <b>[ Out of scope ]</b>
- [line 184](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/builder.rs#L184): Contract parity break: `disable_events` is set based on `cfg(test)` rather than preserved from the source `Client`. This can flip event behavior unexpectedly when rebuilding. <b>[ Out of scope ]</b>
- [line 187](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/builder.rs#L187): Contract parity break: `disable_commit_log_worker` is hardcoded to `false` rather than reflecting the original client's configuration, potentially enabling a worker unexpectedly. <b>[ Out of scope ]</b>
- [line 190](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/builder.rs#L190): Contract parity break: `cursor_store` is set to `None` instead of preserving any existing cursor store from the source `Client`. This can disable or reset cursor-based behavior after rebuild. <b>[ Out of scope ]</b>
- [line 192](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/builder.rs#L192): Contract parity break: `client_mode` is always set to `None` rather than preserving the `Client`'s existing mode. This can change runtime behavior when rebuilding from an existing client. <b>[ Low confidence ]</b>
- [line 333](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/builder.rs#L333): Global events flag `EVENTS_ENABLED` is only ever set to `true` during build when `disable_events` is `false`, but it is never set back to `false` when `disable_events` is `true`. This leaks global state across client instances: once a prior client enables events, constructing a later client with `disable_events = true` will not actually disable event tracking, violating the caller’s expectation and the macro’s contract for being configurable per client. This is a persistent mutation of global state without scoped restoration, which can cause unexpected event writes and logs. Consider explicitly setting `EVENTS_ENABLED.store(false, Ordering::SeqCst)` when `disable_events` is `true`, or scoping the flag per-client instead of global. <b>[ Out of scope ]</b>
- [line 679](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/builder.rs#L679): `enable_api_stats` returns `ClientBuilderError::MissingParameter { parameter: "api_client" }` when either `self.api_client` OR `self.sync_api_client` is `None`. This produces a misleading error message when `sync_api_client` is missing, obscuring the actual missing parameter and hindering correct recovery. The guard should differentiate which parameter is missing or report both. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/client.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 339](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/client.rs#L339): `find_inbox_ids_from_identifiers` silently returns `None` for duplicate identifiers in the input slice because it uses `HashMap::remove` for both `cached_inbox_ids` and `new_inbox_ids`. On the first occurrence the entry is removed and returned; on subsequent occurrences of the same identifier, `remove` returns `None`, causing data loss for valid duplicates. Use `get` (or `get_mut` without removal) instead to preserve repeated lookups. Problematic lines are where `cached_inbox_ids.remove(&cache_key)` and `new_inbox_ids.remove(&ident.into())` are used during the final collection. <b>[ Out of scope ]</b>
- [line 748](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/client.rs#L748): In `delete_message`, events are sent on `self.context.local_events()` whereas other methods in the same impl (e.g., `create_group`, `create_dm_by_inbox_id`) send on `self.local_events`. If these senders are not the exact same channel, deletion notifications may not reach subscribers that listen to the client’s `local_events`, causing inconsistent externally visible behavior. Use a single, consistent sender (`self.local_events`) for client-originated local events, or ensure both are the same and document it. <b>[ Out of scope ]</b>
- [line 849](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/client.rs#L849): `register_identity()` stores a locally generated key package before verifying the provided `SignatureRequest`, and on verification failure it returns early without cleaning up the newly stored package. Specifically, `generate_and_store_key_package(...)` persists state, then `build_identity_update().to_verified(...).await?` may error, but there is no rollback or deletion of the stored key package in this error path. This leaves an unused key package artifact in local storage, violating the single paired cleanup invariant for effects introduced before verification succeeds. <b>[ Low confidence ]</b>
- [line 868](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/client.rs#L868): `register_identity()` performs network-side `upload_key_package(...)` (Step 3) before `publish_identity_update(...)` (Step 4). If upload succeeds but publish fails, the function returns an error without compensating actions. This leaves an orphaned key package on the network and an inconsistent local state (no identity stored/ready), violating the effect sequencing/cleanup invariant for partial failure. There is no retry of publish in-place nor rollback of the uploaded package. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/intents/queue.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 56](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/groups/intents/queue.rs#L56): `queue_for_each` assumes all `MlsGroup<C>` entries share the same `context`, but it only takes the `context` from an arbitrary first element and uses that single storage connection to queue intents for every group. If `groups` contains entries from different contexts/databases, intents for non-first-context groups will be queued against the wrong store, causing misrouting or partial failure without surfacing an error. There is no validation that all groups have the same `context` nor partitioning by context. Fix by either (a) validating all groups share the same `context` and returning an error otherwise, or (b) partitioning groups by `context` and performing a separate transaction per context. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_mls/src/groups/mls_sync.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 405](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/groups/mls_sync.rs#L405): `sync_with_conn` returns `Ok(summary)` when only one of `publish_errors` or `post_commit_errors` is non-empty because `SyncSummary::is_errored()` is defined as `self.other.is_some() || (!self.publish_errors.is_empty() && !self.post_commit_errors.is_empty())`. This AND condition means a single class of errors (publish-only or post-commit-only) will not trigger an error result, causing callers to treat a partially failed sync as success. This can suppress retries and misreport sync status. The check at lines `405..409` relies on this method, so the function can incorrectly return success despite errors. <b>[ Out of scope ]</b>
- [line 2432](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/groups/mls_sync.rs#L2432): `elapsed_ns` is computed using `now_ns` (line 2432) which casts `as_nanos()` (`u128`) to `i64`. In `common/src/time.rs`, `now_ns()` does `duration_since_epoch().as_nanos() as i64`, which will truncate and wrap once the epoch duration exceeds `i64::MAX` nanoseconds (circa year 2262). This can produce negative/incorrect `now_ns`, causing `elapsed_ns` to be wrong and suppressing or triggering updates incorrectly. Use a wider type (e.g., `i128` or `u128`) for time arithmetic, or clamp/handle overflow explicitly before comparisons. <b>[ Out of scope ]</b>
- [line 2435](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/groups/mls_sync.rs#L2435): `update_interval_ns` is accepted as `Option<i64>` without validation. If a caller supplies a negative interval, the comparison `elapsed_ns > interval_ns` on line 2435 will be true for most realistic `elapsed_ns`, causing near-always updates and potentially excessive DB writes. Reject negatives or clamp to `0` before use: e.g., `let interval_ns = update_interval_ns.map_or(SYNC_UPDATE_INSTALLATIONS_INTERVAL_NS, |v| v.max(0));`. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_mls/src/identity.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 622](https://github.com/xmtp/libxmtp/blob/b31d94f26e570595f2f3c59927d66090015f7cb2/xmtp_mls/src/identity.rs#L622): `Identity::queue_key_rotation` claims in its docstring "If no key rotation is scheduled, queue it to occur in the next 5 seconds.", but the implementation unconditionally calls `conn.queue_key_package_rotation()?` without checking whether a rotation is already scheduled. This can result in duplicate scheduling if `conn.queue_key_package_rotation` is not idempotent, and the log message "Last key package not ready for rotation, queued for rotation" may be misleading. Either add a guard to check for an existing scheduled rotation before enqueuing, or update the documentation to match behavior. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -482,7 +482,7 @@ impl FfiSignatureRequest {
 pub enum FfiClientMode {
     #[default]
     Default,
-    Notification,
+    Readonly,
 }
 
 #[derive(uniffi::Object)]
@@ -497,7 +497,7 @@ impl From<FfiClientMode> for ClientMode {
     fn from(mode: FfiClientMode) -> Self {
         match mode {
             FfiClientMode::Default => Self::Default,
-            FfiClientMode::Notification => Self::Readonly,
+            FfiClientMode::Readonly => Self::Readonly,
         }
     }
 }
--- a/xmtp_mls/src/identity.rs
+++ b/xmtp_mls/src/identity.rs
@@ -258,8 +258,8 @@ pub enum IdentityError {
     Bincode,
     #[error(transparent)]
     UninitializedField(#[from] derive_builder::UninitializedFieldError),
-    #[error("Unable to register in notification client")]
-    NotificationClientRegister,
+    #[error("Unable to register in readonly client")]
+    ReadonlyClientRegister,
 }
 
 impl NeedsDbReconnect for IdentityError {


bindings_ffi/src/mls.rs
```diff
@@ -494,6 +493,14 @@ pub struct FfiXmtpClient {
     #[allow(dead_code)]
     account_identifier: FfiIdentifier,
 }
+impl From<FfiClientMode> for ClientMode {
+    fn from(mode: FfiClientMode) -> Self {
+        match mode {
+            FfiClientMode::Default => Self::Default,
+            FfiClientMode::Notification => Self::Readonly,
```
neekolas: This enum needs a rename too

xmtp_mls/src/identity.rs
```diff
@@ -258,6 +258,8 @@ pub enum IdentityError {
     Bincode,
     #[error(transparent)]
     UninitializedField(#[from] derive_builder::UninitializedFieldError),
+    #[error("Unable to register in notification client")]
+    NotificationClientRegister,
```
neekolas: Needs a rename too

xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -455,6 +454,12 @@ where
         intent_id: ID,
     ) -> Result<SyncSummary, GroupError> {
         let mut summary = SyncSummary::default();
+
+        if self.context.readonly_mode() {
```
neekolas, line 458: I'd love to understand a little more about what the callers should do in this situation. We don't really know if the the intent has succeeded or failed. I'd imagine in read only mode we aren't expecting a lot of calls to this (since it's typically triggered by writes), but I want to make sure misuse doesn't lead to someone thinking a thing worked when it didn't or leading to an infinite loop.

PR #2841, opened 2025-11-24
remove unnecessary file, downgrade libcrux-ed25519, use nix for wasm tests

insipx:
downgrading libcrux back to `0.0.3` b/c unifying under 0.0.4 requires updating [`openmls`](https://github.com/xmtp/openmls) which should undergo some more scrutiny & care

this pr fixes the workflow again, which failed silently in the last PR

switches to using nix for webassembly tests, which makes it much easier to test the workflow locally rather than in GH actions. it also makes it easier to deterministically choose the right version of chromium/chromedriver -- the version from `warpbuild` could have been causing builds to fail

### DIFF
--- a/.github/workflows/test-webassembly.yml
+++ b/.github/workflows/test-webassembly.yml
@@ -21,16 +21,13 @@ on:
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
-  CARGO_PROFILE_TEST_DEBUG: 0
-  WASM_BINDGEN_TEST_TIMEOUT: 256
-  WASM_BINDGEN_TEST_ONLY_WEB: 1
-  WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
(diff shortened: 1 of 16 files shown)


.github/workflows/test-webassembly.yml
```diff
@@ -26,7 +26,6 @@ env:
   WASM_BINDGEN_TEST_ONLY_WEB: 1
   WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
   RSTEST_TIMEOUT: 90
-  GECKODRIVER: "${{ env.GECKOWEBDRIVER }}"
```
insipx, line 29: one of the annoying things about actions is that a "lint error" will silently fail making the rest of the workflow green, which is why the last #2832 pr got merged. Apparently env can't be used here b/c its a global which creates a circular reference. bindgen needs the driver to be under `GECKODRIVER` and not geckowebdriver like the warp builder provides

PR #2866, opened 2025-12-01
Sync additional metadata during device sync

rygine:
### Persist message disappearing settings and app_data during Element::Group import in `xmtp_mls::groups::device_sync::archive::insert`
Add `MessageDisappearingSettings` handling and pass computed values into `DMMetadataOptions` and `GroupMetadataOptions`, and restore `app_data` from backup attributes in [archive.rs](https://github.com/xmtp/libxmtp/pull/2866/files#diff-331bdc3141c56959e610be9d90604dac573da97ae895bb92577c334a938eb4dc).

#### 📍Where to Start
Start with the `insert` function in [archive.rs](https://github.com/xmtp/libxmtp/pull/2866/files#diff-331bdc3141c56959e610be9d90604dac573da97ae895bb92577c334a938eb4dc), focusing on computation of `message_disappearing_settings` and construction of `DMMetadataOptions` and `GroupMetadataOptions`.

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 94cefb3.

### DIFF
--- a/xmtp_mls/src/groups/device_sync/archive.rs
+++ b/xmtp_mls/src/groups/device_sync/archive.rs
@@ -13,6 +13,7 @@ use xmtp_db::{
     prelude::*,
 };
 use xmtp_mls_common::group::{DMMetadataOptions, GroupMetadataOptions};
+use xmtp_mls_common::group_mutable_metadata::MessageDisappearingSettings;
 use xmtp_proto::xmtp::device_sync::{BackupElement, backup_element::Element};
 
 pub async fn insert_importer(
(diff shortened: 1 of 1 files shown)


xmtp_mls/src/groups/device_sync/archive.rs
```diff
@@ -51,6 +52,14 @@ fn insert(element: BackupElement, context: &impl XmtpSharedContext) -> Result<()
                 .map(|m| m.attributes)
                 .unwrap_or_default();
 
+            let message_disappearing_settings =
+                match (save.message_disappear_from_ns, save.message_disappear_in_ns) {
+                    (Some(from_ns), Some(in_ns)) => {
```
codabrink, line 57: This is just a naive question on my part. Do we want to match them separately in case one of them is `None` and set a default for the other, or is message disappearing an all-or-nothing kind of feature?

rygine: AFAIK, it's all or nothing given the `MessageDisappearingSettings` struct. not sure what the default would be for either value. seems like both should be explicitly set for this type of feature.

PR #2867, opened 2025-12-01
Archive Group Timestamp Behavior Tweak

codabrink:
### Update archive import to set `groups.last_message_ns` from snapshot timestamps and adjust tests to use `NS_IN_MIN` for timestamp math
Add `ImportContext` to collect per-group timestamps during archive import and persist them via a post-import DB update; modify `insert_importer` and `insert` to populate this context for both existing and new groups; update tests to verify `groups.last_message_ns` after import and replace hardcoded minute nanoseconds with `NS_IN_MIN`.

#### 📍Where to Start
Start with `insert_importer` and `insert` in [archive.rs](https://github.com/xmtp/libxmtp/pull/2867/files#diff-331bdc3141c56959e610be9d90604dac573da97ae895bb92577c334a938eb4dc), then review `ImportContext::post_import` in the same file.

#### Changes since #2867 opened

- Modified group timestamp computation logic in the archive module's insert function [4210dbe]
- Removed redundant timestamp maximization operation in `archive.insert` function [b56dfae]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 5693150. 1 file reviewed, 1 issue evaluated, 1 issue filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_mls/src/groups/device_sync/archive.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 82](https://github.com/xmtp/libxmtp/blob/5693150376650541df5cb140b2dd770b3821ddcd/xmtp_mls/src/groups/device_sync/archive.rs#L82): Incorrect timestamp merge when restoring existing groups: when `save.last_message_ns` is `Some(snap_timestamp)` and the existing group’s `last_message_ns` is `None`, the code sets `timestamp` to `None` via `timestamp.map(|ts| ts.max(snap_timestamp))`, silently discarding the snapshot’s timestamp. This prevents `post_import` from updating the database to the snapshot value and leaves `last_message_ns` as `NULL`. The intended behavior appears to be to take the max of the existing and snapshot timestamps, falling back to the snapshot if the existing is `None`. A correct implementation would handle both `Option` values, e.g. `timestamp = match (existing_group.last_message_ns, save.last_message_ns) { (Some(e), Some(s)) => Some(e.max(s)), (None, Some(s)) => Some(s), (Some(e), None) => Some(e), (None, None) => None }`. <b>[ Already posted ]</b>
</details>

</details>

### DIFF
--- a/xmtp_mls/src/groups/device_sync/archive.rs
+++ b/xmtp_mls/src/groups/device_sync/archive.rs
@@ -84,9 +84,6 @@ fn insert(
                     (Some(e), None) => Some(e),
                     (None, None) => None,
                 };
-                if let Some(snap_timestamp) = save.last_message_ns {
-                    timestamp = timestamp.map(|ts| ts.max(snap_timestamp));
-                }
 
                 import_context
                     .group_timestamps


xmtp_mls/src/groups/device_sync/archive.rs
```diff
@@ -41,7 +77,10 @@ fn insert(element: BackupElement, context: &impl XmtpSharedContext) -> Result<()
             context.db().insert_newer_consent_record(consent)?;
         }
         Element::Group(save) => {
-            if let Ok(Some(_)) = context.db().find_group(&save.id) {
+            if let Ok(Some(existing_group)) = context.db().find_group(&save.id) {
+                import_context
+                    .group_timestamps
+                    .insert(existing_group.id, existing_group.last_message_ns);
```
neekolas: These are always run in a fully sorted order, so we don't need to check if the timestamp is > the saved value?

codabrink: Yeah that is a weird edge case I didn't account for. Where a group exists but is out of date enough that an archive from another phone is newer. I'll add a check for it.

PR #2873, opened 2025-12-02
expose conversation membership state

mchenani:
### Expose conversation membership state by adding `FfiConversation.membership_state` and platform bindings in [mls.rs](https://github.com/xmtp/libxmtp/pull/2873/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f), [conversation.rs](https://github.com/xmtp/libxmtp/pull/2873/files#diff-d87172f473cc087eb00db1b730396f12f232ebf80aa7cb1b60ed661a8b117886), and [conversation.rs](https://github.com/xmtp/libxmtp/pull/2873/files#diff-106fd4741a71a89abefb1837cde5785b14100381cea3bf577416671c49bae37e)
Add `FfiGroupMembershipState` and map to/from internal `GroupMembershipState`; implement `MlsGroup.membership_state` to read stored state; expose `membership_state`/`membershipState` on conversation bindings for FFI, Node, and WASM; add tests asserting creator is Allowed and invited user is Pending.

#### 📍Where to Start
Start with `MlsGroup.membership_state` in [mod.rs](https://github.com/xmtp/libxmtp/pull/2873/files#diff-c29f56a38916c7410eff8091df1a2e43487ffe20646d96827e846e475f4608d3), then follow the mapping to `FfiConversation.membership_state` in [mls.rs](https://github.com/xmtp/libxmtp/pull/2873/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f) and the Node/WASM bindings in [conversation.rs](https://github.com/xmtp/libxmtp/pull/2873/files#diff-d87172f473cc087eb00db1b730396f12f232ebf80aa7cb1b60ed661a8b117886) and [conversation.rs](https://github.com/xmtp/libxmtp/pull/2873/files#diff-106fd4741a71a89abefb1837cde5785b14100381cea3bf577416671c49bae37e).

#### Changes since #2873 opened

- Added membership state transition assertions to existing self-removal tests [9089b94]
- Created new test module and test for self-removal with pending state validation [9089b94]

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 8f52dff.

### DIFF
--- a/bindings_ffi/src/mls/tests/mod.rs
+++ b/bindings_ffi/src/mls/tests/mod.rs
@@ -94,6 +94,7 @@ mod identity;
 mod networking;
 mod static_methods;
 mod streaming;
+mod test_self_removal;
 
 // Constants
 pub(crate) const HISTORY_SYNC_URL: &str = "http://localhost:5558";
--- a/bindings_ffi/src/mls/tests/test_self_removal.rs
+++ b/bindings_ffi/src/mls/tests/test_self_removal.rs
@@ -1,0 +1,59 @@
+//! Test for self-removal and PendingRemove membership state
+
+use super::*;
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_self_removal_with_pending_state() {
+    let alix = new_test_client().await;
+    let bo = new_test_client().await;
+
+    // Alix creates a group and adds Bo
+    let alix_group = alix
+        .conversations()
+        .create_group(
+            vec![bo.account_identifier.clone()],
+            FfiCreateGroupOptions::default(),
+        )
+        .await
+        .unwrap();
+
+    // Bo syncs and gets the group
+    bo.conversations().sync().await.unwrap();
+    let bo_group = bo.conversation(alix_group.id()).unwrap();
+
+    // Verify Bo's membership state is Pending when first invited
+    let bo_state_initial = bo_group.membership_state().unwrap();
+    assert_eq!(bo_state_initial, FfiGroupMembershipState::Pending);
+
+    // Verify Alix's membership state is Allowed (creator)
+    let alix_state = alix_group.membership_state().unwrap();
+    assert_eq!(alix_state, FfiGroupMembershipState::Allowed);
+
+    // Bo leaves the group
+    bo_group.leave_group().await.unwrap();
+
+    // Verify Bo's membership state is PendingRemove after requesting to leave
+    let bo_state_after_leave = bo_group.membership_state().unwrap();
+    assert_eq!(bo_state_after_leave, FfiGroupMembershipState::PendingRemove);
+
+    // Alix syncs to process the leave request
+    alix_group.sync().await.unwrap();
+
+    // Wait for admin worker to process the removal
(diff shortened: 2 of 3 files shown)


bindings_ffi/src/mls/tests/group_management.rs
```diff
@@ -1813,3 +1813,32 @@ async fn test_pagination_of_conversations_list() {
...
+    assert_eq!(state, FfiGroupMembershipState::Allowed);
+
+    // Sync so bola receives the group
+    bola.conversations().sync().await.unwrap();
+    let bola_group = bola.conversation(group.id()).unwrap();
+
+    // Bola should have Pending membership state when first receiving the welcome
+    // (invited members start as Pending until explicitly accepted)
+    let bola_state = bola_group.membership_state().unwrap();
+    assert_eq!(bola_state, FfiGroupMembershipState::Pending);
```
cameronvoell, line 1843: I'm not sure the "Pending" membership state is actually used by the protocol anywhere. I think we use ConsentState for managing if a user has allowed a group. I'm not actually sure what Pending is supposed to signify here. (they are in the group, not sure if this property ever updates)

If this test is demonstrating Pending state it should show how a group member leaves pending state. (again I don't think this code is active anywhere in libxmtp, but maybe I'm wrong?)

I think the only property of MembershipState that is useful at the moment is whether the membership state is PendingRemoval or not. 

Maybe instead of the test here showing Pending (which I think is somewhat broken because I'm not sure we ever leave Pending state), we can write  a test that shows how we can use MembershipState to test whether a user is a regular member, whether they are pending removal, and then how we should tell if the removal has been successful. 

wdyt @mchenani ?

cameronvoell (changes requested):

bindings_ffi/src/mls/tests/group_management.rs
mchenani: This was just a test to check the added properties. Will add a test to cover pending removal status as close to the real world scenario.

mchenani: Also just realized that we check the state in the self removal tests, but not via this specific function, we verify the state by directly getting it from the db, but still I'm updating the tests to instead of getting the value from the db, getting the value from the group itself.

mchenani: updated the self-removal tests to check the value from the db and the exposed function, but kept this test just for a general validation for the exposed function.

cameronvoell (approved):
test updates look good!

I think the current setup is fine for getting the needed functionality for detecting pending removals to be available. 👍  

I think for upcoming sprints however, we should consider how we might clean these functions up to be less confusing. 

For example, if GroupMembershipState is not actually used anywhere other then for pending removals, maybe we should deprecate that property on groups and instead track pending removal state on its own without the confusing Allowed and Pending options alongside it. 

For now, this should be good though 👌

mchenani:
Agree, but the usecase for pending and allowed would also give the devs an option to manage the conversations based on the consent state from the membership state itself instead of relying on the consent states. More freedom for the devs, but I agree that could make more confusion for them too.

PR #2876, opened 2025-12-02
utilize the more formal d14n types in mls streams

insipx:
the proto types have some extra safety rules around ordering sequence ids to apply either max or least. it also removes some code in the stream that was essentially the new proto type but worse/duplicated

adds some ergonomic helpers/implementations to `TopicCursor` like `FromIterator`/`IntoIterator` and a proxy entry api based on topic

PR #2877, opened 2025-12-03
use vector clock logic in the stream for has seen ids

insipx:
this should fix an issue where the stream would always do a sync on a commit message, since the EntityId combo with a Application message and originator of 0 would always return 0. this fixes the stream processing code to be more inline with other d14n code that uses cursors, namely using the Vector Clock `GlobalCursor` type


insipx:

1. `trunk()`
1. **push-vvpvpvktowyy ← this PR**
1. [push-yylolrwovkzt](https://github.com/xmtp/libxmtp/pull/2878)
1. [push-looyptkwtwwu](https://github.com/xmtp/libxmtp/pull/2879)

---

codabrink (approved):

PR #2878, opened 2025-12-03
LCC & GCC on `TopicCursor`, move `VectorClock` into `xmtp_proto`

insipx:
- moves the `VectorClock` trait & implementation into `xmtp_proto`. debating whether the trait should be removed altogether and methods simply moved onto `GlobalCursor`, since it doesn't seem like we want multiple `VectorClock` types anyway. I mostly like the trait for the separation between more normal `HashMap` operations and formal `VectorClock` operations.
- added an `lcc` and `gcc` function to `TopicCursor` which computes lcc & gcc with a iterative fold operation on all VectorClocks in `TopicCursor`.


insipx:

1. `trunk()`
1. [push-vvpvpvktowyy](https://github.com/xmtp/libxmtp/pull/2877)
1. **push-yylolrwovkzt ← this PR**
1. [push-looyptkwtwwu](https://github.com/xmtp/libxmtp/pull/2879)

---

PR #2879, opened 2025-12-03
use `SmallVec` for `Topic` type

insipx:
smallvec backs a vec first with a fixed-size `Copy`-able Array (in  this case `[u8;33]`) and if it grows larger "spills" the contents onto the heap.

`Topics` should not grow beyond 33 bytes, since the largest topic identifier is an `InstallationId` at 32 bytes, so this should reduce cloning wherever `Topic`s are used

moves Ids into their own files (`GroupId` and `InstallationId`)


insipx:

1. `trunk()`
1. [push-vvpvpvktowyy](https://github.com/xmtp/libxmtp/pull/2877)
1. [push-yylolrwovkzt](https://github.com/xmtp/libxmtp/pull/2878)
1. **push-looyptkwtwwu ← this PR**

---

xmtp_proto/src/types/topic.rs
```diff
@@ -3,11 +3,13 @@ use std::{
...
-    types::{GroupId, InstallationId},
-    xmtp::xmtpv4::envelopes::AuthenticatedData,
-};
+use smallvec::SmallVec;
+
+use crate::{ConversionError, types::InstallationId, xmtp::xmtpv4::envelopes::AuthenticatedData};
+
+/// the max size of an item in a [`TopicKind`] is 32 bytes (installation id).
+/// the 1st byte is interpreted as the prefixed [`TopicKind`] byte.
+type InnerTopicBytes = SmallVec<[u8; 33]>;
```
tylerhawkes: Do we need the prefix byte in memory here? This has the unfortunate effect of making this 8 bytes bigger on the stack.

tylerhawkes: I doesn't really matter if these aren't long lived though.

xmtp_proto/src/types/topic.rs
```diff
@@ -106,14 +117,20 @@ impl Topic {
...
+    /// get the full topic bytes, including the identifying [`TopicKind`]
+    pub fn cloned_bytes(&self) -> Vec<u8> {
+        self.inner.clone().to_vec()
     }
 
+    /// consume this [`Topic`] into its bytes as a Vec
     pub fn to_bytes(self) -> Vec<u8> {
-        self.inner
+        self.inner.to_vec()
     }
```
tylerhawkes: Why not return the `SmallVec` here? It's got almost the same api and can be convert to a vec if really needed.

Cargo.toml
```diff
@@ -151,6 +151,7 @@ wasm-bindgen-test = "0.3.51"
 web-sys = "0.3"
 web-time = "1.1"
 zeroize = "1.8"
+smallvec = "1.15"
```
tylerhawkes: I'm pretty sure we want the `union` feature to save 8 bytes on each of these.

xmtp_proto/src/types/topic.rs
insipx: we can store the prefix byte separately but it would require allocation when we want to turn the Topic into a contiguous byte array, for instance when using it in the d14n api queries. not sure which is better: accepting the 8 bytes or allocating on `to_bytes`

tylerhawkes: Definitely better to keep it then so there's only one place to manage construction and we anything that returns `&[u8]`is free.

xmtp_proto/src/types/topic.rs
insipx: I changed `to_bytes` to use SmallVec, but `cloned_bytes` is nice to keep as a vec to use in the api query chains like `topics: topics.iter().map(Topic::cloned_vec).collect()` rather than mapping and then mapping again to a `Vec`

xmtp_proto/Cargo.toml
```diff
@@ -22,6 +22,7 @@ pin-project-lite.workspace = true
 prost = { workspace = true, features = ["derive"] }
 prost-types = "0.14"
 serde = { workspace = true }
+smallvec = { workspace = true, features = ["union"] }
```
tylerhawkes: Where possible, we should unify features in the workspace.

PR #2869, opened 2025-12-01
add child resolution & ice to `CursorStore` trait

insipx:
### Add child resolution and ICE handling by extending `CursorStore` with `ice` and `resolve_children` across in-memory, SQL, and wrapper implementations
Extend `CursorStore` with `ice` and `resolve_children`, add an in-memory icebox in `InMemoryCursorStore`, delegate persistence/lookup in the SQLite store, and switch `GlobalCursor` to `BTreeMap` with merge via `apply`. Update callers to construct and return `GlobalCursor` directly.

#### 📍Where to Start
Start with the trait changes in [cursor_store.rs](https://github.com/xmtp/libxmtp/pull/2869/files#diff-028daad07b57a99602b70e297432f143b84633ff46fa3f2b9851f6b476260eae), then review the in-memory implementation in [in_memory_cursor_store.rs](https://github.com/xmtp/libxmtp/pull/2869/files#diff-81e05c4cbc4c5877da996a1302449b85dde8e85a96338b80953cbc520363c6c2) and the SQLite store in [cursor_store.rs](https://github.com/xmtp/libxmtp/pull/2869/files#diff-92939812065b708096075a08e83053a96bbb0d48d2a13d665eba3336c85740ca).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 0972aaf. 9 files reviewed, 6 issues evaluated, 6 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_db/src/encrypted_store/refresh_state.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 133](https://github.com/xmtp/libxmtp/blob/0972aaf951c6484cd66aee375a5c8819c1079685/xmtp_db/src/encrypted_store/refresh_state.rs#L133): rows_to_global_cursor_map performs unchecked signed-to-unsigned integer casts that can silently wrap negative database values, yielding incorrect OriginatorId and SequenceId. Specifically, `originator_id as u32` and `sequence_id.unwrap_or(0) as u64` will wrap if `originator_id` or `sequence_id` are negative, producing large unsigned values instead of rejecting invalid input. This can corrupt the resulting `GlobalCursor` (e.g., mapping to the wrong originator or an enormous sequence), leading to incorrect cursor computations downstream. Validate non-negativity before casting, or return an error on invalid rows. <b>[ Previously rejected ]</b>
[...]

PR #2868, opened 2025-12-01
`bytes` & `OrphanedEnvelope` extractor, add delegate_visitor macro

insipx:
### Add `BytesExtractor`, `OrphanExtractor`, and `delegate_envelope_visitor!` to enable `Envelope::bytes()` and `Envelope::orphan()` across protocol and collection types
Introduce new extractors and a delegating macro, extend the `Envelope` and `EnvelopeCollection` traits with `bytes()` and `orphans()`, and wire `orphans()` through subscribe types; update DB icebox queries to handle multiple cursors with idempotent inserts.

#### 📍Where to Start
Start with the new trait methods in `Envelope` and `EnvelopeCollection` in [xmtp_api_d14n/src/protocol/traits/envelopes.rs](https://github.com/xmtp/libxmtp/pull/2868/files#diff-974a5b6ad5cd0c15ff488b73bf921d7c4b255f5ff3d4f053ab99f1a7bc477ef5) and [xmtp_api_d14n/src/protocol/traits/envelope_collection.rs](https://github.com/xmtp/libxmtp/pull/2868/files#diff-0af39c7627e651b4a4f836c6082762ca4cb70e8992b62695adcd4cb2fa8bd512), then review the extractors in [xmtp_api_d14n/src/protocol/extractors/bytes.rs](https://github.com/xmtp/libxmtp/pull/2868/files#diff-a6cbfd5d967e5d7345d86635f7f820d05712ea8b41434f1179f1f4ad35e0341d) and [xmtp_api_d14n/src/protocol/extractors/orphaned_envelope.rs](https://github.com/xmtp/libxmtp/pull/2868/files#diff-a15a2ff96d5029829ea707ef220be104b97cdc4c2f7a12f085ff7e1d2c5ea3d4), and finally the delegating macro in [xmtp_api_d14n/src/protocol/macros.rs](https://github.com/xmtp/libxmtp/pull/2868/files#diff-b253d53f5aff5a0ddee62ddccdeda11f5959d4f1c3f7ed130e4f047fecf0c91a).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized e05e925.

PR #2857, opened 2025-11-26
more information to icebox table

insipx:
### Add more information to the icebox by migrating to a group-scoped schema with multi-dependency support and updating query/insert APIs in `xmtp_db` and `xmtp_proto`
Replace single in-row dependency with `icebox_dependencies`, require `group_id` on `icebox`, and add `QueryIcebox` for inserting and querying dependants; apply destructive migrations and update types and schema bindings.

#### 📍Where to Start
Start with the `QueryIcebox` trait and its `DbConnection` implementation in [icebox.rs](https://github.com/xmtp/libxmtp/pull/2857/files#diff-015edee4b874f40f0db8cdb344e1cd3bccf7d162977a16c5d1a0b740365416d0), then review the schema changes in the migrations in [up.sql](https://github.com/xmtp/libxmtp/pull/2857/files#diff-195fca113b520a99ac76f113de0413ad16c60fae52169c53ce37d21b78e88f2a) and [down.sql](https://github.com/xmtp/libxmtp/pull/2857/files#diff-90cdd5fc2dffcd969887be165293765efd0c1f003e7c99d439d6ee067348127a).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized f3f9562.

PR #2882, opened 2025-12-03
add icebox to `OrderedEnvelopeCollection` implementation

insipx:
### Add icebox support by requiring a `CursorStore` in `protocol.order.Ordered` and icing unresolved orphans during `OrderedEnvelopeCollection::order`
Introduce a store-backed ordering flow that recovers children, sorts by timestamp/causal order, requests dependencies as `RequiredDependency`, and ices unresolved orphans in the store; update resolver and traits to accept `RequiredDependency` and propagate store/decode errors. Core changes center on `protocol.order.Ordered` and dependent combinators/resolvers, with `Envelope` dropping `consume` and tests updated accordingly.

#### 📍Where to Start
Start with the `OrderedEnvelopeCollection for Ordered::order` implementation in [order.rs](https://github.com/xmtp/libxmtp/pull/2882/files#diff-78b9f275c691ca4b70a53c2a710b84b741cf86e28ac909dc8b250cbe9b61ed1e), then review `required_dependencies`, `recover_lost_children`, and the resolver changes in [network_backoff.rs](https://github.com/xmtp/libxmtp/pull/2882/files#diff-de7799794fbc3300554d0474e251ec4117a5b0773a085ba0f94dd6f863c2f51e).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 38c66a7. 12 files reviewed, 3 issues evaluated, 3 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_api_d14n/src/protocol/order.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 176](https://github.com/xmtp/libxmtp/blob/38c66a77f1f1dd3dd9d73354b1e6508ad6e1b859/xmtp_api_d14n/src/protocol/order.rs#L176): The filter logic change reduces matching precision: the old code matched envelopes by BOTH `topic` AND `cursor` using `MissingEnvelope::new(topic, cursor)`, but the new code only matches by `cursor` alone. This could cause the resolver to return envelopes with matching cursors but incorrect topics, potentially masking bugs where topic matching is important. If `RequiredDependency` contains topic information that should be used for matching, it's being ignored. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/resolve/network_backoff.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 53](https://github.com/xmtp/libxmtp/blob/38c66a77f1f1dd3dd9d73354b1e6508ad6e1b859/xmtp_api_d14n/src/protocol/resolve/network_backoff.rs#L53): The `resolve` function sleeps before making the first network query. On the first iteration with `attempts = 0`, `self.backoff.backoff(0, time_spent)` returns `Some(duration)` (base duration + jitter), causing an unnecessary delay before any query attempt. The first resolution attempt should be made immediately without sleeping. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/traits.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 88](https://github.com/xmtp/libxmtp/blob/38c66a77f1f1dd3dd9d73354b1e6508ad6e1b859/xmtp_api_d14n/src/protocol/traits.rs#L88): In `is_retryable()` for `EnvelopeError`, `Self::Decode(_) => true` marks `prost::DecodeError` as retryable. However, decode errors indicate malformed protobuf bytes, and retrying a decode operation on the same malformed bytes will produce the same error. This is inconsistent with the rationale documented in `ConversionError`'s `is_retryable()` implementation which explicitly states that conversions are not retryable because "the bytes are static". This could cause infinite retry loops or wasted retry attempts on permanently failing operations. <b>[ Already posted ]</b>
</details>

</details>

### DIFF
--- a/xmtp_api_d14n/src/protocol/order.rs
+++ b/xmtp_api_d14n/src/protocol/order.rs
@@ -1,13 +1,14 @@
 use std::collections::HashSet;
 
 use crate::protocol::{
-    Envelope, EnvelopeError, OrderedEnvelopeCollection, ResolutionError, ResolveDependencies,
-    Resolved, Sort, sort, types::MissingEnvelope,
+    CursorStore, Envelope, EnvelopeError, OrderedEnvelopeCollection, ResolutionError,
+    ResolveDependencies, Resolved, Sort, sort, types::RequiredDependency,
 };
 use derive_builder::Builder;
 use itertools::Itertools;
+use tracing::Level;
 use xmtp_proto::api::VectorClock;
-use xmtp_proto::types::TopicCursor;
+use xmtp_proto::types::{Cursor, OrphanedEnvelope, TopicCursor};
 
 /// Order dependencies of `Self` according to [XIP](https://github.com/xmtp/XIPs/blob/main/XIPs/xip-49-decentralized-backend.md#335-cross-originator-message-ordering)
 /// If dependencies are missing, this ordering will try to resolve them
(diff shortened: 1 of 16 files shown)


insipx:
This PR is part of a stack of 12 bookmarks:

1. `trunk()`
1. [insipx/causal-fix](https://github.com/xmtp/libxmtp/pull/2921)
1. [insipx/batch-queries](https://github.com/xmtp/libxmtp/pull/2922)
1. **insipx/icebox-in-ordering ← this PR**
1. [insipx/paged-stream](https://github.com/xmtp/libxmtp/pull/2931)
1. [insipx/ordered-stream](https://github.com/xmtp/libxmtp/pull/2932)
1. [insipx/ordered-stream-test](https://github.com/xmtp/libxmtp/pull/2933)
1. [push-pnvnmqsyqqqy](https://github.com/xmtp/libxmtp/pull/2936)
1. [push-syomxrnuqsnk](https://github.com/xmtp/libxmtp/pull/2946)

---

xmtp_api_d14n/src/protocol/resolve/network_backoff.rs
```diff
@@ -42,7 +44,7 @@ where
     /// * `HashSet<Self::ResolvedEnvelope>`: The list of envelopes which were resolved.
     async fn resolve(
         &self,
-        mut missing: HashSet<MissingEnvelope>,
```
insipx, line 45: feel like half this PR turned into renaming `MissingEnvelope` -> `RequiredDependency` but it's a rename that makes the intent of the type more clear IMO

the actual critical stuff is in `protocol/order.rs`

xmtp_api_d14n/src/protocol/order.rs
```diff
@@ -1,104 +1,138 @@
...
+            .map(|e| {
+                let dependencies = e.depends_on()?.unwrap_or(Default::default());
+                let topic = e.topic()?;
+                let topic_clock = self.topic_cursor.get_or_default(&topic);
+                let need = topic_clock.missing(&dependencies);
+                let needed_by = e.cursor()?;
+                Ok(need
+                    .into_iter()
+                    .map(|c| RequiredDependency::new(topic.clone(), c, needed_by))
+                    .collect::<HashSet<RequiredDependency>>())
```
codabrink: Why not collect into a Vec here so we're not hashing twice? Should be cheaper for the same outcome.

xmtp_api_d14n/src/protocol/order.rs
```diff
@@ -1,104 +1,138 @@
...
+    }
+
+    /// try to find any lost children and re-apply them to the
+    /// end of the envelopes list before any resolution occurs
+    fn recover_lost_children(&mut self) -> Result<(), EnvelopeError> {
+        let cursors: Vec<_> = self.envelopes.iter().map(|e| e.cursor()).try_collect()?;
+        let children = self.store.resolve_children(&cursors)?;
+        if !children.is_empty() {
+            tracing::info!("recovered {} children", children.len());
+            for child in &children {
```
codabrink: Tracing has an [enabled!](https://docs.rs/tracing/latest/tracing/macro.enabled.html) macro to check if a level is enabled if we want to wrap this loop in that.

xmtp_api_d14n/src/protocol/order.rs
insipx: good catch!

insipx: in fact I don't think we need to collect here at all, and can just leave the `map` for maximum laziness

PR #2890, opened 2025-12-04
Make stream_message_deletions async

rygine:
### Make `Conversations.stream_message_deletions` in [conversations.rs](https://github.com/xmtp/libxmtp/pull/2890/files#diff-305d8a9df912ee15a450791cf1af33e26798a0a4877f6a4dc457d291914bc126) async to add on_close support for the message deletion stream
Change the N-API method `Conversations.stream_message_deletions` to `async`, returning a `Future` that yields `Result<StreamCloser>` after invoking `RustXmtpClient::stream_message_deletions_with_callback`.

#### 📍Where to Start
Start with the `Conversations::stream_message_deletions` N-API method in [conversations.rs](https://github.com/xmtp/libxmtp/pull/2890/files#diff-305d8a9df912ee15a450791cf1af33e26798a0a4877f6a4dc457d291914bc126).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized abda21c.

### DIFF
--- a/bindings_node/src/conversations.rs
+++ b/bindings_node/src/conversations.rs
@@ -729,7 +729,7 @@ impl Conversations {
   }
 
   #[napi]
-  pub fn stream_message_deletions(
+  pub async fn stream_message_deletions(
     &self,
     callback: ThreadsafeFunction<String, ()>,
   ) -> Result<StreamCloser> {


rygine:
> We can add this if it adds more consistency. FWIW, these streams will never close since they are just listening for local events.

ok, in that case, these changes aren't necessary, but `stream_message_deletions` does need to be `async` in the node bindings so that it gets a tokio context.

neekolas (approved):

PR #2863, opened 2025-12-01
fix webassembly test `bindings_wasm` decodeText clash

insipx:
bindings_wasm `decodeText` clashes with another declared function. I'm not sure if it was a function declared in bindgen or just the test runner, but changing `decodeText` to `decodeXmtpText` fixed it. In our code, i only found a single `decodeText`. I changed `encodeText` to `encodeXmtpText` for consistency

Error message ci/local tests failing with:

```
Uncaught SyntaxError: redeclaration of function decodeText
```

it looks like some internal bindgen/js glue code, here is the function its conflicting with for historical purposes (from the sourcemap in console).
mostly i'm surprised these functions are not prefixed with `__` to avoid clashes like this
```js
function getUint8ArrayMemory0() {
    if (cachedUint8ArrayMemory0 === null || cachedUint8ArrayMemory0.byteLength === 0) {
        cachedUint8ArrayMemory0 = new Uint8Array(wasm.memory.buffer);
    }
    return cachedUint8ArrayMemory0;
}

let cachedTextDecoder = new TextDecoder('utf-8', { ignoreBOM: true, fatal: true });

cachedTextDecoder.decode();

const MAX_SAFARI_DECODE_BYTES = 2146435072;
let numBytesDecoded = 0;
function decodeText(ptr, len) {
    numBytesDecoded += len;
    if (numBytesDecoded >= MAX_SAFARI_DECODE_BYTES) {
        cachedTextDecoder = new TextDecoder('utf-8', { ignoreBOM: true, fatal: true });
        cachedTextDecoder.decode();
        numBytesDecoded = len;
    }
    return cachedTextDecoder.decode(getUint8ArrayMemory0().subarray(ptr, ptr + len));
}

function getStringFromWasm0(ptr, len) {
    ptr = ptr >>> 0;
    return decodeText(ptr, len);
}

```

PR #2856, opened 2025-11-26
add orphaned envelope type

insipx:
### Add `OrphanedEnvelope` type and re-export it from `xmtp_proto::types` to define orphaned envelope handling
Introduce `types::orphaned_envelope` with `OrphanedEnvelope` (builder, accessors, `into_payload`) and publicly expose it via `types` re-exports in [types.rs](https://github.com/xmtp/libxmtp/pull/2856/files#diff-3a297549d22d0dc04727c6d75baaf601334568dbba0976b206fe8006805ed202).

#### 📍Where to Start
Start with the `OrphanedEnvelope` definition and its builder in [orphaned_envelope.rs](https://github.com/xmtp/libxmtp/pull/2856/files#diff-fed8448996e2467cc9f39ee6dddc857ff4612566565a78e622bbedf6a2491681), then see the re-export in [types.rs](https://github.com/xmtp/libxmtp/pull/2856/files#diff-3a297549d22d0dc04727c6d75baaf601334568dbba0976b206fe8006805ed202).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 05ff62f.

PR #2895, opened 2025-12-05
test ordering implementation with icebox

insipx:
### Switch ordering to icebox-backed dependency handling and append-pass resolution in `protocol::Ordered::<S,R,T>::order` for test ordering with icebox
Replace vector-based orphan storage with a `HashSet` icebox and update dependency resolution to use `RequiredDependency` with `needed_by`. Update `protocol::Ordered::<S,R,T>::order` to append resolved and missing envelopes per pass and ice all missing when no progress occurs. Adjust hashing/equality for `OrphanedEnvelope` to ignore payload, add `GlobalCursor::cursors`, and update tests and resolvers to the new dependency type.

#### 📍Where to Start
Start with the ordering loop in `protocol::Ordered::<S,R,T>::order` in [order.rs](https://github.com/xmtp/libxmtp/pull/2895/files#diff-78b9f275c691ca4b70a53c2a710b84b741cf86e28ac909dc8b250cbe9b61ed1e), then review the icebox changes in `InMemoryCursorStore` in [in_memory_cursor_store.rs](https://github.com/xmtp/libxmtp/pull/2895/files#diff-81e05c4cbc4c5877da996a1302449b85dde8e85a96338b80953cbc520363c6c2) and the `RequiredDependency` model in [required_dependency.rs](https://github.com/xmtp/libxmtp/pull/2895/files#diff-3fff6e5b097e62e41e3eb04ca952f7b6b64a3e52a8a99126d6c17bb7b616ed60).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 1962291. 7 files reviewed, 27 issues evaluated, 25 issues filtered, 2 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_api_d14n/src/protocol/in_memory_cursor_store.rs — 1 comment posted, 2 evaluated, 1 filtered</summary>

- [line 144](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/in_memory_cursor_store.rs#L144): resolve_children no longer removes resolved children from the icebox. Previously, with a Vec-based icebox, the code used `extract_if` (or equivalent) to remove matched entries. Now it clones matches via `.iter().cloned().filter(...)` and returns them, leaving the originals in the `HashSet`. This changes externally visible behavior: resolved orphans remain and will be repeatedly returned on subsequent calls, causing duplicate processing and unbounded growth over time. Either restore removal (e.g., `extract_if`/`drain_filter` equivalent or collect keys then remove) or document the contract change and ensure downstream deduplicates. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/order.rs — 1 comment posted, 20 evaluated, 19 filtered</summary>

- [line 84](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L84): Potential duplicate insertion: the deduplication only excludes children whose `cursor` already exists in `self.envelopes` via `filter(|o| !cursors.contains(&o.cursor))`, but it does not deduplicate duplicates among the `children` collection itself. If `store.resolve_children(&cursors)` returns multiple `OrphanedEnvelope`s with the same `cursor`, both will be decoded and appended, resulting in duplicate envelopes in `self.envelopes`. This can silently violate uniqueness invariants for `cursor` and may break downstream sorting/causal assumptions. <b>[ Out of scope ]</b>
- [line 114](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L114): Implementation contradicts the NOTE above: the comment describes that when resolution fails, the child is re-iced (step 4). In this code path, when some dependencies remain unresolved but `resolved` is non-empty, the code appends `missing` to `self.envelopes` instead of re-icing those entries. This creates ambiguity about the intended behavior and may violate the documented invariant. <b>[ Low confidence ]</b>
- [line 123](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L123): Potential livelock/infinite loop: If `resolve(...)` returns a non-empty `resolved` set but never fully resolves all dependencies, the loop will continue indefinitely because `missing` items (including those still unresolved) are re-appended to `self.envelopes`, `self.causal_sort()?` will likely produce `Some(missing)` again, and there is no backoff or progress guarantee; the `break` occurs only when `resolved.is_empty()`. <b>[ Low confidence ]</b>
- [line 131](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L131): Over-aggressive icing: When `resolved.is_empty()`, all `missing` are iced (`self.store.ice(orphans)?`) and the loop breaks. This differs from prior behavior that only iced envelopes corresponding to truly unresolved dependencies, risking unnecessary icing of items whose dependencies might have been immediately resolvable or become available with a different batch/ordering. <b>[ Low confidence ]</b>
- [line 136](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L136): Partially unresolved dependencies are ignored: the `unresolved` returned by `self.resolver.resolve(...)` is no longer examined or used to cull `missing`. As a result, envelopes in `missing` whose dependencies remain unresolved are still appended to `self.envelopes` (`self.envelopes.append(&mut missing)`), allowing items with unmet dependencies to be re-queued/applied, violating causal invariants and risking repeated failures or churn. <b>[ Low confidence ]</b>
- [line 138](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L138): Ordering regression: Previously, items were inserted at the front via `splice(0..0, ...)` ensuring immediate processing of `resolved` (and then `missing`) relative to the current queue. Now they are appended with `self.envelopes.append(...)`. This can delay application and allow interleaving with unrelated envelopes already in the queue, potentially violating causal ordering and starving dependent items. <b>[ Low confidence ]</b>
- [line 192](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L192): `extract_if(.., |env| missing_set.contains(&cursor))` will remove and return all entries in `available` whose `cursor` is in `missing_set`. If multiple items in `available` share the same `cursor` (e.g., duplicates or same cursor across topics), this will resolve multiple envelopes for a single required dependency, causing over-removal and potential double-application of effects, violating uniqueness/at-most-once expectations. <b>[ Test / Mock code ]</b>
- [line 194](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L194): Topic scoping was removed when matching available envelopes to `missing` dependencies. Previously the match used both `topic` and `cursor` (`env.topic().unwrap()` with `MissingEnvelope::new(topic, cursor)`), but now it uses only `cursor` via `missing_set.contains(&cursor)`. If `cursor` values are not globally unique across topics, this can resolve and remove the wrong envelope(s) from `available`, misreport items as resolved, and leave the actual required dependency unresolved. It also silently changes externally visible behavior (contract parity) from topic+cursor to cursor-only matching. <b>[ Test / Mock code ]</b>
- [line 218](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L218): `unavailable.retain(|e| !available.contains(e));` computes the difference by value equality. If `envelopes.removed` can contain duplicate elements (equal by `PartialEq`), and the `available` subsequence includes only some of those duplicates, this logic will remove all equal duplicates from `unavailable`, not just the instances actually chosen for `available`. That can silently drop entries and break the intended partition between `available` and `unavailable`. <b>[ Test / Mock code ]</b>
- [line 245](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L245): Inconsistent snapshot/race in logic: `available()` acquires a lock and clones a snapshot, but `unavailable` is read without synchronization (`&self.resolver.unavailable`). If `unavailable` is mutable elsewhere (as suggested by `available` being locked), checks against it can be inconsistent within a single call, causing nondeterministic aborts and violating the intent of evaluating a consistent dependency state. <b>[ Low confidence ]</b>
- [line 258](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L258): Potential infinite loop: the dependency walk lacks cycle detection. If the dependency graph contains a cycle (and none of the involved nodes are marked unavailable), `depends` will continually be repopulated, preventing `depends` from becoming empty and causing the `while` loop to never terminate. <b>[ Low confidence ]</b>
- [line 265](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L265): `find(...).unwrap()` will panic if a dependency cursor is not present in either `available` or `to_be_ordered`. A dependency may be missing (e.g., not yet fetched, filtered out, or otherwise absent), which would cause a runtime panic instead of a graceful rejection of that envelope. Replace the `unwrap()` with explicit handling (abort the envelope or return an error) when `find` returns `None`. <b>[ Low confidence ]</b>
- [line 315](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L315): Uses `now_or_never()` followed by `expect("Future should complete immediately")` and `unwrap()` which will panic if the future doesn’t complete synchronously. There is no visible guarantee that `order()` is always ready; if it ever yields `Pending`, `now_or_never()` returns `None`, causing a panic, making this test flaky and timing-dependent. <b>[ Out of scope ]</b>
- [line 321](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L321): Unconditional `unwrap()` on `topic()` at `env.topic().unwrap()` may panic if an envelope lacks a topic. No visible guard ensures all generated envelopes have topics in this path. <b>[ Test / Mock code ]</b>
- [line 359](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L359): Unconditional `unwrap()` on `topic()` at `available_env.topic().unwrap()` may panic if an envelope lacks a topic. This block has no guard ensuring a topic exists. <b>[ Test / Mock code ]</b>
- [line 398](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L398): Uses `now_or_never()` followed by `expect("Future should complete immediately")` and `unwrap()` which will panic if the future is pending. No guarantee here that the second ordering pass completes synchronously. <b>[ Test / Mock code ]</b>
- [line 449](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L449): Second ordering pass also uses `now_or_never()` followed by `expect("Future should complete immediately")` and `unwrap()`, risking panic if the future is pending. No visible guarantee of immediate completion. <b>[ Test / Mock code ]</b>
- [line 462](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L462): Re-uses name `orphan_count` inside the recovery phase, shadowing the earlier pre-recovery `orphan_count`. The subsequent `let had_children = orphan_count > 0 && had_children;` uses the post-reordering `orphan_count`, which can zero-out the flag even if there were children before, weakening or bypassing assertions. This likely defeats the intent to check that there were orphans prior to recovery and that they were recovered. <b>[ Test / Mock code ]</b>
- [line 499](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/order.rs#L499): Unconditional `unwrap()` on `topic()` at `envelope.topic().unwrap()` in the recovery verification may panic if an envelope lacks a topic. No guard ensures presence. <b>[ Test / Mock code ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/resolve/network_backoff.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 59](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/resolve/network_backoff.rs#L59): Dependencies can be dropped without any network attempt if the backoff returns `None` on the first iteration. The loop sleeps only if backoff yields `Some(...)`, and if it yields `None` at `attempts == 0`, it logs and breaks, returning all as unresolved without ever issuing a query. With configurations like `total_wait_max == 0`, this path is reachable and violates the expectation to attempt resolution at least once. <b>[ Low confidence ]</b>
- [line 70](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/resolve/network_backoff.rs#L70): Resolver aborts on the first API error and does not retry despite `ApiClient::Error: RetryableError`. The call to `.query(&self.client).await.map_err(ResolutionError::api)?` propagates any network/API error immediately, bypassing backoff/retry behavior implied by the resolver name and trait bound. This causes transient errors to fail the whole resolution instead of being retried. <b>[ Low confidence ]</b>
- [line 79](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/resolve/network_backoff.rs#L79): Potential duplicate envelopes in `resolved`. The loop appends all returned envelopes every iteration (`resolved.extend(envelopes);`) without de-duplication. If the server returns previously seen envelopes across retries/pages, duplicates will be emitted, violating uniqueness and risking double-application of effects by callers. <b>[ Low confidence ]</b>
- [line 79](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/resolve/network_backoff.rs#L79): Resolved list may include envelopes that were not requested. The code extends `resolved` with all `envelopes` returned by the server (`resolved.extend(envelopes);`) without filtering to only those that match the currently required dependencies. If the server returns extra envelopes (e.g., within the topic and before/after the requested cursors), these unrelated envelopes will be emitted in the `Resolved.resolved` output, violating the expectation that only the requested dependencies are returned and risking double-application or out-of-order effects downstream. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/traits/dependency_resolution.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 61](https://github.com/xmtp/libxmtp/blob/19622910781a307ba5bc3f8b5a7168dc3bfe6b1d/xmtp_api_d14n/src/protocol/traits/dependency_resolution.rs#L61): `NoopResolver::resolve` always sets `unresolved: Some(m)` even when `m` is empty. Other resolver paths (e.g., `network_backoff::resolve`) use `(!missing.is_empty()).then_some(missing)` so that `unresolved` is `None` when there are no unresolved dependencies. Returning `Some(empty)` can break contract parity and lead callers that check `is_some()` to misinterpret completion as having unresolved dependencies. <b>[ Low confidence ]</b>
</details>

</details>


insipx:
This PR is part of a stack of 7 bookmarks:

1. `trunk()`
1. [push-tkyvzsqvnkyz](https://github.com/xmtp/libxmtp/pull/2863)
1. [insipx/orphaned-envelope](https://github.com/xmtp/libxmtp/pull/2856)
1. **push-zxvxknpsxlmu ← this PR**

---

PR #2899, opened 2025-12-05
reset membership state after re-add to the already left group

mchenani:
### Reset membership to Allowed and set consent to Unknown on re-add by updating `welcomes::xmtp_welcome::welcome` and add tests for group rejoin flow
Update `welcomes::xmtp_welcome::welcome` to detect re-adds via `db.find_group` and set `membership_state` to `Allowed` and consent to `Unknown` for existing groups; add tests covering leave, removal, re-add, and post-sync state in [xmtp_mls/src/groups/tests/mod.rs](https://github.com/xmtp/libxmtp/pull/2899/files#diff-3943e92f9d0be4ce83280dcbb75f2f222fc96342e69cb677bcd2b006e2182c4a) and [bindings_ffi/src/mls/tests/test_self_removal.rs](https://github.com/xmtp/libxmtp/pull/2899/files#diff-e8308ce3114bb2f02b672deb422411959aedd4a678df1709b9b382520e708109).

#### 📍Where to Start
Start with the re-add handling in `welcome` within [xmtp_mls/src/groups/welcomes/xmtp_welcome.rs](https://github.com/xmtp/libxmtp/pull/2899/files#diff-a8d5a5fab05f9277d06cce92de1f0806f13f6d3298cf178dcd67f8599f94220e), then review the new `test_membership_state_after_readd` tests in [xmtp_mls/src/groups/tests/mod.rs](https://github.com/xmtp/libxmtp/pull/2899/files#diff-3943e92f9d0be4ce83280dcbb75f2f222fc96342e69cb677bcd2b006e2182c4a).

#### Changes since #2899 opened

- Modified welcome processing logic to detect re-add scenarios by checking if existing group has `GroupMembershipState::PendingRemove`, setting new member's `membership_state` to `Allowed` only in this case and `Pending` otherwise, updating the existing group's membership to `Allowed` in the database via `update_group_membership` before calling `insert_or_replace_group`, and restricting consent reset to `Unknown` to only occur during re-add-after-leaving scenarios [daead68]
- Added mock expectations in nested transaction test setups to return `Ok(None)` when `find_group` is called during welcome processing [daead68]
- Modified re-add detection logic in `check_if_processed` method within `xmtp_welcome` module [16d31aa]
- Refactored conditional logic in `xmtp_welcome.check_if_processed` method to use let-chain syntax [c7856c4]
- Modified re-add detection logic in `xmtp_welcome.check_if_processed` method to require incoming welcome cursor's `sequence_id` to be strictly greater than the stored group's `sequence_id` for members in `PendingRemove` state [ddf6fd9]
- Modified re-add detection logic in `xmtp_welcome.check_if_processed` method to require explicit `sequence_id` comparison [b1f71af]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized ddf6fd9. 2 files reviewed, 2 issues evaluated, 2 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_mls/src/groups/welcomes/xmtp_welcome.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 437](https://github.com/xmtp/libxmtp/blob/ddf6fd920d932ce9b5947b2d371283b01997030a/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L437): Event ordering/leak: `Sync`-group path enqueues a `SyncWorkerEvent::NewSyncGroupFromWelcome(group_id)` before persistence, at lines where the event is added (prior to `insert_or_replace_group`). If subsequent operations fail with a non-retryable error while `cursor_increment` is true, the savepoint returns `Ok(CommitResult::FailedForever(err))`, the outer transaction commits, and `events.send_all(&self.context)` is still called. This sends a worker event announcing a new sync group that was not actually stored, violating at-most-once/validity guarantees for emitted events. <b>[ Out of scope ]</b>
- [line 512](https://github.com/xmtp/libxmtp/blob/ddf6fd920d932ce9b5947b2d371283b01997030a/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs#L512): Potential integer overflow/semantic corruption when converting `welcome_metadata.message_cursor` from `u64` to `i64` and later back to `u64`. In `let cursor = welcome_metadata.map(|m| m.message_cursor as i64).unwrap_or_default();` a large `u64` value (>= i64::MAX) will wrap when cast to `i64`, producing a negative `i64`. This corrupted `cursor` is then used to (a) persist `StoredGroupMessage.sequence_id` (an `i64`) and (b) update the commit cursor via `Cursor::mls_commits(cursor as u64)`, re-wrapping the negative `i64` to a very large `u64`. This can silently produce incorrect cursors and break ordering/consistency guarantees. A safe bound check or using `u64` end-to-end (with DB/schema alignment) is required. <b>[ Previously rejected ]</b>
</details>

</details>

### DIFF
--- a/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs
+++ b/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs
@@ -377,13 +377,16 @@ where
 
         // Check if this is a re-add scenario (user was in PENDING_REMOVE state)
         // This happens when a user leaves or is removed, then gets re-added with a NEW welcome
-        // We verify it's a NEW welcome by checking that the sequence_id is GREATER (not just different)
-        // This prevents incorrectly treating backup/restore as a re-add:
-        // - True re-add: new welcome has HIGHER sequence_id than the old group
+        // We verify it's a NEW welcome by checking that:
+        // 1. The existing group has a valid sequence_id (Some)
+        // 2. The new welcome's sequence_id is GREATER than the existing one
+        // This prevents incorrectly treating backup/restore or groups without sequence_ids as re-adds:
+        // - True re-add: existing group has sequence_id AND new welcome has HIGHER sequence_id
         // - Backup/restore: welcome might have different but not necessarily higher sequence_id
+        // - Groups without sequence_id: should not be treated as re-adds
         let is_readd_after_leaving = existing_group.as_ref().is_some_and(|g| {
             g.membership_state == GroupMembershipState::PendingRemove
-                && welcome.cursor.sequence_id as i64 > g.sequence_id.unwrap_or(0)
+                && matches!(g.sequence_id, Some(seq) if (welcome.cursor.sequence_id as i64) > seq)
         });
 
         // Determine the membership state


xmtp_mls/src/groups/welcomes/xmtp_welcome.rs
```diff
@@ -373,6 +373,20 @@ where
             }
         });
 
+        let existing_group = db.find_group(group_id.as_slice())?;
+        let is_readd = existing_group.is_some();
```
neekolas: `is_readd` would also appear as `true` if you had previously restored the group from device sync or a backup, right?

In that situation, we'd want to make sure we respect the consent state from the backup

PR #2900, opened 2025-12-05
Add a migration check to startup

codabrink:
We don't want to allow old clients running on mismatched databases.
i.e. a 1.5 client running on a 1.6 database as unexpected behavior will occur, potentially corrupting the database beyond repair.

### DIFF
--- a/xmtp_db/src/encrypted_store/mod.rs
+++ b/xmtp_db/src/encrypted_store/mod.rs
@@ -102,8 +102,12 @@ pub enum ConnectionError {
     ReconnectInTransaction,
     #[error("invalid query: {0}")]
     InvalidQuery(String),
-    #[error("{0}")]
-    Generic(String),
+    #[error(
+        "Applied migrations does not match available migrations.\n\
+    This is likely due to running a database that is newer than this version of libxmtp.\n\
+    Expected: {expected}, found: {found}"
+    )]
+    InvalidVersion { expected: String, found: String },
 }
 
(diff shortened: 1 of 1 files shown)


xmtp_db/src/encrypted_store/mod.rs
```diff
@@ -110,6 +115,7 @@ impl RetryableError for ConnectionError {
             Self::DisconnectInTransaction => true,
             Self::ReconnectInTransaction => true,
             Self::InvalidQuery(_) => false,
+            Self::Generic(_) => false,
```
insipx: Would be nice to make this more descriptive than "Generic" -- ideally form the string according to thiserror, maybe something like 

```rust
#[error(
"Applied migrations does not match available migrations.\n\
This is likely due to running a database that is newer than this version of libxmtp.\n\
Expected: {expected}, found: {found}"
)]
InvalidVersion {
    expected: String,
    found: String
}
```

it just makes grepping for a specific version type much clearer/easier, and also puts all error messages in one centralized place

codabrink: This would be a great discussion piece at the next rust guild. 😄 
We can do that. I just haven't seen much reason for building error types for errors that we don't plan on recovering from, and have found that creating specific error types encourages unfortunately vague error messages. (Since they can be applied in several places, and we never match on them anyway.) Unless we build ginormous error structs.

PR #2908, opened 2025-12-06
Fix lots of typos

Sajjon:
Fix 60+ typos in 40+ files using excellent [`typos`](https://github.com/crate-ci/typos) CLI tool.

Let me know if you are interested in me adding a [pre-commit](https://pre-commit.com/) file with `typos` as a step (e.g. [Sajjon/klirr/.pre-commit-config.yaml](https://github.com/Sajjon/klirr/blob/main/.pre-commit-config.yaml), and CI, as seen [in Sajjon/klirr/.github/../test.yml](https://github.com/Sajjon/klirr/blob/main/.github/workflows/test.yml#L25) - if you are, this is a `_typos.toml` which produces this diff:

```toml
[files]
extend-exclude = [
	"xmtp_id/artifact/CoinbaseSmartWallet.json"
]

[default]
extend-ignore-identifiers-re = [
    "[a-zA-Z0-9]*2nd[a-zA-Z0-9]*",
    "SetRefeshJobRequest",
]
```

As you can see there is a type in the `SetRefeshJobRequest` type, which I chose to not change.

### DIFF
--- a/bindings_ffi/src/fork_recovery.rs
+++ b/bindings_ffi/src/fork_recovery.rs
@@ -29,7 +29,7 @@ pub struct FfiForkRecoveryOpts {
     // After a 'bad' commit is made, fork recovery can be expected to take up to 4x this interval
     // to complete end-to-end, assuming both the super admin and forked installation are online
     // and streaming welcomes.
-    // This can be overriden for end-to-end/manual testing purposes.
+    // This can be overridden for end-to-end/manual testing purposes.
     // If the interval duration is less than the time it takes for a single tick of the worker to
     // complete, the worker will wait for the first tick to complete before the next tick begins.
     // Worth considering if any strange behavior is observed with low intervals.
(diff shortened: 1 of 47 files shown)


xmtp_debug/src/args.rs
```diff
@@ -260,7 +260,7 @@ impl std::fmt::Display for EntityKind {
 /// specify the log output
 #[derive(Args, Debug)]
 pub struct LogOptions {
-    /// Output libxmtp logs into file with a structured, ndJSON format
```
Sajjon, line 263: not sure if `", ndJSON format"` meant something I'm unaware of? I think it is a strange typo, no?

insipx: not a typo, stands for "newline-delmited JSON", its specific to a log format where each line is a JSON object delimited by newlines

Sajjon: Thanks! I will revert this one

Sajjon: reverted

PR #2915, opened 2025-12-08
fix deeply nested dependencies dont make it to the query

insipx:
### Unify cursor construction and change dependency resolution to use `RequiredDependency` with transitive child recovery across ordering and in-memory icebox
Replace ad-hoc `Cursor` literals with `Cursor::new`, switch dependency modeling from `MissingEnvelope` to `RequiredDependency` including a `needed_by` cursor, change `InMemoryCursorStore` icebox to `HashSet` with recursive child resolution, and update the ordering loop to compute required dependencies, ice en masse when none resolve, and append resolved and missing envelopes.

#### 📍Where to Start
Start with `Ordered::order` and `Ordered::required_dependencies` in [order.rs](https://github.com/xmtp/libxmtp/pull/2915/files#diff-78b9f275c691ca4b70a53c2a710b84b741cf86e28ac909dc8b250cbe9b61ed1e), then review `InMemoryCursorStore` changes and child resolution in [in_memory_cursor_store.rs](https://github.com/xmtp/libxmtp/pull/2915/files#diff-81e05c4cbc4c5877da996a1302449b85dde8e85a96338b80953cbc520363c6c2), and the new dependency type in [required_dependency.rs](https://github.com/xmtp/libxmtp/pull/2915/files#diff-3fff6e5b097e62e41e3eb04ca952f7b6b64a3e52a8a99126d6c17bb7b616ed60).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 49f202f. 24 files reviewed, 42 issues evaluated, 40 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>xmtp_api_d14n/src/protocol/extractors/group_messages.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 60](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/extractors/group_messages.rs#L60): Possible panic due to unchecked timestamp conversion: `DateTime::from_timestamp_nanos(envelope.originator_ns)` can panic if `originator_ns` is out of the valid range for `chrono`. Since this value can come from external data, the code should validate or use a fallible conversion (or clamp/fallback) to avoid runtime termination. <b>[ Previously rejected ]</b>
- [line 122](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/extractors/group_messages.rs#L122): Possible panic due to unchecked timestamp conversion: `DateTime::from_timestamp_nanos(message.created_ns as i64)` can panic if `created_ns` is out of the valid range for `chrono`. Input comes from messages; prefer a fallible/validated conversion and handle errors instead of panicking. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/extractors/welcomes.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 53](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L53): `DateTime::from_timestamp_nanos(envelope.originator_ns)` may panic at runtime if `envelope.originator_ns` is out of the valid range for `chrono::DateTime` (nanoseconds since epoch as an `i64`). There is no guard or fallible handling here. Use a fallible constructor (e.g., `from_timestamp_nanos_option` or validate range and return an error) to avoid panics on malformed or extreme input. <b>[ Previously rejected ]</b>
- [line 113](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L113): In `visit_v3_welcome_message`, `message.created_ns as i64` can silently wrap/overflow if `created_ns` exceeds `i64::MAX`, producing a wildly incorrect timestamp. The resulting value is then passed to `DateTime::from_timestamp_nanos`, which may panic if the (wrapped) value is out of range. Use a fallible conversion (`i64::try_from`) and a fallible DateTime constructor to return an error instead of producing incorrect times or panicking. <b>[ Previously rejected ]</b>
- [line 133](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/extractors/welcomes.rs#L133): In `visit_v3_welcome_pointer`, `message.created_ns as i64` can silently wrap/overflow if `created_ns` exceeds `i64::MAX`, yielding incorrect times. The wrapped value is then passed to `DateTime::from_timestamp_nanos`, which may panic for out-of-range values. Use `i64::try_from(message.created_ns)` and a fallible DateTime constructor (or range-check) to avoid overflow and panics. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/in_memory_cursor_store.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 161](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/in_memory_cursor_store.rs#L161): Potential infinite recursion in `resolve_children_inner`: if the icebox contains a cycle of orphaned envelopes (A depends on B, B depends on A, etc.), the recursion will repeatedly compute the same `children` and `cursors` sets and call `resolve_children_inner` again without a visited/termination check, leading to non-termination or stack overflow. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/order.rs — 0 comments posted, 16 evaluated, 16 filtered</summary>

- [line 126](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L126): Possible infinite loop/livelock: the `while let Some(mut missing) = self.causal_sort()? { ... }` loop (lines 126–141) has no cap or forward-progress check beyond `resolved.is_empty()`. If `self.resolver.resolve(needed_envelopes)` continually returns a non-empty `resolved` set that does not ultimately cause `self.causal_sort()` to return `None` (e.g., cyclic dependencies, repeatedly resurfacing or already-present items), the loop can run indefinitely. The code comment on line 124 acknowledges this risk. This can result in unbounded CPU usage and the function never reaching a terminal state. <b>[ Low confidence ]</b>
- [line 126](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L126): Possible infinite loop in `order()`: the `while let Some(mut missing) = self.causal_sort()?` loop has no termination guard and now ignores the resolver's `unresolved` set entirely. If `self.resolver.resolve(needed_envelopes)` repeatedly returns a non-empty `resolved` (e.g., duplicates or items that do not advance causal state) while leaving the same dependencies unresolved, the loop will keep re-appending `missing` and continue indefinitely. The code even notes the risk with a comment about adding a maximum resolution attempts. At runtime this can hang the ordering process and starve progress. <b>[ Low confidence ]</b>
- [line 128](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L128): Ignoring `unresolved` results from the resolver may cause repeated re-processing without progress and contribute to livelock. The code destructures `Resolved { mut resolved, unresolved }` but never reads `unresolved`. If dependencies are permanently missing or cyclic, and `resolved` intermittently contains known or irrelevant items, the loop may continue without ever acting on `unresolved` (e.g., by icing or explicitly failing earlier), increasing the chance of non-termination or unnecessary cycles until `resolved` eventually becomes empty. <b>[ Low confidence ]</b>
- [line 131](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L131): Resolver `unresolved` results are ignored in `order()`. The code destructures `Resolved { mut resolved, unresolved, }` but never uses `unresolved`. Previously, unresolved dependencies were selectively iced and removed from `missing`. Now, when `resolved` is non-empty, all `missing` are re-appended; when `resolved` becomes empty, all remaining `missing` are iced wholesale. This is a contract/behavior change that can cause unnecessary re-queuing and later over-icing of envelopes that could have been specifically identified as permanently unresolved, increasing latency and resource usage. <b>[ Low confidence ]</b>
- [line 137](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L137): Potential unbounded growth and duplication in `self.envelopes`: on success paths the code appends both `resolved` and `missing` back into `self.envelopes` unconditionally (`self.envelopes.append(&mut resolved); self.envelopes.append(&mut missing);`). If any of these elements are already present (e.g., because `causal_sort()` did not remove them or they have been appended in a prior iteration), this can accumulate duplicates and grow memory without bound, especially under the looping condition above. There is no deduplication or removal shown. <b>[ Low confidence ]</b>
- [line 187](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L187): Behavioral regression: resolution now matches only by `cursor` and ignores other fields (e.g., topic) from `RequiredDependency`. `missing_set` is built from `m.cursor` and the predicate uses only `cursor`, whereas the previous logic matched `topic`+`cursor`. If `cursor` is not globally unique across topics or dependency kinds, this can return/consume incorrect envelopes and misreport unresolved items. <b>[ Test / Mock code ]</b>
- [line 190](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L190): Potential deadlock due to lock-ordering: the code acquires `self.available.lock()` into `available` and, while that mutex guard is still in scope, subsequently acquires `self.returned.lock()` into `ret`. If any other code path acquires `returned` first and then `available`, this can deadlock. Release `available` (e.g., by scoping or `drop(available)`) before locking `returned`. <b>[ Test / Mock code ]</b>
- [line 213](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L213): `assert_topic_cursor_seen` uses `env.topic().unwrap()` (line 213). If `env.topic()` returns `None`, this will panic, bypassing `proptest`'s shrinking/reporting and breaking the function's `Result<(), TestCaseError>` contract. Prefer `prop_assert!(env.topic().is_some(), ...)` then safely use the value (e.g., via `if let Some(topic) = env.topic() { ... }`). <b>[ Test / Mock code ]</b>
- [line 213](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L213): `assert_topic_cursor_seen` uses `cursor.get(&env.topic().unwrap()).unwrap()` (line 213). If the `cursor` does not contain the topic key, the second `unwrap()` will panic, bypassing `proptest` failure reporting and the function’s `Result<(), TestCaseError>` contract. Use `prop_assert!(cursor.get(&topic).is_some(), ...)` and then use the reference safely. <b>[ Test / Mock code ]</b>
- [line 223](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L223): Potential panic due to unguarded `unwrap()` on `env.topic()`. If `env.topic()` returns `None`, this will abort the test execution instead of returning a `TestCaseError` via `prop_assert!`, breaking the property-test contract. Prefer using a `prop_assert!(env.topic().is_some(), ...)` or handling the `Option` without panicking to ensure a graceful test failure. <b>[ Test / Mock code ]</b>
- [line 289](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L289): `make_available` removes an element from `unavailable` by index without validating bounds: `v.remove(idx)` will panic at runtime if `idx >= v.len()`. This is not an initialization-time invariant and can be hit by callers passing an out-of-range index, causing unwinding mid-operation. <b>[ Low confidence ]</b>
- [line 323](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L323): `has_unavailable_in_dependency_chain` performs a transitive traversal of dependencies without cycle detection (no visited set). If there are cyclic dependencies, the `while let Some(deps) = to_check.pop()` loop can push dependency sets indefinitely, leading to a non-terminating loop and potential hang. <b>[ Low confidence ]</b>
- [line 345](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L345): Concurrency/consistency risk: `only_valid_dependants` and `has_unavailable_in_dependency_chain` build results by taking multiple independent snapshots (`unavailable()`, then repeated `all_envelopes()` with separate locks) without holding a consistent view across them. Concurrent updates (e.g., via `make_available`) can interleave between these snapshots, yielding inconsistent or incorrect results (false positives/negatives) relative to any single logical state. <b>[ Low confidence ]</b>
- [line 376](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L376): The test assumes `order()`'s future is always immediately ready and calls `now_or_never().expect("Future should complete immediately")`. This will panic if `order()` ever performs any async/await that yields (e.g., I/O, timers, or awaits another future). Rust Futures do not guarantee immediate readiness; using `now_or_never` like this can create flaky tests or false negatives. <b>[ Test / Mock code ]</b>
- [line 438](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L438): Same immediate-readiness assumption: `ordered.order().now_or_never().expect(...)` can panic if the future yields. This makes the proptest flaky if `order()` stops being immediately ready under some generated inputs or internal timing. <b>[ Test / Mock code ]</b>
- [line 483](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/order.rs#L483): Same pattern again in the second ordering pass: `ordered.order().now_or_never().expect(...)` assumes synchronous completion and can panic if the future yields. This can hide real behavior changes and introduce flakiness. <b>[ Test / Mock code ]</b>
</details>

<details>
<summary>xmtp_api_d14n/src/protocol/sort/causal.rs — 1 comment posted, 2 evaluated, 1 filtered</summary>

- [line 36](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_api_d14n/src/protocol/sort/causal.rs#L36): Non-atomic state update: if `self.topic_cursor.apply(env)?` succeeds (mutating external state) and `missing_cursor.apply(env)?` subsequently fails, `sort` returns `Err` while leaving `self.topic_cursor` advanced. This partial mutation on the error path can violate invariants and surprise callers who receive an error but observe updated cursor state. Consider applying to the auxiliary cursor first and only committing to `self.topic_cursor` after success, or implementing rollback/error containment. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/group.rs — 0 comments posted, 3 evaluated, 2 filtered</summary>

- [line 133](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/group.rs#L133): `StoredGroup::cursor` performs lossy, unchecked integer casts (`i64` to `u64` for `sequence_id` and `i64` to `u32` for `originator_id`) via `as` when constructing a `Cursor`. Negative values or values exceeding the target type's range will wrap, yielding incorrect cursors at runtime. Use fallible conversions (e.g., `try_into()` with handling) or validate ranges before casting. <b>[ Previously rejected ]</b>
- [line 956](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/group.rs#L956): Unchecked casts from signed to unsigned can silently wrap. `sequence_id` (`i64`) is cast to `u64` and `originator_id` (`i64`) to `u32` when constructing `Cursor`. If negative values exist in the DB, this will produce large wrapped numbers rather than failing, yielding invalid cursors. Add non-negative guards in SQL (e.g., `>= 0`) or validate before casting and return an error. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/group_intent.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 545](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/group_intent.rs#L545): Unchecked casts from signed DB fields to unsigned types. When building `Cursor`, `sequence_id` (`i64`) is cast to `u64` and `originator_id` (`i32`) to `u32`. Negative values would wrap, creating invalid cursors. Add guards (`>= 0`) or validate before casting and return an error. <b>[ Low confidence ]</b>
- [line 568](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/group_intent.rs#L568): Missing keys are silently dropped in the returned map. The function is given `payload_hashes` to look up, but it builds the map only from join results. Payload hashes with no matching dependency rows never appear in the map, and the intended `NoDependencyFound` error path is unreachable because the grouping map contains no entry (and thus is never iterated). This is silent data loss at a mapping boundary. Either: (a) seed the map with every requested hash and then error or insert a `None`/default when no rows exist, or (b) explicitly document that absent hashes are omitted. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/group_message.rs — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 1268](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/group_message.rs#L1268): Batch filter starts with `dsl::group_id.eq(&[] as &[u8])` as a 'false' seed. If any row legitimately has an empty `group_id`, that seed is actually true and may leak in unrelated rows via the `OR` chain. Use a guaranteed-false predicate instead (e.g., `diesel::dsl::sql::<diesel::sql_types::Bool>("0=1")`) to avoid accidental matches. <b>[ Low confidence ]</b>
- [line 1326](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/group_message.rs#L1326): Unchecked casts from signed to unsigned when constructing `Cursor`. Results selected as `(i64, i64)` are later cast to `u64`/`u32` without validation. If any negative values are present, the returned cursors will be nonsensical due to wrapping. Add SQL guards (`>= 0`) or validate/return error before constructing. <b>[ Low confidence ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/icebox.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 397](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/icebox.rs#L397): Potential panic due to unchecked vector indexing on `orphans` at `orphans[0]`, `orphans[1]`, and `orphans[2]`. If `iced(group_id)` returns fewer than 3 elements, these lines will panic. Consider asserting the length before indexing or pattern-matching to ensure at least three elements. <b>[ Test / Mock code ]</b>
- [line 397](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/icebox.rs#L397): Potential panic due to unchecked HashMap indexing on `depends_on` via `depends_on[&1]` and `depends_on[&2]`. If the keys are absent, these lines will panic. Occurs at lines 397–399 (on `orphans[...]`) and again at lines 407 and 411 (on `dep_chain[...]`). Consider using `get` with assertions or clearer error messages. <b>[ Test / Mock code ]</b>
- [line 422](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/icebox.rs#L422): Potential out-of-bounds indexing on `orphans[2]`. There is no visible guard ensuring `iced(group_id.clone())` returns at least 3 elements. If `orphans.len() < 3`, this will panic at runtime. Consider asserting the length or using `get_mut(2)` and handling `None` with a clear failure. <b>[ Test / Mock code ]</b>
- [line 438](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/icebox.rs#L438): Unchecked map indexing `dep_chain[0].depends_on[&2]` and `dep_chain[1].depends_on[&1]` can panic if the expected keys are absent. Prefer `get(&2)`/`get(&1)` with assertions on `is_some()` to avoid runtime panics and provide clearer failure messages. <b>[ Out of scope ]</b>
</details>

<details>
<summary>xmtp_db/src/encrypted_store/refresh_state.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 329](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/refresh_state.rs#L329): Unchecked downcast of `u32` originator IDs to `i32` when creating missing `RefreshState` rows: `originator_id: *originator as i32`. If an originator ID exceeds `i32::MAX`, this wraps to negative and corrupts data. Validate range before casting or store as `i64` to be safe. <b>[ Low confidence ]</b>
- [line 343](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/refresh_state.rs#L343): Unchecked casts from DB fields to unsigned when constructing the result cursors: `Cursor::new(state.sequence_id as u64, state.originator_id as u32)`. If stored values are negative, they will wrap. Add validation or SQL guards (`>= 0`) before casting. <b>[ Low confidence ]</b>
- [line 804](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/refresh_state.rs#L804): Use of `unwrap()` on results from `conn.get_refresh_state(&id, kind, Originators::MLS_COMMITS)` and `conn.get_last_cursor_for_originators(&id, kind, &[0])` can cause a runtime panic if these database operations return `Err`. The test does not handle errors explicitly. <b>[ Out of scope ]</b>
- [line 808](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_db/src/encrypted_store/refresh_state.rs#L808): Possible panic due to unchecked indexing into the result of `conn.get_last_cursor_for_originators(&id, kind, &[0]).unwrap()` with `[0]`. If the returned `Vec<Cursor>` is empty under any realistic condition, `...[0]` will panic at runtime. The test does not prove or guard that the vector is non-empty. <b>[ Test / Mock code ]</b>
</details>

<details>
<summary>xmtp_mls/src/subscriptions/process_welcome.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 277](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_mls/src/subscriptions/process_welcome.rs#L277): Potential lossy/wrapping cast when constructing `Cursor` for filtered `NewStored` results: `Cursor::new(id as SequenceId, originator as OriginatorId)` casts from `i64` to unsigned/narrower integer types without bounds checks. If `maybe_sequence_id` or `maybe_originator` are negative or exceed the target type’s max (e.g., `i64` > `u32::MAX` for originator), the `as` casts will wrap/truncate, producing an invalid `Cursor`. This can corrupt dedup/tracking semantics (e.g., mislabeling the ignored ID), and yield hard-to-diagnose behavior. Guard against negatives and out-of-range values before constructing the `Cursor`, or change stored types to match `SequenceId`/`OriginatorId`. If invalid, prefer returning `Ignore` or an explicit error. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_mls/src/subscriptions/stream_conversations.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 403](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_mls/src/subscriptions/stream_conversations.rs#L403): `filter_welcome` constructs a `Cursor` using unchecked `as` casts from `i64` optionals: `Cursor::new(id as SequenceId, originator as OriginatorId)`. Negative or out-of-range values will wrap, causing incorrect cursor keys in `known_welcome_ids`. Use fallible conversions (`try_into()`) or validate ranges before casting; only insert when both values are valid. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>xmtp_proto/src/types/global_cursor.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 42](https://github.com/xmtp/libxmtp/blob/49f202f82e2b2c7772bd0b8228aa955a82b07437/xmtp_proto/src/types/global_cursor.rs#L42): `GlobalCursor::difference` claims to "get a map of all keys that are in `self` but not in `other`", but the implementation includes entries where the key exists in `other` with an equal (or lower) sequence (`sid >= other.get(oid)`). This returns more than just keys absent from `other`, and even includes equal entries. If the intended contract is strictly keys absent from `other`, the filter should test for absence (e.g., `!other.inner.contains_key(oid)`), or if the intent is to include strictly newer entries, use `sid > other.get(oid)`. As implemented, results are incorrect per the documented behavior. <b>[ Low confidence ]</b>
</details>

</details>


insipx:
This PR is part of a stack of 9 bookmarks:

1. `trunk()`
1. [push-tkyvzsqvnkyz](https://github.com/xmtp/libxmtp/pull/2863)
1. [insipx/orphaned-envelope](https://github.com/xmtp/libxmtp/pull/2856)
1. [insipx/seal-cursor](https://github.com/xmtp/libxmtp/pull/2917)
1. **push-mmsvvuosvwyr ← this PR**

---

PR #2937, opened 2025-12-11
use offline resolver in query group messages

insipx:
### Use offline resolver to order group message queries in `xmtp_api_d14n` without network dependency resolution
Introduce `OfflineResolver` alias and add `D14nCombinatorExt::offline_ordered` to perform ordering without network resolution; update `ordered_query::OrderedQuery` to support an `offline` mode and switch MLS group message queries to use offline ordering.

#### 📍Where to Start
Start with the `mls` query changes that switch to offline ordering in [mls.rs](https://github.com/xmtp/libxmtp/pull/2937/files#diff-b41f084fe8fd9295035f4b6d73b1318093b6475f1f923e681a880e6ba94a2f20), then review the constructor and mode handling in [ordered_query.rs](https://github.com/xmtp/libxmtp/pull/2937/files#diff-a7f3188e32afea719721f5098f8f9cbde6c110942050e65a7abf44fd2218d0fe), and finally the new combinator in [combinators.rs](https://github.com/xmtp/libxmtp/pull/2937/files#diff-03738f42857f573af59205f3949b8c895ae33c91618506bc8af2478d8033cde6).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized a2b5821.

PR #2933, opened 2025-12-10
proptest d14n-ordered stream

insipx:
### Add proptest regression for ordered stream and expose `crate::protocol::test` during tests to validate stream ordering d14n
Add a regression seed file for the ordered stream property test and introduce a test-only re-export `crate::protocol::test`. Implement a proptest `orders_stream_and_ices_missing` in [xmtp_api_d14n/src/queries/stream/ordered.rs](https://github.com/xmtp/libxmtp/pull/2933/files#diff-59cd7abcfecc1658fc695f54e7da04f764862629dabf0aa0e4fafc684c6e732e) that verifies topic clock dominance and asserts non-empty icebox when dependencies are missing.

#### 📍Where to Start
Start with the `orders_stream_and_ices_missing` property test in [xmtp_api_d14n/src/queries/stream/ordered.rs](https://github.com/xmtp/libxmtp/pull/2933/files#diff-59cd7abcfecc1658fc695f54e7da04f764862629dabf0aa0e4fafc684c6e732e).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 0c8cf67.

PR #2922, opened 2025-12-09
make icebox queries batch

insipx:
modifies the icebox queries to work with lists of envelopes which may have children in the icebox

PR #2921, opened 2025-12-09
internally resolve envelopes in causal sort

insipx:
if there's an envelope sequence like:

```
D -> C -> B -> A
```
where `->` indicates a dependency relationship, i.e `A -> B` means "A depends on B"

a list like:
```
C A B D
```

Would return `C` and `D` as missing without this fix, since the topic cursor did not yet contain necessary dependencies `A` and `B` before processing `C`, which `D` then depends on. I thought this would be a non-issue, but in reality it made ordering deeply nested chains of dependencies much more difficult. in this PR, on every dependency we can resolve, a new functions checks if any of the dependencies in `Missed` can now be processed

PR #2940, opened 2025-12-11
dont start tonic stream if topics are empty

### LINKED ISSUE #2924: dont start a stream connection on empty topic lists in d14n
we could return a "fake" stream instead, so that we save a tcp connection. we don't want to get anything in a stream for an empty topics list

insipx:
fixes #2924 

lots of d14n tests are failing, esp streaming ones, because the streams implicitly subscribe to every single message in every single topic if the topic list is empty. this adds a `fake_stream` fn to the client which returns a stream that type-checks exactly as `stream`, except the underlying stream always returns `Poll::Pending` and is not connected to the network.

I considered to instead do this as a `Client` middleware, but it would have been too janky to decode the body bytes into protobuf and check for empty topics. Ideally I would like a way to start a "fake" stream w/o modifying every `Client` implementation, but this is OK for now

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8624,6 +8624,7 @@ dependencies = [
  "getrandom 0.3.4",
  "h2 0.4.12",
  "http 1.4.0",
+ "http-body-util",
  "hyper 1.8.1",
  "hyper-util",
  "pbjson-types",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -154,6 +154,7 @@ web-sys = "0.3"
 web-time = "1.1"
 zeroize = "1.8"
 smallvec = { version = "1.15", features = ["union"] }
+http-body-util = { version = "0.1" }
 
 # Internal Crate Dependencies
 bindings-wasm = { path = "bindings_wasm" }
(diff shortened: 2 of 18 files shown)


insipx:
This PR is part of a stack of 14 bookmarks:

1. `trunk()`
1. [insipx/causal-fix](https://github.com/xmtp/libxmtp/pull/2921)
1. [insipx/batch-queries](https://github.com/xmtp/libxmtp/pull/2922)
1. [insipx/seal-cursor](https://github.com/xmtp/libxmtp/pull/2917)
1. [insipx/paged-stream](https://github.com/xmtp/libxmtp/pull/2931)
1. [insipx/ordered-stream](https://github.com/xmtp/libxmtp/pull/2932)
1. **insipx/no-empty-streams ← this PR**
1. [insipx/ordered-stream-test](https://github.com/xmtp/libxmtp/pull/2933)
1. [push-pnvnmqsyqqqy](https://github.com/xmtp/libxmtp/pull/2936)
1. [push-tpmlvrluopvt](https://github.com/xmtp/libxmtp/pull/2937)

---

PR #2941, opened 2025-12-11
dont start tonic stream if topics are empty

### LINKED ISSUE #2924: dont start a stream connection on empty topic lists in d14n

insipx:
fixes #2924 

bunch of tests were failing for d14n ordered streams caused by this


insipx:

1. `trunk()`
1. **insipx/fake-streams ← this PR**

---

PR #2946, opened 2025-12-12
support d14n in external stream process message/welcome proxies

### LINKED ISSUE #2565: support processing d14n messages in external process proxies
this function: https://github.com/xmtp/libxmtp/blob/43cb69fb8d2d925fc7985f1cc7420a314d50441c/xmtp_mls/src/subscriptions/mod.rs#L232

For both welcomes/d14n needs to know about d14n envelopes and correctly extract group messages with the correct originator ids. this can be done by chaining the extractor as a tuple:
```rust
let mut extractor = (V3, D14n)
```

must ensure that v3 is processed first then d14n. might be good to extract error handling into api_d14n rather than leaving it in xmtp_mls, so that versioning has a single place of truth for v3/d14n

insipx:
closes #2565 

this is a breaking change since it changes the API of `process_streamed_{message|welcome}`

in d14n, this function may return an empty `Vec` if the message was iceboxed instead of returned. It is also possible it returns more than 1 message as well, although currently I think messages in the stream are singular.

this mostly effects the notification code part of an app. In practice, usually only one envelope will be present in the d14n subscription response. Although, it would possible for this to change in the future and silently break clients if not handled correctly now.

### DIFF
--- a/xmtp_mls/src/groups/subscriptions.rs
+++ b/xmtp_mls/src/groups/subscriptions.rs
@@ -1,43 +1,104 @@
 use super::MlsGroup;
 use crate::{
     context::XmtpSharedContext,
+    cursor_store::SqliteCursorStore,
     subscriptions::{
         Result, SubscribeError,
+        d14n_compat::{V3OrD14n, decode_group_message},
         process_message::{ProcessFutureFactory, ProcessMessageFuture},
-        stream_messages::{MessageStreamError, StreamGroupMessages},
+        stream_messages::StreamGroupMessages,
     },
 };
-use futures::{Stream, StreamExt};
-use prost::Message;
+use futures::{FutureExt, Stream, StreamExt, TryStreamExt, future, stream as future_stream};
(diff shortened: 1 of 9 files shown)


insipx:

1. `trunk()`
1. **push-syomxrnuqsnk ← this PR**

---

xmtp_mls/src/groups/subscriptions.rs
```diff
@@ -1,43 +1,104 @@
...
+                            };
+                            SubscribeError::dyn_err(err)
+                        })
+                    })
+                    .try_collect()?;
+                Ok(s)
+            }
+        }?;
+
+        let mut out = Vec::new();
```
codabrink: We should probably start allocating for size when we know it using `Vec::with_capacity`. Pretty sure this Vec gets moved after it's created during resizing. Preaching to myself too here. 😄

insipx: fair enough, fixed

PR #2952, opened 2025-12-15
move content type fallback text to ffis

mchenani:
### Expose leave request fallback text via `bindings_ffi::mls::leave_request_fallback` and set `xmtp_content_types::leave_request::LeaveRequestCodec::encode` to populate `EncodedContent.fallback`
Add `LeaveRequestCodec::FALLBACK_TEXT` and use it in `encode` to set `EncodedContent.fallback`; add a UniFFI function `leave_request_fallback()` that returns the fallback text via FFI in [bindings_ffi/src/mls.rs](https://github.com/xmtp/libxmtp/pull/2952/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f) and [xmtp_content_types/src/leave_request.rs](https://github.com/xmtp/libxmtp/pull/2952/files#diff-67d8cf386920d12a812ecb868102e2c1c873c3fcc8f2e3925831ed8af0e852d8).

#### 📍Where to Start
Start with the UniFFI export `leave_request_fallback()` in [bindings_ffi/src/mls.rs](https://github.com/xmtp/libxmtp/pull/2952/files#diff-3a24c3e76565487a710ac9863ac05160128f4f90892e07849b555a6de43a6e8f), then review `LeaveRequestCodec::FALLBACK_TEXT` and its use in `encode` in [xmtp_content_types/src/leave_request.rs](https://github.com/xmtp/libxmtp/pull/2952/files#diff-67d8cf386920d12a812ecb868102e2c1c873c3fcc8f2e3925831ed8af0e852d8).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 932d1c1.

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -3139,6 +3139,12 @@ pub fn decode_leave_request(bytes: Vec<u8>) -> Result<FfiLeaveRequest, GenericEr
         .map_err(|e| GenericError::Generic { err: e.to_string() })
 }
 
+// LeaveRequest FFI fallback function
+#[uniffi::export]
+pub fn leave_request_fallback() -> String {
+    LeaveRequestCodec::FALLBACK_TEXT.to_string()
+}
+
 #[uniffi::export]
 pub fn decode_group_updated(bytes: Vec<u8>) -> Result<FfiGroupUpdated, GenericError> {
     let encoded_content = EncodedContent::decode(bytes.as_slice())
--- a/xmtp_content_types/src/leave_request.rs
+++ b/xmtp_content_types/src/leave_request.rs
@@ -18,6 +18,11 @@ impl LeaveRequestCodec {
     pub const MINOR_VERSION: u32 = 0;
 }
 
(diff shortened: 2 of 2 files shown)


mchenani:
@cameronvoell we had this discussion last week, I made this POC PR to check if this is what we want?
1- since we don't have a field for fallback texts in all codecs, that makes it messy since we need an extra function per codec to extract the text, we can make it cleaner by either having a single function to return the associated text for each content type.
2- we already have the fallback texts hardcoded in the sdks, so I would like to extract those if we need to do that as a practice.
3- Since we're having this conversation about the fallback texts, I would like to make it easier for integrators to give them an option to make it easier to translate this texts, so instead of giving them a hardcoded text, I would like to introduce a struct to hold not only the predefined english fallback text, but also gives them a code or a key to help them make a translation based on those keys, for example:
ContentTypeFallbackText{
Key: "LeaveRequest-PendingRemove",
Lan: "En",
Value: "use has requested leaving the group"
}
And so on.
wdyt?

PR #2968, opened 2025-12-17
fix originator sequence id order in icebox

insipx:
on diesel selects, the sequence and originator ids would be swapped

### DIFF
--- a/xmtp_db/migrations/2025-11-25-213223-0000_make_icebox_depending_fields_non_null/up.sql
+++ b/xmtp_db/migrations/2025-11-25-213223-0000_make_icebox_depending_fields_non_null/up.sql
@@ -7,10 +7,10 @@ DELETE FROM icebox;
 
 -- Step 1: Create the new icebox_dependencies table
 CREATE TABLE icebox_dependencies (
-    envelope_sequence_id BIGINT NOT NULL,
     envelope_originator_id BIGINT NOT NULL,
-    dependency_sequence_id BIGINT NOT NULL,
+    envelope_sequence_id BIGINT NOT NULL,
     dependency_originator_id BIGINT NOT NULL,
+    dependency_sequence_id BIGINT NOT NULL,
     PRIMARY KEY (envelope_originator_id, envelope_sequence_id, dependency_originator_id, dependency_sequence_id),
     -- when an envelope is deleted, also delete its dependency records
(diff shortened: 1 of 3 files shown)


xmtp_db/src/encrypted_store/icebox.rs
```diff
@@ -28,7 +28,9 @@ mod types;
 #[diesel(primary_key(originator_id, sequence_id))]
 #[diesel(belongs_to(crate::group::StoredGroup, foreign_key = group_id))]
 pub struct Icebox {
+    #[diesel(column_name = "originator_id")]
```
tylerhawkes: why add the overrides now? Don't they match what is already generated?

xmtp_db/migrations/2025-11-25-213223-0000_make_icebox_depending_fields_non_null/up.sql
```diff
@@ -7,10 +7,10 @@ DELETE FROM icebox;
 
 -- Step 1: Create the new icebox_dependencies table
 CREATE TABLE icebox_dependencies (
-    envelope_sequence_id BIGINT NOT NULL,
     envelope_originator_id BIGINT NOT NULL,
-    dependency_sequence_id BIGINT NOT NULL,
+    envelope_sequence_id BIGINT NOT NULL,
     dependency_originator_id BIGINT NOT NULL,
+    dependency_sequence_id BIGINT NOT NULL,
```
tylerhawkes, line 13: This file should be in the 1.7 release, so we should make sure this makes it into the release, otherwise we shouldn't touch old migrations.

xmtp_db/src/encrypted_store/icebox.rs
insipx: used while testing, but also wanted to be "extra sure". In reality they didn't fix the issue anyway

xmtp_db/migrations/2025-11-25-213223-0000_make_icebox_depending_fields_non_null/up.sql
insipx: yeah I asked @cameronvoell he said it would be OK to cherry-pick into 1.7.0 if it's small

tylerhawkes (approved):

PR #2971, opened 2025-12-17
Add remote attachment encryption methods

rygine:
### Add remote attachment encryption methods and expose Node/WASM APIs for encrypting and decrypting attachments using AES-256-GCM with HKDF-SHA256 in [remote_attachment.rs](https://github.com/xmtp/libxmtp/pull/2971/files#diff-9204bc90bafeaf70f166a6c45a144bbacae90349fc36fa93d7f89683e4e74cac) and bindings
Introduce core encryption utilities and a `remote_attachment` flow with `EncryptedAttachment`; expose `encrypt_attachment` and `decrypt_attachment` in Node and WASM bindings; validate `content_length` to fit in `u32`; add tests for cross-compatibility and error cases.

#### 📍Where to Start
Start with the core flow in `remote_attachment.encrypt_attachment` and `remote_attachment.decrypt_attachment` in [remote_attachment.rs](https://github.com/xmtp/libxmtp/pull/2971/files#diff-9204bc90bafeaf70f166a6c45a144bbacae90349fc36fa93d7f89683e4e74cac), then review the Node exports in [bindings_node/src/content_types/remote_attachment.rs](https://github.com/xmtp/libxmtp/pull/2971/files#diff-6b9d793d8c67322d0d2557dded41bef74983cf826275f36a7d1f6b94245269e2) and the WASM exports in [bindings_wasm/src/content_types/remote_attachment.rs](https://github.com/xmtp/libxmtp/pull/2971/files#diff-00e8ba5ba639bc99f49e101d55f520f0c638e2445a70c6ae5b0bc6a0f7e99c3c).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 81e3d7f.

### DIFF
--- a/xmtp_content_types/src/encryption.rs
+++ b/xmtp_content_types/src/encryption.rs
@@ -34,7 +34,7 @@ pub fn encrypt(plaintext: &[u8], secret: &[u8]) -> Result<EncryptedPayload, Code
     let nonce: [u8; AES_GCM_NONCE_SIZE] = xmtp_common::rand_array();
 
     // Derive AES-256 key using HKDF-SHA256
-    let key = derive_key(secret, &salt)?;
+    let key = derive_key(secret, &salt).map_err(CodecError::Encode)?;
 
     // Create cipher
     let cipher = Aes256Gcm::new_from_slice(&key)
@@ -70,7 +70,7 @@ pub fn decrypt(encrypted: &EncryptedPayload, secret: &[u8]) -> Result<Vec<u8>, C
     }
 
     // Derive AES-256 key using HKDF-SHA256
-    let key = derive_key(secret, &encrypted.salt)?;
+    let key = derive_key(secret, &encrypted.salt).map_err(CodecError::Decode)?;
 
     // Create cipher
     let cipher = Aes256Gcm::new_from_slice(&key)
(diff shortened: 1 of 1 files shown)


bindings_node/src/content_types/remote_attachment.rs
```diff
@@ -88,3 +95,81 @@ pub fn encode_remote_attachment(remote_attachment: RemoteAttachment) -> Result<E
...
+      content_length,
+      filename: encrypted.filename,
+    })
+  }
+}
+
+/// Encrypts an attachment for storage as a remote attachment.
+#[napi]
+pub fn encrypt_attachment(attachment: Attachment) -> Result<EncryptedAttachment> {
+  xmtp_encrypt_attachment(attachment.into())
```
cameronvoell, line 161: just commenting here for knowledge share that I initially tried [adding encryption](https://github.com/xmtp/libxmtp/pull/1609/commits/92eeced9145df32d244707c6b4f7a45ceb65cb1f) for mobile bindings as well, but it turned out that it was noticeably slower calling encrypt from rust then doing the encryption in native android. I think that this specifically applies to passing large in memory values from JVM to rust. though I think doing native swift might have been slightly faster as well. 

I dont think this applies for napi or wasm, but I'll keep an eye out for documenting in the code base, if this is a difference that persists long term, that napi/wasm use rust encryption functions but mobile uses sdk/native side. Think this is definitely makes sense to have this here even if just for napi wasm though 👍. 

Also it's possible we can get around those performance issues on android by sending a reference to a file instead of passing in memory, or something else I missed, but think we should be good in the mean time.

rygine: if performance is an issue, we can definitely move back to native encryption. just wanted something i could share across platforms.

re: your comment below about compatibility, i added tests in the node and wasm bindings for compatibility with the current JS content type.

PR #2978, opened 2025-12-28
Content type updates and bug fixes

rygine:
# Summary

### New features

- Added ability to filter messages by `Actions`, `Intent`, and `MultRemoteAttachments` content types
- Added ability to update `app_data` permission policy
- Added `bindings_wasm_macros` crate to facilitate numeric enums in the WASM bindings
- Added consent updates to preferences stream (node + WASM only)

### Quality of life improvements

- Always use `ConversationType` enum when referencing the conversation type (node + WASM)
- Added more test commands to WASM bindings
- Added `metadata_field_name` function to extract field names from `MetadataField` enum (node + WASM)
- Refactored most WASM enums to be numeric to match the node bindings
- Renamed `*_content_type` functions to `content_type_*`
- Renamed `CreateDMOptions` to `CreateDmOptions` in WASM bindings to match node bindings
- Renamed `UserPreference` to `UserPreferenceUpdate` in WASM bindings to match node bindings

### Fixes

- Fixed `EncodedContent` type in WASM bindings (the `parameters` field was marked as optional)
- Fixed `InboxState` and `KeyPackageStatus` serialization to handle large numbers
- Added `is_secure` argument to `get_inbox_id_for_identifier` in WASM bindings to match node bindings
- Added missing `AppData` variant to `MetadataField` enum, renamed `ImageUrlSquare` to `GroupImageUrlSquare` to match internal enum
- Fixed `i64` precision loss in node bindings
- Fixed ignored `exclude_content_types` query arg values when listing messages

### Maintenance

- Removed unused structs `MarkdownContent` and `TextContent` (WASM)
- Removed automatic `kind` query arg when listing messages (node + WASM)

### Open questions

- Should the preferences stream expose consent updates?
- Should `allowed_states` be exposed as an option when listing conversations?
- Should the original `find_messages` be removed in the bindings?
- Should the `app_data` metadata field default to admin only?
- Should `find_messages_with_reactions` be removed in the bindings?
- Should the `exclude_content_types` message query arg overwrite the defaults or merge them into the passed value?
- Should `ListConversationsOptions` include a sort `direction`?

### DIFF
--- a/bindings_ffi/src/mls.rs
+++ b/bindings_ffi/src/mls.rs
@@ -1448,6 +1448,7 @@ pub struct FfiPermissionPolicySet {
     pub update_group_description_policy: FfiPermissionPolicy,
     pub update_group_image_url_square_policy: FfiPermissionPolicy,
     pub update_message_disappearing_policy: FfiPermissionPolicy,
+    pub update_app_data_policy: FfiPermissionPolicy,
 }
 
 impl From<PreconfiguredPolicies> for FfiGroupPermissionsOptions {
@@ -1488,6 +1489,10 @@ impl TryFrom<FfiPermissionPolicySet> for PolicySet {
             MetadataField::MessageDisappearInNS.to_string(),
             policy_set.update_message_disappearing_policy.try_into()?,
         );
+        metadata_permissions_map.insert(
+            MetadataField::AppData.to_string(),
+            policy_set.update_app_data_policy.try_into()?,
+        );
 
         Ok(PolicySet {
             add_member_policy: policy_set.add_member_policy.try_into()?,
@@ -1505,6 +1510,7 @@ pub enum FfiMetadataField {
     GroupName,
     Description,
     ImageUrlSquare,
+    AppData,
 }
 
 impl From<&FfiMetadataField> for MetadataField {
@@ -1513,6 +1519,7 @@ impl From<&FfiMetadataField> for MetadataField {
             FfiMetadataField::GroupName => MetadataField::GroupName,
             FfiMetadataField::Description => MetadataField::Description,
             FfiMetadataField::ImageUrlSquare => MetadataField::GroupImageUrlSquare,
+            FfiMetadataField::AppData => MetadataField::AppData,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


bindings_wasm/src/conversation.rs
```diff
@@ -75,30 +77,23 @@ impl From<SendMessageOpts> for xmtp_mls::groups::send_message_opts::SendMessageO
...
-      },
+      conversation_type: metadata.conversation_type.into(),
     }
   }
 }
 
-#[derive(Clone, Serialize, Deserialize, Tsify)]
-#[tsify(into_wasm_abi, from_wasm_abi)]
-#[serde(rename_all = "camelCase")]
+#[wasm_bindgen_numbered_enum]
```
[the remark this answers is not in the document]
rygine: this is all handled by this trait and associated macro

bindings_wasm_macros/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+    // Generate match arms for deserialization
+    let deser_arms = variants.iter().map(|(variant_name, _, disc)| {
+        quote! {
+            #disc => Ok(Self::#variant_name)
+        }
+    });
+
+    let expanded = quote! {
+        #[::wasm_bindgen::prelude::wasm_bindgen]
+        #[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
rygine, line 85: not sure all of these traits are necessary here

codabrink: I think Copy is good here actually. For small enough types in most cases it can actually be faster than non-copy.

bindings_wasm_macros/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
codabrink, line 1: I think putting this in it's own crate was a good idea. The existing derive macros crate is getting too inflated and probably needs to be broken out into it's own pieces too.

rygine: i originally tried to add the macro in `bindings_wasm`, but i learned that rust requires procedural macros be in their own crate. 😅

codabrink: Yeah I love a lot of things about Rust. Rust's macros is not one of them.

PR #3012, opened 2026-01-15
Use String for ContentType in Archives

codabrink:
proto https://github.com/xmtp/proto/pull/316

### DIFF
--- a/xmtp_db/src/encrypted_store/group_message/convert.rs
+++ b/xmtp_db/src/encrypted_store/group_message/convert.rs
@@ -11,6 +11,11 @@ impl TryFrom<GroupMessageSave> for StoredGroupMessage {
         let kind = value.kind().try_into()?;
         let delivery_status = value.delivery_status().try_into()?;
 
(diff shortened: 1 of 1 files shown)


xmtp_db/src/encrypted_store/group_message/convert.rs
```diff
@@ -79,20 +82,15 @@ impl TryFrom<ContentTypeSave> for ContentType {
             ContentTypeSave::Reply => Self::Reply,
             ContentTypeSave::Text => Self::Text,
             ContentTypeSave::TransactionReference => Self::TransactionReference,
-            ContentTypeSave::Unknown => Self::Unknown,
-            ContentTypeSave::Unspecified => {
-                return Err(ConversionError::Unspecified("content_type"));
-            }
-        };
-        Ok(content_type)
+            _ => Self::Unknown,
```
codabrink, line 85: This matches the `From<String>` for Content type impl.

neekolas (approved):

PR #3025, opened 2026-01-16
add new client function to get total number of updates for an inbox

### LINKED ISSUE #3024: Attach total number of inbox log updates to `InboxState` return
`InboxState` returns the total current state of updates, but not the # of updates that lead to that state. The total # of updates is useful for callers to take action if nearing the 256 limit

insipx:
fixes #3024

### DIFF
--- a/crates/xmtp_mls/src/client.rs
+++ b/crates/xmtp_mls/src/client.rs
@@ -392,6 +392,39 @@ where
         Ok(state)
     }
 
+    /// Get the total number of inbox updates for `inbox_ids`. `refresh_from_network` will force
+    /// a network refresh. May still access network if an inbox_id does not yet exist in the local
+    /// cache.
+    pub async fn fetch_inbox_updates_count(
+        &self,
+        refresh_from_network: bool,
+        inbox_ids: Vec<InboxIdRef<'_>>,
+    ) -> Result<HashMap<InboxId, u32>, ClientError> {
+        let conn = self.context.db();
+        if refresh_from_network {
+            load_identity_updates(self.context.api(), &conn, &inbox_ids).await?;
+        }
+        let inbox_id_strs = inbox_ids.to_vec();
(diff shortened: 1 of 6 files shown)


crates/xmtp_mls/src/client.rs
```diff
@@ -392,6 +392,39 @@ where
...
+        refresh_from_network: bool,
+    ) -> Result<u32, ClientError> {
+        let conn = self.context.db();
+        let inbox_id = self.inbox_id();
+        if refresh_from_network {
+            load_identity_updates(self.context.api(), &conn, &[inbox_id]).await?;
+        }
+        let counts = conn.count_inbox_updates(&[inbox_id])?;
+        Ok(counts.get(inbox_id).copied().unwrap_or(0) as u32)
+    }
```
rygine, line 426: should this just call the above function?

rygine:
should this info just be included standard in [inbox_state](https://github.com/xmtp/libxmtp/blob/main/crates/xmtp_mls/src/client.rs#L359-L373)?

insipx:
forgot, still need to expose to ffi/node/wasm

insipx:
@rygine 

> should this info just be included standard in [inbox_state](https://github.com/xmtp/libxmtp/blob/main/crates/xmtp_mls/src/client.rs#L359-L373)?

This was my first instinct as well, but doing a ctrl-f for association state references and there are 206. its a quicker win to make it its own function IMO, esp since its only a simple sqlite query. happy to reconsider if other feel strongly it should be in the `AssociationState`

rygine:
> This was my first instinct as well, but doing a ctrl-f for association state references and there are 206. its a quicker win to make it its own function IMO, esp since its only a simple sqlite query. happy to reconsider if other feel strongly it should be in the `AssociationState`

no strong feelings, more just curiosity. as is looks good to me!

PR #3053, opened 2026-01-23
Device Sync Updates

codabrink:
### Add device sync APIs and processing, switch clients to device sync groups, and require server URL with PIN-based archives across Rust core and WASM/Node/mobile bindings
Introduce explicit device sync flows with PIN-based archives and server URL handling, add message attempt/state tracking with a 3-attempt cap, replace history sync with device sync in client methods and tests, and expose device sync operations in WASM, Node, and mobile bindings. Core updates include new listing/processing by PIN, paging of sync messages, refined retryability for `DeviceSyncError`, and proto changes adding `server_url`.

#### 📍Where to Start
Start with device sync message flow in `SyncWorker`—review `process_message`, `send_archive`, `send_sync_archive`, `process_archive_with_pin`, and `list_available_archives` in [worker.rs](https://github.com/xmtp/libxmtp/pull/3053/files#diff-4d3b29d2fdae905c9eebffa4bbee45dced10e61c0609ced6333b251d261b4ab4), then follow the new DB paging/state APIs in [processed_device_sync_messages.rs](https://github.com/xmtp/libxmtp/pull/3053/files#diff-b06a423735a9bb0b7f7dd0e5cef7ee6249f640369dcc7ae6435e8841d08fe4aa) and the client entrypoints in [client.rs](https://github.com/xmtp/libxmtp/pull/3053/files#diff-74086c32ace91ebbc5bca31e480e4e33cfd9f03bb70c4a8fdbcca16ddfcdf52f).

#### Changes since #3053 opened

- Replaced `inner_client()` method calls with direct `inner_client` field access across all `DeviceSync` methods [ab0536e]
- Replaced `Client` import with `RustXmtpClient` and added `std::sync::Arc` import in the `bindings.wasm.device_sync` module [ab0536e]
- Added explicit discriminant values to `BackupElementSelectionOption` enum [ab0536e]
- Reordered import statements in the `bindings/wasm` module [70446f5]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized a386e3b. 14 files reviewed, 12 issues evaluated, 12 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>bindings/mobile/src/mls/device_sync/mod.rs — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 74](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/mobile/src/mls/device_sync/mod.rs#L74): The `create_archive` function returns `BackupMetadata` (the internal type from `xmtp_archive`) instead of `FfiBackupMetadata`. This is inconsistent with `archive_metadata` which returns `FfiBackupMetadata`. If this function is exported via UniFFI, `BackupMetadata` lacks the `#[derive(uniffi::Record)]` attribute and cannot be serialized for FFI. The return type should likely be `FfiBackupMetadata` with a conversion: `Ok(BackupMetadata::from_metadata_save(metadata, BACKUP_VERSION).into())` <b>[ Already posted ]</b>
- [line 171](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/mobile/src/mls/device_sync/mod.rs#L171): The `check_key` function uses a hardcoded value `32` for the minimum key length check, but uses `ENC_KEY_SIZE` for truncation and in the error message. If `ENC_KEY_SIZE` differs from 32, this creates inconsistent behavior: either the validation allows keys that are too short (if `ENC_KEY_SIZE > 32`), or rejects valid keys that meet the actual requirement (if `ENC_KEY_SIZE < 32`). The check on line 171 should use `ENC_KEY_SIZE` instead of the hardcoded `32`. <b>[ Already posted ]</b>
- [line 198](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/mobile/src/mls/device_sync/mod.rs#L198): The `From<BackupMetadata> for FfiBackupMetadata` conversion at line 198 uses `filter_map(|selection| selection.try_into().ok())` which silently drops any `BackupElementSelection` values that fail to convert to `FfiBackupElementSelection`. When users call `archive_metadata` to inspect an archive, they may receive an incomplete list of elements with no indication that some were dropped, potentially leading to confusion about archive contents. <b>[ Low confidence ]</b>
- [line 198](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/mobile/src/mls/device_sync/mod.rs#L198): The `filter_map(|selection| selection.try_into().ok())` pattern silently discards any `BackupElementSelection` elements that fail conversion to `FfiBackupElementSelection`. This could cause data loss where the caller receives an incomplete `elements` list without any indication that elements were dropped. If certain element types are unsupported in the FFI layer, this should either error out or be explicitly documented/logged. <b>[ Already posted ]</b>
</details>

<details>
<summary>bindings/node/src/device_sync.rs — 0 comments posted, 5 evaluated, 5 filtered</summary>

- [line 27](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/node/src/device_sync.rs#L27): In the `From<ArchiveOptions> for BackupOptions` implementation, `n.get_i64().0` silently discards the overflow indicator boolean. When JavaScript passes a `BigInt` value larger than `i64::MAX` or smaller than `i64::MIN`, `get_i64()` returns `(truncated_value, false)` where `false` indicates the conversion was lossy. The code ignores this and uses the potentially incorrect truncated value, which could silently corrupt `start_ns` and `end_ns` timestamp boundaries, causing the archive to include wrong data ranges. <b>[ Already posted ]</b>
- [line 27](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/node/src/device_sync.rs#L27): The `BigInt::get_i64()` method returns a tuple `(i64, bool)` where the boolean indicates whether the conversion was lossless. The code uses `.0` to extract only the `i64` value, ignoring the overflow indicator. If a `BigInt` value exceeds `i64::MAX` or is below `i64::MIN`, the timestamp will be silently truncated to an incorrect value, corrupting the `start_ns` or `end_ns` filter and potentially causing wrong data to be archived. <b>[ Already posted ]</b>
- [line 28](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/node/src/device_sync.rs#L28): Same silent truncation issue for `end_ns` - `BigInt` values outside `i64` range will be silently converted to incorrect values without any error or warning to the caller. <b>[ Already posted ]</b>
- [line 91](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/node/src/device_sync.rs#L91): The conversion from `BackupMetadata` to `ArchiveMetadata` silently drops any `BackupElementSelection` elements that fail to convert via `.filter_map(|selection| selection.try_into().ok())`. If any element's `try_into()` fails, it is silently discarded rather than propagating an error or logging a warning. This could result in data loss where callers receive an incomplete `elements` list without any indication that some elements were dropped during conversion. <b>[ Already posted ]</b>
- [line 120](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/node/src/device_sync.rs#L120): The key length check uses a hardcoded value `32` while the error message and `truncate` call use `ENC_KEY_SIZE`. If `ENC_KEY_SIZE` is not 32, the check and the error message would be inconsistent, and keys could be rejected or accepted incorrectly. The comparison on line 120 should use `ENC_KEY_SIZE` instead of the magic number `32` to ensure consistency. <b>[ Already posted ]</b>
</details>

<details>
<summary>bindings/wasm/src/device_sync.rs — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 105](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/wasm/src/device_sync.rs#L105): The `From<BackupMetadata>` implementation at lines 102-106 uses `filter_map(|selection| selection.try_into().ok())` which silently drops any `BackupElementSelection` variants that fail conversion. When `archive_metadata` is called, the returned `ArchiveMetadata.elements` may contain fewer elements than the actual archive contains, giving callers an incomplete view of the archive contents without any indication that data was omitted. <b>[ Already posted ]</b>
- [line 105](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/wasm/src/device_sync.rs#L105): The `filter_map(|selection| selection.try_into().ok())` pattern silently discards any `BackupElementSelection` elements that fail conversion to `BackupElementSelectionOption`. This could cause data loss where the caller receives an incomplete `elements` list without any indication that elements were dropped. If certain element types are unsupported in the WASM layer, this should either error out or be explicitly documented/logged. <b>[ Already posted ]</b>
- [line 138](https://github.com/xmtp/libxmtp/blob/a386e3b68dba87236c2caffbd18bdd065b5e2b06/bindings/wasm/src/device_sync.rs#L138): The length check uses hardcoded `32` but the error message and truncation use `ENC_KEY_SIZE`. If `ENC_KEY_SIZE` is greater than 32, keys with length between 32 and `ENC_KEY_SIZE` will pass the validation check but `truncate(ENC_KEY_SIZE)` will have no effect, returning a key shorter than `ENC_KEY_SIZE`. This could cause cryptographic failures downstream when the code expects exactly `ENC_KEY_SIZE` bytes. <b>[ Already posted ]</b>
</details>

</details>

### DIFF
--- a/bindings/wasm/src/device_sync.rs
+++ b/bindings/wasm/src/device_sync.rs
@@ -1,8 +1,8 @@
 use crate::client::{GroupSyncSummary, RustXmtpClient};
-use std::sync::Arc;
 use bindings_wasm_macros::wasm_bindgen_numbered_enum;
 use js_sys::Uint8Array;
 use serde::{Deserialize, Serialize};
+use std::sync::Arc;
 use tsify::Tsify;
 use wasm_bindgen::{JsError, prelude::wasm_bindgen};
 use xmtp_id::associations::DeserializationError;


crates/xmtp_mls/src/groups/device_sync/worker.rs
```diff
@@ -455,38 +467,29 @@ where
...
 
-            Ok::<_, DeviceSyncError>(true)
+            Ok::<_, DeviceSyncError>(())
         };
 
         // Acknowledge the sync request
-        if !acknowledge().await? {
-            return Ok(());
-        };
+        acknowledge().await?;
```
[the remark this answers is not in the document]
codabrink: Good point. Thanks.

crates/xmtp_mls/src/groups/device_sync/worker.rs
```diff
@@ -625,24 +686,6 @@ where
             return Ok(());
         }
 
-        tracing::info!("Inspecting sync payload.");
-
-        // Check if this reply was asked for by this installation.
-        if !self.is_reply_requested_by_installation(&reply).await? {
```
codabrink, line 631: Moved this check higher in the stack. This function assumes that you've already checked whether or not you want to import it.

crates/xmtp_mls/src/groups/device_sync/worker.rs
```diff
@@ -608,6 +647,29 @@ where
         Ok(false)
     }
 
+    pub async fn process_sync_payload_with_pin(&self, pin: &str) -> Result<(), DeviceSyncError> {
```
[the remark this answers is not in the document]
codabrink: It's intentional. I don't want to force network bloat on the user.

bindings/mobile/src/mls/device_sync/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        Ok(())
+    }
+
+    /// Manually send a sync archive to the sync group.
+    /// The pin will be later used as a reference when importing.
+    pub async fn send_sync_archive(
+        &self,
+        options: FfiArchiveOptions,
+        server_url: String,
+        pin: String,
```
rygine, line 31: WDYT of this function returning a generated ID used by `process_sync_archive`? i'm just thinking of the case where you can put in a duplicate pin.

bindings/mobile/src/mls/device_sync/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+            .device_sync_client()
+            .process_archive_with_pin(archive_pin.as_deref())
+            .await?;
+        Ok(())
+    }
+
+    /// List the archives available for import in the sync group.
+    /// You may need to manually sync the sync group before calling
+    /// this function to see recently uploaded archives.
+    pub fn list_available_archives(
```
rygine, line 56: what about adding an option to refresh from network? only syncing the sync group is a cumbersome task and calling syncAll to do this is not optimal.

bindings/wasm/src/device_sync.rs
```diff
@@ -0,0 +1,274 @@
...
+      exclude_disappearing_messages: value.exclude_disappearing_messages,
+    }
+  }
+}
+
+/// Selection of what elements to include in a backup
+#[derive(Clone, Serialize, Deserialize, Tsify)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+#[serde(rename_all = "camelCase")]
+pub enum BackupElementSelectionOption {
```
rygine, line 52: use the `#[wasm_bindgen_numbered_enum]` macro here so that an actual enum is used in the  SDK.

bindings/mobile/src/mls/device_sync/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        Ok(available.into_iter().map(Into::into).collect())
+    }
+
+    /// Archive application elements to file for later restoration.
+    pub async fn create_archive(
+        &self,
+        path: String,
+        opts: FfiArchiveOptions,
+        key: Vec<u8>,
+    ) -> Result<(), GenericError> {
```
rygine: this is a personal preference here, but (like the WASM bindings) i'd rather the create and import archive functions return/accept bytes when creating/importing rather than automatically writing/reading to/from files. this will allow the devs to decide how to store it.

bindings/mobile/src/mls/device_sync/mod.rs
codabrink: I'd like to avoid adding an automatic sync tax in an effort to keep the library nimble. If syncing the sync groups is cumbersome, we can add a function for that as well.

bindings/mobile/src/mls/device_sync/mod.rs
codabrink: I had that originally and opted for leaving the control of the pin to the integrator. Having two archives with the same pin generally isn't a problem per-se. When you import by pin, it'll restore the most recent archive.

rygine: it just seems like extra work for the integrator. also, why is a pin necessary when i can get a list of available archives? i don't see the pin exposed in `FfiAvailableArchive`. does that mean it's not possible to process an individual archive from this list?

bindings/mobile/src/mls/device_sync/mod.rs
codabrink: The benefit of taking a path or a file handle is we can stream to that, allowing us to support large files. Do you know how well the bindings handle stream types?

bindings/mobile/src/mls/device_sync/mod.rs
codabrink: The pin is the `request_id` field. I'll rename it to `pin` as it should be. Thanks for that catch. The pin is the key, and allows the integrator to use mnemonic keys or 0..1000 keys, or a custom key from the user, it leaves the design up to them, which I think is best. Generating a random number, stringifying it, and passing it to an arg really isn't that difficult.

bindings/mobile/src/mls/device_sync/mod.rs
rygine: fair point. i don't know how the bindings handle stream types.

rygine: on another note, WDYT of this function returning `FfiBackupMetadata`?

bindings/mobile/src/mls/device_sync/mod.rs
rygine: i think it would be nice to have a way to only sync the sync group

codabrink: Sounds good. I'll add it.

bindings/mobile/src/mls/device_sync/mod.rs
codabrink: Done

bindings/mobile/src/mls/device_sync/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+        let options: BackupOptions = opts.into();
+        let metadata = ArchiveExporter::export_to_file(options, db, path, &check_key(key)?)
+            .await
+            .map_err(DeviceSyncError::Archive)?;
+
+        Ok(BackupMetadata::from_metadata_save(metadata, BACKUP_VERSION))
+    }
+
+    /// Import a previous archive from file.
+    pub async fn import_archive(&self, path: String, key: Vec<u8>) -> Result<(), GenericError> {
```
rygine, line 85: WDYT of returning a report of sorts on successful import?

something like this:

```
pub struct ArchiveImportReport {
  elements_imported: u64 // or usize?
}
```

not sure if anything else would make sense, but the purpose would be to see if any elements weren't imported from the archive. i know it's possible that archives would overlap on some elements so this would allow some insight into that.

codabrink: Sure. Follow-up PR? This PR is getting kind of big.

rygine: for sure!

bindings/mobile/src/mls.rs
```diff
@@ -832,11 +825,10 @@ impl FfiXmtpClient {
...
     }
 
+    #[deprecated(note = "Please call `sync_all_device_sync_groups` instead. \
+        This function will be removed in a future version.")]
     pub async fn sync_preferences(&self) -> Result<FfiGroupSyncSummary, GenericError> {
-        let inner = self.inner_client.as_ref();
-        let summary = inner.sync_all_welcomes_and_history_sync_groups().await?;
-
-        Ok(summary.into())
+        self.sync_all_device_sync_groups().await
```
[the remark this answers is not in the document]
codabrink: We'll just manually sync the welcomes for now inline.

PR #3066, opened 2026-01-28
Add error codes

bennycode:
### Add error codes by introducing `ErrorCode` across crates and switching mobile FFI to `FfiError` that formats errors as `[code] message`
Introduce `ErrorCode` trait and derive macro, apply it to core error enums across crates, and route all mobile FFI errors through `FfiError` with `[code] message` formatting; add a small data-path change to include `creation_signature_kind` in `FfiInboxState` and expose `parse_xmtp_error` to extract codes.

#### 📍Where to Start
Start with the `ErrorCode` trait and derive in [lib.rs](https://github.com/xmtp/libxmtp/pull/3066/files#diff-328b382ffef73a0b3b84b33ab767b44ce725c3ba0a33c2c475cdf4fa26f4560c) and [error_code.rs](https://github.com/xmtp/libxmtp/pull/3066/files#diff-00e6d1180289a9778def0490bf444b7d5d4f484a62b4a9f5df2f176a3ba0dca8), then review the `FfiError` wrapper and `GenericError` changes in [lib.rs](https://github.com/xmtp/libxmtp/pull/3066/files#diff-52a26425218f1a79d7189b55938258637d4b5e3a2c0ece6be150f06631044b59).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 54c2dbe.


bindings/mobile/src/lib.rs
```diff
@@ -95,13 +114,121 @@ impl From<uniffi::UnexpectedUniFFICallbackError> for GenericError {
...
+/// Parse an error string like `[ErrorCode] message` into a structured error info.
+#[uniffi::export]
+pub fn parse_xmtp_error(message: String) -> FfiErrorInfo {
+    parse_error_message(&message)
+}
+
+// Implement From for all error types that GenericError supports.
+// NOTE: When adding a new error type with #[from] to GenericError,
+// also add it here to enable the ? operator with FfiError return types.
+macro_rules! impl_ffi_error_from {
```
tylerhawkes: Does a blanket impl work here instead of a macro and having to remember to add each type?

```rust
impl <T: Into<GenericError>> From<T> for FfiError {
    fn from(err: T) -> Self {
        Self::Error(err.into())
    }
}
```

bindings/node/src/lib.rs
```diff
@@ -21,31 +21,45 @@ xmtp_common::if_test! {
...
+/// } catch (e) {
+///   console.log(e.message); // "[ErrorType::Variant] error message"
+/// }
+/// ```
 #[derive(Debug)]
-pub struct ErrorWrapper<E>(E)
+pub struct ErrorWrapper<E>(pub E)
 where
-  E: std::error::Error;
+  E: std::error::Error + ErrorCode;
```
tylerhawkes: every constraint that is `E: std::error::Error + ErrorCode` can be simplified to `E: ErrorCode` because `ErrorCode` requires that it implements `std::error::Error`

bindings/wasm/src/lib.rs
```diff
@@ -18,12 +18,62 @@ pub mod signatures;
...
+/// try {
+///   await client.doSomething();
+/// } catch (e) {
+///   console.log(e.message); // "[ErrorType::Variant] error message"
+/// }
+/// ```
+#[derive(Debug)]
+pub struct ErrorWrapper<E>(pub E)
+where
+  E: std::error::Error + ErrorCode;
```
tylerhawkes: same as above, this can just be `E: ErrorCode`

bindings/wasm/src/lib.rs
```diff
@@ -18,12 +18,62 @@ pub mod signatures;
...
+
+impl<T: std::error::Error + ErrorCode> std::fmt::Display for ErrorWrapper<T> {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+    write!(f, "{}", self.0)
```
tylerhawkes: Instead of a macro here, I think it is better to do `self.0.fmt(f)`

bindings/wasm/src/lib.rs
```diff
@@ -18,12 +18,62 @@ pub mod signatures;
...
+{
+  fn from(err: T) -> ErrorWrapper<T> {
+    ErrorWrapper(err)
+  }
+}
+
+impl<T: std::error::Error + ErrorCode> From<ErrorWrapper<T>> for JsError {
+  fn from(e: ErrorWrapper<T>) -> JsError {
+    let code = e.0.error_code();
+    let message = e.0.to_string();
```
tylerhawkes: No need to allocate a string here, just change the next line to end with `code, e.0));`

crates/xmtp_common/src/error_code.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+impl<E: ErrorCode> ErrorCode for &E {
+    fn error_code(&self) -> &'static str {
+        (*self).error_code()
+    }
+}
+
+// Manual implementations for xmtp_cryptography errors.
```
tylerhawkes: These impls aren't necessarily needed. You can use the derive macro inside of xmtp_common if the generated code references the trait as coming from `xmtp_common`. I did that when implementing the `async_trait` macro.

bindings/node/src/client/create_client.rs
```diff
@@ -16,32 +16,31 @@ use xmtp_mls::identity::IdentityStrategy;
...
+    } else {
+      tracing_subscriber::registry()
+        .with(fmt::layer())
+        .with(filter)
+        .init();
+    }
+    Ok(())
+  }) {
+    Ok(()) => Ok(()),
+    Err(e) => Err(Error::from_reason(e.reason.clone())),
```
codabrink: `.map_err` is cleaner and can still be used like it was before. No need to match.

crates/xmtp_macro/src/lib.rs
```diff
@@ -419,3 +419,201 @@ impl Attributes {
         Err(meta.error("unknown attribute"))
     }
 }
+
+/// Derive macro for the `ErrorCode` trait.
```
codabrink, line 423: Probably not in this PR, but the macro crate is getting crowded. We should split it up soon.

tylerhawkes (approved):

PR #3072, opened 2026-01-30
add edit message

mchenani:
### Add edit message support across mobile, Node, and WASM bindings
- Introduces `Group::edit_message` in the MLS core, which validates the sender, content type match, and editability, then sends the edit to the network, persists it to a new `message_edits` DB table, and emits a `LocalEvents::MessageEdited` event.
- Adds a `message_edits` Diesel table and migration with a foreign key to `group_messages` and indexed by `(original_message_id, edited_at_ns DESC)`.
- Exposes `editMessage`, `streamMessageEdits`, and `deleteMessage` APIs across Node, WASM, and mobile (iOS/Android) FFI bindings, with corresponding codec (`xmtp.org/editMessage` v1.0) registered on all platforms.
- Enriches `DecodedMessage` with `edited_at_ns` and `is_edited` fields; enrichment logic applies the latest valid edit to messages and their reply parents, and falls back to original content on decode failure.
- Risk: existing clients that do not filter `EditMessage` content type will see these messages surface in conversation lists until they update to the new hidden-types filter.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized 745d4f9.</sup>


mchenani:
> ### Code Quality Issues

> **1. Missing `on_close` callback in Node.js `stream_message_edits`** (`bindings/node/src/conversations/streams.rs:241`)

The stream_message_edits doesn't expose an on_close callback in Node.js bindings. However, stream_message_deletions  follows the same pattern. 

> **2. Potential data loss if `edit.store(&conn)?` fails after network send** (`crates/xmtp_mls/src/groups/mod.rs:905-921`)

The edit_message function sends the message first, then stores the edit record locally. If storage fails, there's a potential inconsistency. But the chances here are rare and nothing harmful would happen.

PR #2458, opened 2025-09-15
build android jniLibs with a nix derivation

insipx:
- removes UDL file, kotlin bindings can be built with `--library`
- adds cargo-ndk commands to `android.nix`
- this enables calling the jniLib output from the xmtp-android repository (or anywhere with nix) include the libxmtp jniLibs as external dependencies in other nix derivations
- adds an easy overlay module to nix flake and a `lib/` directory to house some common functionality that will be passed to all derivations via nix `withPackage`
- release-kotlin script will use the `nix build .#android-bindings` derivation
- `androidEmulator` output to enable running android tests in ci/with a script in libxmtp
- removes need to sed/replace "xmtpv3" with "uniffi_xmtpv3" + renaming of dyLibs by indicating target name in scaffolding macro call

possible todo:
- [ ] each android target should be its own derivation to allow caching of platform-specific crate dependencies, and allow nix flake to specify which targets to build jniLibs for (in order to avoid building each target when debugging for instance)

PR #3082, opened 2026-02-02
Add xmtp-ios to monorepo

neekolas:
## Monorepo Part 1

This Graphite stack has been condensed down to a single PR. 

It includes:

- Merging in the complete history of `xmtp-ios`
- Updating our iOS nix configuration to allow for building the bindings as a package
- Adds a bunch of new release tooling to help create/manage release branches and release notes
- Updates iOS SDK release workflows

### Setup
- `dev/nix-up`
- `dev/up`
- `cd sdks/ios`

### Testing

- Run `dev/build` in the iOS SDK and you should be able to get a clean version of libxmtp built and usable in XCode
- Tests run properly

### Actions
- `create-release-branch.yml`
- `lint-ios.yaml`
- `release-ios.yml`
- `release-notes.yml`
- `test-ios.yml`
- `publish-release.yml`

### Scripts

- `sdks/ios/dev/{build|test}`

### Config

- `Package.swift` now lives in the repo root. I do some funky stuff to make it use the locally built libxmtp version when it's available, and the remote one for published releases. Maybe there is a better way...
- Minimal changes to the nix folder. One change I did make was adding an overlay for swiftformat and SwiftLint. When I hit a cache miss it tried to build Swift from source on my machine, which took forever and then eventually failed. Replaced with a binary download link.

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized a49f299. 76 files reviewed, 5 issues evaluated, 1 issue filtered, 2 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>dev/release-tools/src/commands/bump-version.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 12](https://github.com/xmtp/libxmtp/blob/a49f299a5537f30e8f9a59b30175d1bfd475b4aa/dev/release-tools/src/commands/bump-version.ts#L12): The `bumpVersion` function passes the raw `sdk` string from user input directly to `getSdkConfig`. The referenced `getSdkConfig` implementation performs an unsafe lookup (`SDK_CONFIGS[sdk]`) on a plain object. If a user provides an SDK name matching a standard `Object.prototype` property (e.g., `"toString"`, `"constructor"`, `"valueOf"`), the lookup returns a function or object rather than `undefined`. This bypasses the `if (!config)` check in `getSdkConfig`. Consequently, `config` in `bumpVersion` becomes that function/object, and the property access `config.manifest` evaluates to `undefined`. Accessing `.readVersion` on `undefined` at line 12 causes a runtime `TypeError` crash instead of a handled "Unknown SDK" error. <b>[ Already posted ]</b>
</details>

</details>

### DIFF
--- a/sdks/ios/Sources/XMTPiOS/Conversations.swift
+++ b/sdks/ios/Sources/XMTPiOS/Conversations.swift
@@ -75,7 +75,7 @@ final class ConversationStreamCallback: FfiConversationCallback {
 		onCloseCallback()
 	}
 
-	func onError(error: FfiSubscribeError) {
+	func onError(error: FfiError) {
 		print("Error ConversationStreamCallback \(error)")
 	}
 
@@ -100,7 +100,7 @@ final class MessageDeletionCallback: FfiMessageDeletionCallback {
 		onCloseCallback()
 	}
 
-	func onError(error: FfiSubscribeError) {
+	func onError(error: FfiError) {
 		print("Error MessageDeletionCallback \(error)")
 	}
 
--- a/sdks/ios/Sources/XMTPiOS/Group.swift
+++ b/sdks/ios/Sources/XMTPiOS/Group.swift
@@ -5,7 +5,7 @@ final class MessageCallback: FfiMessageCallback {
 		onCloseCallback()
 	}
 
-	func onError(error: FfiSubscribeError) {
+	func onError(error: FfiError) {
 		print("Error MessageCallback \(error)")
 	}
 
--- a/sdks/ios/Sources/XMTPiOS/Libxmtp/xmtpv3.swift
+++ b/sdks/ios/Sources/XMTPiOS/Libxmtp/xmtpv3.swift
@@ -655,7 +655,7 @@ open func onAuthRequired()async throws  -> FfiCredential  {
             completeFunc: ffi_xmtpv3_rust_future_complete_rust_buffer,
             freeFunc: ffi_xmtpv3_rust_future_free_rust_buffer,
             liftFunc: FfiConverterTypeFfiCredential_lift,
-            errorHandler: FfiConverterTypeGenericError_lift
+            errorHandler: FfiConverterTypeFfiError_lift
         )
 }
     
@@ -709,7 +709,7 @@ fileprivate struct UniffiCallbackInterfaceFfiAuthCallback {
                 makeCall: makeCall,
                 handleSuccess: uniffiHandleSuccess,
                 handleError: uniffiHandleError,
-                lowerError: FfiConverterTypeGenericError_lower
+                lowerError: FfiConverterTypeFfiError_lower
             )
             uniffiOutReturn.pointee = uniffiForeignFuture
         },
@@ -871,7 +871,7 @@ open func set(credential: FfiCredential)async throws   {
             completeFunc: ffi_xmtpv3_rust_future_complete_void,
             freeFunc: ffi_xmtpv3_rust_future_free_void,
             liftFunc: { $0 },
-            errorHandler: FfiConverterTypeGenericError_lift
+            errorHandler: FfiConverterTypeFfiError_lift
         )
 }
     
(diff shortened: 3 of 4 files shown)


nix/lib/packages/swiftlint.nix
```diff
@@ -0,0 +1,33 @@
+# Pre-built SwiftLint binary from GitHub releases
+# This avoids building the Swift compiler from source in Nix
+{ stdenv, fetchurl, unzip }:
+
+stdenv.mkDerivation rec {
```
insipx, line 5: these derivations and the one in [nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-25.11/pkgs/by-name/sw/swiftlint/package.nix#L38) look similar -- in what workflow was the compiler re-compiling? maybe a binary cache was missing, it should normally pull from the default nixos.cache

nix/ios.nix
```diff
@@ -49,8 +47,54 @@ mkShell {
...
+      swiftlint
     ]
     ++ lib.optionals isDarwin [
       darwin.cctools
     ];
+
+  shellHook = ''
+    # Use system Xcode for iOS SDK access (iPhoneOS, iPhoneSimulator platforms)
+    # The Nix apple-sdk only includes MacOSX.platform, so we override at runtime
+    export DEVELOPER_DIR="/Applications/Xcode.app/Contents/Developer"
```
insipx: can set these directly outside the shell, then be able to re-use them recursively

```nix
DEVELOPER_DIR="/Applications/Xcode.app/Contents/Developer"
IOS_SDK="${DEVELOPER_DIR}/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk"
```

sdks/ios/.direnv/flake-profile
```diff
@@ -0,0 +1 @@
+flake-profile-1-link
```
insipx: should add `.direnv/**` to .gitignore everywhere since it's env-specific and auto-generated if a dev has it enabled

nix/lib/packages/swiftlint.nix
neekolas: Maybe I could have left SwiftLint alone. [Swiftformat](https://github.com/NixOS/nixpkgs/blob/nixos-25.11/pkgs/by-name/sw/swiftformat/package.nix) depends on Swift, which was causing a cache miss for me. I replaced both with an overlay at the same time.

Totally possible something was hosed on my machine, but I was using other caches for other things around the same time.

nix/package/ios.nix
```diff
@@ -0,0 +1,239 @@
...
+    doNotPostBuildInstallCargoBinaries = true;
+    installPhaseCommand = ''
+      ${nativeEnvSetup}
+      # Generate Swift bindings using uniffi-bindgen.
+      # This runs the ffi-uniffi-bindgen binary (built above) against the compiled
+      # static library to extract the FFI interface and produce:
+      #   - xmtpv3.swift: Swift source with all public API types and functions
+      #   - xmtpv3FFI.h: C header for the FFI layer
+      #   - xmtpv3FFI.modulemap: Clang module map (renamed to module.modulemap)
+      cargo run -p xmtpv3 --bin ffi-uniffi-bindgen --release --features uniffi/cli generate \
```
insipx, line 200: the `ffi-uniffi-bindgen` CLI is a compilation step we can also add as a package that is cached. i have a pkg working [here](https://github.com/xmtp/libxmtp/pull/2458/changes#diff-f8d6404c9dfbc44029c5ff846b606ecb64546dddb3b41361e4342db18d42b83dR149) but it could be worth adding to the pkg `xmtp` namespace in a future PR

neekolas: Nice. One less thing to wait for.

cameronvoell (approved):
able to run tests and the example app!

PR #3098, opened 2026-02-03
Minor testing updates

codabrink:
### Add `StreamWithStats.spin` to drive streams with notifications and require `Unpin` bounds for test-only `Client.stream_all_messages_owned_with_stats`
Introduce `StreamWithStats.spin` that spawns a task to consume the stream and notify on each item, tighten trait bounds and lifetimes on `StreamStatsWrapper` impls, and move the test-only `Client.stream_all_messages_owned_with_stats` into a bounded impl block. Update test util to publish identity updates via `PublishIdentityUpdateRequest`.

#### 📍Where to Start
Start with the `StreamWithStats` trait and its `spin` implementation in [stream_stats.rs](https://github.com/xmtp/libxmtp/pull/3098/files#diff-708b1c770e851594115eb1fab809298b6420f273367eb2d6813124ff9ef20a93).

#### Changes since #3098 opened

- Gated `StreamWithStats::spin` trait method and its implementation in `StreamStatsWrapper` behind conditional compilation [c653c2e]

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 3079270.

### DIFF
--- a/crates/xmtp_mls/src/subscriptions/stream_messages/stream_stats.rs
+++ b/crates/xmtp_mls/src/subscriptions/stream_messages/stream_stats.rs
@@ -36,6 +36,7 @@ pin_project! {
 
 pub trait StreamWithStats: Stream<Item = Result<StoredGroupMessage>> {
     fn stats(&self) -> Arc<StreamStats>;
+    #[cfg(any(feature = "test-utils", test))]
     fn spin(self) -> Arc<Notify>;
 }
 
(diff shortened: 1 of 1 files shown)


crates/xmtp_mls/src/subscriptions/stream_messages/stream_stats.rs
```diff
@@ -35,16 +36,36 @@ pin_project! {
 
 pub trait StreamWithStats: Stream<Item = Result<StoredGroupMessage>> {
     fn stats(&self) -> Arc<StreamStats>;
+    fn spin(self) -> Arc<Notify>;
```
tylerhawkes, line 40: Why add this if it isn't used anywhere in the PR?

codabrink: It's for testing. Gated it behind the test cfg.

PR #3121, opened 2026-02-05
Add missing indexes

codabrink:
Improve queries on commit log tables and group messages expiry.

Before:
image

After:
image

### DIFF
--- a/crates/xmtp_mls/src/groups/device_sync/tests.rs
+++ b/crates/xmtp_mls/src/groups/device_sync/tests.rs
@@ -50,14 +50,9 @@ async fn basic_sync() {
 #[xmtp_common::test(unwrap_try = true)]
 #[cfg(not(target_arch = "wasm32"))]
 async fn only_one_payload_sent() {
-    use std::time::Duration;
-    use tokio::time::sleep;
-
-    use crate::utils::LocalTesterBuilder;
-
     tester!(alix1, sync_server, sync_worker, with_name: "alix1");
-    let alix2 = alix1.builder.clone().with_name("alix2").build().await;
-    let alix3 = alix1.builder.clone().with_name("alix3").build().await;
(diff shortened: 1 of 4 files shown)


crates/xmtp_db/migrations/2026-02-05-012554-0000_add_missing_indexes/up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX idx_local_commit_log_group_id ON local_commit_log(group_id);
+CREATE INDEX idx_remote_commit_log_group_id ON remote_commit_log(group_id);
```
tylerhawkes, line 2: Do indexes in sqlite point to the rowid or to where they are actually stored on the disk? If it's the latter, it would probably be better to add the rowid as that's used in the queries and could skip even more of the data.

codabrink: It looks like it's just the rowid

crates/xmtp_mls/src/groups/device_sync/tests.rs
```diff
@@ -244,9 +244,11 @@ async fn test_hmac_and_consent_preference_sync() {
     assert_eq!(alix1_group.consent_state()?, ConsentState::Unknown);
     alix1_group.update_consent_state(ConsentState::Allowed)?;
 
+    alix2.sync_all_welcomes_and_groups(None).await?;
+
     alix2
         .worker()
-        .register_interest(SyncMetric::ConsentReceived, 3)
+        .register_interest(SyncMetric::ConsentReceived, 2)
```
codabrink, line 225: Looking at the test, I'm not sure why this was ever 3. 2 is the right answer.

PR #3156, opened 2026-02-11
Nix script improvements

cameronvoell:
Tried to address a few "gotcha's" with the nix-up and nix-down scripts and added a nix-status script for verifying that things are up and running as expected. 

One of the main fixes was to automatically add your logged in user to the trusted user config which is needed for using nix caches during the initial setup, otherwise you download a bunch of dependencies and build from source.

### DIFF
--- a/dev/nix-up
+++ b/dev/nix-up
@@ -16,44 +16,46 @@ else
   echo "detected existing nix installation"
 fi
 
-# Ensure current user is a trusted nix user (required for cachix substituters)
+# Ensure project binary caches are trusted substituters
+# Uses trusted-substituters (whitelist specific caches) instead of trusted-users
+# (which would grant broad Nix privileges to the user account)
 NIX_CUSTOM_CONF="/etc/nix/nix.custom.conf"
-CURRENT_USER="$(whoami)"
+XMTP_CACHE="https://xmtp.cachix.org"
+NIX_COMMUNITY_CACHE="https://nix-community.cachix.org"
+XMTP_KEY="xmtp.cachix.org-1:nFPFrqLQ9kjYQKiWL7gKq6llcNEeaV4iI+Ka1F+Tmq0="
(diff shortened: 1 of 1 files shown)


dev/nix-up
```diff
@@ -16,8 +16,31 @@ else
...
 fi
 
-
-
+# Ensure current user is a trusted nix user (required for cachix substituters)
+NIX_CUSTOM_CONF="/etc/nix/nix.custom.conf"
+CURRENT_USER="$(whoami)"
+if [[ -f /etc/nix/nix.conf ]]; then
+  # Check if user is already trusted in either nix.conf or nix.custom.conf
+  if ! nix show-config 2>/dev/null | grep "trusted-users" | grep -qF "$CURRENT_USER" && \
```
insipx: IMO we should prompt here to ask if the user wants this, as well as explain what it does/why we need to escalate privileges to `sudo`

cameronvoell: updated 1812f15b9efd5d6fc749064e7b2116a67259077a

dev/nix-up
```diff
@@ -16,8 +16,49 @@ else
...
+    echo "  The Nix config ($NIX_CUSTOM_CONF) is owned by root, and the"
+    echo "  Nix daemon must be restarted to pick up the changes."
+    echo ""
+    read -p "Add these caches as trusted substituters? (y/n) " -n 1 -r
+    echo
+    if [[ ! $REPLY =~ ^[Yy]$ ]]; then
+      echo "Skipped. Dependencies will be built from source (slower)."
+      echo "You can re-run this script later to add them."
+    else
+      printf '\nextra-trusted-substituters = %s %s\n' "$XMTP_CACHE" "$NIX_COMMUNITY_CACHE" | sudo tee -a "$NIX_CUSTOM_CONF" > /dev/null
```
cameronvoell, line 50: @insipx found this trusted-substituters config which lets you whitelist specific caches as opposed to the trusted users which lets you install from any cache. used the public keys that are also hard-coded in flake.nix, which seems like should be fine. let me know if there's any concerns

PR #3160, opened 2026-02-11
cleanup craneLib manual imports

### LINKED ISSUE #3150: [nix] use one nixpkg and craneLib config for libxmtp
Currently we are passing craneLib manually to nix functions that require building packages, like in the wasm bindings function call: https://github.com/xmtp/libxmtp/blob/d1d80e5a4154cc29ed96a9d2a4a6aec0a461a9bc/flake.nix#L57

we define package configuration in `nix/lib`, and then define craneLib via that same config. it would be much easier to also move the craneLib definition into `nix/lib` so that dependant derivations may import it through the `xmtp` namespace like `xmtp.craneLib` rather than manually specify an import statement. 

this makes it easier to keep track of imports and keeps versioning in a single spot

insipx:
fixes #3150 

makes accessing filesets and defining new craneLib-dependant derivations a little easier

### DIFF
--- a/.github/workflows/release-android.yml
+++ b/.github/workflows/release-android.yml
@@ -9,8 +9,7 @@ on:
         description: "dev, rc, or final"
       rc-number:
         required: false
-        type: number
-        default: 0
+        type: string
         description: "RC number (required for rc releases)"
       ref:
         required: true
--- a/.github/workflows/release-ios.yml
+++ b/.github/workflows/release-ios.yml
@@ -10,8 +10,7 @@ on:
         description: "dev, rc, or final"
       rc-number:
         required: false
-        type: number
-        default: 0
+        type: string
         description: "RC number (required for rc releases)"
       ref:
         required: true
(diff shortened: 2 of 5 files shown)


insipx:

1. `trunk()`
1. **push-nwxxwzqwnxoq ← this PR**
1. [push-wztqyrwztzqs](https://github.com/xmtp/libxmtp/pull/3163)
1. [push-lssnrpnpzvll](https://github.com/xmtp/libxmtp/pull/3166)
1. [push-mxrnnusmvnut](https://github.com/xmtp/libxmtp/pull/3168)

---

nix/android.nix
```diff
@@ -16,9 +15,7 @@
 , writeShellScriptBin
 }:
 let
-  # Shared Android environment configuration
-  androidEnv = import ./lib/android-env.nix { inherit lib androidenv stdenv; };
-
+  inherit (xmtp) androidEnv;
```
neekolas: Would we want to `inherit` from mobile here (like we do in package/iOS.nix) and get the inputs like `sqlite` and `zlib` that way, rather than listing them at the top of this file?

insipx: yeah that would be cleaner

PR #3163, opened 2026-02-11
optimize nix build cache hits & local compile time with `cargo-hakari`

insipx:
[crane docs](https://crane.dev/examples/quick-start-workspace.html) highly recommend integrating [cargo-hakari](https://docs.rs/cargo-hakari/latest/cargo_hakari/) to reduce cache misses. cargo hakari should also optimize local building and reduce re-builds by optimizing the dependency graph of the cargo workspace.

read about how hakari works [here](https://docs.rs/cargo-hakari/latest/cargo_hakari/about/)

workspace-hack is used on some other large projects as well, like

[rustc](https://github.com/rust-lang/rust/blob/0bfc45aa859b94cedeffcbd949f9aaad9f3ac8d8/src/tools/rustc-workspace-hack/Cargo.toml)
[firefox](https://hg.mozilla.org/mozilla-central/file/cf6956a5ec8e21896736f96237b1476c9d0aaf45/build/workspace-hack/Cargo.toml)

this PR follows the instructions in cargo-hakari docs and is completely auto-generated. it adds a CI workflow to verify the config. you can run hakari locally with `nix develop .# --command cargo hakari verify` or `nix develop .# --command cargo hakari generate`


insipx:

1. `trunk()`
1. [push-nwxxwzqwnxoq](https://github.com/xmtp/libxmtp/pull/3160)
1. **push-wztqyrwztzqs ← this PR**
1. [push-lssnrpnpzvll](https://github.com/xmtp/libxmtp/pull/3166)
1. [push-mxrnnusmvnut](https://github.com/xmtp/libxmtp/pull/3168)

---

.github/workflows/lint-workspace.yml
```diff
@@ -21,3 +21,11 @@ jobs:
         run: cargo clippy --locked --workspace --all-features --all-targets --no-deps --exclude bindings_wasm -- -D warnings
       - name: Run format check
         run: cargo fmt --check
+      - name: Install cargo-hakari
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hakari
+      - name: Check workspace-hack Cargo.toml is up-to-date
+        run: cargo hakari generate --diff
```
neekolas, line 29: So, any time we add a dependency to anything in the workspace we need to run `cargo hakari generate` afterwards?

If that's the case, should include a note in the README.

deny.toml
```diff
@@ -29,20 +29,7 @@ allow = [
...
-    "Unlicense",
-  ], name = "pharos", version = "0.5.3" },
-  { allow = [
-    "Unlicense",
-  ], name = "async_io_stream", version = "0.3.3" },
-  { allow = [
-    "Unlicense",
-  ], name = "ws_stream_wasm", version = "0.7.4" },
-]
+exceptions = []
```
neekolas, line 32: Why this change?

nix/android.nix
```diff
@@ -113,6 +114,7 @@ mkShell {
   ANDROID_NDK_ROOT = androidPaths.ndkHome;
   NDK_HOME = androidPaths.ndkHome;
   EMULATOR = "${androidEmulator}";
+  LD_LIBRARY_PATH = lib.makeLibraryPath [ openssl zlib ];
```
neekolas, line 117: Would be helpful to know when this list of inputs would need to change.

.config/hakari.toml
```diff
@@ -0,0 +1,33 @@
...
+
+# Setting workspace.resolver = "2" or higher in the root Cargo.toml is HIGHLY recommended.
+# Hakari works much better with the v2 resolver. (The v2 and v3 resolvers are identical from
+# hakari's perspective, so you're welcome to set either.)
+#
+# For more about the new feature resolver, see:
+# https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#cargos-new-feature-resolver
+resolver = "3"
+
+platforms = [
```
neekolas, line 17: What would happen if we added _all_ the platforms we build against to this list?

nix/android.nix
insipx: I was getting build errors in android, I think some buildscripts need the `LD_LIBRARY_PATH` to be set in order to work correctly. I've found its mostly an issue with the zstd/openssl dependencies

deny.toml
insipx: `cargo deny` gave warnings that the exceptions were no longer necessary, so removed them. either the dependencies changed license or are no longer used by us

.config/hakari.toml
insipx: I think it would just enable the workspace hack for everything. I went with the platforms that I think were in the `fast ci` route, since we can always add on to them later

.github/workflows/lint-workspace.yml
insipx: I'll add this into the child PR

PR #3169, opened 2026-02-11
Improvements around toggling nix and direnv on/off

cameronvoell:


flake.nix
```diff
@@ -18,16 +18,6 @@
     };
   };
 
-  nixConfig = {
-    extra-trusted-public-keys = [
```
cameronvoell, line 22: @insipx let me know what you think about moving these to the nix-up script. I think the `extra-trusted-public-keys` here is ignored unless you are in the config as a trusted-user, but trusted user lets you download from any cache. 

So it seems like moving this to nix-up was the best way to enable downloading from only the two cachix links here.

dev/direnv-up
```diff
@@ -0,0 +1,34 @@
...
+
+echo "Allowing direnv environment for this repo..."
+direnv allow "$REPO_ROOT"
+
+# Re-add direnv shell hook if missing (e.g. after direnv-down)
+for shell_rc in ~/.bashrc ~/.zshrc; do
+  if [[ -f "$shell_rc" ]] && ! grep -q "direnv hook" "$shell_rc"; then
+    case "$shell_rc" in
+      *bashrc) echo 'eval "$(direnv hook bash)"' >> "$shell_rc" ;;
+      *zshrc) echo 'eval "$(direnv hook zsh)"' >> "$shell_rc" ;;
```
insipx, line 26: coda and I use `fish` shell, would be worth adding

flake.nix
cameronvoell: if we leave this in flake.nix you get extra error noise about not being in the trusted users, even though everything works as expected.

insipx: ~this is useful in CI or otherwise by using nix build .#package --accept-flake-config which will use the flake config, regardless of trusted user status. For local environments --accept-flake-config is technically an anti-pattern, since malicious flakes can put just any cache into the nixConfig, but in CI this is less of an issue~

insipx: after taking a look, it seems we replaced the `--accept-flake-config` with the cachix action so this should be OK

cameronvoell: do you think it's worth the security benefits for us to recommend just adding the user to substituter here instead of trusted-user like how I currently have it in `nix-up,` and then we can maybe make a note to ignore the warnings from the "not a trusted user" errors from flake.nix :

https://github.com/xmtp/libxmtp/blob/0ca60f0947ab1971b560e6645c8cc698d4e160cd/dev/nix-up#L44-L51

Or do you think we should just make it simple for `nix-up` to add you to t`rusted-users`, and just note that we need to be mindful that we can add caches outside the trusted substituters list when operating locally?

insipx: I think the only other issue with this is if someone wants the cache to be local. Modifying user systems may not be 100% reliable, not everyone may want xmtp.cachix.org to be a system-wide cache for their system, esp. if they use NixOs as a daily driver. `./dev/nix-up` is nice for xmtp-internal use but IMO we should remain general here to other setups

insipx (approved):

dev/direnv-up
cameronvoell: added https://github.com/xmtp/libxmtp/pull/3169/changes/228d6b9a4fd5222409026a5cb2e40834c48f0716

flake.nix
cameronvoell: updated to check trusted users for sudo use, and to use `nix run nixpkgs#cachix` as per conversation, thanks for the feedback! => https://github.com/xmtp/libxmtp/commit/c2fd2974835042bcae248fada7bf80cb957d71cc

PR #3179, opened 2026-02-12
add aarch64 support for garnix

insipx:
this should populate the android build cache on each PR to garnix

adds the garnix build cache to `./dev/nix-cache-up`. Splits the cache-adding logic out of `./dev/nix-up` (calls separate script from dev/nix-up)

also cleans up the readme header


nix/package/wasm.nix
```diff
@@ -10,8 +10,8 @@
 , llvmPackages
 , wasm-bindgen-cli
 , xmtp
-, geckodriver
-, firefox
+, chromedriver
```
insipx, line 13: firefox compile fails often on aarch64, as well as takes forever, so its not available via binary cache. switching back to chrome.

garnix.yaml
```diff
@@ -0,0 +1,6 @@
+builds:
+  exclude:
+    - "packages.aarch64-darwin.ios-libs"
```
insipx, line 3: it's possible to build with garnix but requires some custom sandbox options b/c of the `__noChroot` environment. will look into at a later time

PR #3192, opened 2026-02-13
Get iOS API reference docCs working

jhaaaa:
Preview: https://docs.xmtp.org/ios/latest/documentation/xmtpios

Preview of versions page: https://docs.xmtp.org/ios

I asked Claude to assess the accuracy of its code comments:

> Accuracy assessment: ~98%. The comments are derived from code reading and are straightforward descriptions of method signatures, parameters, and return types — areas where it's hard to get wrong. 

I have added comments where Claude believes human review would be helpful.
 

> [!NOTE]
> ### Add DocC-based iOS API reference documentation generation and deployment
> - Replaces Jazzy with DocC for iOS SDK API reference generation; removes [Gemfile](https://github.com/xmtp/libxmtp/pull/3192/files#diff-fd35845574874743fd552ea157aff406782dbd31948bef5514884b7dae9d3aac) and updates [VALIDATION.md](https://github.com/xmtp/libxmtp/pull/3192/files#diff-c2f6853f96419be19145fc8b7823b8c91966e0e717a6d4e6ce23cb08f8d21482) accordingly.
> - Adds a new [hide-internal-symbols](https://github.com/xmtp/libxmtp/pull/3192/files#diff-c8a9092e16f63f6a2410dfc65bd56e689a6acc74fe4336faa82416e78dae76dd) script that injects `@_documentation(visibility: internal/private)` into generated proto and FFI Swift sources to suppress internal symbols from DocC output.
> - Rewrites [docs-ios.yml](https://github.com/xmtp/libxmtp/pull/3192/files#diff-380527fb98afabb8b360c146ad40bf1ab6a908009a1a4da1b2cf9f60e8b03eff) as a reusable workflow that builds docs via `swift package generate-documentation`, stores versioned docs in an `ios-docs-versions` branch, and deploys to Vercel instead of GitHub Pages.
> - Updates [release-ios.yml](https://github.com/xmtp/libxmtp/pull/3192/files#diff-da4efb9d30ae321f09b449630274516e7364d12a9d9b7d4d8185c7a895a6f8de) to trigger the docs workflow after a publish, updating `ios/latest` only for final releases.
> - Adds DocC-style documentation comments across major public API surfaces (`Client`, `Conversations`, `Group`, `Dm`, `Conversation`, `DecodedMessage`, `DecodedMessageV2`, `PublicIdentity`) with no functional or signature changes.
> - Fixes `includeDynamicTarget` in [Package.swift](https://github.com/xmtp/libxmtp/pull/3192/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677e) to only be set when `useLocalDynamicBinary` is true, not when a local binary is absent.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9981774.</sup>
> 
>
>


jhaaaa:
Claude Code on the multiple iterations of feedback: 
  None of this is actionable for our   
  PR. Every item is either pre-existing code (signal
  handler, throws that doesn't throw, logging, struct vs
   class), repeated from earlier rounds (workflow
  permissions, deprecated migration), or pedantic       
  (buffer comment says "3" but keeps 4). The automated  
  reviewer is reviewing the entire files rather than
  just our diff.

  1. Buffer comment — The comment already says "keep    
  current + 3 preceding" which matches > 4. It's clear  
  enough.                                               
  2. Workflow permissions — Reviewer agrees no concern.
  3. Package.swift version — That's from a main merge,
  not our change.
  
  The bot is generating diminishing-value feedback
  because it's re-reviewing the full files each time
  rather than scoping to the diff.

sdks/ios/Sources/XMTPiOS/Client.swift
```diff
@@ -25,9 +35,16 @@ public enum ClientError: Error, CustomStringConvertible, LocalizedError {
 	}
 }
 
+/// Controls which groups are eligible for automatic MLS fork recovery.
+///
+/// An MLS "fork" occurs when group state diverges between members. Recovery re-establishes
+/// a consistent state so messages can flow again.
```
jhaaaa: Claude requests human review:

Fork recovery (ForkRecoveryPolicy, ForkRecoveryOptions) — I described the mechanics based on code, but I don't have deep context on when apps should use .allowlistedGroups vs .all, or the practical trade-offs. Someone who's dealt with forks in production would know if the guidance is useful or misleading.

sdks/ios/Sources/XMTPiOS/Libxmtp/DecodedMessageV2.swift
```diff
@@ -1,47 +1,77 @@
 import Foundation
 
+/// A type representing a structured intent from a message.
 public typealias Intent = FfiIntent
+/// A type representing structured actions from a message.
 public typealias Actions = FfiActions
```
jhaaaa, line 7: Claude requests human review:

DecodedMessageV2 vs DecodedMessage — I described V2 as "enriched" with reactions baked in, but I'm not confident about the full scope of differences or when developers should prefer one over the other. A team member who designed this API would know.

sdks/ios/Sources/XMTPiOS/Group.swift
```diff
@@ -200,42 +296,66 @@ public struct Group: Identifiable, Equatable, Hashable {
...
+	/// - Parameter appData: The new application data string.
+	/// - Important: Requires the appropriate metadata update permission.
 	public func updateAppData(appData: String) async throws {
 		try await ffiGroup.updateAppData(appData: appData)
 	}
 
+	/// Updates the permission policy for adding members.
+	///
+	/// - Parameter newPermissionOption: The new permission level required to add members.
+	/// - Important: Requires super admin permissions.
```
jhaaaa, line 358: Claude requests human review:

Permission policy methods on Group — There are ~7 methods like updateAddMemberPermission, updateRemoveMemberPermission, etc. The comments say "Requires super admin permissions" based on the docs permission table, but I'd want someone to confirm that's universally true (vs. depending on the group's custom policy set).

sdks/ios/Sources/XMTPiOS/PrivatePreferences.swift
```diff
@@ -45,7 +45,11 @@ public struct ConsentRecord: Codable, Hashable {
 	}
 }
```
jhaaaa, line 47: Claude asks for human review:

PrivatePreferences — This got almost no coverage (just 2 comment blocks) because I wasn't confident enough. A team member could fill in what the actor does, what sync() on preferences means, and how consent preferences propagate across installations.

sdks/ios/Sources/XMTPiOS/Client.swift
```diff
@@ -260,6 +331,31 @@ public final class Client {
...
+	/// ``revokeAllOtherInstallations(signingKey:)`` or ``revokeInstallations(signingKey:installationIds:)``.
+	///
+	/// ```swift
+	/// let options = ClientOptions(dbEncryptionKey: myKey)
+	/// let client = try await Client.create(account: wallet, options: options)
+	/// print(client.inboxID) // the newly registered inbox
+	/// ```
+	///
+	/// On subsequent app launches where the identity is already registered, use
+	/// ``build(publicIdentity:options:inboxId:)`` instead -- it does not require a signing key.
```
jhaaaa, line 355: Claude asks for human review:

Deprecated method migration paths — I wrote things like "Use X instead" on deprecated methods. Someone should verify those migration suggestions are actually correct and not subtly different in behavior.

sdks/ios/Sources/XMTPiOS/Client.swift
codabrink: Fork recovery can be a little heavy-handed in some scenarios, so we want to give the integrator options to control when they want to opt-in or out. While we've done quite a bit of security auditing on fork recovery, including only allowing super-admins to perform fork recovery, it is still a mechanism that operates outside of openmls and it's guarantees, and some integrators might not want to take the risk that comes with that and instead just take the fork when it happens.

sdks/ios/Sources/XMTPiOS/PrivatePreferences.swift
```diff
@@ -45,7 +45,11 @@ public struct ConsentRecord: Codable, Hashable {
...
 
-/// Provides access to contact bundles.
+/// Manages consent state and user preferences for conversations and contacts.
+///
+/// Use this actor to allow or deny conversations and inbox IDs,
```
codabrink: Can we change this to "consent to or deny"? I think it connects to the next part of this sentence better.

sdks/ios/Sources/XMTPiOS/PrivatePreferences.swift
```diff
@@ -45,7 +45,11 @@ public struct ConsentRecord: Codable, Hashable {
...
+/// query current consent state, and stream real-time consent updates
+/// that sync across all devices associated with this account.
```
codabrink: "that sync across all devices associated with this account using the device sync feature."

So that way they know when they turn device sync off this won't work.

PR #3196, opened 2026-02-13
support more targets for validation-service docker image

insipx:


crates/xmtp_db/Cargo.toml
```diff
@@ -57,24 +59,17 @@ tokio = { workspace = true, optional = true, features = [
...
+diesel = { workspace = true, features = ["r2d2"] }
+dyn-clone.workspace = true
 libsqlite3-sys = { version = "0.35", features = [
   "bundled-sqlcipher-vendored-openssl",
 ] }
-diesel = { workspace = true, features = ["r2d2"] }
-dyn-clone.workspace = true
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
-sqlite-wasm-rs = { version = "0.4", default-features = false, features = [
```
neekolas, line 75: We don't need this anymore?

nix/musl-docker.nix
```diff
@@ -2,34 +2,67 @@
...
+        CARGO_TARGET_AARCH64_UNKNOWN_LINUX_MUSL_LINKER = "${aarch64Pkgs.stdenv.cc}/bin/${aarch64Pkgs.stdenv.cc.targetPrefix}cc";
+      } // (commonStatic old aarch64Pkgs);
+
+      imageCommon = {
+        name = "ghcr.io/xmtp/mls-validation-service"; # override ghcr images
+        tag = "main";
+        created = "now";
+        config = {
+          Env = [
+            "ANVIL_URL=http://anvil:8545"
```
neekolas: If we remove this from the image build, and instead set it in the docker-compose.yml file we can re-use the same images for local and for deployments. And we never risk accidentally setting the anvil URL in the public image.

insipx: I'll have to try this out -- I think its tricky b/c the ANVIL_URL gets compiled in a separate layer since its actually `xmtp_id` grabbing it. might have to move that to `mls_validation_service` for it to work but should be easy

crates/xmtp_db/Cargo.toml
insipx: It's actually been included in a libsqlite3-sys release: https://github.com/rusqlite/rusqlite/releases/tag/v0.38.0, https://github.com/rusqlite/rusqlite/pull/1769 so it should be automatic. Ran into other issues so i'll leave that for another day. Diesel also needs to update first

nix/musl-docker.nix
insipx: the `test-utils` cfg in `xmtp_id` was causing issues -- I think its safe to remove from the `ANVIL_URL` variable, esp since that's how we publish the image to prod anyway. should also make the binary a bit smaller

crates/xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -188,7 +188,6 @@ impl MultiSmartContractSignatureVerifier {
             };
         }
 
-        #[cfg(feature = "test-utils")]
         if let Ok(url) = std::env::var("ANVIL_URL") {
```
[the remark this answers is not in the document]
insipx: 31337 is described by eip155 to be the official ID for local testnet/anvil/hardhat/GoChain Testnet, so it should not be used outside of a local test context

nix/musl-docker.nix
neekolas: Ahh. That rings a bell.  
  
Don't worry about it for now. We can solve publishing these Nix images later. Great to have something for local dev at least.

crates/xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -188,7 +188,6 @@ impl MultiSmartContractSignatureVerifier {
             };
         }
 
-        #[cfg(feature = "test-utils")]
```
neekolas, line 191: This seems pretty reasonable. We should be definining this behavior at runtime and not compile time anyways.

insipx: it seems we were building with `--test-utils` for the prod image for this variable anyway, so it shouldn't be different. 31337 is also official delegated as the local testnet EIP, so nothing real should be using the URL even if its set

neekolas (approved):
Approved. You should probably rebase on top of main now that some of the other Nix PRs are in

neekolas:
@insipx sorry, my merge queue failed. Changes (like adding the format lint) still aren't in main yet.

PR #3209, opened 2026-02-18
Add error code descriptions as code comments

jhaaaa:
This PR:
- Adds `///` doc comments to all ~327 error variants that derive `ErrorCode`, describing what each error means and its retryability.
- The `ErrorCode` derive macro now enforces doc comments at compile time. A missing comment will result in a build error.
- Adds `#[error_code(internal)]` attribute to mark error types that are not reachable from SDK bindings. These errors will not be included in the error_glossary.
- Creates `apps/error_glossary/`, a tool that parses the source and generates `docs/error_glossary.md` with all public error codes.
- Adds a CI workflow to auto-regenerate the glossary when error types change on main.
- Adds `docs/error-architecture.md` explaining how the error code system works and how to add new codes.

Regarding the accuracy of the error code descriptions, Claude Code said it is "99% confident on technical accuracy for the descriptions outside of the 5 flagged items." You can see these flagged items marked by comments in the diff view. This is what Claude Code did to ensure error code description accuracy:
- Read each error variant’s #[error("...")] message, field types, and #[from] attributes to understand what the error represents.
- Traced each variant’s usage in the codebase — where it is constructed, what triggers it, and what context it carries.
- Reviewed the RetryableError implementation for each type to determine retryability, checking whether each variant returns true, false, or delegates via retryable!() to an inner error.
- Interpreted “May be retryable” to mean the variant delegates to the inner error’s is_retryable() implementation, so retryability depends on runtime behavior.

> [!NOTE]
> ### Add error code glossary generator and doc comments across all error enums
> - Adds a new `error_glossary` binary app ([apps/error_glossary/](https://github.com/xmtp/libxmtp/pull/3209/files#diff-e369806e2d3b321d8b03854dfaa39801bd8383de1ccdb9899c6df939a8609dd4)) that walks the workspace, parses Rust source files with `syn`, and generates a markdown glossary of all `ErrorCode`-deriving types with descriptions and retryability notes.
> - Adds doc comments to error enum variants across 20+ crates to populate the glossary with human-readable descriptions.
> - Updates the `derive_error_code` proc macro in [crates/xmtp_macro/src/lib.rs](https://github.com/xmtp/libxmtp/pull/3209/files#diff-a17f97bd1d5a58e3b8f67819028a0d60be985771f17def93f0af145a0fcda274) to enforce doc comments on non-remote types and support new attributes: `inherit`, `internal`, and `remote = "path::Type"`.
> - Adds a GitHub Actions workflow ([.github/workflows/error-glossary.yml](https://github.com/xmtp/libxmtp/pull/3209/files#diff-d87e174fc57b40a9255b21fc48f4f018e6425affb7fca30bc0524fbbd528024d)) and a dev script ([dev/gen-error-glossary](https://github.com/xmtp/libxmtp/pull/3209/files#diff-80b8e566949440762b7e34f8c56823fb186e1b773d4299c630dfc135691e893c)) to auto-generate and PR updates to [docs/error_glossary.md](https://github.com/xmtp/libxmtp/pull/3209/files#diff-d3efb581db7b0533ee27c6d68c564c77c4b75827275eef9a61cac600da31d3e2).
> - Fixes two error display strings: `NotFound::IntentForToPublish` now says "Published" instead of "Publish", and `VerifierError::MalformedEipUrl` now says "eip155" instead of "eip144".
> - Risk: the updated proc macro now emits compile errors for any non-remote error enum variants missing doc comments, which may break builds if new variants are added without docs.
>
> 
> #### Changes since #3209 opened
>
> - Refactored `error_code` derive macro to enforce doc comment requirements and support internal attribute [1839868]
> - Added doc comments to error enum variants in `xmtp_api_d14n` and `xmtp_mls` crates [1839868]
> - Added procedural macro dependencies to `xmtp-workspace-hack` package [350d5ef]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 0840e82. 36 files reviewed, 2 issues evaluated, 0 issues filtered, 2 comments posted</summary>
>
> ### 🗂️ Filtered Issues
>
> </details>
>
>

### DIFF
--- a/crates/xmtp-workspace-hack/Cargo.toml
+++ b/crates/xmtp-workspace-hack/Cargo.toml
@@ -67,6 +67,7 @@ openmls_rust_crypto = { git = "https://github.com/xmtp/openmls", rev = "382a4649
 openmls_traits = { git = "https://github.com/xmtp/openmls", rev = "382a4649ae4e937ce933adb259e26d6381745731", features = ["test-utils"] }
 p256 = { version = "0.13", features = ["ecdh", "jwk"] }
 percent-encoding = { version = "2" }
+proc-macro2 = { version = "1", features = ["span-locations"] }
 quote = { version = "1" }
 rand-274715c4dabd11b0 = { package = "rand", version = "0.9", features = ["serde"] }
 rand-93f6ce9d446188ac = { package = "rand", version = "0.10" }
@@ -90,6 +91,7 @@ sha3 = { version = "0.10", default-features = false, features = ["asm", "std"] }
 slab = { version = "0.4" }
 smallvec = { version = "1", default-features = false, features = ["const_new", "serde", "union", "write"] }
 subtle = { version = "2", default-features = false, features = ["const-generics", "i128"] }
+syn = { version = "2", features = ["extra-traits", "fold", "full", "visit", "visit-mut"] }
 sync_wrapper = { version = "1", default-features = false, features = ["futures"] }
 time = { version = "0.3", features = ["formatting", "local-offset", "parsing"] }
 tokio-stream = { version = "0.1", features = ["sync"] }
@@ -164,6 +166,7 @@ openmls_traits = { git = "https://github.com/xmtp/openmls", rev = "382a4649ae4e9
 p256 = { version = "0.13", features = ["ecdh", "jwk"] }
 percent-encoding = { version = "2" }
 proc-macro2 = { version = "1", features = ["span-locations"] }
+quote = { version = "1" }
 rand-274715c4dabd11b0 = { package = "rand", version = "0.9", features = ["serde"] }
 rand-93f6ce9d446188ac = { package = "rand", version = "0.10" }
 rand_chacha-274715c4dabd11b0 = { package = "rand_chacha", version = "0.9" }


crates/xmtp_macro/src/lib.rs
```diff
@@ -506,6 +508,9 @@ impl ErrorCodeAttr {
                 if meta.path.is_ident("inherit") {
                     result.inherit = true;
                     Ok(())
+                } else if meta.path.is_ident("internal") {
+                    result.internal = true;
```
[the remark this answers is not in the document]
jhaaaa: But this is intentional. The attribute lives on the source code so   
  the glossary tool can read it via syn. The macro just needs to parse 
  it without erroring — it doesn't need to do anything with it. If the
  macro didn't parse it, you'd get a compile error for an unrecognized
  attribute.

  So it's not dead code — it's deliberately parsed-and-ignored so the
  attribute is valid Rust. No change needed.

apps/error_glossary/src/extraction.rs
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        // Try #[error_code("CustomCode")]
+        if let Ok(lit) = attr.parse_args::<syn::LitStr>() {
+            result.code = Some(lit.value());
+            continue;
+        }
+
+        // Try #[error_code(inherit)] or #[error_code(remote = "...")]
+        let _ = attr.parse_nested_meta(|meta| {
```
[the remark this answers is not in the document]
jhaaaa: If someone writes a malformed #[error_code(remote = ...)], the actual ErrorCode derive   
  macro will give them a compile error. The code won't compile. The    
  glossary tool never runs on code that doesn't compile.

apps/error_glossary/src/extraction.rs
```diff
@@ -0,0 +1,232 @@
...
+    if m.ident == "tests" || m.ident == "test" {
+        return true;
+    }
+    m.attrs.iter().any(|attr| {
+        if !attr.path().is_ident("cfg") {
+            return false;
+        }
+        // Check if the cfg contains "test"
+        let mut found_test = false;
+        let _ = attr.parse_nested_meta(|meta| {
```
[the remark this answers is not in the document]
jhaaaa: No. None of those are ErrorCode-deriving types — they're all test    
  utilities, test modules, and test-utils feature gates. The scenario
  the reviewer is worried about doesn't exist in this codebase.        
                                         
  I'd skip this one. It's a theoretically correct edge case with no    
  practical impact, and the simple fix (contains("test")) introduces   
  its own bugs.

crates/xmtp_id/src/associations/association_log.rs
```diff
@@ -8,38 +8,80 @@ use xmtp_common::ErrorCode;
...
     #[error_code(inherit)]
     Signature(#[from] SignatureError),
+    /// Member not allowed.
+    ///
+    /// Member kind cannot add the specified kind. Not retryable.
     #[error("Member of kind {0} not allowed to add {1}")]
     MemberNotAllowed(MemberKind, MemberKind),
+    /// Missing existing member.
+    ///
+    /// Required signer not found or signer identity mismatch. Not retryable.
```
jhaaaa, line 36: We wrote: “Required signer not found or signer identity mismatch.” The code checks both that the signer exists in the state and that the signer’s identifier matches the expected member. A domain expert may want more precise wording.

crates/xmtp_id/src/associations/association_log.rs
```diff
@@ -8,38 +8,80 @@ use xmtp_common::ErrorCode;
...
     #[error("Missing existing member")]
     MissingExistingMember,
+    /// Legacy signature reuse.
+    ///
+    /// Legacy delegated signature used in disallowed context. Not retryable.
```
jhaaaa, line 41: We wrote: “Legacy delegated signature used in disallowed context.” The actual rule is nuanced: legacy delegated signatures can only appear on CreateInbox actions with nonce 0. Someone familiar with the identity protocol should confirm the description captures the intent.

crates/xmtp_id/src/associations/serialization.rs
```diff
@@ -57,30 +57,69 @@ pub enum DeserializationError {
...
     #[error("Invalid passkey")]
     InvalidPasskey,
+    /// Invalid hash.
+    ///
+    /// Hash must be 32 bytes. Not retryable.
     #[error("Invalid hash (needs to be 32 bytes)")]
     InvalidHash,
+    /// Unspecified value.
+    ///
+    /// An unrecognized or unsupported value was encountered. Not retryable.
```
jhaaaa, line 107: We wrote: “An unrecognized or unsupported value was encountered.” This is used for proto enum fields that deserialize to the UNSPECIFIED (0) variant. A domain expert may want to say something more specific about protobuf zero-values.

crates/xmtp_id/src/scw_verifier/mod.rs
```diff
@@ -17,25 +17,52 @@ static DEFAULT_CHAIN_URLS: &str = include_str!("chain_urls_default.json");
...
+    /// Serialization error.
+    ///
+    /// JSON serialization/deserialization failed. Not retryable.
     #[error(transparent)]
     Serde(#[from] serde_json::Error),
-    #[error("URLs must be preceded with eip144:")]
+    /// Malformed chain ID.
+    ///
+    /// Chain ID string lacks expected eip155: prefix. Not retryable.
+    #[error("Chain IDs must be preceded with eip155:")]
```
jhaaaa, line 51: We renamed the description from “URL” to “Chain ID” and fixed eip144 → eip155. The variant name still says Url, but the actual check is about the eip155: prefix on chain ID strings. Someone may want to rename the variant itself for clarity (with #[error_code("VerifierError::MalformedEipUrl")] to preserve the code).

crates/xmtp_id/src/associations/builder.rs
```diff
@@ -158,13 +158,22 @@ impl SignatureRequestBuilder {
...
+    ///
+    /// Required signature was not provided. Not retryable.
     #[error("Required signature was not provided")]
     MissingSigner,
     #[error("Signature error {0}")]
     #[error_code(inherit)]
     Signature(#[from] SignatureError),
+    /// Unable to get block number.
+    ///
+    /// Block number not returned after successful SCW verification. May be retryable.
```
jhaaaa, line 176: We wrote: “Block number not returned after successful SCW verification.” The code performs SCW verification, then calls get_block_number(), and errors if it returns None. A domain expert should confirm whether “after successful SCW verification” is the right framing.

jhaaaa:
Regarding [Findings](https://github.com/xmtp/libxmtp/pull/3209#issuecomment-3923339003):

  1 - Panic in main.rs: Not an issue. canonicalize() failing is the   
  error case, and we handle it with unwrap_or_else — printing the
  original path the user typed plus the OS error. This is clear and    
  correct.                                                  

  2 - Doc comment whitespace: They answered their own question — it
  matches Rust's doc comment processing. No issue.

  3 - Missing test coverage: Same feedback as before. Valid
  observation, reasonable follow-up, not a blocker.

  4 - Workflow trigger efficiency: They also answered their own
  question — detecting error type changes would require parsing, so
  triggering broadly is the pragmatic choice. The
  peter-evans/create-pull-request action is a no-op when there's no
  diff, so spurious triggers are harmless.

  5 - Error code format not enforced: This is intentional. Custom
  codes exist specifically for backwards compatibility when renaming
  variants. The whole point is you paste the old code string. Adding a
  format lint would make the feature harder to use for no real benefit
  — there are currently zero uses of custom codes in the codebase.

  6 - Silent filtering of internal types: Decent idea but low
  priority. The summary already prints counts. If someone is confused
  about why a type is missing, they can grep for
  #[error_code(internal)]. Not worth adding a --verbose flag to a tool
  this simple.

docs/error_glossary.md
```diff
@@ -0,0 +1,568 @@
+<!-- This file is auto-generated by error_glossary. Do not edit manually. -->
```
neekolas, line 1: Wow. Having all of these in one place is a thing of beauty. ❤️

PR #3213, opened 2026-02-19
Fix dylib install_name

Priva28:
Fixes libxmtpv3.dylib shipping with a hardcoded CI install_name.

---

Apps linking LibXMTPSwiftFFIDynamic crash on launch with:

```
dyld: Library not loaded: /Users/runner/work/libxmtp/libxmtp/target/aarch64-apple-ios/release/deps/libxmtpv3.dylib
```

And App Store Connect rejects uploads with:

```
Invalid Swift Support. The files libswiftCompatibilitySpan.dylib aren't at the expected location /Payload/App.app/Frameworks.
```


insipx:
I think this was fixed via other means

PR #3214, opened 2026-02-20
Bindings builder macros

neekolas:
## Summary

Adds a generic, binding-agnostic builder pattern macro system (`napi_builder`, `wasm_builder`, `uniffi_builder`) for cross-platform bindings in libxmtp.

- **`crates/xmtp_macro/src/builder.rs`** — Core builder expansion engine driven by `AnnotationConfig` (struct/impl/constructor/setter annotations, setter style, prefix). Fully binding-agnostic.
- **`SetterStyle` enum** — `NapiThis` (NAPI `This<'scope>` injection for JS chaining), `Consuming` (wasm_bindgen `mut self -> Self`), `MutRefChain` (`&mut self -> &mut Self` for UniFFI/tests).
- **`setter_prefix`** — `"set_"` for NAPI/WASM (avoids conflicts with auto-generated field property getters), `""` for UniFFI.
- **`setter_impl_ann`** — Splits constructor and setter impl blocks for UniFFI (Arc<Self> incompatible with `&mut self`).
- **Field modes** — `required` (constructor param), `optional` (Option<T> with setter), `default` (literal expression), `skip` (Default::default, no setter).
- Preserves non-builder struct-level attributes (e.g. `#[allow(dead_code)]`).

### Testing

- **12 macro unit tests** (`crates/xmtp_macro/src/builder_test.rs`) — field modes, setter styles, prefix, struct attributes
- **6 UniFFI Rust tests** (`bindings/mobile/src/builder_test.rs`) — constructor, setters, defaults, skip, chaining
- **4 Node TypeScript tests** (`bindings/node/test/Builder.test.ts`) — constructor, chaining, defaults via `yarn build:test`
- **4 WASM TypeScript tests** (`bindings/wasm/test/Builder.test.ts`) — same coverage, Chromium + Firefox via Playwright

## Test plan

- [x] `cargo test -p xmtp_macro` — 12 tests pass
- [x] `cargo test -p xmtpv3 --lib -- builder_test` — 6 tests pass
- [x] `dev/lint` — clean
- [x] Node TS tests pass (`yarn build:test && yarn test`)
- [x] WASM TS tests pass (`yarn build:test && yarn test`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3215, opened 2026-02-20
Build backend separate from client creation for Node and WASM bindings

neekolas:
## Summary

Introduces `BackendBuilder` and `Backend` types for Node and WASM bindings, enabling pre-configured API backend construction via the new builder macro system.

**Stacked on:** #3214 (cross-platform builder pattern macros)

### New types

- **`XmtpEnv`** enum in `xmtp_configuration` — 7 network environments (Local, Dev, Production, TestnetStaging, TestnetDev, Testnet, Mainnet) with centralized URL resolution from constants.
- **`BackendBuilder`** (Node + WASM) — Uses `napi_builder`/`wasm_builder` macros. Required `env` field, optional `api_url`, `gateway_host`, `readonly`, `app_version` setters, plus `authCallback`/`authHandle` methods. Builds a `Backend`.
- **`Backend`** — Wraps `XmtpClientBundle` with env/host getters. Passed to `createClientWithBackend()`.
- **`createClientWithBackend()`** (Node + WASM) — Creates a client from a pre-built Backend, reducing constructor params by extracting API config into the Backend.
- **`validate_and_resolve()`** in `xmtp_api_d14n::config` — Shared validation logic for env→URL resolution, `is_secure` inference, and auth requirement checking.

### Code deduplication

- Extracted shared helpers `build_store`, `parse_nonce`, and `create_client_inner` in Node bindings — both `createClient` and `createClientWithBackend` delegate to the same inner function.
- Extracted shared helpers `build_store` and `create_client_inner` in WASM bindings — same pattern.
- Made `v3_host` optional in `ClientBundleBuilder` (previously required) since `BackendBuilder` resolves it from `XmtpEnv`.

## Test plan

- [x] `cargo test -p xmtp_configuration` — XmtpEnv tests pass
- [x] `cargo check -p bindings_node` — compiles clean
- [x] `nix develop .#wasm -- cargo check -p bindings_wasm --target wasm32-unknown-unknown` — compiles clean
- [x] `dev/lint` — clean
- [x] All macro and mobile builder tests still pass

### DIFF
--- a/bindings/wasm/src/client/backend.rs
+++ b/bindings/wasm/src/client/backend.rs
@@ -9,16 +9,12 @@ pub struct BackendBuilder {
   #[builder(required)]
   pub env: XmtpEnv,
 
-  #[builder(optional)]
   api_url: Option<String>,
 
-  #[builder(optional)]
   gateway_host: Option<String>,
 
-  #[builder(optional)]
   pub readonly: Option<bool>,
 
-  #[builder(optional)]
   app_version: Option<String>,
 
   #[builder(skip)]
@@ -41,7 +37,7 @@ impl BackendBuilder {
   }
 
   #[wasm_bindgen]
-  pub fn build(&mut self) -> Result<Backend, JsError> {
+  pub fn build(mut self) -> Result<Backend, JsError> {
     let config = validate_and_resolve(
       self.env.into(),
       self.api_url.clone(),
--- a/crates/xmtp_api_d14n/src/config/backend_config.rs
+++ b/crates/xmtp_api_d14n/src/config/backend_config.rs
@@ -20,7 +20,8 @@ pub enum BackendConfigError {
 }
 
 fn is_url_secure(url: &str) -> bool {
-    url.starts_with("https://")
+    let lower = url.to_ascii_lowercase();
+    lower.starts_with("https://") || lower.starts_with("grpcs://")
 }
 
 pub fn validate_and_resolve(
@@ -187,6 +188,48 @@ mod tests {
         let config = validate_and_resolve(XmtpEnv::Local, None, None, false, None, true).unwrap();
         assert!(config.gateway_host.is_some());
         assert_eq!(config.gateway_host.as_deref(), Some(GrpcUrlsLocal::GATEWAY));
+    }
+
+    #[test]
+    fn test_is_url_secure_case_insensitive() {
+        let config = validate_and_resolve(
+            XmtpEnv::Dev,
+            Some("HTTPS://grpc.dev.xmtp.network:443".to_string()),
+            None,
+            false,
+            None,
+            false,
+        )
+        .unwrap();
+        assert!(config.is_secure);
+    }
+
+    #[test]
+    fn test_is_url_secure_grpcs() {
+        let config = validate_and_resolve(
+            XmtpEnv::Testnet,
+            None,
+            Some("grpcs://gateway.testnet.xmtp.network:443".to_string()),
+            false,
+            None,
+            false,
+        )
+        .unwrap();
+        assert!(config.is_secure);
+    }
+
+    #[test]
+    fn test_is_url_secure_grpcs_case_insensitive() {
+        let config = validate_and_resolve(
+            XmtpEnv::Testnet,
+            None,
+            Some("GRPCS://gateway.testnet.xmtp.network:443".to_string()),
+            false,
+            None,
+            false,
+        )
+        .unwrap();
+        assert!(config.is_secure);
     }
 
     #[test]
--- a/bindings/mobile/package.json
+++ b/bindings/mobile/package.json
@@ -1,0 +1,18 @@
+{
+  "name": "@xmtp/mobile",
+  "version": "1.0.0",
+  "description": "This crate provides cross-platform Uniffi bindings for XMTP v3.",
+  "license": "ISC",
+  "author": "",
+  "type": "commonjs",
+  "main": "index.js",
+  "directories": {
+    "example": "examples"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "dependencies": {
+    "koffi": "^2.15.1"
+  }
+}
--- a/bindings/node/src/client/backend.rs
+++ b/bindings/node/src/client/backend.rs
@@ -11,16 +11,12 @@ pub struct BackendBuilder {
   #[builder(required)]
   pub env: XmtpEnv,
 
-  #[builder(optional)]
   pub api_url: Option<String>,
 
-  #[builder(optional)]
   pub gateway_host: Option<String>,
 
-  #[builder(optional)]
   pub readonly: Option<bool>,
 
-  #[builder(optional)]
   pub app_version: Option<String>,
 
   #[builder(skip)]
@@ -28,6 +24,9 @@ pub struct BackendBuilder {
 
   #[builder(skip)]
   auth_handle: Option<AuthHandle>,
+
+  #[builder(skip)]
+  consumed: bool,
 }
 
 #[napi]
@@ -44,6 +43,13 @@ impl BackendBuilder {
 
   #[napi]
   pub fn build(&mut self) -> Result<Backend> {
+    if self.consumed {
+      return Err(napi::Error::from_reason(
+        "BackendBuilder has already been consumed by build()",
+      ));
+    }
+    self.consumed = true;
+
     let config = validate_and_resolve(
       self.env.into(),
       self.api_url.clone(),
--- a/bindings/wasm/Cargo.toml
+++ b/bindings/wasm/Cargo.toml
@@ -38,7 +38,11 @@ tsify = { version = "0.5", default-features = false, features = ["js"] }
 wasm-bindgen.workspace = true
 wasm-bindgen-futures.workspace = true
 wasm-streams = { version = "0.4" }
-web-sys = { workspace = true, features = ["Window", "DomException", "ReadableStream"] }
+web-sys = { workspace = true, features = [
+  "Window",
+  "DomException",
+  "ReadableStream",
+] }
 xmtp_api.workspace = true
 xmtp_api_d14n.workspace = true
 xmtp_common = { workspace = true, features = ["logging"] }
--- a/bindings/wasm/src/client.rs
+++ b/bindings/wasm/src/client.rs
@@ -10,7 +10,7 @@ use tracing_subscriber::{filter, fmt::format::Pretty};
 use tsify::Tsify;
 use wasm_bindgen::{JsValue, prelude::*};
 use xmtp_api_d14n::MessageBackendBuilder;
-use xmtp_db::{EncryptedMessageStore, EncryptionKey, StorageOption, WasmDb};
+use xmtp_db::{EncryptedMessageStore, StorageOption, WasmDb};
 use xmtp_id::associations::Identifier as XmtpIdentifier;
 use xmtp_mls::Client as MlsClient;
 use xmtp_mls::builder::DeviceSyncMode as XmtpDeviceSyncMode;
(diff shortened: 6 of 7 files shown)


neekolas:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3215?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3217** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3217?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3215** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3215?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3215?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3214** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3214?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3216** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3216?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=3215).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/xmtp_api_d14n/src/config/backend_config.rs
```diff
@@ -0,0 +1,205 @@
+use thiserror::Error;
+use xmtp_common::ErrorCode;
+use xmtp_configuration::{GrpcUrlsLocal, XmtpEnv};
```
insipx, line 3: how does this differ from `MessageBackendBuilder` and `ClientBundleBuilder` that already exist -- those should allow configuring the network/grpc backend separately from the client, similar to how the mobile bindings do it. A `MessageBackend` can be built from a `ClientBundle`, where a `ClientBundle` is only the network client piece, where a `MessageBackend` just includes the addition of a cursor store.

insipx: Maybe we can unify all the builders for backend + client into one crate that all the bindings use to share more of the code, and clarify which definitions are used where. each builder can have library-specific gotchas behind a feature flag, that way we can share more code between each library-specific builder

bindings/wasm/src/client/backend.rs
```diff
@@ -0,0 +1,163 @@
+use super::XmtpEnv;
+use super::gateway_auth::{AuthCallback, AuthHandle};
+use std::sync::Arc;
+use wasm_bindgen::prelude::*;
+use xmtp_api_d14n::{ClientBundleBuilder, validate_and_resolve};
+
+#[xmtp_macro::wasm_builder]
```
insipx, line 7: with this macro, how do the wasm_builder and node_builder differ? Is it possible that we can unify the builders completely so we have only one definition of `BackendBuilder` for node/wasm/mobile, or is there some quirk in each bindings library that prevents this

further, is this builder something we can re-use [`derive_builder`](https://docs.rs/derive_builder/latest/derive_builder/index.html) or [`bon`](https://docs.rs/bon/latest/bon/) for to avoid maintaining our own macro?

neekolas: The big thing this custom one does is:  
  
1\. Adds all the correct `#[napi]`  `#[uniffi]` etc tags  
2\. Customizes and wraps the return types to play nicely with the restrictions of each bindgen (arcs, etc)  
  
I did try to make it work with `derive_builder` first and got stuck. I was planning on giving it another go before admitting defeat and going the custom route.

crates/xmtp_api_d14n/src/config/backend_config.rs
neekolas: This is effectively a wrapper around the ClientBundleBuilder, just with all the bindings-specific junk in it.  
  
The only feature-addition I made was resolving the host addresses from envs, and some extra validations...but we could move that logic to the ClientBundleBuilder if it was universally desirable.  
  
Will keep fiddling and see if I can find a way to make it more shareable between languages.

bindings/wasm/src/client/backend.rs
neekolas: This wasn't the only problem, but it was one of the challenges with `derive_builder`.  
  
NAPI requires `fn set_x(&mut self, this: This<'scope>, val: T) -> This<'scope> ` if you want to do chaining. There's a magic extra `this` parameter
injected by the runtime. `derive_builder` isn't well set-up to add extra parameters to setter functions so we would have to hack something together.  
  
The other thing that the home-brewed solution has going for it is that there is only one type. `derive_builder` always gives you `Foo` and `FooBuilder` as two structs. But we sometimes need annotations on both if we want to use them over the FFI boundary (not in this particular case, but probably the case if we wanted builders for options structs that become parameters to functions). We need both of them to be Uniffi/Napi/WasmBindgen compatible.

crates/xmtp_api_d14n/src/config/backend_config.rs
neekolas: I had Claude do some deep research on how viable it would be to use something off the shelf like `derive_builder`.

Here's the report:

## Context

The custom builder macro system in `crates/xmtp_macro/src/builder.rs` generates binding-annotated structs with constructors and chainable
setters for three targets: NAPI (Node.js), WASM, and UniFFI (mobile). The question is whether `derive_builder` can replace this bespoke
implementation, even if we allow two structs (builder + target) and a custom `build()` method.

## Re-evaluation with Relaxed Constraints

Accepting that:

- Two structs is OK (builder + produced type)
- Custom `build()` producing a different type is OK
- Method chaining in JS is required
- Both the builder struct AND produced type must be binding-compatible

### How derive_builder works

`derive_builder` is a **derive macro**: you annotate the TARGET struct with `#[derive(Builder)]`, and it generates a `FooBuilder` struct
with `Option<T>` fields, setters, and a `build() -> Result<Foo>` method.

It supports:

- `builder_struct_attr(...)` — add attributes to the generated builder struct
- `builder_impl_attr(...)` — add attributes to the generated impl block
- `builder_setter_attr(...)` — add attributes to individual setter methods
- `build_fn(skip)` — skip auto-generated `build()` for custom implementation
- `pattern = "owned"` — consuming `self -> Self` setters (needed for WASM)
- `pattern = "mutable"` — `&mut self -> &mut Self` setters
- `setter(strip_option)` — setter takes `T` for `Option<T>` fields

### Feature-parity assessment

| Need | derive_builder | Feasibility |
| --- | --- | --- |
| Optional setters take `T`, store `Some(T)` | `strip_option` | Works |
| Default field initialization | `default = "expr"` | Works |
| Skip fields | `setter(skip)` | Works |
| Custom `build()` | `build_fn(skip)` | Works |
| Consuming setters for WASM | `pattern = "owned"` | Works |
| `&mut self` setters for UniFFI | `pattern = "mutable"` | Works (partially) |
| Struct annotations on builder | `builder_struct_attr` | Works |
| Impl block annotations on builder | `builder_impl_attr` | Works |
| Required params in constructor | Not supported | Wrapper macro generates `new()` on builder — **workaround possible** |
| Per-setter `js_name = camelCase` (WASM) | Static per-field only | Wrapper macro computes and adds — **workaround possible** |
| **NAPI** **`This<'scope>`** **setter signature** | **Impossible** | **No way to add extra params to setters** |
| **UniFFI split impl blocks** | **Impossible** | **derive_builder emits one impl block** |

### Two hard blockers remain

#### 1\. NAPI `This<'scope>` setter signatures

NAPI-RS requires setters to accept an injected `This<'scope>` parameter for JS method chaining:

```rust
pub fn set_port<'scope>(
    &'scope mut self,
    this: ::napi::bindgen_prelude::This<'scope>,  // injected by NAPI runtime
    port: u32,
) -> ::napi::bindgen_prelude::This<'scope> {
    self.port = port;
    this
}

derive_builder generates setters with fixed signatures (fn foo(&mut self, val: T) -> &mut Self or fn foo(self, val: T) -> Self). There is
 no hook to inject additional parameters.

Could we use consuming setters instead? Theoretically self -> Self avoids needing This<'scope>, but NAPI-RS wraps Rust objects in
reference-counted pointers. Consuming self invalidates the JS reference. The This<'scope> pattern exists precisely to avoid this.

2. UniFFI split impl blocks

UniFFI wraps #[uniffi::export] objects in Arc<Self>, making &mut self methods incompatible with the export annotation. The current macro
places the constructor in one impl block (with #[uniffi::export]) and setters in a separate plain impl block:

#[uniffi::export]
impl Foo {
    #[uniffi::constructor]
    pub fn new(host: String) -> Self { ... }
}

// Separate block — NO #[uniffi::export]
impl Foo {
    pub fn port(&mut self, port: u32) -> &mut Self { ... }
}

derive_builder puts all methods in a single impl block. There is no configuration to split them. Note: the plain-impl setters are only
callable from Rust, not from Swift/Kotlin — UniFFI can't export &mut self methods on Arc-wrapped objects regardless.

What COULD work: WASM-only

For WASM alone, a wrapper macro approach is feasible:

// User writes:
#[wasm_builder]
pub struct ClientConfig {
    #[builder(required)]  host: String,
    #[builder(optional)]  port: Option<u32>,
    #[builder(default = "true")]  debug: bool,
}

// Wrapper macro emits:
#[derive(Builder)]
#[builder(pattern = "owned", build_fn(skip))]
#[builder_struct_attr(wasm_bindgen(getter_with_clone))]
#[builder_impl_attr(wasm_bindgen)]
pub struct ClientConfig {
    host: String,
    #[builder(setter(strip_option, name = "set_port"))]
    #[builder_setter_attr(wasm_bindgen(js_name = "setPort"))]
    port: Option<u32>,
    #[builder(default = "true", setter(name = "set_debug"))]
    #[builder_setter_attr(wasm_bindgen(js_name = "setDebug"))]
    debug: bool,
}

The wrapper macro would compute and inject js_name camelCase attributes per field. A custom new() constructor on the builder would accept
 required fields. A manual build() would produce whatever type is needed.

But: using derive_builder for WASM while keeping custom code for NAPI and UniFFI is strictly worse than one consistent custom
implementation. It adds a dependency, splits the mental model, and the wrapper macro doing all the attribute transformation is nearly as
complex as just generating the code directly.

Verdict

derive_builder cannot achieve feature parity across all three binding targets. The NAPI This<'scope> setter signature and UniFFI
split-impl-block requirements have no workaround within derive_builder's architecture.

Using it for WASM alone while keeping custom code for NAPI and UniFFI would add complexity without reducing it — the wrapper macro doing
attribute transformation is comparable in complexity to the current code generation.

Recommendation: Keep the current custom macro

The existing builder.rs (~420 lines) is well-structured:
- AnnotationConfig cleanly separates binding concerns from builder logic
- SetterStyle enum handles the three different chaining patterns
- setter_impl_ann option handles UniFFI's split impl blocks
- setter_ann callback handles per-setter dynamic annotations

The bespokeness is the feature — it exists because cross-platform FFI builder generation has requirements no off-the-shelf crate
addresses.
```

neekolas: I do think there is some world where we have a `bindings_common` crate that has single root structs for common pieces of config and we use features to derive the right bindings types depending on the bindgen. That feels very doable, and doesn't pollute our library code with bindings-specific constructs and limitations.  
  
Would still be a little ugly when you need to have different field/argument types for each bindgen, referencing other types from your bindings. For example, Node, WASM, Mobile all are going to have different types of `AuthCallback`

insipx: that makes sense, we should try to bring as much of the shared config out of the bindings itself


my PR stack for the migration client [is also changing some of this](https://github.com/xmtp/libxmtp/pull/3223) to make working with the clients a bit easier (removes the generics from the root `Client` trait) that should make it easier to pass around just a `MessageBackend` in the bindings, rather than a client bundle and message backend

PR #3224, opened 2026-02-21
MLS validation server map gRPC errors

### LINKED ISSUE #3130: MLS validation service always return InvalidArgument gRPC code on error
All methods return an `Status` error, which translates over the wire always to gRPC InvalidArgument code, no matter the nature of the error.
We should be able to determine wether an error is a validation error, network, I/O, etc.

```
impl From<GrpcServerError> for Status {
    fn from(err: GrpcServerError) -> Self {
        Status::invalid_argument(err.to_string())
    }
}

async fn get_association_state(
     &self,
     request: Request<GetAssociationStateRequest>,
) -> Result<Response<GetAssociationStateResponse>, Status> 
```

fbac:
Closes #3130


neekolas:
We'd want to look closely at the node-go and client validation here.

I don't recall the details, and I'm sure it's changed since the last time I looked, but we do look at error responses in the client to decide if it's retryable or not. That has big implications for whether someone's identity has "failed to lookup" or "is invalid".

PR #3266, opened 2026-03-03
Consolidate client monitoring and Rust-based tests back into libxmtp

api-Hypernova:
### Add client monitoring, metrics collection, and e2e test scenarios to xdbg
- Adds a new `test` CLI subcommand in [test.rs](https://github.com/xmtp/libxmtp/pull/3266/files#diff-c0b40ffed873d8c7c7bbed479e13454800a205a441b1d06876ba548380611dca) that runs end-to-end `MessageVisibility` and `GroupSync` scenarios, measuring and recording latencies.
- Adds a [metrics module](https://github.com/xmtp/libxmtp/pull/3266/files#diff-cc69babee1aa06ccc7c5d1252dd416ebf666f1d3401882ca0ce1c70aa9b29401) that records Prometheus metrics (latency, throughput, member count) and pushes to a Pushgateway when `PUSHGATEWAY_URL` is set; CSV lines are always emitted.
- Instruments group creation, message sending, and identity generation in `generate/` with per-operation timing and optional pausing via `XDBG_LOOP_PAUSE`.
- Adds a containerized deployment in [apps/xmtp_debug/docker/](https://github.com/xmtp/libxmtp/pull/3266/files#diff-69d75fe1d5be2f75ee9216333e4917c52a06f08b17ce6bec9cd61b6e2725e88b) with a multi-stage Dockerfile, `entrypoint.sh` loop, `web-healthcheck.sh`, and a local metrics stack (Prometheus + Pushgateway + Grafana).
- Fixes a negation bug in `clear_network()` that caused wrong entries to be removed, and fixes `is_read_only()` on `None` to return `false`.
- Behavioral Change: `XDBG_DB_ROOT` env var now overrides the data directory; the staging backend URL is corrected to use the staging gRPC endpoint instead of dev.

#### Changes since #3266 opened

- Reformatted logging macro invocations across multiple functions in the `xmtp_debug` app to use multi-line style with trailing commas [d4ca102]
- Condensed function signature formatting for `generate::identity::poll_association_readiness` [d4ca102]
- Modified `Cargo.lock` file [d4ca102]
- Expanded and restructured `xmtp_debug` tool documentation [49a19f9]

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized 937e0b8.</sup>

### DIFF
--- a/.github/workflows/test-workspace.yml
+++ b/.github/workflows/test-workspace.yml
@@ -15,7 +15,8 @@ jobs:
         with:
           github-token: ${{ github.token }}
           cachix-auth-token: ${{ secrets.CACHIX_AUTH_TOKEN }}
-          sccache: "true"
+          sccache: "false"
+          gc-max-store-size: "5G"
       - uses: taiki-e/install-action@just
       - name: Start backend
         run: just backend up
@@ -23,9 +24,11 @@ jobs:
         if: failure()
         uses: jwalton/gh-docker-logs@v2
       - name: Run tests (v3 + d14n with coverage)
-        run: just test
-      - name: Generate coverage report
-        run: nix develop .#rust --command cargo llvm-cov report --output-path lcov.info --lcov
+        # use `nix build` instead of `nix flake check` because we need the resulting coverage
+        run: nix build .#checks.x86_64-linux.nextest-v3 .#checks.x86_64-linux.nextest-d14n
+
+      - name: merge coverage files
+        run: nix run nixpkgs#lcov -- --add-tracefile result/coverage --add-tracefile result-1/coverage --output-file lcov.info
       - name: Upload coverage
         uses: codecov/codecov-action@v5
         with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ members = [
 ]
 exclude = ["apps/android"]
 
+# Default members must be part of workspace hack to maximize cache hits in CI/nix/crane
 default-members = [
   # Applications
   "apps/mls_validation_service",
--- a/apps/mls_validation_service/src/cached_signature_verifier.rs
+++ b/apps/mls_validation_service/src/cached_signature_verifier.rs
@@ -70,19 +70,19 @@ mod tests {
         MultiSmartContractSignatureVerifier, SmartContractSignatureVerifier, ValidationResponse,
         VerifierError,
     };
-    use xmtp_id::utils::test::{SignatureWithNonce, SmartWalletContext, smart_wallet};
+    use xmtp_id::utils::test::{SignatureWithNonce, SmartWalletContext, docker_smart_wallet};
 
     #[rstest::rstest]
     #[timeout(Duration::from_secs(60))]
     #[tokio::test]
-    async fn test_is_valid_signature(#[future] smart_wallet: SmartWalletContext) {
+    async fn test_is_valid_signature(#[future] docker_smart_wallet: SmartWalletContext) {
         let SmartWalletContext {
             factory,
             owner0: owner,
             sw,
             sw_address,
             ..
-        } = smart_wallet.await;
+        } = docker_smart_wallet.await;
         let chain_id = factory.provider().get_chain_id().await.unwrap();
         let hash = B256::random();
         let replay_safe_hash = sw.replaySafeHash(hash).call().await.unwrap();
--- a/apps/mls_validation_service/src/handlers.rs
+++ b/apps/mls_validation_service/src/handlers.rs
@@ -318,14 +318,13 @@ mod tests {
     use xmtp_common::{rand_string, rand_u64};
     use xmtp_configuration::CIPHERSUITE;
     use xmtp_cryptography::XmtpInstallationCredential;
-    use xmtp_id::utils::test::smart_wallet;
     use xmtp_id::{
         associations::{
             Identifier,
             test_utils::{MockSmartContractSignatureVerifier, WalletTestExt},
             unverified::{UnverifiedAction, UnverifiedIdentityUpdate},
         },
-        utils::test::{SignatureWithNonce, SmartWalletContext},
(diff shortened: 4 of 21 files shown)


apps/xmtp_debug/concurrent-runner.sh
```diff
@@ -0,0 +1,177 @@
...
+}
+
+function generate_identities() {
+    local db_root=$1
+    mkdir -p "$db_root"
+    log "Generating identities at $db_root"
+    # Fast pass (no delay), then a conservative retry with 5s pause if needed
+    XDBG_LOOP_PAUSE=0 XDBG_DB_ROOT="$db_root" xdbg -d -b "${BACKEND}" generate --entity identity --amount 10 --concurrency 1 \
+        || { log "Identity generation failed, clearing and retrying"; clear_db "$db_root"; XDBG_LOOP_PAUSE=5 XDBG_DB_ROOT="$db_root" xdbg -d -b "${BACKEND}" generate --entity identity --amount 10 --concurrency 1; }
+}
```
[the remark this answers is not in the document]
api-Hypernova: fix it for me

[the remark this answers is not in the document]
api-Hypernova: try again

apps/xmtp_debug/Cargo.toml
```diff
@@ -34,9 +34,13 @@ num_cpus = "1.16.0"
 openmls.workspace = true
 openmls_rust_crypto.workspace = true
 owo-colors.workspace = true
+prometheus = "0.14"
 prost.workspace = true
 rand.workspace = true
 redb = { version = "3.1", features = ["logging"] }
+reqwest = { version = "0.12", features = [
```
insipx: Should use `reqwest.workspace` here

apps/xmtp_debug/src/app/generate.rs
```diff
@@ -38,24 +39,56 @@ impl Generate {
...
+                let elapsed = start.elapsed();
+
+                let total_send_ms: u128 = latencies.iter().map(|d| d.as_millis()).sum();
+                let avg_send_ms = if !latencies.is_empty() {
+                    total_send_ms as f64 / latencies.len() as f64
+                } else {
+                    0.0
+                };
+
+                println!(
```
insipx: Wondering if we need to figure out a better strategy for tracing logs -- why the preference to printing here rather than `tracing`?

apps/xmtp_debug/docker/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM rust:1-bookworm AS builder
+WORKDIR /src
+
+RUN apt-get update \
```
insipx, line 4: can use nix to make the build a bit smaller here

like from this blog post: https://mitchellh.com/writing/nix-with-dockerfiles

or this example dockerfile, and xdbg can be built with `nix build .#xdbg` and available in `result/bin/xdbg` one succeeds

not a blocker though, mostly a bonus

would roughly follow this dockerfile
```dockerfile
# Nix builder
FROM nixos/nix:latest AS builder

# Copy our source and setup our working dir.
COPY . /tmp/build
WORKDIR /tmp/build

# Build our Nix environment
RUN nix \
    --extra-experimental-features "nix-command flakes" \
    --option filter-syscalls false \
    build .#xdbg

# Copy the Nix store closure into a directory. The Nix store closure is the
# entire set of Nix store values that we need for our build.
RUN mkdir /tmp/nix-store-closure
RUN cp -R $(nix-store -qR result/) /tmp/nix-store-closure

# Final image is based on scratch. We copy a bunch of Nix dependencies
# but they're fully self-contained so we don't need Nix anymore.
FROM scratch

WORKDIR /app

# Copy /nix/store
COPY --from=builder /tmp/nix-store-closure /nix/store
COPY --from=builder /tmp/build/result /app
CMD ["/app/bin/xdbg"]
```

apps/xmtp_debug/src/app/generate/groups.rs
```diff
@@ -84,15 +92,122 @@ impl GenerateGroups {
                             );
                             ids.push(inbox_id);
                         }
-                        let client = client.lock().await;
-                        let group = client.create_group(Default::default(), Default::default())?;
+                        let member_count = ids.len();
+
+                        // -- create group --
+                        let t_create = Instant::now();
```
insipx: one thing that would be nice is separating this block into a one/two/three new functions, the generate stuff is becoming a bit long

apps/xmtp_debug/src/app/generate/identity.rs
```diff
@@ -177,21 +193,96 @@ impl GenerateIdentity {
...
                 tokio::spawn({
                     let sem = semaphore.clone();
+                    let network_clone = network.clone();
                     async move {
                         let _permit = sem.acquire().await?;
-                        let id = hex::encode(identity.inbox_id);
-                        trace!(inbox_id = id, "getting association state");
+                        let inbox_id_hex = hex::encode(identity.inbox_id);
+                        trace!(inbox_id = inbox_id_hex, "getting association state");
+
```
insipx, line 206: same here, it might be good to separate into a new fn

tylerhawkes:
Yeah, using tracing spans is the way to go here, because it gives a timestamp of when it is created and when it stops automatically. You can emit it as an event if want when a span closes depending on the subscriber.

apps/xmtp_debug/docker/Dockerfile
api-Hypernova: Good call. The Nix infra for this already exists in the repo (crane + musl-docker.nix). Will follow up in a separate PR to switch to dockerTools.buildLayeredImage, should cut the image from 172MB to under 50MB

apps/xmtp_debug/src/app/generate.rs
api-Hypernova: Done!

PR #3284, opened 2026-03-04
remove Associated types from `Client` trait, add `MigrationClient` to d14n

insipx:
### Make `MultiNodeClient` generic over gateway client type and unify API error types
- Refactors `MultiNodeClient` into `MultiNodeClient<T>` so the gateway client is a generic parameter rather than a fixed `GrpcClient`, enabling different gateway implementations to be injected.
- Removes all generic error parameters from `ApiClientError`, `Client`, and related traits; lower-level errors are now wrapped in a new concrete `NetworkError` type.
- Replaces `ClientBundle<Err>` and `ClientKind` with a concrete `ClientBundle` enum (`V3`, `D14n`, `Migration`) carrying typed `ArcClient` instances; the bundle no longer proxies client operations.
- Adds `build_v3()`, `build_d14n()`, and `build_optional_d14n()` to `ClientBundleBuilder`, with `build_optional_d14n()` selecting d14n or v3 based on gateway host presence; mobile and WASM bindings now call `build_optional_d14n()` instead of `build()`.
- Introduces `MigrationClient<V3, D14n, Store>` (replacing `CombinedD14nClient`) with write-retry logic that detects cutover via a regex match on network errors and re-routes to the d14n backend.
- Risk: `Client::fake_stream()` is now a default trait method returning a `FakeEmptyStream`; previous overrides in `GrpcClient` and `MultiNodeClient` are removed.

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 342416e. 53 files reviewed, 9 issues evaluated, 1 issue filtered, 3 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>apps/xmtp_debug/src/args.rs — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 420](https://github.com/xmtp/libxmtp/blob/342416e7f30c952e30ede78d543e8944c32c9b0d/apps/xmtp_debug/src/args.rs#L420): The `is_secure` variable is derived from `self.network_url().scheme()` at line 416, but it's used to configure TLS for the gateway client which uses `self.xmtpd_gateway_url()` at lines 419-420. These are different URLs that could have different schemes. If the network URL is `http://...` but the gateway URL is `https://...`, the gateway client will incorrectly have TLS disabled, causing connection failures or security issues. The `is_secure` check should be based on `self.xmtpd_gateway_url()?.scheme()` for the gateway client. <b>[ Out of scope ]</b>
</details>

</details>

PR #3285, opened 2026-03-04
[breaking] remove is_secure from all external builders, simplify client builder

insipx:
- Removes "`is_secure`" from all client builders/constructors. Generally feedback has been that `is_secure` is confusing, most expect tls to be enabled with an `https://` url. This PR detects TLS based on `https`/`grpcs` for everything related to gRPC. It also passes the given url through `url::Url::parse` and returns an error if a URL is not a well-formed URL.
- combines the validation/builder logic in `backend_config` with `ClientBundleBuilder`. Eventually would be nice to unify ClientBundleBuilder and MessageBackendBuilder, which should be possible after D14n when we can delete all the separate client impls
gg

> [!NOTE]
> ### Remove `is_secure` from all external builders and infer TLS from URL scheme
> - Removes the explicit `is_secure`/`isSecure` parameter from all public builder APIs (`MessageBackendBuilder`, `ClientBundleBuilder`, `GrpcClient`, mobile/Node/WASM FFI), replacing it with automatic TLS detection based on URL scheme (`https`/`grpcs` = TLS, otherwise plain).
> - Updates `NetConnectConfig` trait: `set_host` now accepts `url::Url` and `set_tls` is removed; all implementors derive TLS from the URL internally.
> - Deprecates `isSecure` in Android and iOS `ClientOptions.Api` for source compatibility, and removes it from API client cache keys.
> - `MigrationClient.upload_key_package` now executes via `write_with_refresh` to retry on failure.
> - Risk: Invalid URLs that were previously accepted will now surface as errors earlier (from `ClientBundleBuilder`), and callers that relied on overriding TLS independently of the URL will break.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 37ea781.</sup>
> 
>


insipx:

1. `trunk()`
1. [insipx/generic-multinode](https://github.com/xmtp/libxmtp/pull/3284)
1. **insipx/remove-tls ← this PR**
1. [insipx/migration-tests](https://github.com/xmtp/libxmtp/pull/3290)
1. [insipx/stream-restart](https://github.com/xmtp/libxmtp/pull/3297)
1. [insipx/migration-client-stream-tests](https://github.com/xmtp/libxmtp/pull/3298)

---

crates/xmtp_api_d14n/src/queries/combined/connected_check.rs
```diff
@@ -0,0 +1,16 @@
...
+
+#[xmtp_common::async_trait]
+impl<V3, D14n, Store> IsConnectedCheck for MigrationClient<V3, D14n, Store>
+where
+    V3: IsConnectedCheck,
+    D14n: IsConnectedCheck,
+    Store: MaybeSend + MaybeSync,
+{
+    async fn is_connected(&self) -> bool {
+        self.v3_client.is_connected().await && self.xmtpd_client.is_connected().await
```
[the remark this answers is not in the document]
insipx: the v3 backend will still be reachable, but read only until the next libxmtp update which will remove the v3 client totally

bindings/mobile/src/mls.rs
```diff
@@ -137,7 +137,6 @@ impl XmtpApiClient {
 pub async fn connect_to_backend(
     v3_host: String,
     gateway_host: Option<String>,
-    is_secure: bool,
```
tylerhawkes, line 140: Instead of breaking this (I think this is the public api), mark it as deprecated.

bindings/node/src/client/create_client.rs
```diff
@@ -184,7 +184,6 @@ async fn create_client_inner(
 pub async fn create_client(
   v3_host: String,
   gateway_host: Option<String>,
-  is_secure: bool,
```
tylerhawkes, line 187: Same here. Mark it as deprecated and do a `let _ = is_secure` in the body if needed.

crates/xmtp_api_grpc/src/grpc_client/client.rs
```diff
@@ -279,22 +273,19 @@ impl ApiBuilder for ClientBuilder {
 }
 
 impl GrpcClient {
-    pub fn create(host: &str, is_secure: bool) -> Result<Self, GrpcBuilderError> {
+    pub fn create(host: &str) -> Result<Self, GrpcBuilderError> {
```
tylerhawkes: Can this take `Url` instead of `&str`? We could change the const values to be static and parse them lazily if needed.

bindings/mobile/src/mls.rs
insipx: since no one depends on the rust SDK, I removed it but did the deprecated in ios/android

[android](https://github.com/xmtp/libxmtp/pull/3285/changes#diff-6a1a4efe485f82ebc6bac85479e71e7de7a9756b0e39be79db3cdf1d6723d937)


[ios](https://github.com/xmtp/libxmtp/pull/3285/changes#diff-fda53e366d262601ad6a3c0d242a4e628b446a19baec1687abfd484b6cb1efc4R80)

crates/xmtp_api_grpc/src/grpc_client/client.rs
insipx: absolutely

PR #3339, opened 2026-03-17
feat: add client-side envelope validation

### LINKED ISSUE #1931: Client-Side Envelope Validation
Envelopes in d14n must be validated, according to the [validation rules in XIP-49](https://github.com/xmtp/XIPs/blob/main/XIPs/xip-49-decentralized-backend.md#334-validation-rules)

Validation can be implemented as a series of extractors or a singular extractor. This extractor must inspect the OriginatorEnvelope, and ensure the AAD and payload is as expected. 

Example Implementation:

```rust
/// Validate Envelopes when querying the network, according to the constraints specified by XIP-49:
/// Verify the originator_sequence_id is strictly increasing per originator.
///    Verify the originator_signature or node_signature on each payload.
///    Verify the client envelope AAD on each payload:
///        The target topic matches the actual topic.
///    Verify the client envelope AAD was correctly authenticated ([3.3.5. Cross-originator message ordering](https://github.com/xmtp/XIPs/blob/main/XIPs/xip-49-decentralized-backend.md#335-cross-originator-message-ordering))
///    Verify that the payloads are causally consistent ([3.3.5. Cross-originator message ordering](https://github.com/xmtp/XIPs/blob/main/XIPs/xip-49-decentralized-backend.md#335-cross-originator-message-ordering)).
///    Perform payload-specific validation, as defined in previous XIP's.
///    (Optional) Query multiple nodes simultaneously to ensure no payloads were omitted.
///    (Optional) Query the blockchain directly to ensure that blockchain payloads were not doctored.
pub struct QueryEnvelopeValidator;

pub enum ValidationError {
   BadAAD,
   InvalidSignature,
   // etc, etc.
}
impl EnvelopeVisitor<'_> for QueryEnvelopeValidator {
    type Error = ValidationError;

    fn visit_unsigned_originator(
        &mut self,
        envelope: &UnsignedOriginatorEnvelope,
    ) -> Result<(), Self::Error> {
        // inspect the unsignedOriginatorEnvelope
        Ok(())
    }

    fn visit_unsigned_originator(
        &mut self,
        _e: &UnsignedOriginatorEnvelope,
    ) -> Result<(), Self::Error> {
        tracing::debug!("visit_unsigned_originator");
        // Validate Blockchain Proof
        Ok(())
    }

    fn visit_client(&mut self, _e: &ClientEnvelope) -> Result<(), Self::Error> {
        tracing::debug!("visit_client");
        // validate the AAD
        Ok(())
    }
}

```

Once these validators are complete, they may be run individual on an envelope, ie

```rust
let mut extractors = (QueryEnvelopeValidator::new(), AADValidator::new(), PayerValidator::new());
self.accept(&mut extractors)?;

```

can also create a higher-level `CompleteEnvelopeValidation` which just contains validators and can be run on its own, or any combination of stuff to validate envelopes.

The validators can be run in. the queries/d14n/mls.rs traits/identity traits or wherever protocol envelopes need to be validated.

vlopes11:
This commit introduces xmtp_api_d14n/src/protocol/validators/mod.rs, which performs client-side validations as defined in XIP-49.

It injects the validation logic into Envelope::client_envelope, so whenever a new ClientEnvelope is created, it is automatically validated.

The test coverage extends the currently existing encode/decode structure, as the validation will always be called by default. It introduces the need to compute appropriate originator ns, topic structure, among others, for some of the test cases. While we aim to have a testing environment with the least possible overhead, it is crucial to consider such validations to achieve type-safety.

Closes #1931

> [!NOTE]
> ### Add client-side envelope validation to `Envelope::client_envelope()`
> - Introduces a new `EnvelopeValidator` visitor in [`validators/mod.rs`](https://github.com/xmtp/libxmtp/pull/3339/files#diff-3eed2d6393d0db55ada504daf5d4b6a0a5827ded513faf200e11b64909bcd23e) that checks originator timestamps are within 30 minutes of now and that `ClientEnvelope` authenticated data contains a non-empty, parseable `target_topic`.
> - Adds a `validate` method to the `Envelope` trait; the blanket impl for `ProtocolEnvelope` runs `EnvelopeValidator` and `client_envelope()` now calls `validate()` before returning.
> - Adds a `Validation(ValidationError)` variant to `EnvelopeError`, marked non-retryable.
> - Updates `TestEnvelopeBuilder` defaults to use a non-zero `originator_ns` (current time) and a default topic, matching the new validation requirements.
> - Risk: `client_envelope()` now returns a `Validation` error for envelopes older than 30 minutes or missing authenticated data, which is a breaking behavioral change for callers that previously accepted such envelopes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4a3e6be.</sup>
> 
>
>


crates/xmtp_api_d14n/src/protocol/traits/envelopes.rs
```diff
@@ -124,7 +126,15 @@ where
...
     }
 
+    fn validate(&self) -> Result<(), EnvelopeError> {
+        let mut validator = EnvelopeValidator::default();
+        self.accept(&mut validator)?;
+        Ok(())
+    }
+
     fn client_envelope(&self) -> Result<ClientEnvelope, EnvelopeError> {
+        self.validate()?;
```
insipx, line 135: after thinking about this a bit more, I think it would be safer to have this as a [combinator](https://xmtp.github.io/libxmtp/xmtp_api_d14n/queries/trait.D14nCombinatorExt.html) rather than on the `Envelope` trait implementation

However, the QueryExt/combinator stuff suffers from some type-system shortcomings in rust but also the implementation could be improved to allow for better nesting. so if it's not possible, we could keep it in `client_envelope`, although I would love your thoughts on the current system:


Here's my reasoning:
- having it in `client_envelope` is implicit -- it runs the possibility of the caller following a different path of querying envelopes and ending up with non-validated envelopes
- it's possible that other places in the stack, callers of `client_envelope` may not expect the envelope validation, for instance in the sort methods

As a combinator:
- Combinators should have available to them an opaque `Envelope` type
- combinator implentations are admittedly type-generic heavy, but they do isolate the validation in a nice way

Here's what I mean --


For context, this API design is loosesly/heavily based on [This blogpost by gitlab api author](https://plume.benboeckel.net/~/JustAnotherBlog/designing-rust-bindings-for-rest-ap-is) where the network `Client` `Endpoint`'s and `Query`'s are decoupled from each other

Article summary is basically:

A `combinator` is an extension to the [`Query`](https://xmtp.github.io/libxmtp/xmtp_proto/api/trait.Query.html) trait. A Query is implemented for all [`Endpoint`](https://xmtp.github.io/libxmtp/xmtp_proto/api/trait.Endpoint.html) types.
- Endpoint: describes the form of a specific endpoint, what it looks like as a Rust type as well as a bag of bytes
- `Query` describes how an endpoint is called. Queries are special, because they can be chained and nested. The implementation of this in libxmtp so far IMO can be improved (esp once Rustlang adds specialization) but the way it is now works OK
- `Client` describes how the request is actually "made" and acted on at the lowest level. it abstracts away network specific details about the transport protocol like HTTP/gRPC/QUIC etc.

There are a few batteries included combinators that come with the xmtp_proto impl [here](https://xmtp.github.io/libxmtp/xmtp_proto/api/index.html)

for instance, the [Ignore](https://xmtp.github.io/libxmtp/xmtp_proto/api/fn.ignore.html) combinator ignores a query response,

```rust
let endpoint = MyEndpoint::build();
ignore(endpoint).query(&client).await; // this transforms MyEndpoint to `Ignore<MyEndpoint>` and now the endpoint returns `Result<(), Error>` instead of `Result<T, Error>` validated through the rust type system
```

the [`EndpointExt`](https://xmtp.github.io/libxmtp/xmtp_proto/api/trait.EndpointExt.html) trait is for convenience -- it allows us to avoid writing `ignore(endpoint)` and instead write it like `endpoint.ignore().query()`

the [`D14nCombinatorExt`](https://xmtp.github.io/libxmtp/xmtp_api_d14n/queries/trait.D14nCombinatorExt.html) does the same as `EndpointExt` but for combinators specific to the d14n implementation. so far, that is just the `Ordered` combinator. the `Ordered` combinator takes a bunch of Envelopes that result from a `Query` and ensures its dependencies are properly ordered according to XIP-49


the [ignore](https://xmtp.github.io/libxmtp/src/xmtp_proto/traits/combinators/ignore.rs.html#46-48) combinator is a good example -- it's probably the simplest. combinators are implemented on QueryRaw or query depending on if they need a formed type or not


this way, we can be more declarative in the flow where all the magic happens (in `query_group_message`). i.e the function goes from what it is now,

```rust
let response = QueryEnvelope::builder()
            .topic(topic)
            .last_seen(lcc)
            .limit(MAX_PAGE_SIZE)
            .build()?
            .ordered(resolver, topic_cursor, &self.cursor_store)
            .query(&self.client)
            .await?;
```

to 
```rust
let response = QueryEnvelope::builder()
            .topic(topic)
            .last_seen(lcc)
            .limit(MAX_PAGE_SIZE)
            .build()?
            .ordered(resolver, topic_cursor, &self.cursor_store)
            .validated() // <--- note the new validated function coming from `D14nCombinatorExt`
            .query(&self.client)
            .await?;
```

This will :
- do a base query for envelopes
- order them
- once ordered, validate them

IMO this is the correct level of abstraction -- it is ultimately part of the D14n spec and the caller of the Query endpoint should decide whether they want validated envelopes or "unvalidated"

vlopes11: The combinator approach sounds great, but I think it would imply we would need to have a type-safety oriented approach. That could imply a bigger refactor; the rationale of the original PR was to cause the minimum impact possible.

I'll list some of the pros & cons that I can think of:

### Visitor
#### Pros
- Minimal impact implementation (traversal already implemented, validation for free)
- Easy to compose validation passes by chaining visitors
- Scales well with big type complexity (I believe our case)
#### Cons
- We will end up performing the same validation multiple times, as the read-only back and forth traversal should not be limited by this new implementation
- The error handling can get ambiguous as we no longer can distinguish if the traversal itself failed, or if the contents are inconsistent
- We could have future trouble with the traversal order if we expect certain validations to be performed before certain type construction. While calling `self.accept` is easy for now, since we are validating just the ClientEnvelope, this will become more convoluted as we validate more types
### Combinators
#### Pros
- The manual call makes it easier to avoid duplicated validations and completely bypass the traversal order problem of the visitor approach
- We unlock type-safety by allowing the user to consume a ClientEnvelope to create a ValidatedClientEnvelope, so the functions can accept a ValidatedClientEnvelope when then are dealing with user data
- Easier error handling: we don't mix whatever validation error with the visitor tree error
#### Cons
- From a quick look, I think this will imply a much bigger refactor of the codebase as we will not be able to reuse the current visitor infrastructure we have setup
- This scales poorly if the type complexity grows. We will have these manual combinator traits that may be implemented for only certain types (it's hard to distinguish as a class-logic which should and should not be validated)

I think it's overall more idiomatic in Rust to have the combinator and type transformation (i.e. `let validated: ValidatedClientEnvelope = envelope.validate()?;`), but I can't see clearly how much of an impact on the current codebase we would have.

Issue #3162, opened 2026-02-11
Flaky test: testSyncConsent in Android HistorySyncTest


neekolas:
## Summary

The `testSyncConsent` test in `sdks/android/library/src/androidTest/java/org/xmtp/android/library/HistorySyncTest.kt` is flaky and has been skipped with `@Ignore`.

## Test Description

The test creates a group via `boClient`, syncs it to `alixClient`, verifies the initial consent state is `UNKNOWN`, then creates a second installation (`alixClient2`). After syncing, it updates the group's consent state to `DENIED` on `alixClient` and expects the state to propagate to `alixClient2` after a delay.

## Failure Details

```
> Task :library:connectedDebugAndroidTest
org.xmtp.android.library.HistorySyncTest > testSyncConsent[emulator-5560 - 14] FAILED 
    java.lang.AssertionError: expected:<UNKNOWN> but was:<DENIED>
    at org.junit.Assert.fail(Assert.java:89)
```

The initial consent state of the group on `alixClient` is expected to be `UNKNOWN` but is already `DENIED`, suggesting the consent state was set earlier than expected during the sync.

## Likely Cause

The consent sync between installations relies on timing-sensitive operations (`preferences.sync()` with fixed delays). The consent state may propagate faster than expected, arriving before the initial assertion checks.

## Current Mitigation

The test has been annotated with `@Ignore("Flaky: consent sync timing is non-deterministic")` to prevent CI failures.

## Resolution

- Investigate why the consent state is `DENIED` before the test expects it to be
- Consider using a callback/stream-based approach instead of fixed delays to detect consent changes
- Ensure the timing assumptions in the test match the actual sync behavior

## File

`sdks/android/library/src/androidTest/java/org/xmtp/android/library/HistorySyncTest.kt`

PR #3368, opened 2026-03-30
Fix flaky consent sync tests in HistorySyncTest

xmtp-coder-agent:
## Summary

Resolves https://github.com/xmtp/libxmtp/issues/3162

- **`testSyncConsent`**: Replaced fixed `delay(1000)` + `delay(4000)` with a polling loop that actively syncs preferences and checks consent state (500ms interval, 15s timeout)
- **`testV3CanMessageV3`**: Added polling for initial ALLOWED consent state propagation from the creating installation, which was previously assumed to be immediate
- Removed both `@Ignore("Flaky: consent sync timing is non-deterministic")` annotations to re-enable the tests
- Removed unused `org.junit.Ignore` import

## Test plan

- [ ] Run `testSyncConsent` on Android emulator — should pass reliably without timing-dependent failures
- [ ] Run `testV3CanMessageV3` on Android emulator — should pass reliably with polling-based consent state checks
- [ ] Run full `HistorySyncTest` suite to verify no regressions in other tests

> [!NOTE]
> ### Fix flaky consent sync tests in `HistorySyncTest` by replacing fixed delays with polling
> - Adds a `waitUntil(timeoutMs, intervalMs, condition)` helper in [HistorySyncTest.kt](https://github.com/xmtp/libxmtp/pull/3368/files#diff-33ca4980924b8d59d5ffc62a9e4dd6a5a30413b2183b9a946f9bba9faa3ac79f) that polls a suspendable condition up to a timeout instead of relying on fixed sleeps.
> - Re-enables `testSyncConsent` and `testV3CanMessageV3` (previously `@Ignore`d) and rewrites both to use `waitUntil` loops that repeatedly call `syncAllConversations` and `preferences.sync` until group discovery and consent propagation are confirmed.
> - Both tests now poll up to 30s for `ConsentState.ALLOWED` and `ConsentState.DENIED` transitions rather than asserting on potentially stale state.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3b878bf.</sup>
> 
>
>

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -29,7 +29,7 @@ jobs:
           submodules: recursive
 
       - name: Setup Pages
-        uses: actions/configure-pages@v5
+        uses: actions/configure-pages@v6
 
       - uses: ./.github/actions/setup-rust
 
@@ -53,4 +53,4 @@ jobs:
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v4
+        uses: actions/deploy-pages@v5
--- a/.github/workflows/docs-ios.yml
+++ b/.github/workflows/docs-ios.yml
@@ -50,11 +50,11 @@ jobs:
         working-directory: sdks/ios
         run: bundle exec jazzy --output ./docs --theme=fullwidth --module=XMTPiOS
       - name: Setup Pages
-        uses: actions/configure-pages@v5
+        uses: actions/configure-pages@v6
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v4
         with:
           path: "sdks/ios/docs"
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v4
+        uses: actions/deploy-pages@v5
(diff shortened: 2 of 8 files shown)


sdks/android/library/src/androidTest/java/org/xmtp/android/library/HistorySyncTest.kt
```diff
@@ -49,21 +47,36 @@ class HistorySyncTest : BaseInstrumentedTest() {
             val state = alixClient2.inboxState(true)
             assertEquals(state.installations.size, 2)
 
-            // Sync all of the first client's conversations to add Alix2
-            alixGroup.sync()
-
-            alixClient2.conversations.sync()
+            // Fully sync both installations to establish device sync groups
+            alixClient.conversations.syncAllConversations()
+            delay(2000)
```
neekolas: Can we make these tests work without any fixed delays at all? Delays are brittle. Use helpers to wait for conditions instead

xmtp-coder-agent: Done. Added a `waitUntil` polling helper that takes a condition lambda and retries with a 500ms interval up to a 30s timeout. Both `testSyncConsent` and `testV3CanMessageV3` now have zero fixed delays — all sync setup and consent propagation waiting is condition-based, calling `syncAllConversations()` and `preferences.sync()` each iteration until the expected state is reached.

PR #3369, opened 2026-03-30
update nix flake & rust version

insipx:
also confirmed the updated `swiftformat` works both on linux & darwin so was able to remove the custom pkg

> [!NOTE]
> ### Update Nix flake and Rust toolchain to 1.94.1 with nix-fast-build integration
> - Bumps Rust toolchain from 1.94.0 to 1.94.1 in [flake.nix](https://github.com/xmtp/libxmtp/pull/3369/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0).
> - Replaces `nix build` invocations across CI workflows and `just` recipes with `nix-fast-build`, which skips cached builds and uses custom out-link names (e.g. `v3-`, `d14n-`, `result-`).
> - Adds `AWS_LC_SYS_TARGET_CC/CXX` env vars in [nix/lib/base.nix](https://github.com/xmtp/libxmtp/pull/3369/files#diff-78e27920e7e7f148052510551d065fae07e34ca9a88df0abd2450dedb25668bc) to use the native host compiler for `aws-lc-sys` build checks during cross-compilation.
> - Bumps `alloy` from 1.3 to 2, `ctor` from 0.6 to 0.10, and `gloo-timers` from 0.3 to 0.4 in [Cargo.toml](https://github.com/xmtp/libxmtp/pull/3369/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542); updates workspace-hack accordingly.
> - Strips vendored cargo source store paths from built NAPI `.node` binaries via `remove-references-to` in [nix/lib/napiBuild.nix](https://github.com/xmtp/libxmtp/pull/3369/files#diff-97105a706805ee749b47f57d475da65a8dd712cc01c754967d45667181d2ad56), reducing closure size.
> - Sets `doInstallCargoArtifacts=false` across most Nix package derivations to use explicit packaging steps instead of automatic artifact installation.
> - Risk: Coverage merge steps in [test-workspace.yml](https://github.com/xmtp/libxmtp/pull/3369/files#diff-bb6ea121a02e5dd13a93185f52a448c5211138550427916dd66d2d75bc04af34) now expect artifacts at `v3-/coverage` and `d14n-/coverage` instead of `result/coverage` and `result-1/coverage`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7e53b7d.</sup>
> 
>

### DIFF
--- a/nix/package/ios.nix
+++ b/nix/package/ios.nix
@@ -61,23 +61,16 @@ let
     let
       envSetup = iosEnv.envSetup target;
     in
-    rust.buildDepsOnly (
-      commonArgs
-      // {
-        pname = "xmtpv3-deps-${target}";
-        CARGO_BUILD_TARGET = target;
-        __noChroot = true;
-        cargoExtraArgs = "--target ${target} -p xmtpv3";
-        # envSetup is inlined in buildPhaseCargoCommand because crane's buildDepsOnly
-        # strips preBuild hooks (it needs full control of the build phase to replace
-        # source files with dummies). envSetup dynamically resolves the Xcode path
-        # via xcode-select and sets DEVELOPER_DIR, SDKROOT, CC/CXX, and bindgen args.
-        buildPhaseCargoCommand = ''
-          ${envSetup}
-          cargo build --locked --release --target ${target} -p xmtpv3
-        '';
-      }
-    );
+    xmtp.base.mkCargoArtifacts rust false {
+      pname = "xmtpv3-deps-${target}";
+      CARGO_BUILD_TARGET = target;
+      __noChroot = true;
+      cargoExtraArgs = "--target ${target} -p xmtpv3";
+      buildPhaseCargoCommand = ''
+        ${envSetup}
+        cargo build --locked --release --target ${target} -p xmtpv3
+      '';
+    };
 
   buildTarget =
     target:
--- a/sdks/ios/Tests/XMTPTests/GroupTests.swift
+++ b/sdks/ios/Tests/XMTPTests/GroupTests.swift
@@ -1177,10 +1177,10 @@ class GroupTests: XCTestCase {
 		try fixtures.cleanUpDatabases()
 	}
 
-	// Formerly testCanSyncManyGroupsInUnderASecond; the hard <1s budget was
-	// not robust to CI runner variance (see issue #3448). The 10s ceiling
-	// still catches order-of-magnitude regressions while tolerating a
-	// loaded macOS runner; strict perf targets belong in dev/bench.
+	/// Formerly testCanSyncManyGroupsInUnderASecond; the hard <1s budget was
+	/// not robust to CI runner variance (see issue #3448). The 10s ceiling
+	/// still catches order-of-magnitude regressions while tolerating a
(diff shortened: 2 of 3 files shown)


quash
```diff
@@ -0,0 +1,37 @@
+[1m[38;5;2m@[0m  [1m[38;5;13mt[38;5;8mzkzvqlo[39m [38;5;3mgithub@andrewplaza.dev[39m [38;5;14m2026-03-30 16:27:33[39m [38;5;12m3[38;5;8m0711f47[39m[0m
```
tylerhawkes: Is this file meant to be included?

tylerhawkes (approved):

nix/lib/default.nix
```diff
@@ -40,7 +40,6 @@
           mkVersion = import ./mkVersion.nix;
         };
         wasm-bindgen-cli = pkgs.callPackage ./packages/wasm-bindgen-cli.nix { };
-        swiftformat = pkgs.callPackage ./packages/swiftformat.nix { };
```
neekolas, line 175: The reason I couldn't use Swiftformat before was that it depended on Swift, and then would try to compile Swift from source (which would then blow up).

Are you sure this works everywhere?

insipx: yeah it works on my linux machine and the macbook. It seems like swift was updated since we last tried. I can keep the derivation if it breaks in the future if that gives us some more security.

what i'm having trouble getting to work are the cargoArtifacts for our derivations, it seems crane buildDepsOnly changed slightly so now there are lockfile errors. might merge one of my previous prs that unifies the cargoartifacts which should solve that

nix/lib/default.nix
cameronvoell: erroring when trying to bring the backend up.

think we can remove this now yea?

https://github.com/NixOS/nixpkgs/pull/507296

.github/workflows/test-workspace.yml
```diff
@@ -25,10 +25,11 @@ jobs:
...
       - name: Run tests (v3 + d14n with coverage)
         # use `nix build` instead of `nix flake check` because we need the resulting coverage
-        run: nix build -L .#nextest.x86_64-linux.v3 .#nextest.x86_64-linux.d14n
-
+        run: |
+          nix-fast-build --no-nom --flake .#nextest.x86_64-linux.v3 --out-link v3 --option sandbox relaxed
+          nix-fast-build --no-nom --flake .#nextest.x86_64-linux.d14n --out-link d14n --option sandbox relaxed
       - name: merge coverage files
-        run: nix run nixpkgs#lcov -- --add-tracefile result/coverage --add-tracefile result-1/coverage --output-file lcov.info
+        run: nix run nixpkgs#lcov -- --add-tracefile v3-/coverage --add-tracefile d14n-/coverage --output-file lcov.info
```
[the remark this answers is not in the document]
insipx: seems like the run finished just fine

nix/lib/default.nix
insipx: yup was able to remove

PR #3377, opened 2026-03-31
docs: add toolchain fix instruction to build successfully on aarch64-apple-darwin

floscodes:
# Description

When building on macOS with aarch64 (Apple Silicon) architecture a build error might occur when trying to use the `clang` compiler which is necessary to build some dependencies. You need to tell `cargo` to use compiler, archiver and linker from `llvm` (see the example below).

## Fix:

- Create a `.cargo` directory in your project root
- In this directory create a `config.toml`file
- Add the following to `config.toml`:

  ```toml
  [target.wasm32-unknown-unknown]
  linker = "/opt/homebrew/opt/llvm/bin/wasm-ld"

  [env]
  CC_wasm32_unknown_unknown = "/opt/homebrew/opt/llvm/bin/clang"
  AR_wasm32_unknown_unknown = "/opt/homebrew/opt/llvm/bin/llvm-ar"
  ```

This should fix the building issues.

> [!NOTE]
> ### Add Apple Silicon build instructions for wasm32-unknown-unknown to WASM README
> Updates [README.md](https://github.com/xmtp/libxmtp/pull/3377/files#diff-2e50bdeb6504dc1fd218c95470a1b0a752a8d4574b7c0c55a10fb966ceed5768) with a new section covering how to build with Cargo when targeting `wasm32-unknown-unknown`. Adds an Apple Silicon subsection detailing the required `.cargo/config.toml` entries to set `wasm-ld` as the linker and point `CC_wasm32_unknown_unknown` and `AR_wasm32_unknown_unknown` to Homebrew-installed LLVM binaries.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 14498fb.</sup>
> 
>
>


neekolas:
Are you seeing these issues even when using `nix` and the `just wasm build` command?

bindings/wasm/README.md
```diff
@@ -28,6 +28,28 @@ These commands require Homebrew and `llvm` to be installed. See above.
...
 - `yarn test:integration:macos`: Run integration tests using vitest
 
+### macOS aarch64 (Apple Silicon) additional configuration
+
+When building on macOS with aarch64 (Apple Silicon) architecture a build error might occur when trying to use the `clang` compiler which is necessary to build some dependencies. As mentioned above you need to install `llvm` and tell `cargo` to use compiler,
+archiver and linker from `llvm` (see the example below).
+
+**Fix:**
+
+- Create a `.cargo` directory in your project root
```
neekolas: Cargo config in the project would be checked in, which is no good for Linux users.

floscodes:
> Are you seeing these issues even when using nix and the just wasm build command?

No, I don’t see any errors with that approach. However, when compiling with Cargo directly or using Dioxus, the build fails on aarch64-apple-darwin without the additional configuration. It would be beneficial to support this use case, as it allows calling functions directly from Rust within Dioxus. What do you think? Shall I update the docs accordingly?​​​​​​​​​​​​​​​​

Sorry for the late reply!

bindings/wasm/README.md
floscodes: That's a valid concern! To clarify: I didn't mean to check the config file into the repository itself — it's only documented as a manual setup step for Apple Silicon users. The idea is that Linux users would never need to touch it at all.

neekolas:
@floscodes I think for us, the answer is: "use Nix when working in this repo". Nix isn't perfect, but it lets us make sure that for all supported platforms things actually work. Without Nix, you're always going to be in a partially-working state. For example, the WASM bindings need a lot of specific dependencies to build successfully. Same for Android or iOS SDKs.

floscodes:
Ok, I see. So should I close this PR?

floscodes:
@neekolas , I refactored the docs a bit. Is it OK like this? With these configurations I could build successfully with Cargo...

PR #3413, opened 2026-04-08
Add component-based permission system for app data

tylerhawkes:
Introduces a ComponentId-based permission and registry system that will
replace the current fixed-field PolicySet. Key pieces:

- ComponentId (u16 newtype) with XMTP/App/Immutable/Reserved ranges
- ComponentRegistry (TlsMap<ComponentId, VLBytes>) storing per-component
  metadata: permissions (insert/update/delete) and data type
- Three-layer validation: immutability -> hardcoded -> registry lookup
- Hardcoded: COMPONENT_REGISTRY and SUPER_ADMIN_LIST (super admin only)
- Constrained: ADMIN_LIST (must be AdminOrSuperAdmin or SuperAdmin)
- Deny by default: no registry entry = write rejected

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

> [!NOTE]
> ### Add component-based permission system for app data in `xmtp_mls_common`
> - Introduces `ComponentId`, a `u16` newtype with well-known constants, range classification helpers (XMTP, app, reserved, immutable, hardcoded), and QUIC variable-length integer TLS serialization/deserialization.
> - Adds `ComponentRegistry`, a `TlsMap`-backed store for component metadata that validates writes, enforces immutability for write-once IDs, and rejects hardcoded/reserved IDs.
> - Implements `validate_component_write` in [validation.rs](https://github.com/xmtp/libxmtp/pull/3413/files#diff-8a1842938cea64924a56d4911d66c58918bf4078c667458b84c5be5a52eb0878), which checks immutability, requires super-admin for hardcoded components, and evaluates insert/update/delete policies from the registry, defaulting to deny.
> - Policy evaluation supports `AndCondition`/`AnyCondition` combinators and base Allow/Deny/Admin/SuperAdmin policies; invalid proto states are treated conservatively as deny.
> - Risk: `from_bytes` eagerly validates all registry entries on deserialization, so malformed or policy-violating persisted data will fail to load.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 34151e2.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls_common/src/app_data/component_registry.rs
+++ b/crates/xmtp_mls_common/src/app_data/component_registry.rs
@@ -1,0 +1,950 @@
+use crate::tls_map::TlsMap;
+use prost::Message;
+use tls_codec::{Deserialize, Serialize, VLBytes};
+use xmtp_proto::xmtp::mls::message_contents::{
+    ComponentMetadata, ComponentPermissions, ComponentType, MetadataPolicy as MetadataPolicyProto,
+    metadata_policy::{Kind as MetadataPolicyKind, MetadataBasePolicy},
+};
+
+use super::component_id::ComponentId;
+
+/// The operation being performed on a component.
+///
+/// Each variant maps to one of the policy fields in
+/// [`ComponentPermissions`](xmtp_proto::xmtp::mls::message_contents::ComponentPermissions):
+/// `Insert` → `insert_policy`, `Update` → `update_policy`, `Delete` → `delete_policy`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ComponentOp {
+    /// Creating a new value (component does not yet exist).
+    Insert,
+    /// Modifying an existing value.
+    Update,
+    /// Removing a value.
+    Delete,
+}
+
+impl std::fmt::Display for ComponentOp {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ComponentOp::Insert => write!(f, "insert"),
+            ComponentOp::Update => write!(f, "update"),
+            ComponentOp::Delete => write!(f, "delete"),
+        }
+    }
+}
+
+/// Helper to construct a [`ComponentMetadata`] from permissions and type.
+pub fn new_component_metadata(
+    permissions: ComponentPermissions,
+    component_type: ComponentType,
+) -> ComponentMetadata {
+    ComponentMetadata {
+        permissions: Some(permissions),
+        component_type: component_type as i32,
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ComponentRegistryError {
+    #[error("component ID {0} is not in the component ID space")]
+    InvalidComponentId(ComponentId),
+    #[error("component ID {0} is in the reserved range")]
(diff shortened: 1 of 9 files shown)


tylerhawkes:
* **#3460** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3460?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3413** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3413?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3413?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3410** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3410?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3409** <a href="https://app.graphite.com/github/pr/xmtp/libxmtp/3409?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=3413).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/xmtp_mls_common/src/app_data/component_registry.rs
```diff
@@ -0,0 +1,950 @@
...
+        }
+    }
+
+    /// Register or update a component's metadata.
+    ///
+    /// For mutable components in the registry, this silently overwrites any
+    /// existing entry — there is no audit log here because the audit trail
+    /// lives in the MLS commit history that produced the change.
+    ///
+    /// Rejects invalid IDs, IDs in the reserved range, hardcoded components
```
neekolas, line 126: It sounds like a hardcoded component can never move to the registry, and a registry component can never become hardcoded (without a hammer release and waiting for everyone to upgrade). Not a bad thing on its own, just want to make sure I understand how to evolve from the initial state.  
  
Would we ever introduce additional hardcoded components?

tylerhawkes: Yeah, I went back and forth on this. The two components it is tied to is the component registry with all the permissions and the super admin list. Both of which should probably ever only be modified by super admins. Everything else is in the registry. If we enforced both of those via the registry we would still have a carve out like we do for the admin list (can only be admin or superadmin) and I think it would still be a hammer release.

tylerhawkes: It might be possible to avoid a hammer release by letting anything proposed by a super admin go through without checks, but that has it's own set of risks.

neekolas: I'd be worried about the "anything the super admin does is correct" loophole coming to bite us on things like user-controlled data fields. Even a super-admin shouldn't be able to modify data meant to be owned by a specific group member. Also application-controlled fields, where the app developer probably isn't thinking about the super-admin having god mode powers.  
  
Having two magic hardcoded fields that are both extremely unlikely to to have their permissions change seems fine.

neekolas (approved):

PR #3430, opened 2026-04-10
fix: repair iOS docs and cargo doc CI builds

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3428

## Summary

Fixes both broken docs CI workflows on main:

### Deploy Docs (cargo doc) — `deploy-docs.yml`
- **Root cause:** `log_parser` depends on `slint` → `fontique` v0.7.0, which fails to compile on Ubuntu due to `fontconfig_sys` API mismatch
- **Fix:** Exclude `log_parser` from `dev/docs` (`--exclude log_parser`). It's a GUI tool, not an API library — no reason to include it in workspace docs

### iOS Docs (jazzy) — `docs-ios.yml`
- **Root cause 1:** `xcode-version: "16.0"` doesn't support `swift-tools-version: 6.1` (introduced in d49d8557). Swift 6.1 requires Xcode 16.3+
- **Root cause 2:** Jazzy runs from `sdks/ios/` (`working-directory`) but `Package.swift` is at the repo root. `sourcekitten` invokes `swift build` from the wrong directory
- **Fix:** Update Xcode to 16.3 and pass `--source-directory ${{ github.workspace }}` so jazzy finds `Package.swift` at the repo root

## Test plan
- [ ] `deploy-docs.yml` workflow passes (cargo doc succeeds without fontique compilation error)
- [ ] `docs-ios.yml` workflow passes (jazzy finds Package.swift and builds with Xcode 16.3)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Fix iOS docs and cargo doc CI builds
> - Updates the iOS docs workflow to use Xcode 16.3, run the bindings script directly (without a Nix dev shell), and write Jazzy output to an absolute workspace path with `--source-directory` specified.
> - Excludes the `log_parser` crate from `cargo doc` generation in [dev/docs](https://github.com/xmtp/libxmtp/pull/3430/files#diff-99f6b915660b496bb9ff7305e7bb589ce2fd1ce07624095a20d3d690703ae0f8) to fix the cargo doc CI build.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 99f5ecc.</sup>
> 
>

PR #3453, opened 2026-04-10
ci: remove broken iOS docs workflow

### LINKED ISSUE #3452: Flaky CI Failure: iOS Docs — jazzy cannot find SPM module 'XMTPiOS'
## Flaky CI Failure: iOS Docs — jazzy cannot find SPM module 'XMTPiOS'

**Workflow:** iOS Docs
**Failed run:** https://github.com/xmtp/libxmtp/actions/runs/24256695991
**Commit:** ee56c56becc9557a838aa86e83a8a7009fb01dce
**Failed jobs:** deploy

### Summary

The `deploy` job failed at the "Generate documentation" step. `jazzy` (v0.14.4) invoked `sourcekitten` to build the Swift package, but `sourcekitten`/`swift build` could not locate the `XMTPiOS` SPM module. The documentation generation process exited with code 1. This appears to be an infrastructure or configuration issue rather than a test or logic failure.

### Error Details

```
Running swift build
Could not find SPM module 'XMTPiOS'. Here are the modules available:

Error: Failed to generate documentation
bundler: failed to load command: jazzy (...)
/...gems/jazzy-0.14.4/lib/jazzy/executable.rb:39:in `execute_command':
  /...bin/sourcekitten ["doc", "--spm", "--module-name", "XMTPiOS", "--"] (RuntimeError)

Process completed with exit code 1.
```

### Analysis

`jazzy` runs `sourcekitten doc --spm --module-name XMTPiOS`, which internally calls `swift build` to resolve the module list. The build did not surface `XMTPiOS` as an available target. Possible root causes:

- A `Package.swift` target named `XMTPiOS` is missing or was recently renamed
- Package dependencies were not fetched before documentation generation ran
- The runner was pointed at the wrong source directory
- A transient SPM resolution failure (network or registry issue) prevented the dependency graph from being fully resolved

This is an infrastructure/configuration failure in the iOS documentation CI workflow, not a test or code logic regression.

---

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3452

## Summary

Remove `.github/workflows/docs-ios.yml` and its associated Ruby/jazzy setup. The workflow has never succeeded on `main` (0/18 runs since the monorepo migration), so it is pure CI noise. A replacement DocC-based pipeline is already in flight in PR #3192 — this clears the way for that to land cleanly.

## Root cause (why just patching it isn't worth it)

`jazzy 0.14.4` shells out to `sourcekitten doc --spm`, which:
1. runs `swift build`
2. parses `.build/debug.yaml` to find the `XMTPiOS` module's compiler arguments via its `swiftModuleName` heuristic (`Source/SourceKittenFramework/Module.swift`)
3. that heuristic understands pre-5.10 and 5.10 SwiftPM llbuild formats — but **not** the 6.x format

With `swift-tools-version: 6.1` + Xcode 16.3 / Swift 6.1, `swift build` succeeds but emits a manifest whose Swift-compile entries don't match either legacy shape, so `commands.compactMap(\.swiftModuleName)` returns empty and sourcekitten prints:

```
Running swift build
Could not find SPM module 'XMTPiOS'. Here are the modules available:
```

…with nothing after `available:`. The empty list is the tell.

PR #3430 tried to fix this by passing `--source-directory ${{ github.workspace }}` so jazzy `chdir`'s to the repo root before running sourcekitten. Verified against jazzy v0.14.4 `lib/jazzy/doc_builder.rb:80` — `Dir.chdir(options.source_directory)` does run before `SourceKitten.run_sourcekitten`, so #3430 was directionally correct. But the real failure is downstream in sourcekitten's parser, so the fix had no effect.

This is an **upstream sourcekitten** incompatibility with SwiftPM 6.x that can't be worked around in the workflow. The only way forward is to stop using jazzy for this repo.

## Why remove instead of migrate

PR #3192 is already doing a full DocC migration — reusable workflow, Vercel deploy, versioning, archive branch, `hide-internal-symbols` script. It's in human review. Rather than landing a stopgap DocC implementation on top of main (which would force a conflict with #3192 and need to be unwound), just remove the broken workflow. When #3192 merges, it brings the new system in one step.

## Changes

- Delete `.github/workflows/docs-ios.yml`
- Delete `sdks/ios/Gemfile` and `sdks/ios/Gemfile.lock` (jazzy was the only gem)
- Remove the `"iOS Docs"` entry from `.github/workflows/flaky-failure-watcher.yml` — the watcher's `workflow_run.workflows` list would otherwise name a workflow that no longer exists

```
.github/workflows/docs-ios.yml              |  60 -------------
.github/workflows/flaky-failure-watcher.yml |   1 -
sdks/ios/Gemfile                            |   2 -
sdks/ios/Gemfile.lock                       | 128 ----------------------------
4 files changed, 191 deletions(-)
```

## Test plan

- [ ] CI on this PR passes (no workflow referenced by the removed paths remains)
- [ ] `flaky-failure-watcher.yml` still parses and runs (it no longer tries to subscribe to a non-existent workflow)
- [ ] Grep confirms zero remaining references to `docs-ios`, `iOS Docs`, `jazzy`, `Gemfile`, or `bundle exec` in `.github/` or dev scripts. The only remaining mention is the historical `sdks/ios/VALIDATION.md` migration snapshot from Jan 2026, left untouched.

## Follow-up

PR #3192 reintroduces iOS docs generation using Swift DocC + `swift package generate-documentation` + Vercel. Merging this removal first lets that PR land without rebase churn against a broken workflow.

> [!NOTE]
> ### Remove broken iOS docs CI workflow
> Deletes the [docs-ios.yml](https://github.com/xmtp/libxmtp/pull/3453/files#diff-380527fb98afabb8b360c146ad40bf1ab6a908009a1a4da1b2cf9f60e8b03eff) GitHub Actions workflow that built iOS bindings, generated Swift docs with Jazzy, and deployed them to GitHub Pages. Also removes the `"iOS Docs"` entry from the `workflow_run` trigger in the flaky-failure-watcher workflow, and deletes the associated [Gemfile](https://github.com/xmtp/libxmtp/pull/3453/files#diff-fd35845574874743fd552ea157aff406782dbd31948bef5514884b7dae9d3aac) and lockfile from the iOS SDK directory.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2a6e8fc.</sup>
> 
>


neekolas:
Let's just turn off the iOS docs workflow for now. The bigger change is the right direction, but we don't need to handle it immediately.

neekolas (approved):

PR #3488, opened 2026-04-13
fix(nix): guard Android NDK cross-compilation outputs on macOS ARM

### LINKED ISSUE #3481: Flaky CI Failure: Android NDK nixpkgs evaluation throws on arm64-apple-darwin runner
Flaky CI Failure - Android NDK nixpkgs evaluation throws on arm64-apple-darwin runner. Workflow: Cache all Nix Outputs. Failed run: https://github.com/xmtp/libxmtp/actions/runs/24360658651. Commit: dbe47a2508266f92de9a1fde704e7aebec8aa599. Failed jobs: build (warp-macos-26-arm64-12x), validation-image-aarch64. The Build all flake outputs step failed when Nix evaluated the Android NDK package on the warp-macos-26-arm64-12x runner. The nixpkgs androidndk-pkgs.nix throws when stdenv.buildPlatform.config is arm64-apple-darwin since the NDK does not support macOS ARM as a build host. Error: Android NDK doesnt support building on arm64-apple-darwin, as far as we know - nix build failed exit code Some(1). Secondary failure: validation-image-aarch64 cachix pin returned HTTP 400 - store path mls-validation-service.tar.gz not in binary cache. This is distinct from issue #3476 (atf cross-compilation); this is a hard throw in the nixpkgs Android NDK package for arm64-apple-darwin. Fix requires guarding Android NDK outputs from evaluation on macOS ARM. Reported by Flaky Failure Watcher.

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3481

## Summary

- Wrap the `perSystem` output in `nix/android-packages.nix` with `lib.optionalAttrs (lib.hasSuffix "-linux" system)` so Android NDK cross-compilation packages are only exposed on Linux systems
- Prevents nixpkgs' `androidndk-pkgs.nix` from being evaluated on `arm64-apple-darwin`, which hard-throws because the NDK doesn't support that build host
- The Android devShell remains available on all platforms (uses pre-built SDK/NDK downloads, not cross-compilation)
- Uses the `system` string instead of `pkgs.stdenv.isLinux` to avoid infinite recursion in flake-parts module evaluation

## Test plan

- [x] `nix flake show` succeeds — Android packages appear on Linux systems, not on aarch64-darwin
- [x] `nix fmt -- --ci` passes (formatting clean)
- [ ] CI: "Cache all Nix Outputs" workflow passes on macOS ARM runner

> [!NOTE]
> ### Guard Android NDK cross-compilation outputs on Linux in `android-packages.nix`
> Wraps all Android package attributes (`android-libs`, `android-libs-fast`, `kotlin-bindings`, and per-ABI `android-bindings-*`) in `lib.optionalAttrs (lib.hasSuffix "-linux" system)` in [android-packages.nix](https://github.com/xmtp/libxmtp/pull/3488/files#diff-8b047e47ceb6f5603985efee1af17c324796e643cb15c21de53b90e3b35fc9fc), so they are only exposed when building on a Linux host. The inner logic (link farms, ABI mappings, cross-compilation targets) is unchanged.
>
> Behavioral Change: On macOS ARM (and any non-Linux system), all Android NDK cross-compilation outputs are now absent from the package set rather than present but broken.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7ca9ade.</sup>
> 
>

### DIFF
--- a/nix/android-packages.nix
+++ b/nix/android-packages.nix
@@ -7,86 +7,77 @@
       lib,
       ...
     }:
-    let
-      platforms = import "${inputs.nixpkgs}/lib/systems/platforms.nix" { inherit lib; };
+    # The Android NDK cross-compilation toolchain in nixpkgs (androidndk-pkgs.nix)
+    # does not support arm64-apple-darwin as a build host and hard-throws during
+    # evaluation. Guard all Android cross-compilation outputs so they are only
+    # exposed on Linux, where the NDK is supported.
+    # Note: use `system` (a plain string) instead of `pkgs.stdenv.isLinux` to
+    # avoid infinite recursion in the flake-parts module system.
+    # See https://github.com/xmtp/libxmtp/issues/3481
+    lib.optionalAttrs (lib.hasSuffix "-linux" system) (
+      let
+        platforms = import "${inputs.nixpkgs}/lib/systems/platforms.nix" { inherit lib; };
 
-      sdkConfig = {
-        androidSdkVersion = "23";
-        androidNdkVersion = "27";
-        useAndroidPrebuilt = true;
-      };
-
-      # Single source of truth: ABI name → target config
-      androidTargets = {
-        "arm64-v8a" = {
-          config = "aarch64-unknown-linux-android";
-          rust.rustcTarget = "aarch64-linux-android";
+        sdkConfig = {
+          androidSdkVersion = "23";
+          androidNdkVersion = "27";
+          useAndroidPrebuilt = true;
         };
-        "armeabi-v7a" = {
-          config = "armv7a-unknown-linux-androideabi";
-          rust.rustcTarget = "armv7-linux-androideabi";
-        }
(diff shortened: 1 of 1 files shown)


insipx:
this should no longer be an issue since https://github.com/xmtp/libxmtp/pull/3482

this happened because i merged an update where I ran `nix flake update`, but forgot that I had pinned nix to a nixpkgs from 3/21, pre android patch. pull 3482 fixed that by removing the pin.

xmtp-coder-agent:
You're right — I verified that with the current nixpkgs (unpinned via #3482), the original `android-packages.nix` evaluates cleanly for `aarch64-darwin` including all `android-*` packages. The root cause was the pinned nixpkgs revision missing arm64-apple-darwin NDK support, which #3482 already resolved.

This PR is unnecessary. Closing.

PR #3490, opened 2026-04-13
fix: propagate PendingRemove state to late-joining installations

### LINKED ISSUE #3475: Fix late same-inbox self-removal recovery for new installations
## Summary

`origin/main` still has an ignored regression test covering late same-inbox welcome / self-removal recovery:
- `crates/xmtp_mls/src/groups/tests/mod.rs`
- `test_self_removal_with_late_installation`
- currently annotated with `#[ignore] // fix after consent sync`

## Current Behavior

The test exercises a same-inbox multi-installation path:
1. one installation joins a group
2. that installation leaves the group and enters `PendingRemove`
3. a later installation for the same inbox comes online
4. the new installation should bootstrap the correct removal-related state from welcomes / sync

On `main`, that behavior is still not standing coverage because the test remains ignored.

## Expected Behavior

A late installation for an inbox that already self-removed should recover consistently:
- bootstrap the right group state for that inbox on the new installation
- preserve the expected `PendingRemove` semantics
- avoid regressing same-inbox recovery around welcome / consent sync ordering

## Acceptance Criteria

- `test_self_removal_with_late_installation` is unignored and passing
- same-inbox late-installation recovery after self-removal is explicitly covered on `main`
- consent-sync interactions required for this path are either fixed or documented in the implementation/tests

## Why This Matters

This is another explicit regression point still visible on `main`, and it sits in a subtle inbox/device recovery path where silent drift would be easy to miss without standing coverage.

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3475

## Summary

- Re-sends LeaveRequest in the new MLS epoch after adding missing installations, so late-joining same-inbox installations can decrypt and process the self-removal
- Unignores `test_self_removal_with_late_installation` which now passes

## Problem

When an inbox has multiple installations and one leaves a group (PendingRemove), a new installation created later cannot decrypt the original LeaveRequest because it was encrypted for an earlier MLS epoch. This caused the new installation to miss the pending self-removal state, leading to inconsistent group state across installations.

## Solution

In `add_missing_installations()`, after the commit that adds new installations to the group resolves, check if the current inbox has a pending self-removal. If so, re-send the LeaveRequest message in the current epoch. This allows newly-added installations (which now have the epoch keys) to decrypt and process it, correctly setting `PendingRemove` state and populating the pending-remove list.

The re-send is safe because:
- `PendingRemove::store_or_ignore` handles duplicate records idempotently
- `update_group_membership` to PendingRemove is idempotent when already in that state
- The re-send only triggers when installations were actually added (not on every sync)

## Test plan

- [x] `test_self_removal_with_late_installation` — unignored and passing
- [x] `test_self_removal_with_multiple_initial_installations` — still passing (no regression)
- [x] `test_clean_pending_remove_list_on_member_removal` — still passing
- [x] All 6 `test_self_removal*` tests pass
- [x] `cargo clippy` clean, `cargo fmt` clean, `hakari` clean

> [!NOTE]
> ### Propagate `PendingRemove` state to late-joining installations via re-sent LeaveRequest
> - Adds `added_installation_inbox_ids` to `ValidatedCommit` to track which inboxes gained new installations in a commit.
> - Adds `maybe_resend_leave_request_for_new_installations` in [`mls_sync.rs`](https://github.com/xmtp/libxmtp/pull/3490/files#diff-160aadcd7de477c1116260a79601a0548842f3bd91bf16bd8443debb66ce6684): if the current inbox is in `PendingRemove` state and a commit adds one of its installations, a `SendMessage` intent carrying a `LeaveRequest` is queued in the same transaction.
> - This runs for both locally originated commits and externally received commits during sync.
> - Enables the previously ignored test `test_self_removal_with_late_installation`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0360080.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/groups/mls_sync.rs
+++ b/crates/xmtp_mls/src/groups/mls_sync.rs
@@ -2313,6 +2313,11 @@ where
     ///
     /// Runs within the same transaction as commit processing, so the queued intent is
     /// rolled back if the commit fails to apply.
+    ///
+    /// Only queues the SendMessage intent — no optimistic StoredGroupMessage is created
+    /// because this is an internal re-send (not user-initiated) and the sender already
+    /// has the original LeaveRequest in their local history. Newly-added installations
+    /// will store the re-sent message when they receive it via the external-message path.
     fn maybe_resend_leave_request_for_new_installations(
         &self,
         added_installation_inbox_ids: &HashSet<String>,
@@ -2339,32 +2344,7 @@ where
         })?;
         let message_bytes = encoded_content_to_bytes(encoded_content);
 
-        let now = now_ns();
-        let queryable_content_fields = Self::extract_queryable_content_fields(&message_bytes);
-        let message_id = calculate_message_id(&self.group_id, &message_bytes, &now.to_string());
-        let stored_message = StoredGroupMessage {
-            id: message_id,
-            group_id: self.group_id.clone(),
-            decrypted_message_bytes: message_bytes.clone(),
-            sent_at_ns: now,
-            kind: GroupMessageKind::Application,
-            sender_installation_id: self.context.installation_id().into(),
-            sender_inbox_id: own_inbox_id.to_string(),
-            delivery_status: DeliveryStatus::Unpublished,
-            content_type: queryable_content_fields.content_type,
-            version_major: queryable_content_fields.version_major,
-            version_minor: queryable_content_fields.version_minor,
-            authority_id: queryable_content_fields.authority_id,
-            reference_id: queryable_content_fields.reference_id,
-            sequence_id: 0,
-            originator_id: 0,
-            expire_at_ns: None,
-            inserted_at_ns: 0,
-            should_push: false,
-        };
-        stored_message.store(&db)?;
-
-        let plain_envelope = Self::into_envelope(&message_bytes, now);
+        let plain_envelope = Self::into_envelope(&message_bytes, now_ns());
         let mut encoded_envelope = vec![];
         plain_envelope.encode(&mut encoded_envelope)?;
 
--- a/crates/xmtp_mls/src/groups/mls_sync.rs
+++ b/crates/xmtp_mls/src/groups/mls_sync.rs
@@ -85,7 +85,10 @@ use xmtp_configuration::{
     SYNC_BACKOFF_TOTAL_WAIT_MAX_SECS, SYNC_BACKOFF_WAIT_MS, SYNC_JITTER_MS,
     SYNC_UPDATE_INSTALLATIONS_INTERVAL_NS,
 };
-use xmtp_content_types::{CodecError, ContentCodec, group_updated::GroupUpdatedCodec};
+use xmtp_content_types::{
+    CodecError, ContentCodec, encoded_content_to_bytes, group_updated::GroupUpdatedCodec,
+    leave_request::LeaveRequestCodec,
+};
 use xmtp_db::message_deletion::{QueryMessageDeletion, StoredMessageDeletion};
 use xmtp_db::{
     Fetch, MlsProviderExt, StorageError, StoreOrIgnore,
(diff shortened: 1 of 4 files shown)


crates/xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -3298,6 +3301,27 @@ where
             .queue(self)?;
 
         let _ = self.sync_until_intent_resolved(intent.id).await?;
+
+        // After adding new installations (which creates a new epoch), if this inbox
+        // has a pending self-removal, re-send the LeaveRequest. Newly-added
+        // installations joined in a later epoch and cannot decrypt the original
+        // LeaveRequest, so this ensures they can process it and set PendingRemove.
+        if self.is_in_pending_remove(self.context.inbox_id())? {
```
neekolas: I don't think this function is the right place to handle this check. What if someone else is the one to add the missing installations? Then it would never be re-sent.

xmtp-coder-agent: Good catch — moved the check into commit processing (`save_transcript_message`), so it fires whenever any client processes a commit that adds an installation of our inbox, not only when we initiated the add.

Implementation notes:
- Added `added_installation_inbox_ids: HashSet<String>` to `ValidatedCommit`, populated from each Add proposal's credential. This captures new installations of an existing inbox (which `added_inboxes` does not — it only lists newly-joined inboxes).
- Run the check before `is_empty()` in `save_transcript_message` since an installation-only `UpdateGroupMembership` commit has no added/removed inboxes or metadata changes and would otherwise be skipped.
- The intent is queued using the same storage transaction as the commit, so it's rolled back if the commit fails to apply.

`test_self_removal_with_late_installation` still passes; all `self_removal` and `pending_remove` tests green.

neekolas: `save_transcript_message`also seems like the wrong place. That functionn should do one thing and one thing only: save a transcript message. It should happen after the commit has been processed.

xmtp-coder-agent: Good call. Moved the re-send out of `save_transcript_message` — it now runs right after `save_transcript_message` in both post-commit paths (own intent and external message). `save_transcript_message` is back to doing just that one thing.

Still within the commit-processing transaction, so the queued intent remains rolled-back-safe if anything downstream fails.

Pushed `35b6de4b6`; local self_removal (9/9) and pending_remove (4/4) suites green, lint clean.

crates/xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -2288,6 +2305,80 @@ where
...
+
+        let encoded_content = LeaveRequestCodec::encode(LeaveRequest {
+            authenticated_note: None,
+        })?;
+        let message_bytes = encoded_content_to_bytes(encoded_content);
+
+        let now = now_ns();
+        let queryable_content_fields = Self::extract_queryable_content_fields(&message_bytes);
+        let message_id = calculate_message_id(&self.group_id, &message_bytes, &now.to_string());
+        let stored_message = StoredGroupMessage {
```
neekolas: Do I need to pre-create the StoredGroupMessage as an optimistic send? What would happen if we skipped this and just waited for the actual send to go through as part of intent sync?

xmtp-coder-agent: Good point — removed it in `9f899dd46`. The re-send is internal, and we already have the original LeaveRequest locally, so the optimistic store just created a duplicate entry in the sender's message list. Newly-added installations still pick up the re-send via the external-message path (which calls `store_or_ignore` and `process_leave_request_message`), so the test still passes.

Local `test_self_removal_with_late_installation`, `self_removal` (9/9), and `pending_remove` (4/4) suites green, lint clean.

neekolas:
I think this is a sensible change that fixes a real bug. @cameronvoell is there a better way we'd want to resolve #3475 that you had in mind?

xmtp-coder-agent:
CI failed on d14n: `xmtp_mls client::tests::should_reconnect` — a toxiproxy/keepalive reconnect test, unrelated to this PR's changes in commit processing.

This is tracked as a known flaky failure in #3438 ("Flaky CI Failure: xmtp_mls::client::tests::should_reconnect fails all retries in d14n suite"). All my change's tests (`test_self_removal_with_late_installation` included in the run summary) passed.

I don't have permissions to re-run the workflow — could someone kick it off again? Thanks.

PR #3500, opened 2026-04-14
fix(nix): robustly rewrite /nix/store libiconv refs in darwin node binary

### LINKED ISSUE #3479: node-bindings@1.10.0 darwin-arm64 binary linked against Nix libiconv
## Problem

The `@xmtp/node-bindings@1.10.0` native binary for `darwin-arm64` has a hardcoded dependency on a Nix store path for `libiconv`, making it unusable on standard macOS systems without Nix installed.

```
$ otool -L node_modules/@xmtp/node-bindings/dist/bindings_node.darwin-arm64.node
  /nix/store/7h6icyvqv6lqd0bcx41c8h3615rjcqb2-libiconv-109.100.2/lib/libiconv.2.dylib
```

Runtime error:

```
Error: dlopen(...bindings_node.darwin-arm64.node, 0x0001):
  Library not loaded: /nix/store/7h6icyvqv6lqd0bcx41c8h3615rjcqb2-libiconv-109.100.2/lib/libiconv.2.dylib
  Reason: tried: '/nix/store/...' (no such file)
```

## Expected

The binary should link against system `libiconv` (`/usr/lib/libiconv.2.dylib`) rather than a Nix store path, so it works on any macOS machine.

## Environment

- macOS (darwin-arm64)
- Node.js v25.9.0
- `@xmtp/node-bindings@1.10.0`
- No Nix installed

## Workaround

Can be patched locally with `install_name_tool`:

```
install_name_tool -change \
  /nix/store/7h6icyvqv6lqd0bcx41c8h3615rjcqb2-libiconv-109.100.2/lib/libiconv.2.dylib \
  /usr/lib/libiconv.2.dylib \
  node_modules/@xmtp/node-bindings/dist/bindings_node.darwin-arm64.node
```

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3479

## Summary

`@xmtp/node-bindings@1.10.0`'s `bindings_node.darwin-arm64.node` embeds `/nix/store/.../libiconv.2.dylib` as a runtime dep, so `require('@xmtp/node-bindings')` fails to dlopen on any macOS host without Nix.

## Root cause

Two factors:

1. **1.10.0 shipped with no install_name rewrite.** The release tag `node-bindings-1.10.0` points at commit `1481f4bfe`, which pre-dates any `install_name_tool` handling in `nix/package/node.nix`.
2. **The rewrite added afterwards is fragile.** The current darwin `postFixup` uses a single hardcoded `install_name_tool -change "${darwin.libiconv}/lib/libiconv.2.dylib" ...`. That only works if `${darwin.libiconv}` evaluates at expression time to the exact store path the linker embedded. nixpkgs drift (we follow `nixos-unstable`, unpinned since #3482) or cross-compile splicing can make the two diverge, at which point `install_name_tool -change` silently becomes a no-op.

Secondary latent defect: `install_name_tool` invalidates the ad-hoc signature applied by `darwin.autoSignDarwinBinariesHook`, and the old `postFixup` did not re-sign after modification.

## Fix

In `nix/package/node.nix` darwin `postFixup`:

1. Keep the `install_name_tool -id @loader_path/<basename>` for `LC_ID_DYLIB`.
2. Replace the hardcoded `-change` with a loop over `otool -L` output that rewrites every `/nix/store/.../libiconv.<ver>.dylib` reference to `/usr/lib/<basename>`. Using `otool -L` as the source of truth is drift-proof — we rewrite what the linker actually recorded, not what Nix evaluation says.
3. Re-run `codesign --force --sign -` after modification.
4. Assert that no `/nix/store` reference remains in the final binary — fail the build loudly if anything leaks (catches regressions at build time rather than via consumer reports).

Restricting the rewrite to libiconv (not a blanket `/nix/store` → `/usr/lib`) is intentional: other Nix-provided libs may not have `/usr/lib` counterparts, and a blanket rewrite would silently ship a broken binary. The assert in step 4 catches any future dep that starts leaking Nix paths.

## Test plan

- [x] `nixfmt --check nix/package/node.nix`
- [x] `just lint-nix` passes
- [x] `just lint-config` (treefmt + nixfmt + taplo) passes
- [x] Flake evaluates on Linux: `nix eval .#packages.aarch64-linux.node-bindings-linux-arm64-musl.drvPath` returns a valid derivation path
- [x] awk/bash pipeline validated against simulated `otool -L` output (matches libiconv refs, ignores `/usr/lib`; assertion catches non-libiconv `/nix/store` refs)
- [ ] **Pending macOS CI:** `nix build .#node-bindings-darwin-arm64` on `warp-macos-latest-arm64-6x` — the build-time assertion is the end-to-end test
- [ ] **Post-release validation:** `require('@xmtp/node-bindings')` loads on a Nix-free macOS host against the next published version

## Out of scope

- Cutting a 1.10.1 release (release-management concern).
- Linux / Windows / iOS / Android binaries — different build paths, no `/nix/store` rpath issues.

> [!NOTE]
> ### Fail the Darwin node binary build if `/nix/store` libiconv refs survive `install_name_tool` rewrite
> In [nix/package/node.nix](https://github.com/xmtp/libxmtp/pull/3500/files#diff-fa9b6ba547ea5ae35048580f634e5edb6375afb0138fd6fb6e77fef6ee8c357b), the `postFixup` hook now runs `otool -L` on the produced `.node` file after the `install_name_tool` rewrite and exits non-zero if any `/nix/store` paths remain. This catches cases where the rewrite step silently fails to remove all store references, preventing broken binaries from being installed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57e1a64.</sup>
> 
>

### DIFF
--- a/nix/package/node.nix
+++ b/nix/package/node.nix
@@ -95,11 +95,9 @@ rust.napiBuild (
       # loads on hosts without Nix.
       install_name_tool -change "${darwin.libiconv}/lib/libiconv.2.dylib" "/usr/lib/libiconv.2.dylib" "$NODE_LIB"
 
-      # Assert no /nix/store references remain. This would have caught
-      # https://github.com/xmtp/libxmtp/issues/3479 at build time (1.10.0
-      # shipped from a commit that had no install_name_tool rewrite at
-      # all); it also guards against any future silent no-op — e.g. if a
-      # new dynamic dep is introduced that isn't handled above.
+      # Assert no /nix/store references remain — guards against silent
+      # no-ops in the rewrites above and catches the 1.10.0 regression.
+      # See https://github.com/xmtp/libxmtp/issues/3479.
       remaining=$(otool -L "$NODE_LIB" | awk '$1 ~ /^\/nix\/store\// { print $1 }')
       if [ -n "$remaining" ]; then
         echo "error: $NODE_LIB still references /nix/store after postFixup:" >&2
--- a/nix/package/node.nix
+++ b/nix/package/node.nix
@@ -86,13 +86,9 @@ rust.napiBuild (
     postFixup = ''
       NODE_LIB=$(echo $out/dist/bindings_node.*.node)
 
-      # Fix the dylib's own install name (LC_ID_DYLIB) — it embeds the
-      # ephemeral Nix build path.
+      # Replace /nix/store paths (the dylib's own id and its libiconv dep)
+      # with macOS system paths so the .node loads on hosts without Nix.
       install_name_tool -id "@loader_path/$(basename $NODE_LIB)" "$NODE_LIB"
-
-      # Rewrite the Nix-store libiconv reference to the system libiconv,
-      # which macOS resolves via the dyld shared cache so the .node file
-      # loads on hosts without Nix.
       install_name_tool -change "${darwin.libiconv}/lib/libiconv.2.dylib" "/usr/lib/libiconv.2.dylib" "$NODE_LIB"
 
       # Assert no /nix/store references remain — guards against silent
--- a/nix/package/node.nix
+++ b/nix/package/node.nix
@@ -83,13 +83,6 @@ rust.napiBuild (
     nativeBuildInputs = commonArgs.nativeBuildInputs ++ [
       darwin.autoSignDarwinBinariesHook
     ];
-    # Rewrite /nix/store dylib references to macOS system paths so the .node
-    # file loads on hosts without Nix. We iterate `otool -L` output rather
-    # than using a fixed `install_name_tool -change` against
-    # `${darwin.libiconv}`, because nixpkgs drift (we follow nixos-unstable,
-    # unpinned since #3482) or cross-compile splicing can make a hardcoded
-    # path match silently a no-op. See
-    # https://github.com/xmtp/libxmtp/issues/3479.
     postFixup = ''
       NODE_LIB=$(echo $out/dist/bindings_node.*.node)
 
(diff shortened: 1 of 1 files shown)


insipx:
nixpkgs is deterministic and the path referred to in the `node.nix` derivation by darwin libiconv _must_ deterministically match the libiconv dependency linked with. if it doesn't, there is something wrong with nixpkgs or our derivation. Nixpkgs splicing / nixpkgs "package drift" isn't real, if nixpkgs updates then `darwin.libiconv` in node.nix would still point to the correct store path. and libiconv is the only shared library being replaced in this derivation so far. the libiconv library path is constant/stable mac abi that is safe to hardcode

An otool loop or similar just serves to obscure the actual function of postFixup and makes it more difficult to debug if needed


 rather than adding this loop, maybe add a nixpkgs assert or similar that errors if it discovers any /nix/store references

nix/package/node.nix
```diff
@@ -83,12 +83,43 @@ rust.napiBuild (
...
+      # any remaining /nix/store reference is caught by the assert below.
+      otool -L "$NODE_LIB" \
+        | awk '$1 ~ /^\/nix\/store\// && $1 ~ /\/libiconv\.[0-9.]+\.dylib$/ { print $1 }' \
+        | while read -r nixlib; do
+          install_name_tool -change "$nixlib" "/usr/lib/$(basename "$nixlib")" "$NODE_LIB"
+        done
+
+      # install_name_tool invalidates the ad-hoc signature applied by
+      # autoSignDarwinBinariesHook, so re-sign after modification.
+      codesign --force --sign - "$NODE_LIB"
```
insipx: darwin auto sign binaries hook already takes care of this

xmtp-coder-agent: You're right — verified by reading the hook source:

- `autoSignDarwinBinariesHook`'s `auto-sign-hook.sh` registers `signDarwinBinariesInAllOutputs` via `postFixupHooks+=`.
- stdenv's `runHook postFixup` runs the user's `postFixup` variable first, then iterates `postFixupHooks`, so the hook signs *after* our modifications.
- `signIfRequired` → `sign` explicitly handles the `install_name_tool`-invalidated case with a copy-sign-move workaround (the file-comment even calls that scenario out by name).

Dropped the `codesign --force --sign -` call in 181445e47 and replaced it with a note explaining why no explicit re-sign is needed. `nixfmt --check` clean.

xmtp-coder-agent:
Fair point — you're right that 1.10.0's failure was simply the absence of any rewrite (the tagged commit `1481f4bfe` has no `install_name_tool` at all), not nixpkgs drift. The otool loop was defensive programming against a scenario I couldn't actually verify exists, and it obscured what postFixup is doing.

Restructured in c549ed2e6:

- Restored the simple `install_name_tool -change "${darwin.libiconv}/lib/libiconv.2.dylib" ...` form.
- Kept the `/nix/store` assert as the safety net — that's the piece that would have caught 1.10.0 shipping broken, and it covers the "silent no-op" case if nixpkgs or our derivation ever misbehaves.

If the assertion ever fires, that's a real bug to investigate rather than work around.

xmtp-coder-agent:
Noted — the hardcoded `/usr/lib/libiconv.2.dylib` target in the `-change` rewrite aligns with what c549ed2e6 already does (after dropping the otool loop). Nothing additional to change.

nix/package/node.nix
```diff
@@ -86,9 +86,26 @@ rust.napiBuild (
...
       install_name_tool -id "@loader_path/$(basename $NODE_LIB)" "$NODE_LIB"
+
+      # Rewrite the Nix-store libiconv reference to the system libiconv,
+      # which macOS resolves via the dyld shared cache so the .node file
+      # loads on hosts without Nix.
       install_name_tool -change "${darwin.libiconv}/lib/libiconv.2.dylib" "/usr/lib/libiconv.2.dylib" "$NODE_LIB"
+
+      # Assert no /nix/store references remain. This would have caught
+      # https://github.com/xmtp/libxmtp/issues/3479 at build time (1.10.0
+      # shipped from a commit that had no install_name_tool rewrite at
```
insipx: can we get a more succinct comment here, maybe limit to 3 lines (that includes the github link)

nix/package/node.nix
```diff
@@ -86,9 +86,26 @@ rust.napiBuild (
     postFixup = ''
       NODE_LIB=$(echo $out/dist/bindings_node.*.node)
 
-      # Fix the dylib's own install name (LC_ID_DYLIB) — it embeds the ephemeral Nix build path
+      # Fix the dylib's own install name (LC_ID_DYLIB) — it embeds the
+      # ephemeral Nix build path.
       install_name_tool -id "@loader_path/$(basename $NODE_LIB)" "$NODE_LIB"
+
+      # Rewrite the Nix-store libiconv reference to the system libiconv,
+      # which macOS resolves via the dyld shared cache so the .node file
```
insipx: this comment is mostly unnecessary, maybe remove this comment and the one above and replace with a single 1-2 line comment explaining the need to replace /nix/store paths

nix/package/node.nix
xmtp-coder-agent: Trimmed to 3 lines in af65aa667.

nix/package/node.nix
xmtp-coder-agent: Collapsed both comments into a single 2-line header above the two `install_name_tool` calls in 57e1a64f0.

insipx (approved):

PR #3511, opened 2026-04-15
fix(nix): skip otool -L header line in darwin node binary /nix/store check

### LINKED ISSUE #3503: Flaky CI Failure: bindings_node.darwin-arm64.node still references /nix/store after postFixup
## Flaky CI Failure: bindings_node.darwin-arm64.node still references /nix/store after postFixup

**Failed run:** https://github.com/xmtp/libxmtp/actions/runs/24459579250
**Commit:** 5afaa99034bc9cda6fadf70034b278ed3dfb3d1f

### Summary

The Nix build for `bindings-node-js-napi` failed during the `fixupPhase` because `bindings_node.darwin-arm64.node` still contains `/nix/store` references after `postFixup`. This is notable because commit `5afaa99` ("fix(nix): robustly rewrite /nix/store libiconv refs in darwin node binary") was specifically intended to address this exact class of failure, yet the same error occurs on that same commit. The fix appears to be incomplete — not all `/nix/store` references are being rewritten.

### Error Details

```
> checking for references to /nix/var/nix/builds/nix-9230-2572057328/ in /nix/store/snr9fyz0z764ivcdybdhxlhbffqf0ars-bindings-node-js-napi-1.10.0...
> patching script interpreter paths in /nix/store/snr9fyz0z764ivcdybdhxlhbffqf0ars-bindings-node-js-napi-1.10.0
> error: /nix/store/snr9fyz0z764ivcdybdhxlhbffqf0ars-bindings-node-js-napi-1.10.0/dist/bindings_node.darwin-arm64.node still references /nix/store after postFixup:
> /nix/store/snr9fyz0z764ivcdybdhxlhbffqf0ars-bindings-node-js-napi-1.10.0/dist/bindings_node.darwin-arm64.node:

error: Cannot build '/nix/store/8a9pkl7d9c56ry7w6c1fisb7381ygrfh-bindings-node-js-napi-1.10.0.drv'.
       Reason: builder failed with exit code 1.
error: Cannot build '/nix/store/l9plknff6y7279hrmz5sj3xyv1nrsqis-devour-output.json.drv'.
       Reason: 1 dependency failed.
```

### Analysis

- **Root cause hypothesis:** The `postFixup` script introduced in `5afaa99` rewrites `/nix/store` references for `libiconv` in the darwin arm64 `.node` binary, but the binary embeds additional `/nix/store` paths (beyond `libiconv`) that are not covered by the rewrite logic. The Nix `disallowedReferences` / sandbox check triggers because at least one `/nix/store` path survives fixup.
- **Affected area:** `bindings/node` Nix derivation — `bindings_node.darwin-arm64.node` (cross-compiled for `aarch64-apple-darwin` from the macOS arm64 runner). The `x86_64-apple-darwin` variant also appears to be building in parallel and may be affected.
- **Failure type:** Build/code failure, not infrastructure. The failure is deterministic given the same source — the `/nix/store` reference is embedded at link time and not fully scrubbed by the current fixup hook.
- **Next step:** Inspect which `/nix/store` paths remain in the binary after fixup (e.g., `strings bindings_node.darwin-arm64.node | grep /nix/store`) and extend the rewrite script in the Nix derivation to cover all of them, or use `allowedReferences` / `noDefaultExcludes` to suppress the check if the remaining references are benign runtime paths.

---

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3503

## Summary

The `postFixup` assertion in `nix/package/node.nix` introduced by #3500 fails deterministically on every macOS arm64 run. Root cause: the awk filter `$1 ~ /^\/nix\/store\//` matches the **first line of `otool -L`'s output**, which is the file's own path (with trailing colon) — and `$NODE_LIB` itself lives under `/nix/store`. CI was reporting `$NODE_LIB:` as the "leaked" reference, not a library.

The `install_name_tool -change ... libiconv.2.dylib` rewrite on the preceding line is actually working. Only the assertion is broken.

## Fix

Two small changes in `nix/package/node.nix` Darwin `postFixup`:

1. **Fix the false positive** — add `NR > 1` to the assertion awk filter so it skips `otool -L`'s header line (the file's own path).
2. **Re-sign after modification** — `install_name_tool` invalidates the ad-hoc signature applied by `darwin.autoSignDarwinBinariesHook`; add `codesign --force --sign -` so the `.node` loads under Gatekeeper on end-user macOS hosts. (The #3500 commit message promised this but did not ship it.)

## Why the original PR missed this

#3500's test plan listed "awk/bash pipeline validated against simulated `otool -L` output" — but the simulation almost certainly omitted the header line. There was no macOS CI gate on that PR, and the failure only shows up on the actual Darwin build.

## Test plan

- [x] `nix fmt` — formatting clean
- [x] `just lint-nix` passes
- [x] `just lint-config` (treefmt + nixfmt + taplo) passes
- [x] `nix eval .#packages.aarch64-linux.node-bindings-linux-arm64-musl.drvPath` — flake evaluates on Linux
- [x] `nix eval` for the `aarch64-darwin.node-bindings-darwin-arm64` derivation — flake evaluates for Darwin
- [x] Offline awk validation against three simulated `otool -L` shapes:
  - Post-rewrite (libiconv rewritten to `/usr/lib`): assertion emits nothing ✓
  - Real leak (synthetic `/nix/store/.../libfoo.dylib`): assertion flags it ✓
  - Old buggy awk on clean post-rewrite output: confirms header line was the false positive ✓
- [ ] **Pending macOS CI:** the `build (warp-macos-26-arm64-12x)` job in "Cache all Nix Outputs" is the end-to-end signal — the build-time assertion runs against real `otool -L` output.

> [!NOTE]
> ### Fix false-positive /nix/store check by skipping `otool -L` header line in Darwin node binary
> - Fixes the awk filter in the `postFixup` script in [node.nix](https://github.com/xmtp/libxmtp/pull/3511/files#diff-fa9b6ba547ea5ae35048580f634e5edb6375afb0138fd6fb6e77fef6ee8c357b) to skip the first line of `otool -L` output (`NR > 1 &&`), preventing false positives when the binary's own path contains `/nix/store`.
> - Adds an ad-hoc `codesign --force --sign -` step after `install_name_tool` adjustments so the `.node` binary remains loadable on macOS after its library paths are rewritten.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6db38a8.</sup>
> 
>

### DIFF
--- a/nix/package/node.nix
+++ b/nix/package/node.nix
@@ -89,17 +89,7 @@ rust.napiBuild (
       # Replace /nix/store paths (the dylib's own id and its libiconv dep)
       # with macOS system paths so the .node loads on hosts without Nix.
       install_name_tool -id "@loader_path/$(basename $NODE_LIB)" "$NODE_LIB"
-
-      # Discover the actually-embedded libiconv store path from otool -L
-      # rather than trusting ${darwin.libiconv} to match the store hash the
-      # linker recorded — nixpkgs drift / cross-compile splicing can make the
-      # two diverge, at which point a hardcoded -change is a silent no-op.
-      # NR > 1 skips otool -L's header line (which is the file's own path).
-      otool -L "$NODE_LIB" \
-        | awk 'NR > 1 && $1 ~ /^\/nix\/store\/.*\/libiconv\.[0-9]+\.dylib$/ { print $1 }' \
-        | while read -r old; do
-            install_name_tool -change "$old" "/usr/lib/$(basename "$old")" "$NODE_LIB"
-          done
+      install_name_tool -change "${darwin.libiconv}/lib/libiconv.2.dylib" "/usr/lib/libiconv.2.dylib" "$NODE_LIB"
 
       # install_name_tool invalidates the ad-hoc signature applied by
       # darwin.autoSignDarwinBinariesHook; re-sign so the .node loads under
--- a/nix/package/node.nix
+++ b/nix/package/node.nix
@@ -89,12 +89,28 @@ rust.napiBuild (
       # Replace /nix/store paths (the dylib's own id and its libiconv dep)
       # with macOS system paths so the .node loads on hosts without Nix.
       install_name_tool -id "@loader_path/$(basename $NODE_LIB)" "$NODE_LIB"
-      install_name_tool -change "${darwin.libiconv}/lib/libiconv.2.dylib" "/usr/lib/libiconv.2.dylib" "$NODE_LIB"
+
+      # Discover the actually-embedded libiconv store path from otool -L
+      # rather than trusting ${darwin.libiconv} to match the store hash the
+      # linker recorded — nixpkgs drift / cross-compile splicing can make the
+      # two diverge, at which point a hardcoded -change is a silent no-op.
+      # NR > 1 skips otool -L's header line (which is the file's own path).
+      otool -L "$NODE_LIB" \
+        | awk 'NR > 1 && $1 ~ /^\/nix\/store\/.*\/libiconv\.[0-9]+\.dylib$/ { print $1 }' \
+        | while read -r old; do
+            install_name_tool -change "$old" "/usr/lib/$(basename "$old")" "$NODE_LIB"
+          done
+
+      # install_name_tool invalidates the ad-hoc signature applied by
+      # darwin.autoSignDarwinBinariesHook; re-sign so the .node loads under
+      # Gatekeeper on end-user macOS hosts.
+      codesign --force --sign - "$NODE_LIB"
 
       # Assert no /nix/store references remain — guards against silent
       # no-ops in the rewrites above and catches the 1.10.0 regression.
       # See https://github.com/xmtp/libxmtp/issues/3479.
-      remaining=$(otool -L "$NODE_LIB" | awk '$1 ~ /^\/nix\/store\// { print $1 }')
+      # NR > 1 skips otool -L's header line (the file's own /nix/store path).
+      remaining=$(otool -L "$NODE_LIB" | awk 'NR > 1 && $1 ~ /^\/nix\/store\// { print $1 }')
       if [ -n "$remaining" ]; then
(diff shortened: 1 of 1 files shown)


insipx:
> Drift-proof the libiconv rewrite. Replace the hardcoded install_name_tool -change "${darwin.libiconv}/lib/libiconv.2.dylib" ... with a loop over /nix/store/.../libiconv.<N>.dylib refs actually discovered in otool -L output. The hardcoded form silently becomes a no-op under nixpkgs drift or cross-compile splicing — if ${darwin.libiconv} evaluates to a different store hash than the one the linker embedded, -change matches nothing and the binary ships with a /nix/store libiconv reference. Using otool -L as the source of truth closes that gap.

i still don't understand how nix could insert a _different_ store hash during the derivation build than the linker, if the build for node is pure? what is "store path drift" and how does it happen

if the linker is linking against darwin.libiconv in buildinputs, and we have the _same_ darwin.libiconv in postFixup, of the _same_ derivation, how could that differ?

neekolas (approved):

xmtp-coder-agent:
Fair point — you're right. In a pure sandboxed build `${darwin.libiconv}` resolves to the exact store path the linker saw, and the CI log actually confirms the hardcoded `-change` is doing its job: the "leak" the assertion flagged was the `otool -L` header line (the file's own `/nix/store` path), not an unrewritten libiconv ref. "Drift / cross-compile splicing" was speculative and I don't have evidence of it happening here.

Pushed 6db38a8 to narrow the PR back to what's actually needed:

- **Kept:** `NR > 1` on the assertion awk filter — this is the real bug.
- **Kept:** `codesign --force --sign -` after `install_name_tool` — #3500's commit message promised this (since `install_name_tool` invalidates `autoSignDarwinBinariesHook`'s ad-hoc signature) but didn't ship it.
- **Reverted:** the `otool -L | awk | while read` rewrite loop. Back to the original single `install_name_tool -change "${darwin.libiconv}/..."` line.

Final diff is +7/-1.

PR #3515, opened 2026-04-15
fix(mls): don't burn publish-attempt budget on retriable send failures

### LINKED ISSUE #3474: Fix retriable published-intent sends getting stuck as already-published
## Summary

`origin/main` still carries a regression where a retriable send failure can leave an earlier intent marked as "published" even though the payload never actually made it out.

The standing coverage for this is present but disabled:
- `crates/xmtp_mls/src/groups/tests/test_commit_log_local.rs`
- `test_commit_log_retriable_error`
- current comments include:
  - `TODO(rich): Fix intent publishing on bad network conditions`
  - `This currently fails with error SyncFailedToWait, because the intent has been marked as 'published' despite not being published.`

## Current Behavior

In the bad-network path exercised by `test_commit_log_retriable_error`, we queue a `KeyUpdate` intent followed by a `SendMessage` intent while the proxied client is offline. After re-enabling the network, `sync_until_last_intent_resolved()` should recover, but on `main` the test is still ignored because the earlier intent is treated as already published and the flow fails with `SyncFailedToWait`.

## Expected Behavior

Retriable publish failures should not permanently poison the earlier intent. Once connectivity returns:
- `sync_until_last_intent_resolved()` should succeed
- the pending commit should be published exactly once
- the local commit logs should converge as expected

## Acceptance Criteria

- `test_commit_log_retriable_error` is unignored and passing on the relevant runtime(s)
- the intent-publishing flow no longer marks retriable failed sends as durably published
- the fix preserves idempotent recovery when connectivity is restored

## Why This Matters

This is one of the remaining explicit regression points still visible on `main`, and it affects recovery from transient network failures in a high-risk transaction path.

xmtp-coder-agent:
Resolves #3474

## Summary

When `publish_intents` saw a transport-level error from `send_group_messages` (e.g. while the client was offline), the bookkeeping treated it the same as a content-level failure: it bumped `publish_attempts` and, after `MAX_INTENT_PUBLISH_ATTEMPTS` (3), transitioned the intent to `Error`. A sustained outage spanning a few sync cycles would shoot every queued intent dead even though nothing about the intents themselves was wrong — the network was just down.

This change threads `err.is_retryable()` into `handle_published_intent_send_failure`. On a retriable failure the intent is reset to `ToPublish` **without** incrementing `publish_attempts`, so the next sync — after connectivity returns — can re-encrypt and republish at the current epoch. Non-retriable failures keep today's bounded-retry behavior unchanged.

## Why this design

- **Minimal blast radius.** No new enum variants, no new DB migrations, no change to `set_group_intent_to_publish`. We just stop counting retriable errors against a budget that was meant for content-level failures.
- **Correct retry semantics.** Retriable errors are, by definition, ones the caller can safely retry; burning a 3-attempt budget on them effectively turned a recoverable outage into a permanent intent failure, contradicting the existing `RetryableError` trait contract.
- **Loop bounds still hold.** `publish_intents` still returns `Err(err)` after handling a retriable failure, so `sync_with_conn` still records the publish error in its `SyncSummary` and `sync_until_intent_resolved_inner` still gives up after `MAX_GROUP_SYNC_RETRIES`. The difference is that the *intent* is now left in a publishable state, so a later sync (or `sync_until_last_intent_resolved` once the network returns) picks it up and pushes it through.

## Changes

- `crates/xmtp_mls/src/groups/mls_sync.rs`
  - `handle_published_intent_send_failure` now takes a `bool is_retryable`. The retriable branch only calls `set_group_intent_to_publish` and emits a `tracing::warn!`.
  - Call site in `publish_intents` computes `let is_retryable = err.is_retryable();` once and threads it through (also surfaces `is_retryable` in the existing `GroupSyncPublishFailed` log event).
  - Updates the existing `send_failures_for_published_intents_revert_to_to_publish` unit test to the new signature, and adds `retriable_send_failures_revert_to_to_publish_without_incrementing_attempts` covering the new branch.
- `crates/xmtp_mls/src/groups/tests/test_commit_log_local.rs`
  - Removes `#[ignore]` and the `TODO(rich)` comment on `test_commit_log_retriable_error`.

## Test plan

- [x] `just test v3 test_commit_log_retriable_error` — passes (was previously `#[ignore]`)
- [x] `just test v3 test_commit_log_non_retriable_error` — passes (no regression on the bounded-retry path)
- [x] All 30 `test_commit_log_*` and `send_failures_*` / `retriable_send_failures_*` tests in `xmtp_mls` pass via `cargo nextest`
- [x] `just lint-rust` clean (clippy `--all-features --all-targets -Dwarnings`, fmt, hakari)

> [!NOTE]
> ### Fix MLS intent publish to use a higher attempt budget for retriable send failures
> - Adds `MAX_INTENT_RETRIABLE_PUBLISH_ATTEMPTS` (12) alongside the existing lower cap, and a `cap_for_publish_attempts(is_retryable: bool)` helper in [`mls_sync.rs`](https://github.com/xmtp/libxmtp/pull/3515/files#diff-160aadcd7de477c1116260a79601a0548842f3bd91bf16bd8443debb66ce6684) to select between them.
> - Retriable failures during intent data preparation and send now reset the intent to `ToPublish` and consume the larger cap before poisoning; non-retriable failures keep the original smaller cap.
> - Behavioral Change: previously, all publish failures counted against the same budget and could poison an intent after just a few transient network errors; now transient failures get up to 12 attempts before the intent is marked as errored.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 08274f0.</sup>
> 
>

### DIFF
--- a/crates/xmtp_configuration/src/common/mls.rs
+++ b/crates/xmtp_configuration/src/common/mls.rs
@@ -15,6 +15,18 @@ pub const MAX_GROUP_SYNC_RETRIES: usize = 3;
 
 pub const MAX_INTENT_PUBLISH_ATTEMPTS: usize = 3;
 
+/// Fail-safe cap on publish attempts for *retriable* errors.
+///
+/// Retriable publish failures (e.g. a client that is temporarily offline)
+/// use a separate budget from content-level failures so that a normal
+/// outage cannot poison a perfectly good intent. The generous cap here
+/// exists as defense-in-depth against errors that the API layer marks
+/// retriable but which would never actually recover — for example, a
+/// malformed payload that the server persistently rejects with a
+/// retriable gRPC status. In that case the intent will still eventually
+/// be poisoned rather than retried forever.
+pub const MAX_INTENT_RETRIABLE_PUBLISH_ATTEMPTS: usize = 30;
+
 pub const GROUP_KEY_ROTATION_INTERVAL_NS: i64 = NS_IN_30_DAYS;
 
 pub const KEY_PACKAGE_QUEUE_INTERVAL_NS: i64 = 5 * NS_IN_SEC; // 5 secs
--- a/crates/xmtp_configuration/src/common/mls.rs
+++ b/crates/xmtp_configuration/src/common/mls.rs
@@ -19,13 +19,14 @@ pub const MAX_INTENT_PUBLISH_ATTEMPTS: usize = 3;
 ///
 /// Retriable publish failures (e.g. a client that is temporarily offline)
 /// use a separate budget from content-level failures so that a normal
-/// outage cannot poison a perfectly good intent. The generous cap here
-/// exists as defense-in-depth against errors that the API layer marks
-/// retriable but which would never actually recover — for example, a
-/// malformed payload that the server persistently rejects with a
-/// retriable gRPC status. In that case the intent will still eventually
-/// be poisoned rather than retried forever.
-pub const MAX_INTENT_RETRIABLE_PUBLISH_ATTEMPTS: usize = 30;
+/// outage cannot poison a perfectly good intent. The cap here exists as
+/// defense-in-depth against errors that the API layer marks retriable but
+/// which would never actually recover — for example, a malformed payload
+/// that the server persistently rejects with a retriable gRPC status. In
+/// that case the intent will eventually be poisoned rather than retried
+/// forever, but the larger budget tolerates a real offline window before
+/// the app surfaces a hard failure to the user.
+pub const MAX_INTENT_RETRIABLE_PUBLISH_ATTEMPTS: usize = 12;
 
 pub const GROUP_KEY_ROTATION_INTERVAL_NS: i64 = NS_IN_30_DAYS;
 
--- a/crates/xmtp_mls/src/groups/mls_sync.rs
+++ b/crates/xmtp_mls/src/groups/mls_sync.rs
@@ -81,9 +81,9 @@ use xmtp_common::{
 };
 use xmtp_configuration::{
     GRPC_PAYLOAD_LIMIT, HMAC_SALT, MAX_GROUP_SIZE, MAX_GROUP_SYNC_RETRIES,
-    MAX_INTENT_PUBLISH_ATTEMPTS, MAX_PAST_EPOCHS, PROPOSAL_SUPPORT_EXTENSION_ID,
-    SYNC_BACKOFF_TOTAL_WAIT_MAX_SECS, SYNC_BACKOFF_WAIT_MS, SYNC_JITTER_MS,
-    SYNC_UPDATE_INSTALLATIONS_INTERVAL_NS,
+    MAX_INTENT_PUBLISH_ATTEMPTS, MAX_INTENT_RETRIABLE_PUBLISH_ATTEMPTS, MAX_PAST_EPOCHS,
+    PROPOSAL_SUPPORT_EXTENSION_ID, SYNC_BACKOFF_TOTAL_WAIT_MAX_SECS, SYNC_BACKOFF_WAIT_MS,
+    SYNC_JITTER_MS, SYNC_UPDATE_INSTALLATIONS_INTERVAL_NS,
 };
 use xmtp_content_types::{CodecError, ContentCodec, group_updated::GroupUpdatedCodec};
 use xmtp_db::message_deletion::{QueryMessageDeletion, StoredMessageDeletion};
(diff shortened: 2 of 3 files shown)


crates/xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -3898,7 +3900,25 @@ pub(crate) fn decode_staged_commit(
 fn handle_published_intent_send_failure<Db: QueryGroupIntent>(
     db: &Db,
     intent: &StoredGroupIntent,
+    is_retryable: bool,
 ) -> Result<(), GroupError> {
+    if is_retryable {
```
neekolas: This is a little scary since it can lead to infinite retries if the network is rejecting the message for valid reasons (maybe it is malformed).

1. Confirm that there is some sort of backoff before the next attempt
2. Check to see if there are any "retriable" errors that _may_ never recover. If so, we may want to gate on specific error codes rather than the generic `err.is_retryable`

xmtp-coder-agent: Good callout — pushed 082aa85 to address both.

**1. Backoff exists.** Between `sync_with_conn` attempts inside `sync_until_intent_resolved_inner` (`crates/xmtp_mls/src/groups/mls_sync.rs` L515–519 and the sleep at L587): `ExponentialBackoff` with `SYNC_BACKOFF_WAIT_MS` base, `SYNC_BACKOFF_TOTAL_WAIT_MAX_SECS` cap, and jitter. So a single sync-until call does at most `MAX_GROUP_SYNC_RETRIES = 3` publish attempts with backoff, then returns `SyncFailedToWait`.

**2. You're right about the classification gap.** `GrpcError::is_retryable()` in `crates/xmtp_api_grpc/src/error.rs:109–113` unconditionally returns `true`, so a `GrpcError::Status` wrapping something permanent (say `INVALID_ARGUMENT` on a malformed payload) is classified the same as a transient `UNAVAILABLE`. Tightening that classification cleanly means either extending `RetryableError` with an `Any` bound to downcast `Box<dyn RetryableError>` → `tonic::Status` and inspect `.code()`, or adding a publish-specific `is_publish_retryable()` on `ApiError` — both felt like they belonged in a separate PR touching the API layer rather than being bolted onto this one.

In the meantime I've added defense-in-depth inside `handle_published_intent_send_failure`:

- Always bump `publish_attempts` (retriable and non-retriable).
- Two caps instead of one:
  - Non-retriable: `MAX_INTENT_PUBLISH_ATTEMPTS = 3` (unchanged).
  - Retriable: `MAX_INTENT_RETRIABLE_PUBLISH_ATTEMPTS = 30`.

A typical offline window recovers well before hitting 30 (one sync-until call contributes ≤ 3), but a server that persistently rejects a message with a retriable status now poisons the intent after ~10 sync-until calls instead of looping forever. No migration — reuses the existing `publish_attempts` column with a second threshold. Added a unit test that walks an intent past the retriable cap and asserts it transitions to `Error`. Happy to open a follow-up PR to tighten `GrpcError::is_retryable()` itself if you'd prefer that to the cap approach.

crates/xmtp_mls/src/groups/mls_sync.rs
```diff
@@ -3898,19 +3900,49 @@ pub(crate) fn decode_staged_commit(
...
+    db.increment_intent_publish_attempt_count(intent.id)?;
+    let next_attempts = (intent.publish_attempts + 1) as usize;
+
+    // Non-retriable failures keep the tight MAX_INTENT_PUBLISH_ATTEMPTS cap
+    // (guards against content-level failures). Retriable failures get a much
+    // higher cap so a normal offline window can recover; the cap exists as
+    // defense-in-depth against errors that the API layer marks retriable
+    // but which would never actually succeed (e.g. a malformed payload the
+    // server persistently rejects with a retriable gRPC status).
+    let cap = if is_retryable {
```
neekolas: Are we sure that the intent is not validated against MAX_INTENT_PUBLISH_ATTEMPTS anywhere else that might short circuit on retries?

crates/xmtp_configuration/src/common/mls.rs
```diff
@@ -15,6 +15,18 @@ pub const MAX_GROUP_SYNC_RETRIES: usize = 3;
...
+///
+/// Retriable publish failures (e.g. a client that is temporarily offline)
+/// use a separate budget from content-level failures so that a normal
+/// outage cannot poison a perfectly good intent. The generous cap here
+/// exists as defense-in-depth against errors that the API layer marks
+/// retriable but which would never actually recover — for example, a
+/// malformed payload that the server persistently rejects with a
+/// retriable gRPC status. In that case the intent will still eventually
+/// be poisoned rather than retried forever.
+pub const MAX_INTENT_RETRIABLE_PUBLISH_ATTEMPTS: usize = 30;
```
neekolas: 30 seems very high. Could leave the app inoperable for quite some time. What about 12 as a compromise?

crates/xmtp_mls/src/groups/mls_sync.rs
xmtp-coder-agent: Good catch — the same check at `mls_sync.rs:2474` (inside `publish_intents`'s `get_publish_intent_data` error handler) was still gating on the tight non-retriable cap. Pushed 08274f05b: extracted a `cap_for_publish_attempts(is_retryable)` helper and call it from both sites, so a transient network failure during data preparation (e.g. fetching identity updates / key packages for `UpdateGroupMembership`) gets the same generous budget instead of poisoning the intent in three tries.

crates/xmtp_configuration/src/common/mls.rs
xmtp-coder-agent: Agreed — dropped to 12 in 08274f05b. Updated the doc comment to reflect that 12 still tolerates a real offline window without leaving the app inoperable for excessively long if a misclassified-retriable error never recovers.

xmtp-coder-agent:
The Android Integration Tests failure on `08274f05b` is `org.xmtp.android.library.DmTest.testDmConsent` (`expected:<DENIED> but was:<ALLOWED>`) — appears to be a flake in the consent-state set/read path:

```kotlin
fixtures.boClient.preferences.setConsentState(
    listOf(ConsentRecord(dm.id, EntryType.CONVERSATION_ID, ConsentState.DENIED)),
)
assertEquals(
    fixtures.boClient.preferences.conversationState(dm.id),
    ConsentState.DENIED,
)
```

This PR only changes the publish-attempt cap logic in `crates/xmtp_mls/src/groups/mls_sync.rs` and `crates/xmtp_configuration/src/common/mls.rs` — there's no path by which lowering `MAX_INTENT_RETRIABLE_PUBLISH_ATTEMPTS` or extracting a `cap_for_publish_attempts` helper could affect a consent-state DB write/read race. All other CI checks (`test-workspace`, `test-android Unit Tests`, `test-wasm`, `test-node`, `Lint`, `cargo-deny`, `codecov`) are green.

I don't have rerun permission on the upstream — could you re-run the failed `test-android / Integration Tests (Android)` job? Happy to push an empty commit instead if you'd prefer to retrigger that way.

PR #3526, opened 2026-04-20
ci(fh-cache): keep-going and fail-fast=false for full cache coverage

### LINKED ISSUE #3525: use 'keep-going' in cache nix outputs. don't cancel other matrix builds if one matrix fails
use keep-going to build everything despite a single package failure. also, don't cancel other matrix builds if one matrix build fails. extra commands may be passed to `om ci run -- --keep-going`

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3525

## Summary

Harden the `Cache all Nix Outputs` workflow (`.github/workflows/fh-cache.yml`) so that a
single failing flake output or a single failing OS no longer prevents the cache from
being refreshed for the rest.

- **`-- --keep-going`** is forwarded from `om ci run` to the underlying `nix build`
  (supported via `om ci run`'s `--` passthrough — see
  [omnix docs](https://omnix.page/om/ci.html)). Remaining outputs continue building after
  a single-package failure, and the existing cachix-pin step then uploads every output
  that *did* build (it reads `om.json → .result.omnix.build.byName`, which lists only
  successful results).
- **`strategy.fail-fast: false`** on the `build` matrix prevents a failure on one OS from
  cancelling the still-running sibling OS, so both `warp-ubuntu-latest-x64-16x` and
  `warp-macos-26-arm64-12x` always run to completion independently.

If any output fails, the step still exits non-zero and the job is still marked failed, so
regressions remain visible.

## Test plan

- [ ] Next push to `main` (or manual `workflow_dispatch`) shows both matrix legs running
      to completion and each leg attempts every flake output even on failure.
- [ ] `actionlint` (local + the repo's `lint-config` job) reports no new warnings for the
      edited file.

> [!NOTE]
> ### Keep CI cache builds running on failure with `--keep-going` and `fail-fast: false`
> Updates [fh-cache.yml](https://github.com/xmtp/libxmtp/pull/3526/files#diff-ce3dd25bde306abc234d96f51a92cb524edcabd8e9e5eeed5eb121f73b32b9e2) so all matrix jobs and build targets are attempted even when some fail. Sets `strategy.fail-fast: false` on the build matrix and appends `-- --keep-going` to the `om ci run` command, ensuring full cache coverage across all flake outputs regardless of individual failures.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0500f46.</sup>
> 
>

### DIFF
--- a/.github/workflows/fh-cache.yml
+++ b/.github/workflows/fh-cache.yml
@@ -12,6 +12,7 @@ concurrency:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         os: [warp-ubuntu-latest-x64-16x, warp-macos-26-arm64-12x]
     runs-on: ${{ matrix.os }}
@@ -27,7 +28,7 @@ jobs:
       - name: Install omnix
         run: nix profile install nixpkgs#omnix
       - name: Build all flake outputs
-        run: om ci run --include-all-dependencies --results=om.json
+        run: om ci run --include-all-dependencies --results=om.json -- --keep-going
       - name: Pin outputs to cachix
         # if: github.ref_name == 'main'
         run: |


insipx (approved):

insipx:
@xmtp-coder-agent ensure this works by running the nix-cache-action on the gh repo

neekolas (approved):

xmtp-coder-agent:
I tried running the workflow — here's what I found and what I verified locally.

## Ran on the fork

I synced my fork's default branch to register all workflows (they weren't active before), then dispatched `fh-cache.yml` on `fix/issue-3525`:
<https://github.com/xmtp-coder-agent/libxmtp/actions/runs/24691756244>

The dispatch produced **three jobs** as expected:
- `build (warp-ubuntu-latest-x64-16x)`
- `build (warp-macos-26-arm64-12x)`
- `validation-image-aarch64`

The matrix expanded correctly with `fail-fast: false` — both `build` legs queued independently. However all three sat in `queued` indefinitely because the `warp-*` WarpBuild runners are org-scoped to `xmtp` and aren't addressable from my fork. I cancelled the run rather than leave it dangling. A full end-to-end execution has to happen from inside `xmtp/libxmtp` (either by merging this draft or by letting me open it non-draft and running the workflow there).

## Verified locally

To validate the two mechanisms without the warp runners:

**1. `om ci run -- --keep-going` forwards the flag to `nix build` / `nix flake lock`.**
Running `om ci run --systems x86_64-linux . -- --keep-going -v` against this repo shows omnix parsing it into `extra_nix_args: ["--keep-going"]` and emitting:

```
❄️  nix flake lock --keep-going . --no-update-lock-file
❄️  nix build --keep-going '/nix/store/…#json' -L --print-out-paths --no-link …
```

So the flag reaches `nix build` for every output omnix dispatches.

**2. `--keep-going` is a recognised Nix setting** (`nix config show | grep keep-going` returns `keep-going = false` by default, and `nix build --keep-going nixpkgs#hello` is accepted).

**3. `actionlint` reports no new warnings** introduced by the diff (the pre-existing warnings about custom `warp-*` labels and `github.head_ref` in a shell are unchanged).

**4. Partial-success pinning already works unchanged.** The pin step iterates `.result.omnix.build.byName` from `om.json`, which omnix only populates with successfully-built outputs — so whatever keeps going gets pinned.

Happy to open this PR non-draft (or add a tiny GitHub-hosted smoke workflow that builds a deliberately-broken tiny flake to show the behaviour end-to-end on a standard runner) if you'd like additional proof before merging.

PR #3576, opened 2026-04-28
fix(nix): bind openssl-sys against prebuilt openssl in base.nix

### LINKED ISSUE #3575: Flaky CI Failure: openssl-sys build script fails on macOS 26 runner for musl cross-compilation targets
## Flaky CI Failure: openssl-sys build script fails on macOS 26 runner for musl cross-compilation targets

**Failed run:** https://github.com/xmtp/libxmtp/actions/runs/25076737710
**Commit:** dec1eab31f29b50c1b8e115387e173022c71f92d

### Summary

The `Build all flake outputs` step failed on the macOS 26 ARM runner (`warp-macos-26-arm64-12x`) because the `openssl-sys v0.9.114` build script exited with status 1 when cross-compiling for both `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` targets. The failure occurs in OpenSSL's `util/mkinstallvars.pl` install script, which reports an uninitialized `LIBDIR` variable. Both musl targets failed identically, cascading to fail `mls-validation-service` and `bindings-node-js-napi-*-linux-musl` derivations, which ultimately caused the top-level `devour-output.json.drv` to fail with 9 dependent failures.

(On the Linux runner, there was an unrelated transient `No space left on device` failure that killed the Android target builds — that is an independent infrastructure issue.)

### Error Details

```
  cargo-package-deps-x86_64-unknown-linux-musl> error: failed to run custom build command for `openssl-sys v0.9.114`
  
  Caused by:
    process didn't exit successfully: `.../build/openssl-sys-c921086b11814554/build-script-main` (exit status: 1)
    --- stderr
    DEBUG: all keys: APPLINKDIR, BINDIR, CMAKECONFIGDIR, COMMENT, ENGINESDIR, INCLUDEDIR, LDLIBS, LIBDIR, MODULESDIR, PKGCONFIGDIR, PREFIX, VERSION, libdir
    Use of uninitialized value in concatenation (.) or string at util/mkinstallvars.pl line 55.
    DEBUG: LIBDIR = , libdir =  => Use of uninitialized value in concatenation (.) or string at util/mkinstallvars.pl line 57.
    LIBDIR =
    No value given for CMAKECONFIGDIR
    No value given for PKGCONFIGDIR
    No value given for libdir
    Use of uninitialized value in concatenation (.) or string at util/mkinstallvars.pl line 72.
    ...
    Use of uninitialized value in join or string at util/mkinstallvars.pl line 157.

error: Cannot build '/nix/store/lhakfyf8nx1002k3r2cci0i6r9rsw12b-cargo-package-deps-aarch64-unknown-linux-musl-1.10.0.drv'.
error: Cannot build '/nix/store/isvvy88ln7l38nrkp03c6z6xr3nzk4hg-cargo-package-deps-x86_64-unknown-linux-musl-1.10.0.drv'.
...
error: Cannot build '/nix/store/y4wmj748zza5jjykafq37b5ap11lh7hl-devour-output.json.drv'.
```

The cc invocations during the OpenSSL build use `-arch arm64 --target=arm64-apple-macosx -mmacosx-version-min=11.0` (host compilation for the build script runner). The install-phase Perl script then fails with uninitialized `LIBDIR`.

### Analysis

The failure is specific to the `warp-macos-26-arm64-12x` runner (macOS 26 / Tahoe) and affects both musl cross-compilation targets (`x86_64` and `aarch64`) identically. This is not a code regression in libxmtp itself — it is a build environment issue in which the `openssl-sys` crate's source-compiled OpenSSL fails its install-phase Perl script (`mkinstallvars.pl`) on macOS 26.

Root cause hypothesis: macOS 26 may ship a newer version of Perl that changes the behavior of uninitialized variable handling in `mkinstallvars.pl`, or the macOS 26 SDK changes something in the cross-compilation environment that causes OpenSSL's `LIBDIR` to be empty during the install step. The `openssl-sys` crate compiles OpenSSL from source when system headers are not present for the cross-compilation target.

This is a **build/infrastructure failure** at the intersection of the macOS 26 runner environment and the Nix cross-compilation setup for musl targets.

---

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3575

## Summary

Tell `openssl-sys` to bind against the prebuilt nixpkgs `openssl` instead of falling through to its vendored `openssl-src` path. The vendored source build's `util/mkinstallvars.pl` aborts on the macOS 26 runner with empty `LIBDIR` (`Use of uninitialized value` errors), breaking `cargo-package-deps-{x86_64,aarch64}-unknown-linux-musl` and cascading through `mls-validation-service-*-musl`, `bindings-node-js-napi-*-linux-musl`, and `devour-output.json.drv`.

## Implementation

`nix/lib/base.nix` exposes a separate `opensslEnv` attribute set with `OPENSSL_NO_VENDOR=1` plus `OPENSSL_DIR` / `OPENSSL_LIB_DIR` / `OPENSSL_INCLUDE_DIR`. Consumers that keep `buildInputs` intact (`nix/package/mls_validation_service.nix` and `nix/package/node.nix`) merge `opensslEnv` into both `commonArgs` and the `cargoArtifacts` overrides.

The first attempt at this fix folded `opensslEnv` directly into `commonArgs`, which broke Android: `nix/package/android.nix` deliberately clears `buildInputs` to use the NDK sysroot, but a literal `OPENSSL_DIR = "${openssl.dev}"` reintroduced cross-compiled `openssl-x86_64-unknown-linux-android-3.6.1` into the closure, and that derivation cannot build (`internal/ktls.h` needs `struct msghdr`/`struct cmsghdr` from the Linux kernel headers, which the Android sysroot doesn't satisfy). A `lib.optionalAttrs (!stdenv.hostPlatform.isAndroid)` guard inside `base.nix` doesn't help — nixpkgs splicing evaluates `stdenv` to the build-platform variant for the env-var string-interpolation path, so the predicate reads `false` in the cross-android pkgset anyway.

Keeping `opensslEnv` separate sidesteps both issues:

- Android (and iOS) keep the vendored `openssl-src` path they have always used. iOS is also explicitly excluded because the vendored path was working there before issue #3575 was opened.
- `mls_validation_service` and `node` (musl + native) get the bound-against-nixpkgs path that resolves the macOS 26 failure.

## Verification

`nix derivation show` on the realised derivations:

```
cargo-package-deps-x86_64-unknown-linux-musl-1.10.0
  OPENSSL_NO_VENDOR=1
  OPENSSL_DIR=/nix/store/...-openssl-x86_64-unknown-linux-musl-3.6.1-dev
  OPENSSL_LIB_DIR=/nix/store/...-openssl-x86_64-unknown-linux-musl-3.6.1/lib
  OPENSSL_INCLUDE_DIR=/nix/store/...-openssl-x86_64-unknown-linux-musl-3.6.1-dev/include

mls-validation-service-x86_64-unknown-linux-musl-1.10.0
  (same OPENSSL_* values)

bindings-node-js-napi-x86_64-unknown-linux-musl-1.10.0
  (same OPENSSL_* values)

xmtpv3-x86_64-linux-android-x86_64-unknown-linux-android-1.10.0
  (no OPENSSL_* env vars)
  buildInputs=
  no openssl drv in inputDrvs closure

cargo-package-deps-x86_64-unknown-linux-android-1.10.0
  (no OPENSSL_* env vars)
  buildInputs=
  no openssl drv in inputDrvs closure
```

## Test plan

- [ ] Re-run the `Cache all Nix Outputs` workflow on `warp-macos-26-arm64-12x` and confirm the musl derivations build cleanly.
- [ ] Confirm `test-android / Integration Tests (Android)` and `test-android / Unit Tests (Android)` build green again.
- [ ] Confirm `mls-validation-service-{x86_64,aarch64}-unknown-linux-musl` and `bindings-node-js-napi-*-linux-musl` build on the Linux runner.

### DIFF
--- a/nix/lib/base.nix
+++ b/nix/lib/base.nix
@@ -70,6 +70,15 @@ let
     "AWS_LC_SYS_TARGET_CC_${buildPlatformSuffix}" = "cc";
     "AWS_LC_SYS_TARGET_CXX_${buildPlatformSuffix}" = "c++";
 
+    # Tell openssl-sys to bind against the prebuilt nixpkgs openssl already in
+    # buildInputs instead of falling through to its vendored `openssl-src` path,
+    # which compiles OpenSSL from source. The source build's `mkinstallvars.pl`
+    # fails on macOS 26 cross-compiles to musl with empty `LIBDIR`.
+    # See https://github.com/xmtp/libxmtp/issues/3575.
+    OPENSSL_NO_VENDOR = "1";
+    OPENSSL_DIR = "${openssl.dev}";
+    OPENSSL_LIB_DIR = "${lib.getLib openssl}/lib";
+    OPENSSL_INCLUDE_DIR = "${openssl.dev}/include";
   };
 
   # Make cargo artifacts for a derivation building rust code
--- a/nix/lib/base.nix
+++ b/nix/lib/base.nix
@@ -69,12 +69,21 @@ let
     # https://crane.dev/faq/cross-compiling-aws-lc-sys.html
     "AWS_LC_SYS_TARGET_CC_${buildPlatformSuffix}" = "cc";
     "AWS_LC_SYS_TARGET_CXX_${buildPlatformSuffix}" = "c++";
+  };
 
-    # Tell openssl-sys to bind against the prebuilt nixpkgs openssl already in
-    # buildInputs instead of falling through to its vendored `openssl-src` path,
-    # which compiles OpenSSL from source. The source build's `mkinstallvars.pl`
-    # fails on macOS 26 cross-compiles to musl with empty `LIBDIR`.
-    # See https://github.com/xmtp/libxmtp/issues/3575.
+  # Tell openssl-sys to bind against the prebuilt nixpkgs openssl already in
+  # buildInputs instead of falling through to its vendored `openssl-src` path,
+  # which compiles OpenSSL from source. The source build's `mkinstallvars.pl`
+  # fails on macOS 26 cross-compiles to musl with empty `LIBDIR`.
+  # See https://github.com/xmtp/libxmtp/issues/3575.
+  #
+  # Not folded into `commonArgs` because `package/android.nix` deliberately
+  # clears `buildInputs` to use the NDK sysroot — we let openssl-sys vendor
+  # there since the cross-compiled openssl for `*-unknown-linux-android`
+  # fails to build (kernel TLS headers in `internal/ktls.h` aren't satisfied
+  # by the Android sysroot). Consumers that keep `buildInputs` intact merge
+  # this set into their args explicitly.
+  opensslEnv = {
     OPENSSL_NO_VENDOR = "1";
     OPENSSL_DIR = "${openssl.dev}";
     OPENSSL_LIB_DIR = "${lib.getLib openssl}/lib";
@@ -107,5 +116,6 @@ in
     bindingsFileset
     commonArgs
     mkCargoArtifacts
+    opensslEnv
     ;
 }
--- a/docs/plans/2026-04-28-issue-3575-design.md
+++ b/docs/plans/2026-04-28-issue-3575-design.md
@@ -28,16 +28,23 @@ The dev-shell config already sets these vars (see `nix/lib/shell-common.nix:72-7
 
 ## Implementation
 
-In `nix/lib/base.nix`, extend the `commonArgs` attrset with:
+In `nix/lib/base.nix`, expose a separate `opensslEnv` attribute set:
 
 ```nix
-OPENSSL_NO_VENDOR = "1";
-OPENSSL_DIR = "${openssl.dev}";
-OPENSSL_LIB_DIR = "${lib.getLib openssl}/lib";
-OPENSSL_INCLUDE_DIR = "${openssl.dev}/include";
+opensslEnv = {
+  OPENSSL_NO_VENDOR = "1";
+  OPENSSL_DIR = "${openssl.dev}";
+  OPENSSL_LIB_DIR = "${lib.getLib openssl}/lib";
+  OPENSSL_INCLUDE_DIR = "${openssl.dev}/include";
+};
 ```
 
-`openssl` is already a parameter of `base.nix` and is included in `buildInputs`, so no new inputs are needed. The form mirrors `shell-common.nix:72-74`, with `OPENSSL_INCLUDE_DIR` added explicitly per the maintainer's suggestion in the issue thread.
+The first attempt folded these into `commonArgs`. That broke `nix/package/android.nix`, which deliberately clears `buildInputs` (it links against the NDK sysroot). The `OPENSSL_DIR = "${openssl.dev}"` interpolation forced the cross-compiled `openssl-x86_64-unknown-linux-android-3.6.1` derivation into the closure, where it fails to build — `internal/ktls.h` references `struct msghdr` / `struct cmsghdr` that aren't satisfied by the Android sysroot. Splicing also evaluates `stdenv.hostPlatform.isAndroid` inconsistently inside `base.nix`, so a simple `lib.optionalAttrs (!stdenv.hostPlatform.isAndroid)` guard does not actually skip the env vars on Android.
+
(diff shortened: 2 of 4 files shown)


nix/lib/base.nix
```diff
@@ -69,7 +69,25 @@ let
...
+  # fails on macOS 26 cross-compiles to musl with empty `LIBDIR`.
+  # See https://github.com/xmtp/libxmtp/issues/3575.
+  #
+  # Not folded into `commonArgs` because `package/android.nix` deliberately
+  # clears `buildInputs` to use the NDK sysroot — we let openssl-sys vendor
+  # there since the cross-compiled openssl for `*-unknown-linux-android`
+  # fails to build (kernel TLS headers in `internal/ktls.h` aren't satisfied
+  # by the Android sysroot). Consumers that keep `buildInputs` intact merge
+  # this set into their args explicitly.
+  opensslEnv = {
```
insipx, line 86: No need to create a separate env. You can just add them to base it's ok to apply to all the builds

insipx (changes requested):

nix/lib/base.nix
xmtp-coder-agent: I tried that first (commit b55495dad). It broke `test-android` because `${lib.getLib openssl}/lib` pulls the cross-compiled `openssl-x86_64-unknown-linux-android-3.6.1.drv` into the closure of every consumer of `commonArgs`, including `package/android.nix` — and that drv fails to build (`internal/ktls.h` references `struct msghdr`/`struct cmsghdr` that aren't in the Android NDK sysroot).

Verified locally just now: with the env vars in `commonArgs`, `nix derivation show -r .#android-bindings-x86_64` lists `5dm6jmmlz3mxhbnh2yx1a6mlhgkscgaq-openssl-x86_64-unknown-linux-android-3.6.1.drv` in the closure (4509 drvs total). Splicing also picks different `openssl` versions for `${openssl.dev}` vs `${lib.getLib openssl}` in the cross-android pkgset, so I couldn't gate it cleanly with `lib.optionalAttrs (!stdenv.hostPlatform.isAndroid)` either.

A separate `opensslEnv` keeps `android.nix`'s NDK-sysroot strategy intact (it never references `xmtp.base.opensslEnv`) while letting `mls_validation_service.nix` and `node.nix` opt in. Happy to fold it back into `commonArgs` if you have a way to keep cross-android-openssl out of the closure that I missed.

insipx: Ok stay tuned I have better solution for this in one of my branches will push tomorrow

insipx: why use lib.getLib? try wihtout that, just

```
    OPENSSL_NO_VENDOR = "1";
    OPENSSL_DIR = "${openssl.dev}";
    OPENSSL_LIB_DIR = "${openssl.out}/lib";
    OPENSSL_INCLUDE_DIR = "${openssl.dev}/include";

```

xmtp-coder-agent: Tried that. `lib.getLib` is a red herring — `${openssl.out}/lib` produces the same closure. Test with the env vars in `commonArgs` using the snippet you posted:

```
$ nix derivation show -r .#packages.x86_64-linux.android-bindings-x86_64
Total drvs in closure: 4509
android openssl drvs in closure: ['5dm6jmmlz3mxhbnh2yx1a6mlhgkscgaq-openssl-x86_64-unknown-linux-android-3.6.1.drv']
```

The direct referrer is the cargo-package-deps for android plus the xmtpv3 dylib itself, both via `OPENSSL_DIR=${openssl.dev}` / `OPENSSL_INCLUDE_DIR=${openssl.dev}/include`. Even with `buildInputs = []` from `android.nix`'s specialArgs, those env-var interpolations alone are enough to put the cross-android openssl into the closure.

For the musl side `${openssl.out}/lib` works fine and it's a cleaner expression than `${lib.getLib openssl}`, so happy to swap that — but to fold into `commonArgs` we'd also need to drop `OPENSSL_DIR` / `OPENSSL_INCLUDE_DIR`'s `${openssl.dev}` references on the Android target. Should I sit on the current `opensslEnv` arrangement until your branch lands, or take a stab at conditionally clearing `OPENSSL_*` in `android.nix`'s specialArgs?

insipx:
closing in favor of #3584

PR #3584, opened 2026-04-30
fix(nix): use system openssl for non-android builds to avoid macOS-26…

### LINKED ISSUE #3575: Flaky CI Failure: openssl-sys build script fails on macOS 26 runner for musl cross-compilation targets

insipx:
fixes #3575 

this began happening b/c one of my prs pulled libsqlite into workspace hack so it is used for builds which previously did not require it (mls-validation-service).

> [!NOTE]
> ### Fix macOS-26 OpenSSL build failures by using system OpenSSL for non-Android builds
> - Sets `OPENSSL_NO_VENDOR`, `OPENSSL_DIR`, `OPENSSL_LIB_DIR`, and `OPENSSL_INCLUDE_DIR` in [nix/lib/base.nix](https://github.com/xmtp/libxmtp/pull/3584/files#diff-78e27920e7e7f148052510551d065fae07e34ca9a88df0abd2450dedb25668bc) for non-Android platforms, directing `openssl-sys` to link against nixpkgs' OpenSSL instead of building a vendored copy.
> - iOS builds strip the `OPENSSL_*` vars from `commonArgs` to avoid accidentally linking against macOS libssl when targeting iOS/iOS-simulator.
> - Refactors the Nix library structure in [nix/lib/default.nix](https://github.com/xmtp/libxmtp/pull/3584/files#diff-6fd175d36064b2e9b9371596932bf201514494fc02c317f3f4526243d72991f1) to expose base utilities as a top-level `xmtp-base` attribute instead of `xmtp.base`, and updates all package derivations and shells accordingly.
> - Behavioral Change: consumers referencing `xmtp.base` in Nix expressions must switch to `xmtp-base`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ea5f03a.</sup>
> 
>

### DIFF
--- a/nix/lib/base.nix
+++ b/nix/lib/base.nix
@@ -35,6 +35,25 @@ let
     builtins.replaceStrings [ "-" ] [ "_" ]
       pkgsBuildHost.stdenv.hostPlatform.rust.rustcTarget;
 
+  # macOS 26 vendored openssl-src build aborts in util/mkinstallvars.pl
+  # (empty LIBDIR) when cross-compiling to *-unknown-linux-musl. Point
+  # openssl-sys at nixpkgs openssl to skip vendored build.
+  # See https://github.com/xmtp/libxmtp/issues/3575.
+  #
+  # Skip on android: cross openssl fails on internal/ktls.h (NDK sysroot
+  # missing msghdr/cmsghdr fields). android.nix clears buildInputs;
+  # OPENSSL_DIR here would re-pull cross-android openssl or link
+  # android-arm64 against native x86_64 openssl libs.
+  # ios.nix strips these keys itself since it runs through native pkgs
+  # (CARGO_BUILD_TARGET=*-apple-ios) and the inherited macOS libssl.dylib
+  # would fail to link against an ios-simulator slice.
+  opensslEnv = lib.optionalAttrs (!stdenv.hostPlatform.isAndroid) {
+    OPENSSL_NO_VENDOR = "1";
+    OPENSSL_DIR = "${openssl.dev}";
+    OPENSSL_LIB_DIR = "${openssl.out}/lib";
+    OPENSSL_INCLUDE_DIR = "${openssl.dev}/include";
+  };
+
   # Common build arguments shared between iOS and Android derivations.
   # Platform-specific args (like ANDROID_HOME or __noChroot) are added by each derivation.
   commonArgs = {
@@ -70,7 +89,8 @@ let
     "AWS_LC_SYS_TARGET_CC_${buildPlatformSuffix}" = "cc";
     "AWS_LC_SYS_TARGET_CXX_${buildPlatformSuffix}" = "c++";
 
-  };
+  }
+  // opensslEnv;
 
   # Make cargo artifacts for a derivation building rust code
   # "rust" is the rust toolchain to use (native or host)
--- a/nix/lib/default.nix
+++ b/nix/lib/default.nix
@@ -4,167 +4,180 @@
   lib,
   ...
 }:
+let
+  mkToolchain = import ./mkToolchain.nix { inherit inputs lib; };
+  craneConfig = final: prev: {
+    # add napi builder to crane scope
+    napiBuild = final.callPackage ./napiBuild.nix { };
+    uniffiGenerate = final.callPackage ./uniffiGenerate.nix { };
+  };
+
+  # `xmtp` namespace via `lib.makeScope` — same pattern as
+  # python311Packages, darwin.apple_sdk, haskell.packages. `p.newScope`
+  # (splice.nix) makes `xself.callPackage` auto-resolve cross-spliced inputs.
+  mkXmtpLib =
+    p:
+    p.lib.makeScope p.newScope (xself: {
+      inherit mkToolchain;
+      mkNativeToolchain = mkToolchain p;
+      filesets = xself.callPackage ./filesets.nix { };
+      craneLib = (inputs.crane.mkLib p).overrideScope craneConfig;
+      androidEnv = xself.callPackage ./android-env.nix { };
+      iosEnv = xself.callPackage ./ios-env.nix { };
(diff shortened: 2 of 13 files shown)


alceops:
I did a public-code-only pass on #3584 at `ea5f03a` as a validation checklist for #3575.

What looks good:
- the main fix direction is sensible: `OPENSSL_NO_VENDOR=1` + explicit nixpkgs OpenSSL paths avoids the macOS-26 vendored `openssl-src` install path that is failing in the recurrence logs;
- current head appears to have removed the stale `xmtp.base` Nix references called out earlier;
- iOS package args now strip the inherited `OPENSSL_*` values before target linking.

One boundary I would explicitly verify before merge: `nix/shells/android.nix` still sets `OPENSSL_DIR = shellCommon.rustBase.env.OPENSSL_DIR` and pulls `xmtp-base.commonArgs.buildInputs`. That may be fine for the dev shell/host tools, but it is adjacent to the patch’s Android comments, so I’d include `nix develop .#android` (plus the failing musl outputs) in the acceptance run to prove the OpenSSL vars don’t leak into Android target-link steps.

I can turn this into a narrow validation pack if useful: pre-fix vs PR-head logs for the two failed musl outputs, Android/iOS eval checks, and a merge-readiness note. No secrets, private infra, release credentials, or paid runners needed unless you explicitly want external CI.

nix/lib/base.nix
```diff
@@ -35,6 +35,25 @@ let
     builtins.replaceStrings [ "-" ] [ "_" ]
       pkgsBuildHost.stdenv.hostPlatform.rust.rustcTarget;
 
+  # macOS 26 vendored openssl-src build aborts in util/mkinstallvars.pl
+  # (empty LIBDIR) when cross-compiling to *-unknown-linux-musl. Point
+  # openssl-sys at nixpkgs openssl to skip vendored build.
+  # See https://github.com/xmtp/libxmtp/issues/3575.
+  #
+  # Skip on android: cross openssl fails on internal/ktls.h (NDK sysroot
```
tylerhawkes, line 43: Can we ignore libsqlite for workspace hack instead of doing all this? Having all this to build openssl (which will hopefully be going away soon is) is overkill compared to just leaving the crate out of the hack.

insipx: yeah we could do that as well, will change

PR #3611, opened 2026-05-07
xmtp_debug: add --fail-on-error flag for fail-fast bisect sessions

### LINKED ISSUE #3610: Add `fail-on-error` flag to xdbg
Xdbg currently does not fail on error during client operation (when syncing, sending messages, etc). however, this may be desirable during debug/bisect sessions. brainstorm how to add that to libxmtp then update this issue with the plan before starting the work.

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3610

## Summary

Adds a top-level `--fail-on-error` flag to xdbg that turns the existing log-and-continue error sites into fail-fast errors. Intended for `git bisect run` sessions, where a single failed send/sync should mark the commit bad rather than be lost in logs. Default behavior is unchanged.

The plan was [posted to the issue](https://github.com/xmtp/libxmtp/issues/3610#issuecomment-4399751878) per the request to brainstorm first.

## Sites updated

| File | Behavior change when `--fail-on-error` is set |
| --- | --- |
| `generate/messages.rs` | `send_many_messages` returns `Err` on the first per-task failure (was: log + return successes only). |
| `stream.rs` | Conversation/Message stream loops propagate the first `Err` item (was: skip with `while let Some(Ok(...))`). |
| `generate/groups.rs` | `check_member_visibility` now returns `Result<()>`; visibility-poll `sync_welcomes` failures surface (was: warn-and-continue forever). |
| `test.rs` | `MessageVisibility` scenario propagates stream errors (was: warn-and-continue). |

## Plumbing

A process-global `OnceLock<bool>` is initialized in `main.rs` immediately after `AppOpts::parse()`, read via a `crate::fail_on_error()` accessor at each call site. This mirrors the existing `static FDLIMIT: OnceLock<usize>` pattern in `app.rs` and avoids threading a bool through ~10 subcommand constructors.

## Out of scope

- `metrics.rs` push warnings (observability, not client operations).
- Library-side changes — libxmtp already returns `Result`; the policy decision (log vs. fail) lives in the CLI.

## Test plan

- [x] `cargo check -p xdbg` clean
- [x] `cargo test -p xdbg --bin xdbg args::` — 17/17 passing, including 3 new tests:
  - `fail_on_error_defaults_false`
  - `fail_on_error_parses_when_present`
  - `fail_on_error_has_no_short_alias` (asserts `xdbg -f` does not parse, leaving `-f` free for future subcommand options)
- [x] `just lint-rust` — clippy + fmt + hakari all clean
- [ ] Manual smoke against a backend (not gated; requires running `dev/up`): `xdbg generate -e message -a 1 --backend nonexistent` with and without `--fail-on-error` to confirm exit-code difference.

> [!NOTE]
> ### Add `--fail-fast` flag to `xmtp_debug` for terminating on first error
> - Adds a `--fail-fast` CLI flag to [`args.rs`](https://github.com/xmtp/libxmtp/pull/3611/files#diff-625330612e40d918eef6d9785cc1b9adbc3c444cc2666c0d632bb216d3b2b4f5) and stores it in a global `OnceLock<bool>` in [`main.rs`](https://github.com/xmtp/libxmtp/pull/3611/files#diff-1402fc2e065b4dfc49f0c88fdd0afcbab0b1ea61526ee56858b46f0f95be79e1) accessible via `fail_fast()`.
> - When `--fail-fast` is set, errors in group creation, message generation, stream processing, and message visibility tests now return immediately instead of logging a warning and continuing.
> - `check_member_visibility` in [`groups.rs`](https://github.com/xmtp/libxmtp/pull/3611/files#diff-2a926685e9318ce1dc82bb85f365db546e52a68aca168cf9ac4fe69aa9b702d1) is updated to propagate errors via `?` rather than silently swallowing them.
> - Behavioral Change: without `--fail-fast`, behavior is unchanged; with the flag, the first per-operation error terminates the process with a non-zero exit code.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d709e9d.</sup>
> 
>

### DIFF
--- a/apps/xmtp_debug/src/app/generate/groups.rs
+++ b/apps/xmtp_debug/src/app/generate/groups.rs
@@ -90,7 +90,7 @@ impl GenerateGroups {
                         .await;
 
                         if let Some(ref invitee) = first_invitee {
-                            check_member_visibility(&group_id, invitee, &network_clone).await;
+                            check_member_visibility(&group_id, invitee, &network_clone).await?;
                         }
 
                         // -- total group create + add latency --
@@ -218,10 +218,19 @@ async fn create_group_on_network(
 }
 
 /// Check whether an invitee can see the group via sync_welcomes (read-your-own-writes).
-async fn check_member_visibility(group_id: &[u8], invitee: &Identity, network: &args::BackendOpts) {
+async fn check_member_visibility(
+    group_id: &[u8],
+    invitee: &Identity,
+    network: &args::BackendOpts,
+) -> Result<()> {
     let reader_client = match app::client_from_identity(invitee, network) {
         Ok(c) => c,
-        Err(_) => return,
+        Err(e) => {
+            if crate::fail_on_error() {
+                return Err(e);
+            }
+            return Ok(());
+        }
     };
 
     let t_visibility = Instant::now();
@@ -231,6 +240,9 @@ async fn check_member_visibility(group_id: &[u8], invitee: &Identity, network: &
 
     loop {
         if let Err(e) = reader_client.sync_welcomes().await {
+            if crate::fail_on_error() {
+                return Err(eyre!("sync_welcomes failed during visibility poll: {e}"));
+            }
             tracing::warn!(error = %e, "sync_welcomes failed during visibility poll");
         }
         if reader_client.group(group_id).is_ok() {
@@ -260,4 +272,5 @@ async fn check_member_visibility(group_id: &[u8], invitee: &Identity, network: &
         &[("phase", "member_visibility"), ("success", vis_ok)],
     );
     push_metrics("xdbg_debug").await;
+    Ok(())
 }
--- a/apps/xmtp_debug/src/app/generate/groups.rs
+++ b/apps/xmtp_debug/src/app/generate/groups.rs
@@ -89,8 +89,14 @@ impl GenerateGroups {
                         )
                         .await;
 
-                        if let Some(ref invitee) = first_invitee {
-                            check_member_visibility(&group_id, invitee, &network_clone).await?;
+                        if let Some(ref invitee) = first_invitee
+                            && let Err(e) =
+                                check_member_visibility(&group_id, invitee, &network_clone).await
+                        {
+                            if crate::fail_on_error() {
+                                return Err(e);
+                            }
+                            tracing::warn!(error = %e, "member visibility check failed");
                         }
 
                         // -- total group create + add latency --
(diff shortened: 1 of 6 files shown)


apps/xmtp_debug/src/app/generate/groups.rs
```diff
@@ -218,10 +218,19 @@ async fn create_group_on_network(
...
 /// Check whether an invitee can see the group via sync_welcomes (read-your-own-writes).
-async fn check_member_visibility(group_id: &[u8], invitee: &Identity, network: &args::BackendOpts) {
+async fn check_member_visibility(
+    group_id: &[u8],
+    invitee: &Identity,
+    network: &args::BackendOpts,
+) -> Result<()> {
     let reader_client = match app::client_from_identity(invitee, network) {
         Ok(c) => c,
-        Err(_) => return,
```
insipx, line 224: no need to check for fail on error, this should be a normal `?` propagation

apps/xmtp_debug/src/app/generate/groups.rs
```diff
@@ -231,6 +240,9 @@ async fn check_member_visibility(group_id: &[u8], invitee: &Identity, network: &
 
     loop {
         if let Err(e) = reader_client.sync_welcomes().await {
+            if crate::fail_on_error() {
+                return Err(eyre!("sync_welcomes failed during visibility poll: {e}"));
```
insipx: this hsould be a normal ? propagation. to silence errors from stopping xdbg execution should happen at a higher level

apps/xmtp_debug/src/app/generate/groups.rs
xmtp-coder-agent: Good catch — fixed in d047fe365. `client_from_identity` failure is a setup error, so it propagates with `?` regardless of `--fail-on-error`. The flag still gates the per-poll `sync_welcomes` failure inside the loop, which is the per-operation case the flag is meant for.

apps/xmtp_debug/src/app/generate/groups.rs
xmtp-coder-agent: Done in fd87bc622. `sync_welcomes` now propagates with `?`, and the log-vs-fail decision lives at the caller in `create_groups`. Note this is a slight behavior change: a transient `sync_welcomes` error now fails the whole visibility check rather than retrying inside the poll deadline. Let me know if you'd rather keep the retry-within-deadline behavior — that'd push the warn back inside the loop.

insipx (approved):

insipx:
@xmtp-coder-agent actually i think `fail-fast` may be a more appropriate flag name here

xmtp-coder-agent:
Renamed to `--fail-fast` in d709e9d99 (flag, static, accessor `crate::fail_fast()`, tests, and the help text).

PR #3613, opened 2026-05-07
xmtp_mls: enforce registry membership policy on migrated commits

tylerhawkes:
## Summary

Two coupled fixes that together unblock the last 3 of 8 previously-ignored bootstrap-flow proposal tests in `test_proposals.rs`. Stacks on #3612.

**1. Use the verified message sender as the committer.** Previously `extract_committer_and_proposers` derived the committer from `StagedCommit::update_path_leaf_node()`, with a "single proposer becomes the committer" fallback. Both branches were workarounds for not having the actual sender — but OpenMLS already verifies the framing signature against the leaf at `ProcessedMessage::sender()` during `process_message`, so that's the authoritative committer (same way `proposal.sender()` is used for proposals). `extract_committer_and_proposers` now takes an explicit `committer_leaf_index: LeafNodeIndex`. Receive path passes the index from `processed_message.sender()`; own-intent path passes `mls_group.own_leaf_index()`. The path-update + single-proposer fallbacks are gone.

This fixes `ActorCouldNotBeFound` on migrated commits without a path. Migrated `CommitPendingProposals` commits with only `{Add, AppDataUpdate}` proposals don't require a path under RFC 9420 §17.4 (unlike the unmigrated path where the always-emitted GCE proposal forced one), so they were hitting that error.

**2. Replace the `allow_if_super_admin` migrated-policy stub with a real registry-derived `PolicySet`.** Two stub sites had `MembershipPolicies::allow_if_actor_super_admin()` hardcoded for both add and remove member policies. New helper `app_data::policy::membership_policy_set_from_registry` (new file `crates/xmtp_mls/src/groups/app_data/policy.rs`) translates the registry's `GROUP_MEMBERSHIP` insert/delete `MetadataBasePolicy` (Allow / Deny / AllowIfAdmin / AllowIfSuperAdmin) into `MembershipPolicies` for the add/remove slots, with `tracing::warn!` breadcrumbs on every deny-all fallback path. Other slots stay on the conservative super-admin stub because `validate_app_data_update_proposals_in_commit` is the authoritative gate for non-membership components.

The two fixes are interlocked: relaxing the policy stub alone makes admin-only group tests fall over (peer proposals accepted on receive); fixing committer identification alone leaves the migrated path rejecting non-super-admin proposers in default-permission groups.

## Test plan
- [x] `just test crate xmtp_mls` — 539/539 pass (1 flaky, 5 platform-skipped)
- [x] `cargo clippy -p xmtp_mls --all-targets --no-deps -- -Dwarnings` clean
- [ ] CI green

### DIFF
--- a/crates/xmtp_mls/src/groups/app_data/mod.rs
+++ b/crates/xmtp_mls/src/groups/app_data/mod.rs
@@ -17,6 +17,7 @@
 pub(crate) mod bootstrap_validator;
 pub mod component_source;
 pub mod migration;
+pub(crate) mod policy;
 pub(crate) mod sender_intents;
 pub(crate) mod typed_facade;
 
--- a/crates/xmtp_mls/src/groups/app_data/policy.rs
+++ b/crates/xmtp_mls/src/groups/app_data/policy.rs
@@ -1,0 +1,136 @@
+//! Translate the AppData `COMPONENT_REGISTRY`'s membership policies into
+//! the legacy [`GroupMutablePermissions`] shape so the receive-side
+//! validator and standalone-proposal validator can evaluate Add /
+//! Remove proposals on migrated groups against the same policy that
+//! `validate_app_data_update_proposals_in_commit` enforces for the
+//! companion `AppDataUpdate(GROUP_MEMBERSHIP)` proposal.
+//!
+//! The mapping is:
+//! - `GROUP_MEMBERSHIP.insert_policy` → `add_member_policy`
+//! - `GROUP_MEMBERSHIP.delete_policy` → `remove_member_policy`
+//!
+//! Other slots on the synthesized `PolicySet` (metadata field updates,
+//! admin add/remove, permissions update) keep the conservative
+//! "super admin only" stub — the AppDataUpdate path enforces those
+//! per-component when a real change comes through, and the legacy
+//! callers only consult them as part of structural sanity checks.
+
+use openmls::group::MlsGroup as OpenMlsGroup;
+use xmtp_mls_common::app_data::component_id::ComponentId;
+use xmtp_proto::xmtp::mls::message_contents::{
+    MetadataPolicy as MetadataPolicyProto, metadata_policy::Kind as MetadataPolicyKindProto,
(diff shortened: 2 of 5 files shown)


tylerhawkes:
Addressed the warranted observability bits from the review. The bot's three concerns were really about the silent deny-all fallbacks; here's what changed in 9a2d74993:

**1 + minor (75): every deny-all branch now logs.** The four distinct `(deny, deny)` outcomes were indistinguishable in production:
- missing `GROUP_MEMBERSHIP` entry → `tracing::warn!`
- entry with no `permissions` block → `tracing::warn!`
- missing `insert_policy` field → `tracing::warn!`
- missing `delete_policy` field → `tracing::warn!`
- decode error (already logged before) → unchanged

Each is a distinct "shouldn't happen post-bootstrap" signal, so they now leave per-cause breadcrumbs.

**Minor (104-107): inline comment near the stub slots.** Repeats the file-level rationale: non-membership slots stay on super-admin-only because `validate_app_data_update_proposals_in_commit` is the authoritative gate.

**Concern 3 (roach motel groups):** Not addressed at this layer — `MetadataPolicy` itself is the source of expressivity here, and asymmetric add/remove is a legitimate shape (e.g., open-join + admin-only-eject). Validating registry coherence is a separate concern that belongs in the registry construction path, not in the translator.

**Verified:** `just test v3 test_proposals` 52/52 pass, `cargo clippy -p xmtp_mls --all-targets --no-deps -- -Dwarnings` clean.

.claude/scheduled_tasks.lock
```diff
@@ -0,0 +1 @@
+{"sessionId":"23a740a2-e43c-4934-8b4b-1696d4c9d350","pid":90060,"procStart":"Fri May  1 18:56:01 2026","acquiredAt":1778187545036}
```
insipx: should this be ignored

insipx (approved):

.claude/scheduled_tasks.lock
tylerhawkes: Yeah, I'll add it

PR #3626, opened 2026-05-12
ci(release-wasm): build with nix to benefit from cachix

### LINKED ISSUE #3625: Make wasm release workflow use nix
the wasm release workflow is not using nix. we can change that to use nix to be like the other workflows, benefiting from cachix. we build wasm release bindings on every commit so it should speed it up a bit

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3625

## Summary
- Replace the bespoke setup-rust + emsdk-clone + yarn-build path in `.github/workflows/release-wasm.yml` with `nix build .#wasm-bindings`, matching how `release-node.yml` builds its bindings.
- Drop the `setup-rust`, `setup-node`, `taiki-e/install-action wasm-bindgen`, manual emsdk clone, and pre-build `xmtp-release set-manifest-version` steps. `npm-publish.yml` already sets the version before publishing, so the duplicate version-set step is unnecessary.
- Run the build on `warp-ubuntu-latest-x64-16x` so the WarpBuild Nix store cache and Cachix can serve cached store paths. Since wasm release bindings are built on every commit, repeat builds with unchanged inputs should be substantially faster.
- Upload `result/dist` as the `wasm_build` artifact and change the publish job's `artifact-path` to `bindings/wasm/dist` so the downloaded files land in the same location the publish job expects.

## Why this works
- `flake.nix` already exposes `wasm-bindings = (pkgs.callPackage ./nix/package/wasm.nix { }).bin;` which runs `wasm-pack build --target web --out-dir <dist> --no-pack --release ./bindings/wasm` inside Nix.
- `nix/package/wasm.nix` already provides `wasm-pack`, `wasm-bindgen-cli`, `emscripten`, `binaryen`, and the llvm/clang toolchain pinned via nixpkgs. RUSTFLAGS for `wasm32-unknown-unknown` come from `.cargo/config.toml`, so the produced dist matches the previous `yarn build` output.
- After #3620 dropped the npm `wasm-pack` devDependency, the workflow no longer had any step that installed `wasm-pack`, making the legacy `yarn build` path fragile. Nix becomes the single source of truth for the wasm toolchain.

## Test plan
- [ ] `just lint-config` passes locally.
- [ ] `nix eval .#wasm-bindings.outPath` resolves to a valid store path (confirmed in design discussion on the issue).
- [ ] Trigger `release-wasm.yml` end-to-end on a dev/rc release after merge to verify the artifact publishes and the dist contents match a baseline from the previous workflow.

> [!NOTE]
> ### Build WASM release artifacts using Nix and Cachix instead of manual toolchain setup
> - Replaces the manual Rust/Node/emscripten/yarn build steps in [release-wasm.yml](.github/workflows/release-wasm.yml) with a single `nix-fast-build --flake .#packages.x86_64-linux.wasm-bindings` step backed by Cachix for caching.
> - Switches the build runner from `ubuntu-latest` to `warp-ubuntu-latest-x64-16x`.
> - Artifact upload now sources only from `result-/dist`; the publish job adjusts its download path to `bindings/wasm/dist` accordingly.
> - Behavioral Change: `package.json` and `yarn.lock` are no longer included in the uploaded artifact.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2612d7e.</sup>
> 
>

### DIFF
--- a/.github/workflows/release-wasm.yml
+++ b/.github/workflows/release-wasm.yml
@@ -51,67 +51,26 @@ jobs:
 
   build:
     needs: setup
-    runs-on: ubuntu-latest
+    runs-on: warp-ubuntu-latest-x64-16x
     steps:
       - name: Checkout
         uses: actions/checkout@v6
         with:
           ref: ${{ inputs.ref }}
 
-      - uses: ./.github/actions/setup-rust
+      - uses: ./.github/actions/setup-nix
         with:
-          workspaces: |
-            .
-            bindings/wasm
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          cachix-auth-token: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
-      - uses: ./.github/actions/setup-release-tools
-
-      - name: Set version in package.json
-        env:
-          VERSION: ${{ needs.setup.outputs.version }}
-        run: xmtp-release set-manifest-version --sdk wasm-bindings --version "$VERSION"
-
-      - name: Install wasm-bindgen
-        uses: taiki-e/install-action@v2
-        with:
-          tool: wasm-bindgen@0.2.108
-
-      - name: Install wasm-pack
-        uses: taiki-e/install-action@v2
-        with:
-          tool: wasm-pack
-
-      - name: Setup node
-        uses: ./.github/actions/setup-node
-        with:
-          cache-dependency-path: "bindings/wasm/yarn.lock"
-
-      - name: Install dependencies
-        working-directory: bindings/wasm
-        run: |
-          yarn
-
-      - name: Install emscripten toolchains
-        run: |
-          git clone https://github.com/emscripten-core/emsdk.git
-          cd emsdk
-          ./emsdk install latest
-          ./emsdk activate latest
-
-      - name: Build
-        working-directory: bindings/wasm
-        run: |
-          source ../../emsdk/emsdk_env.sh
-          yarn build
+      - name: Build wasm-bindings
+        run: nix-fast-build --flake .#packages.x86_64-linux.wasm-bindings
 
       - name: Upload build output
         uses: actions/upload-artifact@v7
         with:
           name: wasm_build
-          path: |
-            bindings/wasm/dist
-            bindings/wasm/package.json
-            bindings/wasm/yarn.lock
+          path: result-/dist
           retention-days: 1
 
   publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-wasm.yml
```diff
@@ -51,67 +51,26 @@ jobs:
...
-          ./emsdk install latest
-          ./emsdk activate latest
-
-      - name: Build
-        working-directory: bindings/wasm
-        run: |
-          source ../../emsdk/emsdk_env.sh
-          yarn build
+      - name: Build wasm-bindings
+        run: nix build .#wasm-bindings -L
```
insipx: use `nix-fast-build`

xmtp-coder-agent: Done — switched to `nix-fast-build --no-nom --flake .#packages.x86_64-linux.wasm-bindings` in 1f48c5649. The flake URI uses the explicit per-system path because nix-eval-jobs (used internally by nix-fast-build) doesn't apply the `packages.<currentSystem>` shorthand. nix-fast-build appends `-<attr>` to its out-link, so the upload path is now `result-/dist` (same pattern used by `test-workspace.yml` with `v3-/coverage`).

insipx:
@xmtp-coder-agent rebase on main and resolve the conflicts. the conflicting commit is a change to the old workflow so it can be discarded

.github/workflows/release-wasm.yml
insipx: the github actions environment should be automatically detected by nix fast build so --no-nom should be unnecessary. look at other invocatinos of nix fast build in the repo to see how its working, specifically maybe the webassembly tests

xmtp-coder-agent:
Rebased onto main. The conflicting commit #3623 (which added an `Install wasm-pack` step to the pre-nix workflow) was discarded as you noted — the nix toolchain provides wasm-pack, so that fix is no longer needed. The follow-up nix-fast-build commit was auto-dropped during rebase since its patch was already incorporated into the resolved diff. Final state: one commit, same as before, now clean against main.

.github/workflows/release-wasm.yml
xmtp-coder-agent: You're right — checked nix-fast-build's source: when `--no-nom` is omitted, it falls back to `shutil.which("nom") is None`, and `nom` isn't installed by `setup-nix`, so the default already evaluates to no-nom in CI. Dropped the flag. (`test-xdbg.yml` / `test-workspace.yml` still pass it explicitly — happy to clean those up in a follow-up if you want.)

insipx (approved):

PR #3627, opened 2026-05-12
xmtp_mls: capability-aware GroupMutableMetadata reads + single-field getter refactor

tylerhawkes:
### Add capability-aware `GroupMutableMetadata` reads for migrated MLS groups
- Introduces `extract_group_mutable_metadata_capability_aware` in [component_source.rs](https://github.com/xmtp/libxmtp/pull/3627/files#diff-47acce7b46d2cc4f9669dd42120ee90e488ceb0b15817b67b7757e631b05d113) to read `GroupMutableMetadata` from the AppData dictionary on migrated groups and from the legacy MLS extension on unmigrated groups.
- Refactors single-field getters (`group_name`, `group_description`, `group_image_url_square`, `app_data`, `admin_list`, `super_admin_list`) in [mod.rs](https://github.com/xmtp/libxmtp/pull/3627/files#diff-c27dd00cf700fd888b75fdbd19738462c2549fdc881dad70f8e61c979d440966) to use `read_single_component` or `read_admin_set_preserving_legacy_order`, routing reads through the AppData dict on migrated groups.
- Fixes `get_message_expire_at_ns` and welcome processing in [mls_sync.rs](https://github.com/xmtp/libxmtp/pull/3627/files#diff-160aadcd7de477c1116260a79601a0548842f3bd91bf16bd8443debb66ce6684) and [xmtp_welcome.rs](https://github.com/xmtp/libxmtp/pull/3627/files#diff-7ad79af03cc69f17888ed323b42554f52c5c706674e76a981c4fd4663464bb4e) so disappearing-message settings and `paused_for_version` are correctly sourced post-migration.
- Renames `extract_group_mutable_metadata` to `extract_legacy_group_mutable_metadata` in [group_mutable_metadata.rs](https://github.com/xmtp/libxmtp/pull/3627/files#diff-ed696106212fb67c35890c8338ccfd437eff9d62595cc6d1b73139430893a8ed) to make its unmigrated-only contract explicit.
- Behavioral Change: on migrated groups, all metadata reads now source from the AppData dictionary; legacy MLS extension reads will return `MissingExtension` and are no longer used post-bootstrap.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized 30c20db.</sup>

PR #3628, opened 2026-05-12
xmtp_mls: relax unknown-component handling at the three receive-side sites

tylerhawkes:
### Relax unknown-component rejection at receive-side validation, accumulation, and apply sites
- Unknown MLS app data components are no longer categorically rejected; if the on-dict `ComponentRegistry` provides a `ComponentType`, the update is applied and validated using type-aware dispatch instead of being immediately rejected.
- Adds a new `type_dispatch` module in `xmtp_mls_common` with `apply_update_payload_for_type` and `expand_to_changes_for_type`, covering `Bytes`, `String` (UTF-8 validated), `TlsSetInboxId`, TLS maps, and byte-sets.
- `validate_one_app_data_update_with_old_value`, `accumulate_app_data_updates`, and `apply_app_data_update_payload` all gain a `&ComponentRegistry` parameter; the registry is snapshotted at pre-commit and passed through the call chain.
- Per-component invariant checks still only run when a known per-id `Component` impl exists; unknown components pass with policy enforcement only.
- Risk: callers that previously received an `UnknownComponent` error for unregistered components will now succeed if the registry entry is present and policy allows it, changing runtime behavior for forward-compatible components.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized c6f6e02.</sup>

### DIFF
--- a/crates/xmtp_mls/src/groups/app_data/component_source.rs
+++ b/crates/xmtp_mls/src/groups/app_data/component_source.rs
@@ -39,7 +39,11 @@ use openmls::{
 use tls_codec::{Deserialize, Serialize};
 use xmtp_mls_common::{
     app_data::{
-        component_id::ComponentId, registry_table::lookup_component, typed::ComponentTypedError,
+        component_id::ComponentId,
+        component_registry::ComponentRegistry,
+        components::type_dispatch::{apply_update_payload_for_type, expand_to_changes_for_type},
+        registry_table::lookup_component,
+        typed::ComponentTypedError,
     },
     group_mutable_metadata::{
         GroupMutableMetadata, GroupMutableMetadataError, MetadataField,
@@ -167,6 +171,10 @@ impl From<ComponentTypedError> for ComponentSourceError {
             ComponentTypedError::TlsCodec(e) => Self::TlsCodec(e),
             ComponentTypedError::TlsSetApply(e) => Self::TlsSetApply(e),
             ComponentTypedError::TlsMapApply(e) => Self::TlsMapApply(e),
+            ComponentTypedError::UnspecifiedType(id) => Self::MalformedComponentValue {
+                component_id: id,
+                reason: "registered ComponentType is Unspecified".to_string(),
+            },
         }
     }
 }
@@ -446,9 +454,9 @@ pub(crate) fn encode_app_data_update_payload(
 }
 
 // `ExpandedComponentChange` lives in `xmtp_mls_common::app_data::typed`
-// so the `Component` trait there can return it. Re-exported here for
-// backward compatibility while the dispatch refactor (subsequent jj
-// changes) shifts call sites onto the trait directly.
+// so the `Component` trait there can return it. Re-exported here so
+// in-crate callers can construct the change list without pulling the
+// xmtp_mls_common path in directly.
 pub(crate) use xmtp_mls_common::app_data::typed::ExpandedComponentChange;
 
 /// Expand an `AppDataUpdate` proposal payload into the per-element changes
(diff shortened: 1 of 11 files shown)


tylerhawkes:
This PR is part of a stack of 2 bookmarks:

1. `trunk()`
1. **tyler/appdata-unknown-component-tolerance ← this PR**

---

tylerhawkes:
Addressed review feedback in latest push:

**On the immutability concern.** Looked closely at the scenario where a newer release defines a new immutable component and a malicious peer tries to write it on an old client.

The immutability gate (`apply_app_data_update_payload`) is **range-by-id**, not per-`Component`-impl:

```rust
if id.is_immutable() && old_value.is_some() {
    return Err(ComponentSourceError::ImmutableUpdate(id));
}
```

`is_immutable()` is hardcoded to `0xBE00-0xBFFF` (XMTP) / `0xFD00-0xFEFF` (app), so it fires the same way for unknown ids in those ranges as for known ones. First write is allowed (bootstrap-style insert), subsequent overwrites are rejected — exactly the contract the existing well-known immutable components observe.

The "first write by a malicious peer" angle is gated upstream by registry policy: `validate_component_write` runs `insert_policy` against the actor's role, so an unauthorized peer can't insert in the first place. A super-admin doing a first insert on an unknown immutable id is the bootstrap path by design.

What old clients DO skip — and what the WELL_KNOWN docblock now explicitly calls out — is `Component::validate_invariant`. Components shipped via this path are restricted to type-shape semantics; they MUST NOT carry receive-side invariants beyond registry policy.

Added three tests pinning the unknown-immutable-id contract:
- `test_apply_unknown_id_in_xmtp_immutable_range_first_write_allowed`
- `test_apply_unknown_id_in_xmtp_immutable_range_overwrite_rejected`
- `test_apply_unknown_id_in_app_immutable_range_overwrite_rejected`

These pin the contract for both XMTP (`0xBE00-0xBFFF`) and application (`0xFD00-0xFEFF`) immutable ranges, demonstrating that range-based immutability survives the type-aware dispatch refactor.

tylerhawkes:
Round 2 feedback addressed:

**1. Error context in `type_dispatch` TLS decode operations (medium priority).** Added `attach_component_id` at the dispatcher entry points (`apply_update_payload_for_type` / `expand_to_changes_for_type`). Bare `tls_codec::Error` leaking out of the per-type helpers now gets re-wrapped as:

```
ComponentTypedError::MalformedValue {
    component_id,
    reason: "type-aware apply (TlsSetBytes) tls codec error: <inner>",
}
```

Carries the component id, the `ComponentType` it dispatched as, AND apply-vs-expand site distinction in a single error. The wrapping happens only at the unknown-id dispatcher boundary, so existing known-id tests asserting `ComponentSourceError::TlsCodec(_)` are unaffected.

**2. Test for malformed prior on an unknown component (low priority).** Added `unknown_component_update_with_malformed_prior_rejected` in `test_validate_app_data_update.rs`: feeds a well-formed `TlsSetDelta<VLBytes>` with a `RemoveByHash` mutation, but the prior dict bytes are `\xDE\xAD\xBE\xEF` — the expander's prior-set decode trips, the validator collapses to `InsufficientPermissions`, and the warn-level log line at `validated_commit.rs` carries the component id for triage.

**3. Telemetry to distinguish payload vs. prior failures (high priority).** Validator already logs `component_id = %component_id`, `error = %err` at warn-level on every unknown-component expansion failure. With the error-wrapping change in #1 above, the error message now includes the `ComponentType` and which site (apply/expand) failed, which is enough to distinguish payload-decode from prior-decode for the existing decoders (since prior decode only happens in expand-on-Update paths that build a hash index, the apply path always fails on payload). No additional telemetry plumbing needed.

**Skipped:** rollout strategy documentation — these are libxmtp-internal protocol changes that ship as a single migration, not feature-flagged.

**Skipped:** duplicate hash-index logic between `expand_bytes_set_changes` and `expand_inbox_id_set_changes` (the inbox_id version uses raw `into_bytes` for InboxId vs. `as_slice` for VLBytes, which is the per-type point of departure; extracting a shared helper would need a trait abstraction that doesn't pay for the duplication today).

insipx (approved):

PR #3629, opened 2026-05-12
xmtp_mls: split stage_inline_app_data_commit into standalone propose + commit

tylerhawkes:
### Split `stage_inline_app_data_commit` into separate propose and commit steps in MLS group sync
- Replaces `stage_inline_app_data_commit` with `stage_app_data_propose_and_commit`, which first enqueues a standalone `AppDataUpdate` proposal via `propose_app_data_update`, then builds a commit that references it by `ProposalRef` using `consume_proposal_store(true)`.
- Callers in [sender_intents.rs](https://github.com/xmtp/libxmtp/pull/3629/files#diff-09753ccd507e9327f4b4e1df1c02244081bdeeecb5088bb6d21ddb89a349513a) and [mls_sync.rs](https://github.com/xmtp/libxmtp/pull/3629/files#diff-160aadcd7de477c1116260a79601a0548842f3bd91bf16bd8443debb66ce6684) now publish two payloads per intent: the proposal message first, followed by the commit.
- Adds a `Propose(ProposalError)` variant to `GroupAppDataError` and implements `RetryableError`, making propose/stage failures retryable while create-commit and payload-apply failures remain non-retryable.
- `GroupError::AppDataCommit` now delegates retryability to the inner error instead of always returning false.
- Behavioral Change: receivers must now handle a separate proposal message before the commit for `AppDataUpdate`, `UpdateAdminList`, and `UpdatePermission` intents.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1a1c164.</sup>

### DIFF
--- a/crates/xmtp_mls/src/groups/app_data/component_source.rs
+++ b/crates/xmtp_mls/src/groups/app_data/component_source.rs
@@ -361,7 +361,7 @@ pub(crate) fn read_component_bytes(
 /// into [`expand_app_data_update_to_changes`] as `old_value` — the
 /// validator uses that to resolve `RemoveByHash` mutations back to
 /// their underlying inbox id. The parent `app_data` module also uses
-/// it from `process_message_with_app_data`, `stage_inline_app_data_commit`,
+/// it from `process_message_with_app_data`, `stage_app_data_propose_and_commit`,
 /// and `pending_app_data_updates`.
 pub(crate) fn read_from_app_data_dict(
     id: ComponentId,
--- a/crates/xmtp_mls/src/groups/app_data/mod.rs
+++ b/crates/xmtp_mls/src/groups/app_data/mod.rs
@@ -25,8 +25,8 @@ use std::collections::BTreeMap;
 
 use openmls::{
     component::ComponentData,
-    framing::{ProcessedMessage, ProtocolMessage},
-    group::{AppDataUpdates, MlsGroup as OpenMlsGroup, ProcessMessageError},
+    framing::{MlsMessageOut, ProcessedMessage, ProtocolMessage},
+    group::{AppDataUpdates, MlsGroup as OpenMlsGroup, ProcessMessageError, ProposalError},
     messages::proposals::{AppDataUpdateOperation, Proposal},
     messages::proposals_in::{ProposalIn, ProposalOrRefIn},
     // `CommitMessageBundle` lives in `prelude` because the natural path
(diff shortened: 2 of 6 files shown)


tylerhawkes:

1. `trunk()`
1. [tyler/appdata-unknown-component-tolerance](https://github.com/xmtp/libxmtp/pull/3628)
1. **tyler/appdata-standalone-proposals ← this PR**
1. [tyler/appdata-well-known-change-control](https://github.com/xmtp/libxmtp/pull/3630)
1. [tyler/appdata-two-step-bootstrap](https://github.com/xmtp/libxmtp/pull/3631)

---

insipx (approved):

tylerhawkes:
Addressed Macroscope's high-severity finding on `app_data/mod.rs:272-284` (the AppDataUpdate-only precondition that silently folded non-AppDataUpdate proposals into metadata commits).

**Fix:** the precondition now rejects pre-existing **non-`AppDataUpdate`** proposals (`Add`/`Remove`/`Update`/PSK/ReInit/ExternalInit/GroupContextExtensions/SelfRemove/AppEphemeral/Custom). Multiple pre-existing `AppDataUpdate` proposals are **allowed** and get accumulated into the commit's dict updates by step 2 — that's the legitimate concurrent-update case:

> Alix invites Carol via `AppDataUpdate(GROUP_MEMBERSHIP)` (propose-only), then Bo invites Dave and commits everything. Bo's commit folds both proposals; the accumulator chains them via the in-batch map so the final dict bytes match what `process_message_with_app_data` produces on receive. Same logic for two queued `group_name` updates — last-write-wins.

Diagnostic now reports the kinds of pre-existing proposals (`Vec<&'static str>`) rather than just `AppDataUpdate` component ids. Comment + error message + docblock updated to reflect the new precondition. Rebased onto main (currently `#3638`).

tylerhawkes:
**Update: precondition removed entirely** (force-pushed `1a1c164`).

After further design discussion, the right answer for Macroscope's high-severity finding is to **embrace lazy batching, not police it**. The precondition was rejecting things the design should welcome:

- Pending proposals already freeze message-send (MLS invariant), so any pending proposal commits at the next opportunity regardless of which intent's commit body carries it.
- The producer of each pending proposal chose to leave it pending. If they needed it committed immediately (to send a message, grant access right away, etc.), they would have issued their own commit. By leaving it pending, they accepted that it rides into the next commit — whoever's commit that turns out to be.
- Receiver-side state convergence is guaranteed by `consume_proposal_store(true)` sweeping the same set on both sides. Confirmation tag covers it.
- Forcing callers to "proactively flush before metadata changes" introduces commit churn that fights the bandwidth goal of this PR (smaller, batched commits).

**Final shape:**
- `stage_app_data_propose_and_commit` queues its `AppDataUpdate` and lets `consume_proposal_store(true)` sweep everything else (concurrent `AppDataUpdate`s, leaf-node `Update`s, `Add`/`Remove`/`SelfRemove`, PSK, etc.) into one commit.
- Step 2 (`accumulate_app_data_updates`) still only iterates `AppDataUpdate` proposals — the non-AppData proposals don't contribute to dict updates; they ride into the commit natively via OpenMLS.
- `GroupAppDataError::PendingProposalsExist` variant deleted along with its retry arm.

Intent-ledger "divergence" is a non-issue: the intent's commit *did* succeed; it just happens to also carry other proposals the producer of those proposals already authorized.

PR #3630, opened 2026-05-12
xmtp_mls_common: document WELL_KNOWN change-control rule + drop stale migration-plan refs

tylerhawkes:
### Document WELL_KNOWN change-control rules and remove stale migration-plan references
- Adds extensive documentation to [`registry_table.rs`](https://github.com/xmtp/libxmtp/pull/3630/files#diff-dac151bd9d59cbdc7d59c932a5319a050a31ef5b4a4748c73b8112a7a77b47a4) covering component-id ranges, receiver-side tolerance behavior, and requirements for adding new components to the WELL_KNOWN registry.
- Clarifies module and inline comments in [`policy.rs`](https://github.com/xmtp/libxmtp/pull/3630/files#diff-e77c387fd3922591e80936163b4fd57b0ca8f190aeafe5a62162278b5ba22207) around non-membership slot behavior and AppDataUpdate as the authoritative gate for component writes.
- Updates [`component_source.rs`](https://github.com/xmtp/libxmtp/pull/3630/files#diff-47acce7b46d2cc4f9669dd42120ee90e488ceb0b15817b67b7757e631b05d113) docs to explicitly note that GROUP_MEMBERSHIP is unsupported in `read_from_legacy` and directs consumers to the dedicated extension.
- Removes obsolete comments in [`test_proposals.rs`](https://github.com/xmtp/libxmtp/pull/3630/files#diff-fb8ffb096fea904de9f05e2ad66734c897c3a9ceb42099e25f894c256f1f5f8e) about stale migration plans and failed GCE proposal flow tests. No code or logic was changed in any file.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized c511936.</sup>

PR #3631, opened 2026-05-12
xmtp_mls: bump legacy MIN_SUPPORTED_PROTOCOL_VERSION before bootstrap commit

tylerhawkes:
### Bump legacy MIN_SUPPORTED_PROTOCOL_VERSION before bootstrap commit in `Group.enable_proposals`
- Before issuing the bootstrap commit, `enable_proposals` now checks the group's existing `MinimumSupportedProtocolVersion` via legacy `GroupMutableMetadata` and, if it is below the caller's binary version, publishes a metadata update intent to raise the floor and waits for it to resolve via `sync_until_intent_resolved`.
- The method skips the bump if the group is already migrated or if the current floor is already >= the caller's version, and early-returns if proposals are already enabled.
- Member support for proposals is verified twice: once before the optional bump and once before the bootstrap commit.
- A new test in [test_proposals.rs](https://github.com/xmtp/libxmtp/pull/3631/files#diff-fb8ffb096fea904de9f05e2ad66734c897c3a9ceb42099e25f894c256f1f5f8e) validates that after the bump, older clients enter `paused_for_version` and do not apply the bootstrap commit, while the newer client completes migration.
- Behavioral Change: callers on groups needing a version bump now experience an additional commit and blocking sync wait before the bootstrap commit.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized c150483.</sup>

### DIFF
--- a/crates/xmtp_mls/src/groups/mod.rs
+++ b/crates/xmtp_mls/src/groups/mod.rs
@@ -504,6 +504,118 @@ where
     /// - Not all existing members support proposals
     /// - The group context extension update fails
     pub async fn enable_proposals(&self) -> Result<(), GroupError> {
+        // Two-step migration so old clients (any peer running a
+        // libxmtp release predating this code's `pkg_version`) get a
+        // pause hint they can read BEFORE the legacy
+        // GroupMutableMetadata extension that carries it is stripped
+        // by the bootstrap commit. XIP §3.2.
+        //
+        // Step A: bump MIN_SUPPORTED_PROTOCOL_VERSION in legacy GMM to
+        // `pkg_version()`. Pre-bootstrap groups have no AppData
+        // dictionary registry, so the `MetadataUpdate` handler's
+        // migrated branch is false and the write goes through the
+        // legacy GCE path — exactly the extension old clients know how
+        // to read for `paused_for_version`. Any peer below the version
+        // floor processes this commit, sees the version mismatch in
+        // `validate_one_commit`, and lands in `paused_for_version`. It
+        // never observes step B.
+        //
+        // Step B: bootstrap commit. Strips legacy extensions, seeds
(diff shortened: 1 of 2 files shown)


tylerhawkes:

1. `trunk()`
1. [tyler/appdata-unknown-component-tolerance](https://github.com/xmtp/libxmtp/pull/3628)
1. [tyler/appdata-standalone-proposals](https://github.com/xmtp/libxmtp/pull/3629)
1. [tyler/appdata-well-known-change-control](https://github.com/xmtp/libxmtp/pull/3630)
1. **tyler/appdata-two-step-bootstrap ← this PR**

---

insipx (approved):

tylerhawkes:
Addressed both code findings from the Claude review.

**Finding 1 — race between locks 1 and 2:** Collapsed the two `load_mls_group_with_lock_async` calls into one. All three gating signals (`all_members_support_proposals`, `proposals_enabled`, `needs_min_version_bump`) now read under a single lock, so a concurrent migrator's bootstrap commit can't land between our reads and trick us into publishing a legacy GCE bump against a group whose legacy extensions are gone. Comment block rewritten to explain the consolidated check and what each signal gates.

**Finding 2 — `LibXMTPVersion::parse` error context:** Wrapped both `parse(current_str)?` and `parse(&pkg_version)?` with `.map_err(|e| GroupError::ProposalsNotSupported(format!(...)))` so the error message identifies the source ("legacy GMM MinimumSupportedProtocolVersion" vs "this binary's pkg_version") in addition to the bad string. These parses should never fail in practice (both inputs are internally controlled), so the new error path is purely for diagnostic clarity if something goes very wrong.

Skipping Finding 3 (Codecov "1 missing line") — the happy path is covered by `test_enable_proposals_pauses_old_client_via_legacy_gmm_bump`; the uncovered line is most likely a defensive error branch that's not worth a dedicated test.

Force-pushed `ffcb8b1`.

PR #3634, opened 2026-05-13
comprehensively check libxmtp api surface with `xdbg healthcheck`, compat across versions.

insipx:
### Add `xdbg healthcheck` command to validate libxmtp API surface across versions
- Adds a `Healthcheck` subcommand to the `xdbg` CLI that runs a series of registered ops and validators against the libxmtp API, exiting with code 1 on any failure.
- Ops cover the full group lifecycle: identity creation, group/DM creation, member management, message sending, metadata/permission/consent updates, and message disappearing settings. Each op is inventory-registered with declared dependencies and executed in topological order.
- Adds two post-op validators: `NoForkedGroups` (detects MLS commit log forks per client/group) and `NoMissingMessages` (detects per-client gaps in received Application messages post-join).
- Introduces `HealthContext` which bootstraps persisted identities and a transient identity on first run, loading existing groups for use across ops.
- Adds `missing_messages` to the `QueryGroupMessage` trait and `DbConnection` implementation to support the missing-messages validator.
- Risk: first run auto-creates and persists multiple bootstrap identities to the DB; subsequent runs reuse them, so a corrupted DB may cause persistent failures.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0e9cf4a.</sup>


apps/xmtp_debug/src/app/health/ops/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+#[async_trait]
+pub trait HealthOp: Send + Sync {
+    fn name(&self) -> &'static str;
+    async fn execute(&self, ctx: &mut HealthContext) -> Vec<OpResult>;
+}
+
+/// Self-registration entry for an op. Each op submits one of these from
+/// its own module via `inventory::submit!`.
+pub struct OpEntry {
+    pub op_name: &'static str,
```
tylerhawkes, line 43: Is this needed if the make value also has the `name` fn from `HealthOp`?

insipx: it avoids constructing the box each time name is needed, but I can also use a static `&'a Trait` here since all the types implementing are unit structs.

PR #3642, opened 2026-05-14
cross-version cross-talk tests

insipx:
same energy as cross version tests, except clients never upgrade to new versions

> [!NOTE]
> ### Add cross-talk test harness and version-partitioned identity storage to xdbg
> - Introduces a new `cross-talk-test` driver alongside `cross-version-test`, orchestrating multi-phase interop tests where different xdbg binary versions create identities, share a group, and exchange messages using `--strict-versioning` mode.
> - Adds a `--strict-versioning` CLI flag to xdbg that restricts identity reads to the current binary's version partition; SQLite paths are now bucketed by a per-binary xxh3 version hash via `db_directory_for`.
> - Introduces `IdentityKey` with a 3-part `(network, version_hash, inbox)` schema in the redb identity store, enabling version-partitioned reads (`load_for_version`, `find_by_inbox`, `sample_n` with strict mode).
> - Adds a new `EnsurePerVersionMembership` health op that adds missing per-version representatives to newly created groups, and a condition system (`Conditions`/`RegistryBuild`) that gates ops and validators based on active runtime flags.
> - Adds a new `sync` CLI subcommand that syncs libxmtp SQLite state and reconciles the local redb `GroupStore` and `MessageStore` with the network for all loaded identities.
> - Rewrites the `cross-version-test` and `cross-talk-test` Nix packages as Python applications backed by a shared `xdbg-driver-lib`, replacing the previous shell script approach.
> - Risk: identity store version is bumped to 3 (groups to 2); existing redb data under old namespaces is not migrated and startup aborts if legacy identity tables are detected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3821e8e.</sup>
> 
>


apps/xmtp_debug/src/args.rs
```diff
@@ -726,4 +778,79 @@ mod tests {
...
+        let opts = AppOpts::try_parse_from(["xdbg", "-s"]).expect("-s parses via --show-fields");
+        assert!(
+            !opts.strict_versioning,
+            "--strict-versioning should not have a short alias"
+        );
+    }
+
+    #[test]
+    fn parses_sync_subcommand() {
+        let opts = AppOpts::try_parse_from(["xdbg", "sync"]).expect("parses `xdbg sync`");
```
insipx: this test is unnecessary. eliminate tests from this pr which are testing simple functionality, they just serve to bloat lines. we can trust that clap is correct

tylerhawkes (approved):

PR #3645, opened 2026-05-14
xmtp_proto: represent GroupId as fixed-size [u8; 16] Copy newtype

### LINKED ISSUE #3639: Represent `GroupId` as fixed-length `[u8;16]`
# GroupId: Fixed-Size, Copy Representation

**Date:** 2026-05-14
**Status:** Proposed

## Goal

Change the internal representation of `xmtp_proto::types::ids::GroupId` from `bytes::Bytes` to `[u8; 16]` and make the type `Copy`. Update all callsites to match the new ergonomics (drop redundant `.clone()`, switch fallible constructions, pass by value where appropriate).

## Motivation

MLS group IDs in XMTP are always 16 bytes. The current `Bytes` representation:

- Allocates heap storage for every group ID (16 bytes payload + `Bytes` overhead per instance).
- Requires `.clone()` at every move site, even though the underlying data is trivially copyable.
- Silently accepts wrong-length input (e.g. `GroupId::from(vec![1, 2, 3])`), creating IDs that violate the protocol invariant. The audit found this in production paths and tests.
- Hides the fixed-size invariant from the type system. Downstream code cannot rely on `id.len() == 16` without runtime checks.

A `Copy` `[u8; 16]` representation:

- Eliminates the heap allocation and the per-instance `Bytes` overhead.
- Removes ~363 `.clone()` calls (mechanical cleanup).
- Encodes the protocol invariant in the type. Wrong-length inputs become a compile-time-narrowed `TryFrom` failure at the boundary instead of silent garbage downstream.
- Aligns `GroupId` with its sibling `InstallationId` (already `Copy<[u8; 32]>`).

## Non-Goals

- Changing the on-wire or on-disk format. The protobuf and SQLite encoding remain raw bytes of length 16.
- Changing semantic length validation in production. We are *adding* validation where today there is none, but not changing the legal length.
- Refactoring unrelated identifier types (`InstallationId` is already correct; nothing else moves).
- Performance tuning beyond what falls out of removing the heap allocation. No benchmarks are required to land this; it is primarily a correctness and ergonomics change.

## Invariants

1. A `GroupId` is exactly 16 bytes. No code path can construct a `GroupId` with any other length.
2. Legacy data with non-16-byte group IDs does **not** exist in production. Any encountered non-16-byte data is corruption and must surface as an error, not be silently accepted.
3. `GroupId` is `Copy`. Passing it by value never requires `.clone()`.

### Internal representation

```rust
#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
pub struct GroupId([u8; 16]);
```

`Default` produces `[0u8; 16]` (the all-zero ID). This is a semantic change from today (where `Default` is empty `Bytes`). The audit found 3 `GroupId::default()` callers; all are placeholder / not-found values where the zero ID is acceptable.

### Construction (infallible)

```rust
impl From<[u8; 16]> for GroupId { ... }
impl From<&[u8; 16]> for GroupId { ... }
```

Plus a randomness constructor (preserved):

```rust
impl GroupId {
    pub fn random<R: openmls_traits::random::OpenMlsRand>(rand: &R) -> Self { ... }
}
```

### Construction (fallible)

```rust
impl TryFrom<&[u8]> for GroupId { type Error = ConversionError; ... }
impl TryFrom<Vec<u8>> for GroupId { type Error = ConversionError; ... }
impl TryFrom<&openmls::group::GroupId> for GroupId { type Error = ConversionError; ... }
impl TryFrom<openmls::group::GroupId> for GroupId { type Error = ConversionError; ... }
```

`ConversionError` is the existing error type used by `InstallationId` in the same module. Consistency wins.

The previous infallible `From<Vec<u8>>` / `From<&[u8]>` impls are **removed**. Callsites that fed arbitrary-length input must now propagate or `.expect()` the `TryFrom` result. The audit identified ~40 such sites.

### Parsing from hex string

```rust
impl FromStr for GroupId {
    type Err = GroupIdParseError;
    fn from_str(s: &str) -> Result<Self, Self::Err> { ... }
}

#[derive(Debug, thiserror::Error)]
pub enum GroupIdParseError {
    #[error(transparent)]
    Hex(#[from] hex::FromHexError),
    #[error(transparent)]
    Length(#[from] ConversionError),
}
```

A new error type because parsing can fail on either hex decoding or length, and callers benefit from distinguishing.

### Accessors (outward)

```rust
impl GroupId {
    pub fn as_slice(&self) -> &[u8]            { &self.0 }
    pub fn as_bytes(&self) -> &[u8; 16]        { &self.0 }
    pub fn into_bytes(self) -> [u8; 16]        { self.0 }
    pub fn to_vec(self) -> Vec<u8>             { self.0.to_vec() }
    pub fn to_openmls(self) -> openmls::group::GroupId { ... }
}

impl AsRef<[u8]> for GroupId { ... }
impl Borrow<[u8]> for GroupId { ... }
```

**No `Deref` impl.** The previous `Deref<Target = Bytes>` is removed. The audit found Deref-via-`Bytes`-methods usage is essentially nil outside tests that touched `.0` directly; those tests are rewritten.

`From<GroupId> for Vec<u8>` is preserved (now via `to_vec`).

### Equality conveniences

Mirror `InstallationId`'s pattern:

```rust
impl PartialEq<[u8; 16]> for GroupId { ... }   // + reverse direction
impl PartialEq<[u8]>     for GroupId { ... }   // + reverse
impl PartialEq<Vec<u8>>  for GroupId { ... }   // + reverse
impl PartialEq<&Vec<u8>> for GroupId { ... }   // + reverse
```

Length-mismatched comparisons return `false`, never panic. These are convenient in tests and assertion paths.

### Display & Debug

Unchanged in spirit:

- `Display`: hex of the 16 bytes (`{:x}`-style).
- `Debug`: `GroupId(<hex>)` via `xmtp_common::fmt::debug_hex`.

### Serde

```rust
impl Serialize for GroupId {
    // serialize as &[u8] (length-16). Wire-compatible with existing format.
}
impl<'de> Deserialize<'de> for GroupId {
    // deserialize Vec<u8>, then TryFrom — error on length mismatch.
}
```

Wire format unchanged; deserialization now rejects malformed input.

### Diesel

```rust
impl ToSql<Binary, Sqlite> for GroupId   { ... } // writes 16 bytes
impl FromSql<Binary, Sqlite> for GroupId { ... } // length-validates; errors on mismatch
```

A row whose blob is not 16 bytes becomes a load error. Per the invariants section, such rows are not expected to exist; surfacing the corruption is correct behavior.

### Phase 1: type change + boundary fixes

1. Replace the type definition and trait impls in `crates/xmtp_proto/src/types/ids/group_id.rs`.
2. Rewrite the in-file tests using valid 16-byte inputs.
3. Fix all compile errors crate-by-crate, in dependency order (xmtp_proto → xmtp_db → xmtp_api → xmtp_mls → bindings/apps).
4. At each fallible construction site, decide:
   - **Boundary** (decoding bytes from a proto, DB row, network frame): propagate the error with `?` and the crate's error type.
   - **Test helper or known-good source** (just got from `random()`, just read from a known-length struct field): `.try_into().expect("GroupId from N-byte source")` — assertions document the invariant.
   - **Test fixture with wrong-length intent**: replace the fixture with a valid 16-byte value. Prefer `GroupId::from([1u8; 16])` or `xmtp_common::Generate::generate()`.

### Phase 2: cleanup pass

1. Remove redundant `.clone()` on `GroupId` values. Mechanical (clippy `redundant_clone` will flag many; the rest fall out of compile-clean code).
2. Optionally tighten function signatures: `fn(group_id: &GroupId)` → `fn(group_id: GroupId)` where the callee does not need a reference for lifetime reasons. Apply only to internal APIs in this PR; do not touch FFI boundaries (`bindings_*`) in the same change.
3. Drop intermediate `Vec<u8>` allocations where `as_slice()` / `to_openmls()` will do.

### Phase 3: verification

1. `just lint` clean.
2. `just test` (v3 + d14n) green.
3. `just wasm test`, `just node test`.
4. Spot-check the binding crates (`just android check`, `just ios check`) compile.

## Risks

- **Serde wire compatibility:** existing serialized data uses `serialize as &[u8]` and deserializes `Vec<u8>`. The new code preserves both, so cross-version data round-trips. Cap: a payload encoded as a non-`Vec`/non-byte-array (e.g. a serde_json string) would have already been broken; we are not making that case worse.
- **Diesel data on disk:** if any sled/sqlite database in CI fixtures or developer machines holds wrong-length group IDs, loading will fail. The invariants section asserts this case does not exist; if Phase 3 turns up failures, we resolve them as data corruption (delete fixture, regenerate) rather than re-introducing flexibility into the type.
- **Test churn:** ~20+ test files use synthetic short group IDs (`vec![1, 2, 3]`). All rewrite to use either `Generate::generate()` or a fixed 16-byte literal. The rewrites are mechanical but numerous.
- **FFI surface:** `bindings_mobile`, `bindings_wasm`, `bindings_node` expose `Vec<u8>` group IDs to their host languages, not `GroupId` directly. The internal conversion at the FFI layer must `try_into` and surface a sensible error to the host. This is the highest-risk seam; Phase 1 verifies these by-crate.

## Testing

The existing in-file tests are rewritten to use 16-byte inputs and to exercise the new fallible-conversion error paths:

- `From<[u8; 16]>` and `From<&[u8; 16]>` round-trip.
- `TryFrom<&[u8]>` and `TryFrom<Vec<u8>>`: success at length 16, error at every other length (parameterized).
- `TryFrom<&openmls::GroupId>`: success when the inner length is 16, error otherwise.
- `FromStr`: success for 32-hex-char input, error for bad hex, error for hex of wrong byte length.
- Serde round-trip with valid input; deserialization error on wrong-length input.
- Diesel round-trip with valid input; `FromSql` error on wrong-length blob.
- `PartialEq` against `[u8; 16]`, `[u8]`, `Vec<u8>` — true on match, false on mismatch, false on length mismatch.
- `Generate::generate()` always produces 16-byte values.

No new integration tests are required; existing integration tests exercise `GroupId` extensively and will validate the change implicitly.

## Out of Scope (Explicitly)

- Adding a `Bytes`-backed escape hatch for legacy data.
- Changing `InstallationId` (already correct).
- Renaming the type or moving the module.
- Tuning binding-layer error reporting beyond "surface as an error to the host language."

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3639

## Summary

- Replace `GroupId`'s `bytes::Bytes` backing with a `Copy` newtype over `[u8; 16]`, encoding the protocol's 16-byte invariant in the type system.
- Migrate all callsites: `TryFrom` at decode boundaries (propagating `ConversionError`), 16-byte literals or `Generate::generate()` for test fixtures.
- Mechanical cleanup: drop redundant `.clone()` and reference patterns now that `GroupId: Copy`. Mirror the existing `InstallationId` pattern.

## API surface changes

**New:**
- `From<[u8; 16]>`, `From<&[u8; 16]>` (infallible)
- `TryFrom<&[u8]>`, `TryFrom<Vec<u8>>`, `TryFrom<&openmls::group::GroupId>`, `TryFrom<openmls::group::GroupId>` returning `ConversionError`
- `as_bytes() -> &[u8; 16]`, `into_bytes() -> [u8; 16]` accessors
- `GroupIdParseError` (`FromStr` distinguishes hex vs length errors)

**Removed:**
- Infallible `From<Vec<u8>>`, `From<&[u8]>`, `From<openmls::group::GroupId>` (replaced by `TryFrom`)
- `Deref<Target = bytes::Bytes>` impl

**Behavior change (intentional, per spec):**
- `Default` now produces `[0u8; 16]` instead of empty bytes.
- Serde and Diesel `FromSql` now reject wrong-length data (was silently truncated/padded).

## Wire compatibility

- Serde format is unchanged (16 raw bytes).
- Diesel `ToSql`/`FromSql` still read/write a 16-byte `Binary` blob; rows with wrong-length data now surface as load errors (per spec § \"Invariants 2\": such data is corruption and should not be silently accepted).

## Ripple

| Crate | Type of change |
|-------|---------------|
| `xmtp_proto` | Type rewrite + 18 new in-file tests |
| `xmtp_db` | Diesel boundary `try_from`; test fixtures to 16-byte literals; helper signatures tightened |
| `xmtp_api`, `xmtp_api_d14n` | Proto decode boundary `try_from`; test fixtures |
| `xmtp_mls` | Boundary `try_from` (welcomes, sync, commit log, validated commit, oneshot, worker); new `Conversion` error variants where they didn't exist |
| `xmtp_archive` | One redundant projection dropped |
| `bindings/{mobile,node,wasm}` | FFI entry-point `try_from`, mapping `ConversionError` to host error type. Public FFI signatures unchanged. |
| `apps/*` | Same pattern at CLI/debug-tool entry points |
| Workspace-wide | ~160 `hex::encode(&group_id)` → `hex::encode(group_id)` and similar `.clone()` removals now that `GroupId: Copy` |

## Verification

- ✅ `just lint-rust` clean (clippy `-D warnings`, fmt, hakari).
- ✅ `cargo nextest` hermetic suites pass:
  - `xmtp_proto`: 30 new tests for the new type (round-trip, length error paths, serde, Diesel, hex parse).
  - `xmtp_db`: 188 tests pass.
  - `xmtp_archive`: 5 tests pass.
  - `xmtp_proto + xmtp_db + xmtp_archive`: 300 total tests pass.
- ✅ Workspace `cargo check --workspace --all-features --all-targets --exclude log_parser` clean (including `xmtpv3`, `bindings_node`, `bindings_wasm`).
- ✅ Spec invariant audit: zero remaining `GroupId::from(vec![...])` / `GroupId::from(&[...])` / `impl Deref for GroupId` / infallible `From<Vec<u8>>` for GroupId callsites or impls.
- ⏸️ `just test v3` / `just test d14n` / `just wasm test` / `just node test` not run locally (require Docker XMTP node / browser env not available in build sandbox); leaving to CI.
- ⏸️ `just android check` / `just ios check` not run locally; leaving to CI.

## Plan & design

The spec, multi-phase plan, and per-phase task files are checked in under `docs/plans/2026-05-14-issue-3639*`.

## Test plan

- [ ] CI green on `just test v3` and `just test d14n`
- [ ] CI green on `just wasm test`, `just node test`
- [ ] CI green on `just android check`, `just ios check`
- [ ] Reviewer confirms FFI error mapping is the right variant in each binding crate (mobile/wasm/node)

> [!NOTE]
> ### Represent `GroupId` as a fixed-size `[u8; 16]` Copy newtype with fallible constructors
> - Changes `GroupId` internal representation from `bytes::Bytes` to `[u8; 16]`, making it `Copy` and enforcing a 16-byte invariant at construction time.
> - Removes infallible `From<Vec<u8>>` and `From<&[u8]>` conversions; all paths that previously accepted arbitrary-length byte buffers now use `TryFrom` and return `ConversionError` on invalid length.
> - Adds convenience constants (`ZERO`, `ONE`, `TWO`, etc.), `as_slice`, `to_vec`, `into_bytes`, and a `Generate` impl; `to_openmls` now takes ownership.
> - Propagates `ConversionError` through all layers that accept group IDs: MLS sync, commit log, welcome processing, subscriptions, FFI bindings (Node/WASM/mobile), CLI, and DB queries.
> - Risk: Any stored or network-supplied group ID that is not exactly 16 bytes will now produce a hard error instead of being silently accepted; deserialization via serde/Diesel will also fail on wrong-length blobs.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e61fed3.</sup>
> 
>

### DIFF
--- a/apps/db_tools/src/tasks/clear_messages.rs
+++ b/apps/db_tools/src/tasks/clear_messages.rs
@@ -53,10 +53,11 @@ mod tests {
         // Commit and application msg
         assert_eq!(alix_msgs.len(), 2);
 
+        let dm_group_id_vec = alix_bo_dm.group_id.to_vec();
         clear_all_messages_confirmed(
             &alix.db(),
             None,
-            Some(std::slice::from_ref(&alix_bo_dm.group_id)),
+            Some(std::slice::from_ref(&dm_group_id_vec)),
         )?;
         let alix_msgs = alix_bo_dm.find_messages(&MsgQueryArgs::default())?;
         // Commit and application msg
--- a/apps/db_tools/src/tasks/clear_messages.rs
+++ b/apps/db_tools/src/tasks/clear_messages.rs
@@ -1,12 +1,14 @@
 use anyhow::Result;
-use xmtp_db::{ConnectionExt, DbConnection, group_message::QueryGroupMessage};
+use xmtp_db::{
+    ConnectionExt, DbConnection, group_message::QueryGroupMessage, proto::types::GroupId,
+};
 
 use crate::confirm_destructive;
 
 pub fn clear_all_messages<C: ConnectionExt>(
     conn: &C,
     limit_days: Option<u32>,
-    group_ids: Option<&[Vec<u8>]>,
+    group_ids: Option<&[GroupId]>,
 ) -> Result<()> {
     confirm_destructive()?;
     clear_all_messages_confirmed(conn, limit_days, group_ids)
@@ -15,7 +17,7 @@ pub fn clear_all_messages<C: ConnectionExt>(
 pub fn clear_all_messages_confirmed<C: ConnectionExt>(
     conn: &C,
     limit_days: Option<u32>,
-    group_ids: Option<&[Vec<u8>]>,
+    group_ids: Option<&[GroupId]>,
 ) -> Result<()> {
     let db = DbConnection::new(conn);
     db.clear_messages(group_ids, limit_days)?;
@@ -53,11 +55,10 @@ mod tests {
         // Commit and application msg
         assert_eq!(alix_msgs.len(), 2);
 
-        let dm_group_id_vec = alix_bo_dm.group_id.to_vec();
         clear_all_messages_confirmed(
             &alix.db(),
             None,
-            Some(std::slice::from_ref(&dm_group_id_vec)),
+            Some(std::slice::from_ref(&alix_bo_dm.group_id)),
         )?;
         let alix_msgs = alix_bo_dm.find_messages(&MsgQueryArgs::default())?;
         // Commit and application msg
--- a/apps/db_tools/src/tasks/db_bench.rs
+++ b/apps/db_tools/src/tasks/db_bench.rs
@@ -316,17 +316,17 @@ where
         bench!(self, primary_sync_group())?;
         bench!(self, find_group(&group.id))?;
         bench!(self, find_sync_group(&group.id))?;
-        bench!(self, get_rotated_at_ns(group.id.clone()))?;
-        bench!(self, update_rotated_at_ns(group.id.clone()))?;
-        bench!(self, get_installations_time_checked(group.id.clone()))?;
-        bench!(self, update_installations_time_checked(group.id.clone()))?;
+        bench!(self, get_rotated_at_ns(&group.id))?;
+        bench!(self, update_rotated_at_ns(&group.id))?;
+        bench!(self, get_installations_time_checked(&group.id))?;
+        bench!(self, update_installations_time_checked(&group.id))?;
         bench!(
             self,
-            update_message_disappearing_from_ns(group.id.clone(), Some(1000))
+            update_message_disappearing_from_ns(&group.id, Some(1000))
         )?;
         bench!(
             self,
-            update_message_disappearing_in_ns(group.id.clone(), Some(1000))
+            update_message_disappearing_in_ns(&group.id, Some(1000))
         )?;
 
         // Fork status queries
@@ -347,7 +347,7 @@ where
         let Some(group) = self.group_or_dm() else {
             bail!("No group to lookup DMs on");
         };
-        let group_id = hex::encode(&group.id);
+        let group_id = hex::encode(group.id);
 
         let new_consent = StoredConsentRecord {
             consented_at_ns: 0,
@@ -471,7 +471,7 @@ where
         )?;
 
         // Get last cursor for IDs
-        let ids = vec![group.id.clone()];
+        let ids = vec![group.id];
         let entities = vec![EntityKind::ApplicationMessage, EntityKind::CommitMessage];
         bench!(self, get_last_cursor_for_ids(&ids, &entities))?;
 
@@ -479,7 +479,7 @@ where
         bench!(
             self,
             update_cursor(
-                group.id.clone(),
+                group.id,
                 EntityKind::ApplicationMessage,
                 Cursor::new(0, 0u32)
             )
@@ -489,7 +489,8 @@ where
         bench!(self, latest_cursor_for_id(&group.id, &entities, None))?;
 
         // Get remote log cursors
-        let conv_ids = vec![&group.id];
+        let conv_id_vec = group.id.to_vec();
+        let conv_ids = vec![&conv_id_vec];
         bench!(self, get_remote_log_cursors(&conv_ids))?;
 
         Ok(())
--- a/apps/db_tools/src/tasks/db_bench.rs
+++ b/apps/db_tools/src/tasks/db_bench.rs
@@ -489,9 +489,7 @@ where
         bench!(self, latest_cursor_for_id(&group.id, &entities, None))?;
 
         // Get remote log cursors
-        let conv_id_vec = group.id.to_vec();
-        let conv_ids = vec![&conv_id_vec];
-        bench!(self, get_remote_log_cursors(&conv_ids))?;
+        bench!(self, get_remote_log_cursors(&[group.id.as_slice()]))?;
 
         Ok(())
     }
--- a/apps/xmtp_debug/src/app/generate/messages.rs
+++ b/apps/xmtp_debug/src/app/generate/messages.rs
@@ -170,13 +170,11 @@ impl GenerateMessages {
             .get(&group.created_by)
             .ok_or(eyre!("group has no owner"))?;
         let owner = owner.lock().await;
-        let owner_group = owner
-            .group(&xmtp_proto::types::GroupId::from(group.id))
-            .wrap_err(format!(
-                "owner {} of group {} failed to look up in sqlite db",
-                hex::encode(group.created_by),
-                hex::encode(group.id)
-            ))?;
+        let owner_group = owner.group(&group.group_id()).wrap_err(format!(
+            "owner {} of group {} failed to look up in sqlite db",
+            hex::encode(group.created_by),
+            group.group_id()
+        ))?;
         owner_group
             .add_members(&[hex::encode(not_in_group)])
             .await
@@ -213,7 +211,7 @@ impl GenerateMessages {
                 .ok_or(eyre!("client does not exist"))?;
             let client = client.lock().await;
             client.sync_welcomes().await?;
-            let mls_group = client.group(&xmtp_proto::types::GroupId::from(group.id))?;
+            let mls_group = client.group(&group.group_id())?;
             mls_group.sync_with_conn().await?;
             mls_group.maybe_update_installations(None).await?;
             let words = rng.random_range(0..10);
@@ -304,14 +302,14 @@ impl GenerateMessages {
         let group = group_store
             .random(&network, rng)?
             .ok_or(eyre!("no group in local store"))?;
-        info!(time = ?Instant::now(), group = hex::encode(group.id), "sending message");
+        info!(time = ?Instant::now(), group = %group.group_id(), "sending message");
         if let Some(inbox_id) = group.members.choose(rng) {
             let client = clients
                 .get(inbox_id.as_slice())
                 .ok_or(eyre!("client does not exist"))?;
             let client = client.lock().await;
             client.sync_welcomes().await?;
-            let group = client.group(&xmtp_proto::types::GroupId::from(group.id))?;
+            let group = client.group(&group.group_id())?;
             group.sync_with_conn().await?;
             group.maybe_update_installations(None).await?;
             let words = rng.random_range(0..*max_message_size);
--- a/apps/xmtp_debug/src/app/generate/messages.rs
+++ b/apps/xmtp_debug/src/app/generate/messages.rs
@@ -170,11 +170,13 @@ impl GenerateMessages {
             .get(&group.created_by)
             .ok_or(eyre!("group has no owner"))?;
         let owner = owner.lock().await;
-        let owner_group = owner.group(group.id.as_ref()).wrap_err(format!(
-            "owner {} of group {} failed to look up in sqlite db",
-            hex::encode(group.created_by),
-            hex::encode(group.id)
-        ))?;
+        let owner_group = owner
+            .group(&xmtp_proto::types::GroupId::from(group.id))
+            .wrap_err(format!(
+                "owner {} of group {} failed to look up in sqlite db",
+                hex::encode(group.created_by),
+                hex::encode(group.id)
+            ))?;
         owner_group
             .add_members(&[hex::encode(not_in_group)])
             .await
@@ -211,7 +213,7 @@ impl GenerateMessages {
                 .ok_or(eyre!("client does not exist"))?;
             let client = client.lock().await;
             client.sync_welcomes().await?;
-            let mls_group = client.group(&group.id)?;
+            let mls_group = client.group(&xmtp_proto::types::GroupId::from(group.id))?;
             mls_group.sync_with_conn().await?;
             mls_group.maybe_update_installations(None).await?;
             let words = rng.random_range(0..10);
@@ -309,7 +311,7 @@ impl GenerateMessages {
                 .ok_or(eyre!("client does not exist"))?;
             let client = client.lock().await;
             client.sync_welcomes().await?;
-            let group = client.group(&group.id)?;
+            let group = client.group(&xmtp_proto::types::GroupId::from(group.id))?;
             group.sync_with_conn().await?;
             group.maybe_update_installations(None).await?;
             let words = rng.random_range(0..*max_message_size);
--- a/crates/xmtp_api/src/mls.rs
+++ b/crates/xmtp_api/src/mls.rs
@@ -69,7 +69,7 @@ impl<ApiClient> ApiClientWrapper<ApiClient>
 where
     ApiClient: XmtpApi,
 {
-    #[tracing::instrument(level = "trace", skip_all, fields(group_id = hex::encode(group_id)))]
+    #[tracing::instrument(level = "trace", skip_all, fields(group_id = %group_id))]
     pub async fn query_group_messages(&self, group_id: GroupId) -> Result<Vec<GroupMessage>> {
         self.api_client
             .query_group_messages(group_id)
@@ -78,13 +78,13 @@ where
     }
 
     /// Query for the latest message on a group
-    #[tracing::instrument(level = "trace", skip_all, fields(group_id = hex::encode(group_id)))]
+    #[tracing::instrument(level = "trace", skip_all, fields(group_id = %group_id))]
     pub async fn query_latest_group_message(
         &self,
         group_id: GroupId,
     ) -> Result<Option<GroupMessage>> {
         tracing::debug!(
-            group_id = hex::encode(group_id),
+            group_id = %group_id,
             inbox_id = self.inbox_id,
             "query latest group message"
         );
--- a/crates/xmtp_api/src/mls.rs
+++ b/crates/xmtp_api/src/mls.rs
@@ -389,7 +389,7 @@ pub mod tests {
 
     pub fn build_group_messages(
         num_messages: usize,
-        group_id: Vec<u8>,
+        group_id: GroupId,
     ) -> Vec<mls_v1::GroupMessage> {
         let mut out: Vec<mls_v1::GroupMessage> = vec![];
         for i in 0..num_messages {
@@ -397,7 +397,7 @@ pub mod tests {
                 version: Some(GroupMessageVersion::V1(GroupMessageV1 {
                     id: i as u64,
                     created_ns: i as u64,
-                    group_id: group_id.clone(),
+                    group_id: group_id.to_vec(),
                     data: MlsMessageOut::from(FakeMlsApplicationMessage::generate())
                         .to_bytes()
                         .unwrap(),
@@ -472,17 +472,16 @@ pub mod tests {
     async fn test_read_group_messages_single_page() {
         let mock_api = MockNetworkClient::default();
         let mut v3_client = V3Client::new(mock_api, NoCursorStore);
-        let group_id = rand_vec::<16>();
-        let group_id_clone = group_id.clone();
+        let group_id = GroupId::generate();
         // Set expectation for first request with no cursor
         v3_client
             .client_mut()
             .expect_request()
             .returning(move |_, _, mut body| {
                 let req = QueryGroupMessagesRequest::decode(&mut body).unwrap();
-                assert_eq!(req.group_id, group_id.clone());
+                assert_eq!(req.group_id, group_id);
 
-                let msgs = build_group_messages(10, group_id.clone());
+                let msgs = build_group_messages(10, group_id);
                 let mut bytes = prost::bytes::BytesMut::new();
                 let res = QueryGroupMessagesResponse {
                     messages: msgs,
@@ -498,10 +497,7 @@ pub mod tests {
 
         let wrapper = ApiClientWrapper::new(v3_client, exponential().build());
 
-        let result = wrapper
-            .query_group_messages(group_id_clone.try_into().unwrap())
-            .await
-            .unwrap();
+        let result = wrapper.query_group_messages(group_id).await.unwrap();
         assert_eq!(result.len(), 10);
     }
 
@@ -509,17 +505,16 @@ pub mod tests {
     async fn test_read_group_messages_single_page_xactly_100_results() {
         let mock_api = MockNetworkClient::default();
         let mut v3_client = V3Client::new(mock_api, NoCursorStore);
-        let group_id = rand_vec::<16>();
-        let group_id_clone = group_id.clone();
+        let group_id = GroupId::generate();
         // Set expectation for first request with no cursor
         v3_client
             .client_mut()
             .expect_request()
             .returning(move |_, _, mut body| {
                 let req = QueryGroupMessagesRequest::decode(&mut body).unwrap();
-                assert_eq!(req.group_id, group_id.clone());
+                assert_eq!(req.group_id, group_id);
 
-                let msgs = build_group_messages(100, group_id.clone());
+                let msgs = build_group_messages(100, group_id);
                 let mut bytes = prost::bytes::BytesMut::new();
                 let res = QueryGroupMessagesResponse {
                     messages: msgs,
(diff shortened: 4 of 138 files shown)


apps/db_tools/src/tasks/clear_messages.rs
```diff
@@ -53,10 +53,11 @@ mod tests {
         // Commit and application msg
         assert_eq!(alix_msgs.len(), 2);
 
+        let dm_group_id_vec = alix_bo_dm.group_id.to_vec();
         clear_all_messages_confirmed(
             &alix.db(),
             None,
-            Some(std::slice::from_ref(&alix_bo_dm.group_id)),
+            Some(std::slice::from_ref(&dm_group_id_vec)),
```
insipx: Why not just use as slice on Groupid here

apps/db_tools/src/tasks/db_bench.rs
```diff
@@ -489,7 +489,8 @@ where
         bench!(self, latest_cursor_for_id(&group.id, &entities, None))?;
 
         // Get remote log cursors
-        let conv_ids = vec![&group.id];
+        let conv_id_vec = group.id.to_vec();
+        let conv_ids = vec![&conv_id_vec];
```
insipx: Does this have to be a vec reference or could it be a slice?

apps/xmtp_debug/src/app/generate/messages.rs
```diff
@@ -224,11 +170,13 @@ impl GenerateMessages {
             .get(&group.created_by)
             .ok_or(eyre!("group has no owner"))?;
         let owner = owner.lock().await;
-        let owner_group = owner.group(group.id.as_ref()).wrap_err(format!(
-            "owner {} of group {} failed to look up in sqlite db",
-            hex::encode(group.created_by),
-            hex::encode(group.id)
-        ))?;
+        let owner_group = owner
+            .group(&xmtp_proto::types::GroupId::from(group.id))
```
insipx: Could we use the group id type throughout xdbg too? Iirc xdbg also just uses [u8 ;16] for group id

crates/xmtp_api_d14n/src/middleware/multi_node_client/client.rs
```diff
@@ -204,7 +204,7 @@ mod tests {
     #[xmtp_common::test]
     async fn d14n_request_latest_group_message() {
         let client = create_d14n_client();
-        let id: GroupId = GroupId::from(vec![]);
```
insipx, line 207: Could also use GroupId::random for cases where we don't care about group and/or create `GroupId::one', `GroupId::two` helpers

crates/xmtp_db/src/encrypted_store/group.rs
```diff
@@ -1357,7 +1357,7 @@ pub(crate) mod tests {
         state: Option<GroupMembershipState>,
         created_at_ns: i64,
     ) -> StoredGroup {
-        let id = rand_vec::<24>();
+        let id = GroupId::from(xmtp_common::rand_array::<16>());
```
insipx: Couldn't this just use the random method already on group id. Prefer that over this function nesting everywhere else as well

crates/xmtp_api/src/mls.rs
```diff
@@ -545,7 +546,7 @@ pub mod tests {
     async fn test_read_topic_multi_page() {
         let mock_api = MockNetworkClient::new();
         let mut v3_client = V3Client::new(mock_api, NoCursorStore);
-        let group_id = vec![1, 2, 3, 4];
+        let group_id = vec![1u8; 16];
```
insipx: A GroupId::one helper could be useful here and elsewhere for instance

crates/xmtp_api_d14n/src/protocol/utils/test/test_envelope.rs
```diff
@@ -150,7 +150,7 @@ impl Envelope<'_> for TestEnvelope {
             .cursor(self.cursor())
             .depends_on(self.depends_on())
             .payload(buf)
-            .group_id(vec![0, 1, 2])
+            .group_id([0u8; 16])
```
insipx: Group id zero helper too

insipx (changes requested):

crates/xmtp_api_d14n/src/middleware/multi_node_client/client.rs
xmtp-coder-agent: Good call — swapped to `GroupId::generate()` in the test (commit 4dd738945). `GroupId::random` requires an `OpenMlsRand` provider; `generate()` is the test-only zero-arg helper that fits this case (opaque value, no crypto provider in scope). Also applied the same swap across the other ~23 `GroupId::from(xmtp_common::rand_array::<16>())` callsites in xmtp_db / xmtp_mls.

crates/xmtp_db/src/encrypted_store/group.rs
xmtp-coder-agent: Done in 4dd738945 — swept all 24 `GroupId::from(xmtp_common::rand_array::<16>())` sites across `xmtp_db` and `xmtp_mls` to `GroupId::generate()`.

apps/db_tools/src/tasks/clear_messages.rs
xmtp-coder-agent: Better idea — in 4dd738945 I changed the `clear_messages` parameter (trait method, both impls, mock, and the `clear_all_messages*` wrappers) from `Option<&[Vec<u8>]>` to `Option<&[GroupId]>`. That eliminates the `.to_vec()` + temp slice dance at this callsite (now just `Some(std::slice::from_ref(&alix_bo_dm.group_id))`) and tightens the API to the protocol id type. The db_tools CLI `group_ids` parser now parses each hex token via `<GroupId as FromStr>` directly.

crates/xmtp_mls/src/groups/welcomes/xmtp_welcome.rs
```diff
@@ -501,10 +498,10 @@ where
         // before calling insert_or_replace_group
         if is_readd_after_leaving && let Some(ref existing) = existing_group {
             tracing::info!(
-                group_id = hex::encode(&existing.id),
+                group_id = hex::encode(existing.id),
```
insipx: now that the GroupId type has a nice Display/Debug impl, do we need hex::encode here or could we use tracing sugar, i.e `tracing::info!(group_id = %existing.id)`? because that's much more preferable

xmtp-coder-agent: Agreed and much nicer — done in 671651b9e. Swept 78 sites across xmtp_mls / xmtp_api / d14n subscriptions, including the same fix inside `#[tracing::instrument(... fields(...))]` attributes. Left alone the handful of sites where the source is a `Vec<u8>` / `&[u8]` / `openmls::group::GroupId` / proto wire-format field (no `Display`), and the few CLI `let x = hex::encode(...)` locals used by `println!`.

apps/db_tools/src/tasks/db_bench.rs
xmtp-coder-agent: Good catch — done in 346e1c75e. Trait signature changed to `&[&[u8]]` (forwarding impl + mock + Diesel impl). Callers updated: `commit_log.rs` maps `keys()` through `Vec::as_slice`; this `db_bench` site collapses to `bench!(self, get_remote_log_cursors(&[group.id.as_slice()]))?;`.

apps/xmtp_debug/src/app/generate/messages.rs
xmtp-coder-agent: Done in 346e1c75e. Added `Group::group_id() -> GroupId` on xdbg`s internal cache type and migrated every `client.group(&GroupId::from(group.id))` callsite to `client.group(&group.group_id())`. Kept `Group.id` itself as `[u8; 16]` because the type is serialized via redb through `speedy::{Readable, Writable}` and adding speedy to `xmtp_proto` would balloon the dep tree for a single-crate downstream concern.

crates/xmtp_api/src/mls.rs
xmtp-coder-agent: Added `pub const GroupId::{ZERO, ONE, TWO, THREE, FOUR}` in 346e1c75e (plus a `test_const_helpers` test). This test now reads `let group_id = GroupId::ONE;` — and the trio of paged-read tests is refactored to use `GroupId::generate()` directly (which let me drop the `_clone` / `_clone2` aliases now that `GroupId: Copy`). `build_group_messages` also takes `GroupId` instead of `Vec<u8>` now.

crates/xmtp_api_d14n/src/protocol/utils/test/test_envelope.rs
xmtp-coder-agent: Done — `GroupId::ZERO` added in 346e1c75e. This site is now `.group_id(GroupId::ZERO)`.

crates/xmtp_api_d14n/src/endpoints/v3/mls/query_group_messages.rs
```diff
@@ -67,7 +67,7 @@ mod test {
         let client = NodeGoClient::create();
         let client = client.build().unwrap();
         let mut endpoint = QueryGroupMessages::builder()
-            .group_id(GroupId::from(vec![1, 2, 3]))
+            .group_id(GroupId::from([1u8; 16]))
```
insipx: this hsould be `GroupId::one()`

crates/xmtp_db/src/encrypted_store/migration_test/originator_id_refresh_state.rs
```diff
@@ -142,14 +142,14 @@ async fn up_groups() {
     let db = crate::TestDb::create_database(None).await;
     migrate_before(db.conn(), "2025-08-19-141841_originator_id_groups");
 
-    group(db.conn(), &[1, 2, 3], Some(100));
-    group(db.conn(), &[3, 4, 5], Some(150));
+    group(db.conn(), &[1u8; 16], Some(100));
```
insipx: `GroupId::ONE here and everywhere else (GroupId::TWO, GroupId::THREE etc.)

crates/xmtp_api_d14n/src/endpoints/v3/mls/query_group_messages.rs
xmtp-coder-agent: Done in c5e2ded2d. Note: implemented as `pub const GroupId::ONE` (SCREAMING_SNAKE_CASE per Rust convention), so this line is now `.group_id(GroupId::ONE)`. Also swept 22 sibling `GroupId::from([Nu8; 16])` sites across xmtp_api / xmtp_db / xmtp_mls to the matching `ZERO`/`ONE`/`TWO`/`THREE`/`FOUR` constants. (Self-tests in `group_id.rs` keep the explicit constructor since they verify it.)

crates/xmtp_db/src/encrypted_store/migration_test/originator_id_refresh_state.rs
xmtp-coder-agent: Done in 607ca7b01. Sweep covers this `originator_id_refresh_state` site plus 17 sibling `let group_id: GroupId = [Nu8; 16].into();` patterns in `pending_remove`, `readd_status`, and `group::tests` → the matching `GroupId::ZERO/ONE/TWO/THREE/FOUR` constants. The `group()` helper here stays on `&[u8]` because sibling tests in the same file deliberately pass 3-byte fixtures to exercise pre-migration invalid-length data paths; the `up_groups` callers now read `group(db.conn(), GroupId::ONE.as_slice(), Some(100))` so the constant is still surfaced.

insipx (approved):

insipx:
@xmtp-coder-agent pull in latest and resolve any possible merge conflicts

xmtp-coder-agent:
Merged `origin/main` in 42234d101. Five conflicts (all in xdbg `app/health/*` and `app/generate/messages.rs`) — these were exactly the files where the healthcheck rewrite on main collided with this PR's `GroupId` tightening. Resolved by taking main's structural changes (the `record_generated_message` flow, the `MessageStore`-backed `expected_messages` map, the `is_active`-gated validators, etc.) and re-applying the `GroupId`-typed call signatures so the new code goes through `client.group(&gid)` / `stored_group.group_id()` rather than `client.group(gid.as_slice())` / `GroupId::from(msg.group_id.as_slice())`. Workspace `cargo check --workspace --all-features --all-targets --exclude log_parser` clean; `just lint-rust` clean; 301 hermetic tests pass.

xmtp-coder-agent:
The lint-wasm / lint-config failures should now be resolved by e61fed3f1 (drop redundant `&` on `conversation.group_id` in the wasm hmac_keys paths) and d7b18074e (nix fmt re-flow).

The cross-version-test failure is independent and not from this PR — it's a release/main divergence on xdbg `Identity`:

* `release/v1.9` (tip e18cf2b2b, today 12:02 EDT) and `release/v1.10` (tip 6e522de99) both extend `Identity` with `version_hash: u64` + `version_string: [u8; 64]` and bump the redb namespace `xdbg:1 → xdbg:3`. After that change a v1.9/v1.10 xdbg binary writes 224-byte fixed-width `Identity` rows under the `xdbg:3//identity` table.
* `main` (42c6bd139) still has the 3-field 96-byte `Identity` at `xdbg:1//identity`. `KeyValueStore::load` silently returns `None` when the table type doesn't match, so HEAD's xdbg sees the redb as empty, bootstraps fresh identities, and fails `AddPrimaryToExistingGroups` because no existing client is a member of the v1.9/v1.10 groups (root cause of the cascade of "group with id … not found").

This will also fail on every other PR opened against `main` until that commit is forward-ported. Want me to port it as part of this PR, or leave it for a separate change so this PR stays focused on the `GroupId` refactor? The forward-port is non-trivial — the v1.10 commit also reshapes `IdentityKey` to include the `version_hash` partition and adds an `App::current_version_hash()` plumbing — so I lean toward a separate PR.

Last-passing main run: 25905050159 on 2026-05-15 (before e18cf2b2b landed).

insipx:
@xmtp-coder-agent investigate why the cross version tests are failing. these are passing on main so it must be something in this PR that changed

xmtp-coder-agent:
Took another careful look, and the evidence is unambiguous that this is the `release/v1.9`/`release/v1.10` → `main` divergence — nothing in this PR could cause it. Concrete proof:

1. **Tyler's PR `tyler/concurrent-enable-proposals` fails the exact same way** ([run 25952979816](https://github.com/xmtp/libxmtp/actions/runs/25952979816)), with the same `AddPrimaryToExistingGroups → no active member found for group` cascade. That PR doesn't touch `apps/xmtp_debug/` at all — confirmed with `git diff main..tyler/concurrent-enable-proposals -- apps/xmtp_debug/` (empty). Same for every Tyler PR running cross-version-test since 2026-05-15 19:38 UTC — all fail with this pattern.

2. **No main run since 2026-05-15 07:01 UTC.** `gh run list --workflow=cross-version-test --branch=main` shows the last main run is 25905050159 on 2026-05-15 — three days ago. So "passes on main" is true for the May 15 run, but main hasn't been re-tested since the layout changes landed on release/v1.9.

3. **Timeline of the layout breakage on release/v1.9** (you authored both commits):
   * 2026-05-15 20:24 UTC — `1609211e3` "fix msg query": bumps `Identity` to 4 fields (`+version_hash: u64`) and the redb namespace to `xdbg:2//identity`. Tyler's PRs start failing 75 minutes later, 21:38 UTC.
   * 2026-05-18 16:02 UTC — `e18cf2b2b3` "xdbg: record xdbg version_string on Identity": bumps `Identity` to 5 fields (`+version_string: [u8; 64]`) and namespace to `xdbg:3//identity`. My PR's cross-version-test runs began failing 20 minutes later.

4. **`main`'s `Identity` is still 3 fields at `xdbg:1//identity`** — `git show origin/main:apps/xmtp_debug/src/app/store/identity.rs | head -10` confirms `pub const VERSION: u16 = 1`. So when main HEAD's xdbg (built from this PR) reads the redb file that `release/v1.9` (xdbg:3) wrote, `KeyValueStore::load` hits the silent type-mismatch fallback at `apps/xmtp_debug/src/app/store.rs:316` (`if let Ok(table) = r.open_table(...) { ... } else { Ok(None) }`), reports the identity store as empty, bootstraps fresh identities, and `AddPrimaryToExistingGroups` then has no active member for any of the existing groups (created by the older xdbgs).

5. **This PR's xdbg diff is purely additive on top of main** — `git diff main..HEAD -- apps/xmtp_debug/src/app/types.rs` adds `Group::group_id()` and tweaks the `Display` impl to use it. The `Group` struct layout, the redb `Value` impl, the `Identity` struct, the table NAMESPACE — all unchanged from main.

Recommend triggering cross-version-test on main directly (via workflow_dispatch — I don't have admin rights to do that) to confirm. Until the v1.10 commit's `Identity` layout change is forward-ported to main, the test will fail on every PR.

PR #3647, opened 2026-05-15
clamp min_version <= own version when enabling/bumping proposals

tylerhawkes:
### Clamp min_version to own package version when enabling or bumping proposals
- `enable_proposals` and `update_group_min_version` now reject any `min_version` value greater than the caller's own `CARGO_PKG_VERSION`, returning `GroupError::MinVersionExceedsOwnVersion`.
- `update_group_min_version` also rejects downgrades below the current floor (`MinVersionDowngrade`) and malformed inputs (`InvalidMinVersion`); receive-side commit validation enforces the same monotonicity rules via a new `enforce_min_version_monotonicity` helper in [`validated_commit.rs`](https://github.com/xmtp/libxmtp/pull/3647/files#diff-233994c0ec18d23d0636e849849a3500e44611a4f64c6fffb1d0931aa1461ed6).
- `PROPOSALS_MIN_PROTOCOL_VERSION` is changed from `"1.11.0"` to `"1.11.0-dev"` so the default floor stays within the workspace's own pre-release version.
- `enable_proposals` is now idempotent even when a forward-looking `min_version` is supplied on an already-migrated group, and skips queuing a redundant bootstrap intent if a concurrent migration completed between lock acquisitions.
- Risk: callers that previously passed `min_version` values above the current package version will now receive a non-retryable error instead of succeeding.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3e85f6a.</sup>

### DIFF
--- a/crates/xmtp_configuration/src/common/mls.rs
+++ b/crates/xmtp_configuration/src/common/mls.rs
@@ -51,10 +51,17 @@ pub const MIN_RECOVERY_REQUEST_VERSION: &str = "1.6.0";
 /// uses this value to gate them out of migrated groups before they
 /// fork.
 ///
+/// **Invariant**: must be `<= CARGO_PKG_VERSION`. The send-side clamp
+/// in `enable_proposals` refuses any `min_version > own pkg_version`
+/// (you can't legally write a floor you yourself don't satisfy), so
+/// a constant ahead of the workspace version would brick every
+/// production call to `enable_proposals` that takes the default. Bump
+/// this in lockstep with the workspace version, never independently.
+///
 /// Callers that need a different floor (testing, dev nightlies,
 /// staged rollouts) pass `EnableProposalsOptions::min_version` instead
 /// of relying on this default.
-pub const PROPOSALS_MIN_PROTOCOL_VERSION: &str = "1.11.0";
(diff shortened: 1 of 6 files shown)


insipx (approved):

crates/xmtp_configuration/src/common/mls.rs
```diff
@@ -51,10 +51,17 @@ pub const MIN_RECOVERY_REQUEST_VERSION: &str = "1.6.0";
...
+/// (you can't legally write a floor you yourself don't satisfy), so
+/// a constant ahead of the workspace version would brick every
+/// production call to `enable_proposals` that takes the default. Bump
+/// this in lockstep with the workspace version, never independently.
+///
 /// Callers that need a different floor (testing, dev nightlies,
 /// staged rollouts) pass `EnableProposalsOptions::min_version` instead
 /// of relying on this default.
-pub const PROPOSALS_MIN_PROTOCOL_VERSION: &str = "1.11.0";
+pub const PROPOSALS_MIN_PROTOCOL_VERSION: &str = "1.10.0";
```
tylerhawkes: I would like to have this bumped before we merge to either 1.10.1 or 1.11.0. @cameronvoell Do you know why we don't bump a version until release? It feels like we should be doing 1.11.0-dev.xxxx releases right now and then the actual 1.11.0 release is greater than them now that I have the semver change in main.

PR #3655, opened 2026-05-15
xmtp_mls: telemetry events for AppData migration lifecycle

tylerhawkes:
## Summary
Adds two log events to the \`Event\` enum so a fleet dashboard can observe AppData migration rollout health without scraping ad-hoc tracing strings:

- \`Event::EnableProposalsStart\` — fires once at the top of \`enable_proposals\` after pre-flight parsing. Carries \`group_id\`, \`min_version\`, \`force\`.
- \`Event::EnableProposalsCompleted\` — fires on every success path with \`already_migrated\` distinguishing the no-op fast-path from a real bootstrap commit.

Without these events, "flying blind during rollout" — no way to know what fraction of groups are migrated, which floors were chosen, or whether a regression silently breaks the migration path.

## Test plan
- [x] Existing enable_proposals tests pass (events are no-op observers; their absence in the test fixture doesn't affect assertions).
- [x] cargo clippy -D warnings clean.
- [x] cargo fmt --check clean.

> [!NOTE]
> ### Add telemetry events for the `enable_proposals` AppData migration lifecycle
> Adds two new event variants, `EnableProposalsStart` and `EnableProposalsCompleted`, to the [`Event` enum](https://github.com/xmtp/libxmtp/pull/3655/files#diff-b892bd1dce497466fd9fa05752e7d5858b05185df4cfbc2acd1e03982cfcca14). Emits these events in [`Group::enable_proposals`](https://github.com/xmtp/libxmtp/pull/3655/files#diff-c27dd00cf700fd888b75fdbd19738462c2549fdc881dad70f8e61c979d440966) at the start of the call and at each exit path, including an `already_migrated` flag to distinguish no-op calls from those that publish a bootstrap commit.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9d5a864.</sup>
> 
>

### DIFF
--- a/crates/xmtp_common/src/event_logging.rs
+++ b/crates/xmtp_common/src/event_logging.rs
@@ -165,4 +165,15 @@ pub enum Event {
     /// Cannot send sync archive. No server_url present.
     #[context(pin)]
     DeviceSyncNoServerUrl,
+
+    // ===================== AppData Migration =====================
+    /// `enable_proposals` started — pre-flight passed, about to publish
+    /// the legacy-GMM bump (step A) and/or bootstrap commit (step B).
+    #[context(group_id, min_version, force, icon = "🌱")]
+    EnableProposalsStart,
+    /// `enable_proposals` completed. `already_migrated = true` means
+    /// the call was a no-op fast-path; `false` means a bootstrap
+    /// commit was actually published.
+    #[context(group_id, already_migrated, min_version, icon = "🌳")]
+    EnableProposalsCompleted,
 }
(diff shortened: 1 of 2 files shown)


insipx:
Do we have a dashboard setup for these events? Afaiu they only work with the log parser . For telemetry we could probably just setup a open telemetry subscriber, since each root action should have a span now

tylerhawkes:
> Do we have a dashboard setup for these events? Afaiu they only work with the log parser . For telemetry we could probably just setup a open telemetry subscriber, since each root action should have a span now

We don't. They only work with the log parser as far as I'm aware, but they will give operators who have logs insight into how many groups have made the switch.

PR #3658, opened 2026-05-18
xmtp_mls: add Client::close for coordinated shutdown (PR1/3)

mkysel:
Part of #3659 (PR1 of 3).

## Review feedback addressed (most recent push)

- **@insipx — workers should be force-killable, not cooperative**: each worker future now wrapped in \`futures::future::Abortable\`; \`WorkerRunner::shutdown\` aborts each handle before draining the supervisor. Cancellation token stays as the cooperative-shutdown signal; Abortable is the hard kill for workers whose loops don't yield promptly (e.g. ones sitting on long interval sleeps).
- **Macroscope idempotency bug**: \`Client::close\` now gates its early-return on a new \`shutdown_complete\` flag set only after \`disconnect()\` succeeds, not on \`is_cancelled()\`. A mid-close disconnect failure leaves \`shutdown_complete\` false so a retry actually re-attempts the disconnect. Distinct semantic from Tyler's removed-atomic case (cancellation tracking) — this atomic tracks fully-landed teardown.
- **Claude bot — missing on_close in \`stream_message_deletions_with_callback\`**: the one variant of five that wasn't firing \`on_close\` after its loop. Now parameterized like the others; binding sites pass \`|| {}\` for now (real handlers wire through in PR2/PR3).

## Summary

First of three PRs adding \`Client::close()\` to libxmtp. This PR lands the core plumbing in \`xmtp_mls\`; no public binding surface change.

- \`CancellationToken\` on \`XmtpMlsLocalContext\`, exposed via \`XmtpSharedContext\`. \`is_closed()\` delegates to \`cancellation_token().is_cancelled()\`.
- \`shutdown_complete\` \`AtomicBool\` on the context — distinct semantic from cancellation; set only after \`disconnect()\` succeeds, used as \`close()\`'s idempotency check.
- Each worker future wrapped in \`Abortable\`. \`WorkerRunner::shutdown\` aborts every worker and drains the supervisor under a 2s timeout.
- Cancellation token also races each \`stream_*_with_callback\` inner loop, the \`cleanup_duplicate_updates\` spawn, and each worker's outer loop (cooperative path).
- Idempotent \`Client::close(&self) -> Result<(), ClientError>\` and \`ClientError::AlreadyClosed\`. \`reconnect_db\` refuses after close; \`release_db_connection\` is unchanged for callers that want release-then-reconnect cycles.
- New lifecycle \`Event::ClientClosed\` mirroring \`ClientDropped\`.

## Why

Full consumer ask is in #3659. Short version: the Convos / Herald team needs **an awaitable shutdown handshake** for their Node SDK client. They need to:

1. Stop accepting new work.
2. Wait until in-flight workers and streams have actually stopped.
3. Release the DB pool.
4. Get a single \`Ok(())\` back so they can confidently delete the SQLite file and drop their JS references.

\`Drop\` cannot deliver that contract. \`Drop::drop\` is sync, so it can signal workers to stop but cannot \`.await\` their completion. JS finalizers (napi) are also non-deterministic — the JS host decides when to run them. Even a perfect Drop story leaves Herald guessing about *when* teardown is complete, which is the exact thing they're trying to stop guessing about.

\`await client.close()\` provides that signal. Drop-based cleanup remains the fallback for callers that forget; \`close()\` is the contract for callers that care.

(The log spam Herald sees today is one downstream symptom of not having this handshake — workers and streams fire against the disconnected DB before they notice. Fixing log noise alone via per-worker Drop hygiene or finer-grained log levels would help, but neither gives Herald the awaitable signal they actually need.)

PR2 will expose \`close()\` on \`bindings/node\`; PR3 on \`bindings/mobile\` (uniffi).

## Notable change for reviewers

\`Worker::spawn\` (trait method in \`crates/xmtp_mls/src/worker.rs\`) takes a \`CancellationToken\` parameter. Internal trait, no external surface change. The new Abortable wrapping is purely on the supervisor side and doesn't change the trait signature again.

## Test plan

- [x] \`cargo nextest run -p xmtp_mls\` — 574 tests pass (2 flakes pre-existing on main). Five new tests: \`close_stops_workers\`, \`close_is_idempotent\`, \`close_disconnects_db\`, \`close_cancels_callback_stream\`, \`reconnect_after_close_errors\`.
- [x] \`cargo fmt --check\`, \`cargo clippy -Dwarnings\`, \`cargo hakari generate --diff\` — all clean.
- [x] \`test-wasm\` CI — passed, confirming \`tokio_util::sync::CancellationToken\` works on \`wasm32-unknown-unknown\`.
- [ ] \`just lint\` in nix shell (needs reviewer verification).

> [!NOTE]
> Part of #3659 (PR1 of 3).
>
> ## Review feedback addressed (most recent push)
>
> - **@insipx — workers should be force-killable, not cooperative**: each worker future now wrapped in \`futures::future::Abortable\`; \`WorkerRunner::shutdown\` aborts each handle before draining the supervisor. Cancellation token stays as the cooperative-shutdown signal; Abortable is the hard kill for workers whose loops don't yield promptly (e.g. ones sitting on long interval sleeps).
> - **Macroscope idempotency bug**: \`Client::close\` now gates its early-return on a new \`shutdown_complete\` flag set only after \`disconnect()\` succeeds, not on \`is_cancelled()\`. A mid-close disconnect failure leaves \`shutdown_complete\` false so a retry actually re-attempts the disconnect. Distinct semantic from Tyler's removed-atomic case (cancellation tracking) — this atomic tracks fully-landed teardown.
> - **Claude bot — missing on_close in \`stream_message_deletions_with_callback\`**: the one variant of five that wasn't firing \`on_close\` after its loop. Now parameterized like the others; binding sites pass \`|| {}\` for now (real handlers wire through in PR2/PR3).
>
> ## Summary
>
> First of three PRs adding \`Client::close()\` to libxmtp. This PR lands the core plumbing in \`xmtp_mls\`; no public binding surface change.
>
> - \`CancellationToken\` on \`XmtpMlsLocalContext\`, exposed via \`XmtpSharedContext\`. \`is_closed()\` delegates to \`cancellation_token().is_cancelled()\`.
> - \`shutdown_complete\` \`AtomicBool\` on the context — distinct semantic from cancellation; set only after \`disconnect()\` succeeds, used as \`close()\`'s idempotency check.
> - Each worker future wrapped in \`Abortable\`. \`WorkerRunner::shutdown\` aborts every worker and drains the supervisor under a 2s timeout.
> - Cancellation token also races each \`stream_*_with_callback\` inner loop, the \`cleanup_duplicate_updates\` spawn, and each worker's outer loop (cooperative path).
> - Idempotent \`Client::close(&self) -> Result<(), ClientError>\` and \`ClientError::AlreadyClosed\`. \`reconnect_db\` refuses after close; \`release_db_connection\` is unchanged for callers that want release-then-reconnect cycles.
> - New lifecycle \`Event::ClientClosed\` mirroring \`ClientDropped\`.
>
> ## Why
>
> Full consumer ask is in #3659. Short version: the Convos / Herald team needs **an awaitable shutdown handshake** for their Node SDK client. They need to:
>
> 1. Stop accepting new work.
> 2. Wait until in-flight workers and streams have actually stopped.
> 3. Release the DB pool.
> 4. Get a single \`Ok(())\` back so they can confidently delete the SQLite file and drop their JS references.
>
> \`Drop\` cannot deliver that contract. \`Drop::drop\` is sync, so it can signal workers to stop but cannot \`.await\` their completion. JS finalizers (napi) are also non-deterministic — the JS host decides when to run them. Even a perfect Drop story leaves Herald guessing about *when* teardown is complete, which is the exact thing they're trying to stop guessing about.
>
> \`await client.close()\` provides that signal. Drop-based cleanup remains the fallback for callers that forget; \`close()\` is the contract for callers that care.
>
> (The log spam Herald sees today is one downstream symptom of not having this handshake — workers and streams fire against the disconnected DB before they notice. Fixing log noise alone via per-worker Drop hygiene or finer-grained log levels would help, but neither gives Herald the awaitable signal they actually need.)
>
> PR2 will expose \`close()\` on \`bindings/node\`; PR3 on \`bindings/mobile\` (uniffi).
>
> ## Notable change for reviewers
>
> \`Worker::spawn\` (trait method in \`crates/xmtp_mls/src/worker.rs\`) takes a \`CancellationToken\` parameter. Internal trait, no external surface change. The new Abortable wrapping is purely on the supervisor side and doesn't change the trait signature again.
>
> ## Test plan
>
> - [x] \`cargo nextest run -p xmtp_mls\` — 574 tests pass (2 flakes pre-existing on main). Five new tests: \`close_stops_workers\`, \`close_is_idempotent\`, \`close_disconnects_db\`, \`close_cancels_callback_stream\`, \`reconnect_after_close_errors\`.
> - [x] \`cargo fmt --check\`, \`cargo clippy -Dwarnings\`, \`cargo hakari generate --diff\` — all clean.
> - [x] \`test-wasm\` CI — passed, confirming \`tokio_util::sync::CancellationToken\` works on \`wasm32-unknown-unknown\`.
> - [ ] \`just lint\` in nix shell (needs reviewer verification).
>
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)
> #### Changes since #3658 opened
>
> - Added force-abort capability to `WorkerRunner` using `AbortHandle` and `Abortable` futures [639c009]
> - Modified `Client::close` idempotency logic to use `shutdown_complete` flag instead of cancellation token [639c009]
> - Added `shutdown_complete` tracking infrastructure to `XmtpMlsLocalContext` and `XmtpSharedContext` trait [639c009]
> - Added `on_close` callback parameter to `stream_message_deletions_with_callback` methods across all bindings and core subscription system [639c009]
> - Implemented on-close callback invocation for `Conversations::stream_message_deletions` method in the `xmtp_mls` wasm bindings [b1dc195]
> - Reordered `AtomicBool` import statement [3f4728f]
> 
>

### DIFF
--- a/crates/xmtp_mls/src/test/mock.rs
+++ b/crates/xmtp_mls/src/test/mock.rs
@@ -93,6 +93,7 @@ impl Clone for NewMockContext {
             task_channels: self.task_channels.clone(),
             worker_metrics: self.worker_metrics.clone(),
             cancellation_token: self.cancellation_token.clone(),
+            shutdown_complete: self.shutdown_complete.clone(),
         }
     }
 }
@@ -174,5 +175,15 @@ impl XmtpSharedContext for NewMockContext {
 
     fn cancellation_token(&self) -> &CancellationToken {
         &self.cancellation_token
+    }
+
+    fn shutdown_complete(&self) -> bool {
+        self.shutdown_complete
+            .load(std::sync::atomic::Ordering::Acquire)
+    }
+
+    fn mark_shutdown_complete(&self) {
+        self.shutdown_complete
+            .store(true, std::sync::atomic::Ordering::Release);
     }
 }
--- a/crates/xmtp_mls/src/test/mock.rs
+++ b/crates/xmtp_mls/src/test/mock.rs
@@ -178,12 +178,10 @@ impl XmtpSharedContext for NewMockContext {
     }
 
     fn is_closed(&self) -> bool {
-        self.closed
-            .load(std::sync::atomic::Ordering::Acquire)
+        self.closed.load(std::sync::atomic::Ordering::Acquire)
     }
 
     fn mark_closed(&self) -> bool {
-        self.closed
-            .swap(true, std::sync::atomic::Ordering::AcqRel)
+        self.closed.swap(true, std::sync::atomic::Ordering::AcqRel)
     }
 }
--- a/crates/xmtp_mls/src/test/mock.rs
+++ b/crates/xmtp_mls/src/test/mock.rs
@@ -21,6 +21,7 @@ use crate::{
 use alloy::signers::local::PrivateKeySigner;
 use mockall::mock;
 use tokio::sync::broadcast;
+use tokio_util::sync::CancellationToken;
 use xmtp_api::ApiClientWrapper;
 use xmtp_api_d14n::MockApiClient;
 use xmtp_cryptography::XmtpInstallationCredential;
@@ -91,6 +92,8 @@ impl Clone for NewMockContext {
             fork_recovery_opts: self.fork_recovery_opts.clone(),
             task_channels: self.task_channels.clone(),
             worker_metrics: self.worker_metrics.clone(),
+            cancellation_token: self.cancellation_token.clone(),
+            closed: self.closed.clone(),
         }
     }
 }
@@ -168,5 +171,19 @@ impl XmtpSharedContext for NewMockContext {
 
     fn sync_api(&self) -> &ApiClientWrapper<Self::ApiClient> {
         &self.sync_api_client
+    }
+
+    fn cancellation_token(&self) -> &CancellationToken {
+        &self.cancellation_token
+    }
+
+    fn is_closed(&self) -> bool {
+        self.closed
+            .load(std::sync::atomic::Ordering::Acquire)
+    }
+
+    fn mark_closed(&self) -> bool {
+        self.closed
+            .swap(true, std::sync::atomic::Ordering::AcqRel)
     }
 }
--- a/crates/xmtp_mls/src/test/mock.rs
+++ b/crates/xmtp_mls/src/test/mock.rs
@@ -93,7 +93,6 @@ impl Clone for NewMockContext {
             task_channels: self.task_channels.clone(),
             worker_metrics: self.worker_metrics.clone(),
             cancellation_token: self.cancellation_token.clone(),
-            closed: self.closed.clone(),
         }
     }
 }
@@ -175,13 +174,5 @@ impl XmtpSharedContext for NewMockContext {
 
     fn cancellation_token(&self) -> &CancellationToken {
         &self.cancellation_token
-    }
-
-    fn is_closed(&self) -> bool {
-        self.closed.load(std::sync::atomic::Ordering::Acquire)
-    }
-
-    fn mark_closed(&self) -> bool {
-        self.closed.swap(true, std::sync::atomic::Ordering::AcqRel)
     }
 }
--- a/crates/xmtp_mls/src/worker.rs
+++ b/crates/xmtp_mls/src/worker.rs
@@ -14,10 +14,16 @@ use std::fmt::Debug;
 use std::pin::Pin;
 use std::{any::Any, collections::HashMap, hash::Hash, sync::Arc};
 use tasks::TaskWorkerChannels;
+use tokio_util::sync::CancellationToken;
 use tracing::Instrument;
 use tracing::instrument::Instrumented;
 use xmtp_common::{MaybeSend, MaybeSync, StreamHandle, if_native, if_wasm, time::Duration};
 use xmtp_configuration::WORKER_RESTART_DELAY;
+
+/// Hard cap on how long `WorkerRunner::shutdown` waits for the supervisor
+/// task to drain after cancellation. Anything beyond this gets logged and
+/// the task is allowed to detach — keeps `Client::close` bounded.
+const WORKER_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(2);
 
 #[derive(PartialEq, Eq, Copy, Clone, Hash, Debug)]
 pub enum WorkerKind {
@@ -89,6 +95,7 @@ impl WorkerRunner {
         }
 
         let this = self.clone();
+        let cancel = ctx.cancellation_token().clone();
         let handle = xmtp_common::spawn(
             None,
             async move {
@@ -110,7 +117,7 @@ impl WorkerRunner {
                         let mut m = this.metrics.lock();
                         m.insert(worker.kind(), metrics);
                     }
-                    futs.push(worker.spawn());
+                    futs.push(worker.spawn(cancel.clone()));
                 }
 
                 while let Some(kind) = futs.next().await {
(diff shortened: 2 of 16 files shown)


mkysel:
Part of #3659 (this is PR1 of 3). See the issue for the consumer ask and the full 3-PR rollout plan.

neekolas:
Think @insipx and @tylerhawkes are the best to review this workstream

crates/xmtp_mls/src/test/mock.rs
```diff
@@ -91,6 +92,8 @@ impl Clone for NewMockContext {
             fork_recovery_opts: self.fork_recovery_opts.clone(),
             task_channels: self.task_channels.clone(),
             worker_metrics: self.worker_metrics.clone(),
+            cancellation_token: self.cancellation_token.clone(),
```
tylerhawkes, line 95: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.is_cancelled there is a method on the cancellation tokent hat does what the closed Arc is being used for.

tylerhawkes (dismissed):
The `is_closed` and `mark_closed` functions are redundant if we can get by without the need for the atomic. The cancellation_token will return true for `is_closed` once the function to cancel it returns, so that should be good enough.

insipx:
shouldn't the workers close on drop? If that's not happening this is a bandaid over a bigger issue

Is there a specific worker that is causing more logs than usual on drop? there's also a new `set_native_libxmtp` log level that isn't exposed in node yet, we should get that there to allow finer-grained log control. Yesterday a quick look at `PendingSelfRemoveWorker` seems to not respect the db error drop as well and may stay alive. do you remember if you see logs from that worker?

Edit:
-- sorry fatfingered close PR by accident

mkysel:
@insipx 

Logs aren't the actual ask — sorry, the PR body led with that and it was misleading. (Just updated it.) Herald's real need is an awaitable shutdown handshake: they want to await client.close() and have it return only after workers/streams have actually stopped and the DB pool is released, so @neekolas  can confidently delete the SQLite file and drop his JS references. Drop can't deliver that — Drop::drop is sync so it can signal but not .await, and JS finalizers (napi) are non-deterministic on top. Even perfect Drop hygiene on every worker leaves consumers guessing about when teardown is done, which is the thing they're trying to stop guessing about.

So workers-closing-on-drop is necessary but not sufficient. This method still earns its keep by giving callers a deterministic "fully torn down" signal that they can await.

crates/xmtp_mls/src/worker.rs
```diff
@@ -167,24 +199,40 @@ pub trait Worker: MaybeSend + MaybeSync + 'static {
...
+                            xmtp_common::time::sleep(WORKER_RESTART_DELAY).await;
+                            tracing::info!("Restarting {:?} worker...", self.kind());
+                        }
                     }
                 }
+                self.kind()
+            };
+
+            tokio::select! {
+                k = run => k,
```
insipx, line 255: This is still dependent on workers respecting the loop here & the db error, which i'm not sure all do yet + according to PR we want to kill workers before the DB. I think we can do better

ext: [this worker runs on an interval](https://github.com/xmtp/libxmtp/blob/2a8b1ee03b61242aa2eeeb5d6613e0f91a67edbb/crates/xmtp_mls/src/worker/pending_self_remove.rs#L98), and should only yield to this function only if it hits a specific db reconnect error, otherwise it sleeps. that db error will only fire once the db conn is released

to 100% ensure the task is killed as fast as possible independant of the db Pool, we could wrap the run future in [Abortable](https://docs.rs/futures/latest/futures/future/struct.Abortable.html#method.new) and call handle.abort() on the cancellation signal

mkysel: addressed

crates/xmtp_mls/src/worker.rs
```diff
@@ -110,11 +136,17 @@ impl WorkerRunner {
                         let mut m = this.metrics.lock();
                         m.insert(worker.kind(), metrics);
                     }
-                    futs.push(worker.spawn());
+                    let (abort_handle, reg) = AbortHandle::new_pair();
+                    new_handles.push(abort_handle);
+                    futs.push(Abortable::new(worker.spawn(cancel.clone()), reg));
```
tylerhawkes, line 141: Instead of spawning workers, we could use a futures unordered which would keep all the workers in a single task that could then manage the cancellation in a single place.

There's probably more to it than that, but it should be feasible as long as none of the workers have long cpu critical sections.

insipx: they're already in futures unordered (line 124), the `spawn` here is misleading, probably b/c it used to spawn, but now `spawn` just returns a Boxed future polled by the single detached spawn orchestrated at line 92.

crates/xmtp_mls/src/worker.rs
```diff
@@ -110,11 +136,17 @@ impl WorkerRunner {
...
                 }
+                *this.abort_handles.lock() = new_handles;
 
-                while let Some(kind) = futs.next().await {
-                    tracing::warn!("Worker {kind:?} completed unexpectedly")
+                while let Some(outcome) = futs.next().await {
```
insipx, line 145: sorry for churn, but rather than this i was thinking we keep the cancellation token here (I think that's what previous iteration had?) then iterate through the abort handles when we select cancellation token path, then can abort each handle here before ending the task its nested in. that would avoid a mutex over abort handles and keep logic central.

insipx (approved):

PR #3693, opened 2026-05-29
Reduce log noise: drop steady-state logs + remove API stats-on-error spam

insipx:
Two focused changes to cut production log volume. Driven by an audit of a 1000-line VictoriaLogs sample from 5 herald instances, where **53% of lines were TRACE** and a single steady-state message accounted for **~40% of all log lines**.

## Commit 1 — `chore(logs): remove steady-state log noise`
Deletes logs that fire on every poll/tick when *nothing happened* (the largest volume sources):

- **key_package_cleaner**: drop per-poll `"No expired key packages to delete"` trace (~40% of sampled prod lines).
- **device_sync worker**: drop the 20s `"New event: Tick"` heartbeat trace; log only real events, with `installation_id` as a structured field (not a `[..]` message prefix).
- **sql_key_store**: drop the bare `"own_leaf_nodes"` trace (fires on every leaf-node read).
- **worker metrics**: downgrade per-increment `"firing {metric}"` from `info` → `trace`.

## Commit 2 — `refactor(api): remove ApiDebugWrapper stats-on-error spam`
On **every** API error, `ApiDebugWrapper` snapshotted all per-endpoint request counters into `AggregateStats` and embedded its multi-line `Debug` dump into `ApiClientError::{ErrorWithStats, ClientWithEndpointAndStats}` — which then got logged, and re-logged on each retry, as a verbose stats block.

This removes the error-dump path so API errors are clean one-liners:
- Delete `ApiDebugWrapper` and its module re-export.
- Remove `ClientBuilder::enable_api_debug_wrapper` and drop the call from the mobile/node/wasm client builders.
- `WrappedXmtpApiClient` is now `TrackedStatsClient<XmtpApiClient>`.
- Remove the now-unused `ErrorWithStats` / `ClientWithEndpointAndStats` variants and their match arms.

**Kept unchanged:** the `TrackedStatsClient` request counters and the FFI `api_statistics()` / `api_aggregate_statistics()` / `clear_all_statistics()` surface — only the error-message spam is removed.

## Verification
- `cargo check` + `cargo clippy` clean on `xmtp_proto`, `xmtp_api`, `xmtp_mls`, `xmtp_db` (the only warning, `identity.rs:842` unused `mut`, is pre-existing and unrelated).
- `just node check` and `just wasm check` both pass after the binding edits.

## Follow-ups (not in this PR)
- Replace the in-RAM `TrackedStatsClient` counters with OTEL metrics (per-API-call latency/rate at the retry combinator), then retire the stats layer once the FFI is migrated.
- Broader log standardization + OTEL traces/metrics rollout (separate audit docs exist locally).

> [!NOTE]
> ### Remove API debug wrapper and reduce steady-state log noise across MLS workers
> - Deletes the `ApiDebugWrapper` and `enable_api_debug_wrapper` builder method; the `WrappedXmtpApiClient` type alias now resolves to `TrackedStatsClient<XmtpApiClient>` without the debug layer.
> - Removes `ClientWithEndpointAndStats` and `ErrorWithStats` variants from `ApiClientError`, so errors no longer carry embedded aggregate stats.
> - Suppresses high-frequency log output: tick events in `DeviceSyncWorker`, no-op key package cleaner runs, repeated DB connection-outage errors, and per-metric info logs are all silenced or downgraded to trace.
> - Behavioral Change: API errors previously wrapped by the debug adapter are now surfaced as plain `ClientWithEndpoint` or `Client` variants with no attached stats.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 480bc19.</sup>
> 
>


crates/xmtp_mls/src/worker/metrics.rs
```diff
@@ -132,7 +132,7 @@ where
 
     pub(crate) fn increment_metric(&self, metric: Metric) {
         self.info(metric).increment();
-        tracing::info!("[{}] firing {metric:?}", hex::encode(self.installation_id));
```
insipx, line 135: these worker metrics could just become OTEL exports, but we still depend on some of these for tests

crates/xmtp_proto/src/traits/error.rs
```diff
@@ -1,29 +1,12 @@
...
-use crate::api_client::AggregateStats;
 use crate::{ApiEndpoint, ProtoError};
 use thiserror::Error;
 use xmtp_common::{BoxDynError, RetryableError, retryable};
 
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum ApiClientError {
-    #[error(
-        "api client at endpoint \"{}\" has error {}. \n {:?} \n",
```
insipx, line 12: these weren't really super useful as a log line. its also lends itself much better to OTEL export

PR #3694, opened 2026-05-29
Dedupe DB pool-failure error spam across workers

insipx:
When the SQLite connection pool cannot hand out a connection (exhausted / `timed out waiting for connection` / `Unable to open the database file`), a **single** root cause was logged dozens of times: each background worker logged the error itself and then propagated it to the supervisor, which logged it again on every poll/restart. In a production sample this turned one pool outage into a flood of duplicate `ERROR` lines.

## Approach: the supervisor is the single logger for propagated worker errors
The worker supervisor already receives every error a worker returns and logs it with context (`{kind} worker error: {err}`), and routes pool failures to a quiet reconnect. So workers should **not** log errors they propagate.

- **Remove the per-worker error logs** that just logged-then-propagated (`disappearing_messages`, `pending_self_remove` outer fetch, `key_package_cleaner` fetch). The supervisor is now the sole source. Per-group *swallow-and-continue* logs (e.g. one malformed `group_id`) are kept, since those never reach the supervisor.
- **Fix a latent bug in `key_package_cleaner`**: it previously *logged and swallowed* a fetch error (returned `Ok`), so a pool outage was logged every 5s and never triggered the supervisor’s reconnect/restart path (silent infinite retry). It now propagates the error like the other workers.
- **Classify pool-acquisition failures as needing a reconnect**: `StorageError::db_needs_connection()` now matches `PlatformStorageError::Pool(_)` (the r2d2 acquisition/timeout error) in addition to `PoolNeedsConnection`, so the supervisor takes the quiet reconnect path for a pool outage instead of the noisy error+restart path. This is the one place the predicate is consulted.
- **Drop a redundant log**: `"Key package deletion successful"` immediately followed `"Deleted {N} expired key packages (up to ID …)"` — same meaning, removed.

## Tests
- `pool_errors_need_connection` — unit test for the `StorageError::db_needs_connection()` classification (`PoolNeedsConnection` directly and wrapped in `Connection`; a non-pool error is not). The `Pool(_)` arm can’t be unit-constructed (`r2d2::PoolError` has no public ctor) so it’s covered end-to-end below.
- `pool_failure_needs_connection` — end-to-end: disconnect a persistent store’s pool, run a real query, assert the resulting error reports `db_needs_connection()`.

## Incidental CI fix
`xmtp_db`’s `test-utils` feature did not forward `xmtp_proto/test-utils`, so `cargo test -p xmtp_db` only compiled under workspace-wide feature unification (which CI uses). This masked a broken standalone per-crate test build. Added the missing feature forward.

## Verification
- `cargo clippy --locked --all-features --all-targets --no-deps -- -Dwarnings` ✓
- `cargo fmt --check` ✓ · `cargo hakari generate --diff` ✓
- Both new tests pass.

> [!NOTE]
> ### Deduplicate DB pool-failure error spam across MLS workers
> - Extends `StorageError::db_needs_connection()` and `ConnectionError::db_needs_connection()` to classify generic `Pool(_)` acquisition failures (not just `PoolNeedsConnection`) as needing a reconnect.
> - Implements `NeedsDbReconnect` for `MlsStoreError` so worker restart logic can inspect wrapped DB errors.
> - Refactors `DisappearingMessagesCleaner`, `KeyPackagesCleaner`, and `PendingSelfRemoveWorker` to propagate structured errors instead of logging and swallowing them, so pool failures surface once via the reconnect path rather than flooding logs.
> - Behavioral Change: workers now abort and return an error on the first DB failure per cycle instead of continuing past errors; this reduces log spam but changes per-cycle recovery behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ffe1014.</sup>
> 
>

### DIFF
--- a/crates/xmtp_db/Cargo.toml
+++ b/crates/xmtp_db/Cargo.toml
@@ -96,4 +96,5 @@ test-utils = [
   "dep:ascii_table",
   "xmtp_configuration/test-utils",
   "xmtp_cryptography/test-utils",
+  "xmtp_proto/test-utils",
 ]
--- a/crates/xmtp_mls/src/worker/key_package_cleaner.rs
+++ b/crates/xmtp_mls/src/worker/key_package_cleaner.rs
@@ -52,6 +52,10 @@ pub enum KeyPackagesCleanerError {
     Identity(#[from] IdentityError),
     #[error("metadata error: {0}")]
     Metadata(StorageError),
+    #[error("failed to fetch expired key packages: {0}")]
+    Fetch(StorageError),
+    #[error("failed to delete key package: {0}")]
+    DeleteKeyPackage(IdentityError),
     #[error("deletion error: {0}")]
     Deletion(StorageError),
     #[error("rotation error: {0}")]
@@ -64,6 +68,8 @@ impl NeedsDbReconnect for KeyPackagesCleanerError {
             Self::Storage(s) => s.db_needs_connection(),
             Self::Identity(s) => s.needs_db_reconnect(),
             Self::Metadata(s) => s.db_needs_connection(),
+            Self::Fetch(s) => s.db_needs_connection(),
+            Self::DeleteKeyPackage(s) => s.needs_db_reconnect(),
             Self::Deletion(s) => s.db_needs_connection(),
             Self::Rotation(s) => s.needs_db_reconnect(),
         }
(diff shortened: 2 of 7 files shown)


crates/xmtp_db/Cargo.toml
```diff
@@ -96,4 +96,9 @@ test-utils = [
   "dep:ascii_table",
   "xmtp_configuration/test-utils",
   "xmtp_cryptography/test-utils",
+  # Tests use xmtp_proto test helpers (e.g. `GroupId::generate`), which live behind
```
insipx: get rid of these long comments, pretty useless

crates/xmtp_mls/src/worker/pending_self_remove.rs
```diff
@@ -152,11 +154,13 @@ where
                 }
             }
             Err(e) => {
-                tracing::error!(
-                    "Failed to get groups with pending leave requests, error: {:?}",
-                    e
-                );
-                return Err(PendingSelfRemoveWorkerError::Storage(e.into()));
+                // Propagate to the worker supervisor, which is the single place that logs
```
insipx: make comments about 2 lines max

crates/xmtp_db/Cargo.toml
insipx: Done — removed the comment entirely (the feature line is self-explanatory). 166bdef

crates/xmtp_mls/src/worker/pending_self_remove.rs
insipx: Done — trimmed to one line here and across the other workers' propagation/classification comments. 166bdef

crates/xmtp_mls/src/worker/key_package_cleaner.rs
```diff
@@ -145,45 +148,41 @@ where
...
+            .get_expired_key_packages()
+            .map_err(KeyPackagesCleanerError::Fetch)?;
+        if expired_kps.is_empty() {
+            return Ok(());
+        }
+
+        tracing::info!("Deleting {} expired key packages", expired_kps.len());
+        // Delete from local db
+        for kp in &expired_kps {
+            if let Err(err) = self.delete_key_package(
```
insipx: This is something that should be turned into an error variant and propagated rather than a manual match

insipx: Done — added a `DeleteKeyPackage(IdentityError)` variant and replaced the manual match with `.map_err(KeyPackagesCleanerError::DeleteKeyPackage)?`. This removes the per-KP log and propagates to the supervisor (the sole logger). Note: this also changes behavior from best-effort-continue to abort-and-restart on the first failed delete — which is safer, since a failed local delete shouldn't be followed by the DB batch `delete…up_to_id`.

crates/xmtp_mls/src/worker/pending_self_remove.rs
insipx: rather than match using ? would cleanup some code. if returning anyway matching is redundant

insipx: can use `?` here too. returning an error will just repeat the pending remove on the next turn. we should prefer ? here to matching on the errors.

in the future the db could just do one query to do the action in one batch if perf is concerning.

insipx: Done — collapsed the outer match to `?` via the `GetPendingLeaveGroups` variant.

insipx: Done — both the group load and `react_to_…` now use `?`. Added a `LoadGroup(#[from] MlsStoreError)` variant and implemented `NeedsDbReconnect` for `MlsStoreError` so pool failures still route to the quiet reconnect. The whole loop is now propagate-and-retry-next-turn; removed the two inline error logs (supervisor is the sole logger).

insipx:
Converted `react_to_group_has_pending_leave_request` to a span (`ffe1014`): replaced the bracketing `Processing…`/`Completed…` info logs with `#[tracing::instrument(skip_all, fields(group_id = %mls_group.group_id), name = "process_pending_leave_requests")]`. The span carries `group_id` (the old `Completed…` log was missing it) and its enter/exit mark the begin/end of the unit of work. Used the attribute macro rather than `Span::enter` since it's async.

PR #3695, opened 2026-05-29
Propagate DB-disconnect errors out of workers to the supervisor

insipx:
## Summary

Workers in `xmtp_mls` run under `WorkerRunner`. Its outer loop (`Worker::spawn`) stops a worker **only** when an error returned from `run_tasks` reports `needs_db_reconnect() == true` — that's the signal that the connection pool was dropped and the worker should stop until reconnect.

Several workers violated this contract: they caught per-item errors with `tracing::error!` and continued, so a dropped pool could never reach the supervisor and the worker hot-looped against a gone pool. The signal was also silently lost in wrapper error types that mapped `GroupError(_) => false`.

This cleans the workers up to follow the contract, using the recently-cleaned `disappearing_messages` worker as the model.

## Root fix (error-type forwarding)

- `ConnectionError` gets a `db_needs_connection()` helper mirroring `StorageError`.
- `GroupError`, `MlsStoreError`, and `SubscribeError` now implement `NeedsDbReconnect`, forwarding their storage / connection / client / wrapped variants.
- `DeviceSyncError::needs_db_reconnect` now forwards `Db` / `Group` / `MlsStore` / `Subscribe` instead of only `Client`.
- `CommitLogError::needs_db_reconnect` forwards `GroupError` and narrows `Connection` to `db_needs_connection()` (was unconditionally `true`); removes the stale TODOs.

## Workers

- **key_package_cleaner** — propagate the expired-KP fetch error (was swallowed → `Ok`); bubble a disconnect caught while deleting a single key package.
- **pending_self_remove** — capture the group-load (`MlsStoreError`) path (previously never converted into the worker error) and bubble a disconnect from either per-group operation; fix `GroupError(_) => false`.
- **commit_log** — bubble a disconnect from the readd-request and readd-validation loops.
- **device_sync** — bubble a disconnect from the per-message processing loop.

Per-item **non-DB** failures still log and continue, as before — only a dropped pool now stops the worker.

## Other

- Removes the dead `WorkerKind::Event` variant (no impl, no construction, not in bindings/FFI/serde).
- Adds native-only unit tests pinning the disconnect-forwarding contract for every worker error type (`disconnect_propagation_tests`).

## Test plan

- `cargo nextest run -p xmtp_mls disconnect_propagation` — 7/7 pass.
- `cargo check -p xmtp_mls --tests` clean.
- `just lint-rust` clean (clippy / fmt / hakari).

> [!NOTE]
> ### Propagate DB-disconnect errors out of workers to the supervisor
> - Adds or fixes `needs_db_reconnect()` implementations across multiple error types (`GroupError`, `ClientError`, `IdentityError`, `SubscribeError`, `DeviceSyncError`, `PendingSelfRemoveWorkerError`) so dropped-pool signals are forwarded up the call stack rather than swallowed.
> - Errors in `publish_commit_logs_to_remote` and `prepare_publish_commit_log_info` now propagate instead of being logged and ignored, so cursors only advance on success.
> - `send_outgoing_readd_requests` now fails the operation when any readd request fails, instead of logging and continuing.
> - `handle_incoming_pending_readds` now aborts early on retryable validation errors to avoid hot loops.
> - Device-sync and archive import paths abort on DB errors instead of logging warnings and continuing.
> - Behavioral Change: workers that previously swallowed DB-disconnect errors will now surface them to the supervisor, triggering reconnect handling.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4c78768.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/groups/commit_log.rs
+++ b/crates/xmtp_mls/src/groups/commit_log.rs
@@ -103,8 +103,15 @@ pub enum CommitLogError {
     GroupReaddValidationError(String),
     #[error("sync error: {0}")]
     SyncError(#[from] SyncSummary),
-    #[error("error: {0}")]
-    GenericError(String),
+    #[error("failed to send readd request for group {group_id}: {source}")]
+    FailedToSendReadd {
+        group_id: GroupId,
+        source: Box<CommitLogError>,
+    },
+    #[error("{count} readd request(s) failed: {errors:?}", count = errors.len())]
+    FailedReadds { errors: Vec<CommitLogError> },
+    #[error("no latest commit sequence id found for forked group {group_id}")]
+    MissingLatestCommitSequenceId { group_id: GroupId },
 }
 
 impl RetryableError for CommitLogError {
@@ -122,7 +129,9 @@ impl RetryableError for CommitLogError {
             Self::Conversion(_) => false,
             Self::GroupReaddValidationError(_group_readd_validation_error) => false,
             Self::SyncError(sync_error) => sync_error.is_retryable(),
-            Self::GenericError(_generic_error) => false,
+            Self::FailedToSendReadd { source, .. } => source.is_retryable(),
+            Self::FailedReadds { errors } => errors.iter().any(|e| e.is_retryable()),
+            Self::MissingLatestCommitSequenceId { .. } => false,
         }
     }
 }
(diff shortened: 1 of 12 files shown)


crates/xmtp_mls/src/worker/device_sync/worker.rs
```diff
@@ -295,6 +296,13 @@ where
             );
 
             if let Err(err) = self.process_message(handle, &msg, content).await {
+                // A failed message is non-fatal: log it, bump its attempt
+                // counter, and move on. But a dropped DB pool must stop the
+                // worker so the supervisor restarts it on reconnect — bubble it
```
insipx: ensure all comments are 1-2 lines

insipx: Done — trimmed every comment I added in this PR (docstrings + inline) down to 1-2 lines. Pushed as 39675be.

crates/xmtp_mls/src/groups/commit_log.rs
```diff
@@ -620,7 +618,7 @@ where
         let (group, stored_group) = MlsGroup::new_cached(self.context.clone(), group_id)?;
         if stored_group.conversation_type == ConversationType::Dm {
             let Some(dm_id) = stored_group.dm_id.clone() else {
-                tracing::error!("DM group {} has no dm_id", hex::encode(group_id));
+                tracing::error!("DM group {} has no dm_id", group_id);
```
insipx: group_id should be a tracing field, not in the message, i.e `tracing::error(group_id, "DM has no dm_id");`. do this for the other log lines as well. ensure we standardize on `group_id` so that for instance `conversation.id` becomes `group_id`

insipx: Done in c101cd3 — every single-group log line now uses a structured `group_id` field instead of putting the id in the message, and `conversation.id` is logged as `group_id`. GroupId-typed values use `%` (its Display is hex); the three remaining `hex::encode` calls are on raw `Vec<u8>`/`&[u8]` (proto/cursor bytes, not the GroupId type), which have no Display.

insipx:
Addressed the review feedback in 2a28b2d:

**`needs_db_reconnect` forwarding bug (good catch).** `DeviceSyncError::Storage` fell into `_ => false` and wouldn't trigger a reconnect. Fixed — and found the same class of bug in two more error types in the forwarding chain:
- `DeviceSyncError::Storage(StorageError)` → now forwarded
- `IdentityError::Db(ConnectionError)` → now forwarded (was only `StorageError`)
- `ClientError::Db(ConnectionError)` → now forwarded (was only `Storage`)

Added test coverage for `DeviceSyncError::Storage` and `IdentityError::Db` disconnects.

**`send_outgoing_readd_requests` perf.** A single transient readd failure no longer blocks the remaining groups until the next interval. Each group is processed **in order**; per-group failures are collected and surfaced together via a new `CommitLogError::FailedReadds { errors: Vec<CommitLogError> }` variant, so the worker processes as much as possible per tick. A dropped pool is the exception — it short-circuits and bubbles immediately rather than hammering a dead pool. (Used a sequential loop rather than a `Stream`; `.then()` gives the same in-order sequential semantics, but the loop makes the disconnect short-circuit cleaner.)

crates/xmtp_mls/src/groups/commit_log.rs
```diff
@@ -564,9 +565,9 @@ where
             .is_err()
         {
             tracing::warn!(
-                "Invalid signature for commit log entry {} on group {}, skipping",
-                serialized_entry.sequence_id,
-                hex::encode(&entry.group_id),
+                group_id = hex::encode(group_id),
```
tylerhawkes, line 568: inconsistent group_id formatting here. Everywhere else it is `group_id = %group_id`

tylerhawkes: Also, with these group ids everywhere, should we just set it at a higher level on a span and then in the formatter let the group id through even if the env vars to log a span aren't set? That would make some of this a lot cleaner (and we'd get group id in other logs for free.

tylerhawkes (approved):

crates/xmtp_mls/src/groups/commit_log.rs
insipx: yeah it makes sense, i can do that in a followup during one of the upcoming OTEL prs since it will be easier to test there anyway

PR #3703, opened 2026-06-02
perf(grpc): cache TLS endpoints so native root certs load once per host

mkysel:
## Problem

Loading clients in xdbg was dominated by TLS setup, not DB or MLS work. Phase timing while loading 100 identities showed ~100% of the time inside `network.client_bundle()`, with a tell-tale monotonic +42ms-per-call stair-step — i.e. a global lock serializing ~42ms of work per call, paid **twice per client** (api + sync channel).

Root cause: `create_tls_channel` builds a fresh tonic `Endpoint` on every call, and `ClientTlsConfig::with_enabled_roots()` makes tonic call `rustls_native_certs::load_native_certs()` and parse the **entire OS trust store** into a rustls `ClientConfig` each time. On macOS that read is serialized through Security.framework (~40ms). 200 calls × ~40ms ≈ the ~9–19s we were seeing.

## Fix

**`xmtp_api_grpc`:** cache the built `Endpoint` per `(host, rate_limit)`. The endpoint owns an `Arc<ClientConfig>` with the already-parsed roots, so `connect_lazy()` on a cheap clone still hands **every client its own connection** — no change to connection topology or trust semantics — while the native-cert load happens **once per host**. The lock is held across the one-time build so concurrent first-callers load certs exactly once. This benefits every platform that creates clients (node/mobile too), not just xdbg.

**`xdbg`:** parallelize identity loading (was a sequential `for` loop). Each client open is blocking SQLite/MLS work, so they're handed to Tokio's blocking pool via `tokio::task::spawn_blocking` and awaited together; `load_all_identities` / `load_n_identities` and their callers become `async`.

## Result

`cargo xdbg -b dev generate -e message -a 1`, 100 clients:

| | load time |
|---|---|
| before | ~19s |
| after | **~185ms** |

No new dependencies. No trust-store/behavior change. Each client still gets its own connection.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy --locked --all-features --all-targets --no-deps -- -Dwarnings` (full workspace, clean)
- [x] manual: 100-client xdbg load ~19s → ~185ms, sends still succeed

> [!NOTE]
> ## Problem
>
> Loading clients in xdbg was dominated by TLS setup, not DB or MLS work. Phase timing while loading 100 identities showed ~100% of the time inside `network.client_bundle()`, with a tell-tale monotonic +42ms-per-call stair-step — i.e. a global lock serializing ~42ms of work per call, paid **twice per client** (api + sync channel).
>
> Root cause: `create_tls_channel` builds a fresh tonic `Endpoint` on every call, and `ClientTlsConfig::with_enabled_roots()` makes tonic call `rustls_native_certs::load_native_certs()` and parse the **entire OS trust store** into a rustls `ClientConfig` each time. On macOS that read is serialized through Security.framework (~40ms). 200 calls × ~40ms ≈ the ~9–19s we were seeing.
>
> ## Fix
>
> **`xmtp_api_grpc`:** cache the built `Endpoint` per `(host, rate_limit)`. The endpoint owns an `Arc<ClientConfig>` with the already-parsed roots, so `connect_lazy()` on a cheap clone still hands **every client its own connection** — no change to connection topology or trust semantics — while the native-cert load happens **once per host**. The lock is held across the one-time build so concurrent first-callers load certs exactly once. This benefits every platform that creates clients (node/mobile too), not just xdbg.
>
> **`xdbg`:** parallelize identity loading (was a sequential `for` loop). Each client open is blocking SQLite/MLS work, so they're handed to Tokio's blocking pool via `tokio::task::spawn_blocking` and awaited together; `load_all_identities` / `load_n_identities` and their callers become `async`.
>
> ## Result
>
> `cargo xdbg -b dev generate -e message -a 1`, 100 clients:
>
> | | load time |
> |---|---|
> | before | ~19s |
> | after | **~185ms** |
>
> No new dependencies. No trust-store/behavior change. Each client still gets its own connection.
>
> ## Test plan
> - [x] `cargo fmt --check`
> - [x] `cargo clippy --locked --all-features --all-targets --no-deps -- -Dwarnings` (full workspace, clean)
> - [x] manual: 100-client xdbg load ~19s → ~185ms, sends still succeed
>
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)
> #### Changes since #3703 opened
>
> - Converted `load_clients_parallel`, `load_all_identities`, and `load_n_identities` utilities to async functions [69570ca]
> - Updated `GenerateMessages::new` constructor and `GenerateGroups::create_groups` workflow to async [69570ca]
> - Updated `Generate::run` and `Sync::run` command methods to await async operations [69570ca]
> 
>

### DIFF
--- a/apps/xmtp_debug/src/app/clients.rs
+++ b/apps/xmtp_debug/src/app/clients.rs
@@ -208,6 +208,50 @@ fn existing_client_inner_for(
     Ok(client)
 }
 
+/// Build clients for `identities` across many threads.
+///
+/// Each open is blocking SQLite + MLS work (no network — channels are
+/// lazy), so we heavily oversubscribe cores: `num_cpus * 10`. Identities
+/// are spread round-robin into per-thread buckets (moved, so we only need
+/// `Identity: Send`, not `Sync`, and need no shared lock). Each worker
+/// enters the current Tokio runtime context so `client_from_identity`
+/// behaves exactly as it did on the calling worker thread. Returns one
+/// `(inbox_id, result)` per identity in arbitrary order; callers decide
+/// whether to skip or propagate failures.
+fn load_clients_parallel(identities: Vec<Identity>) -> Vec<(InboxId, Result<crate::DbgClient>)> {
+    let n = identities.len();
+    if n == 0 {
+        return Vec::new();
+    }
+    let fanout = (num_cpus::get() * 10).clamp(1, n);
+    let handle = tokio::runtime::Handle::try_current().ok();
+
+    let mut buckets: Vec<Vec<Identity>> = (0..fanout).map(|_| Vec::new()).collect();
+    for (i, identity) in identities.into_iter().enumerate() {
+        buckets[i % fanout].push(identity);
+    }
+
+    std::thread::scope(|scope| {
+        let workers: Vec<_> = buckets
+            .into_iter()
+            .map(|bucket| {
+                let handle = handle.clone();
+                scope.spawn(move || {
+                    let _guard = handle.as_ref().map(|h| h.enter());
+                    bucket
+                        .into_iter()
+                        .map(|identity| (identity.inbox_id, client_from_identity(&identity)))
+                        .collect::<Vec<_>>()
+                })
+            })
+            .collect();
+        workers
+            .into_iter()
+            .flat_map(|w| w.join().expect("client load thread panicked"))
+            .collect()
+    })
+}
+
 /// Loads all identities. Honors `App::strict_versioning()`: when set,
 /// reads only this binary's version partition; otherwise reads every
 /// version on the network.
@@ -230,9 +274,8 @@ pub fn load_all_identities(
     let now = std::time::Instant::now();
     let mut clients = HashMap::new();
     let mut skipped = 0u32;
-    for identity in identities {
-        let inbox_id = identity.inbox_id;
-        match client_from_identity(&identity) {
+    for (inbox_id, result) in load_clients_parallel(identities) {
+        match result {
             Ok(client) => {
                 clients.insert(inbox_id, Mutex::new(client));
             }
@@ -292,14 +335,9 @@ pub fn load_n_identities(
         now.elapsed()
     );
     let now = std::time::Instant::now();
-    let clients = identities
+    let clients = load_clients_parallel(identities)
         .into_iter()
-        .map(|id| {
-            Ok::<_, eyre::Report>((
-                id.inbox_id,
-                Arc::new(Mutex::new(client_from_identity(&id)?)),
-            ))
-        })
+        .map(|(inbox_id, result)| Ok::<_, eyre::Report>((inbox_id, Arc::new(Mutex::new(result?)))))
         .collect::<Result<HashMap<_, _>, _>>()?;
     tracing::info!(
         "took {:?} to load {} xmtp clients",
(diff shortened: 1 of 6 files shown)


apps/xmtp_debug/src/app/clients.rs
```diff
@@ -208,6 +208,50 @@ fn existing_client_inner_for(
...
+///
+/// Each open is blocking SQLite + MLS work (no network — channels are
+/// lazy), so we heavily oversubscribe cores: `num_cpus * 10`. Identities
+/// are spread round-robin into per-thread buckets (moved, so we only need
+/// `Identity: Send`, not `Sync`, and need no shared lock). Each worker
+/// enters the current Tokio runtime context so `client_from_identity`
+/// behaves exactly as it did on the calling worker thread. Returns one
+/// `(inbox_id, result)` per identity in arbitrary order; callers decide
+/// whether to skip or propagate failures.
+fn load_clients_parallel(identities: Vec<Identity>) -> Vec<(InboxId, Result<crate::DbgClient>)> {
```
tylerhawkes: Why not make this function async and then use `tokio::spawn_blocking()`?

neekolas:
Wow. This is a good find.

crates/xmtp_api_grpc/src/grpc_client/native.rs
```diff
@@ -76,10 +78,36 @@ pub(crate) fn apply_channel_options(endpoint: Endpoint, limit: u64) -> Endpoint
...
+
 #[tracing::instrument(level = "trace", skip_all)]
 pub fn create_tls_channel(address: String, limit: u64) -> Result<Channel, GrpcBuilderError> {
-    let channel = apply_channel_options(Channel::from_shared(address)?, limit)
-        .tls_config(ClientTlsConfig::new().with_enabled_roots())?
-        .connect_lazy();
-    Ok(channel)
+    // Hold the lock across the (one-time, per-key) build so concurrent callers
+    // for the same host load native certs exactly once instead of racing.
+    let mut endpoints = TLS_ENDPOINTS.lock().unwrap_or_else(|e| e.into_inner());
```
tylerhawkes, line 102: May as well use a parking_lot mutex if you are going to bypass the panic propagation.

apps/xmtp_debug/src/app/clients.rs
insipx: Could just use the rayon crate which would make this cleaner

crates/xmtp_api_grpc/src/grpc_client/native.rs
insipx, line 102: I'm not sure we want a global cache at this level. We have a separate cache already in android and iOS sdks

Instead the client builder in xdbg could build with a prebuilt grpc client shared across all instances. Cloning wouldn't reunite in that case

apps/xmtp_debug/src/app/clients.rs
mkysel: Good call — switched to `tokio::task::spawn_blocking` and made the loaders `async`, which also let me drop the manual `Handle::enter()` context restoration. Load time unchanged (~185ms). Done in 69570ca4b.

crates/xmtp_api_grpc/src/grpc_client/native.rs
mkysel: I had that option and it multiplexes over the same GRPC client which defeats the purpose of a load testing tool. There is a difference between 100 connections and 1 connection with 100 clients.

insipx: Ah makes sense

insipx (approved):

PR #3708, opened 2026-06-03
feat(observability): xmtp_logging crate (sole owner of tracing-subscriber + otel) + node migration

insipx:
## What

Introduces a new `crates/xmtp_logging` crate as the **sole owner** of the
`tracing-subscriber` and `opentelemetry` dependencies, and migrates the **node**
binding onto it. All production logging/tracing pipeline construction, the OTLP
trace exporter, and the runtime-control handle now live in one place.

## Why

Logging/telemetry setup was scattered across `xmtp_common` (which sat near the
bottom of the dep graph, so every crate that touched it pulled in
`tracing-subscriber` + the full OpenTelemetry stack) and re-implemented per
binding. Consolidating into a dedicated crate gives us:

- one builder + one `LoggingConfig` record at the FFI boundary
- a `LoggingHandle` for runtime control (set level, enable/disable file logging,
  enable telemetry, flush) without re-installing the subscriber
- `xmtp_common` no longer carries `opentelemetry*` / `tracing-subscriber` as
  production deps (test/bench-only now)

## What's in the crate

- `config.rs` — canonical `Level` / `Rotation` / `ProcessType` enums and
  `FileConfig` / `TelemetryConfig` / `LoggingConfig` records.
- `filter.rs` — `filter_directive` (moved from `xmtp_common`).
- `telemetry.rs` — OTLP span exporter + `TelemetryGuard` (native-only,
  `cfg(not(wasm32))`; always compiled, enabled at runtime when an endpoint is set
  — no Rust feature flag).
- `layers/` — stdout/json fmt, rolling file, per-OS native (logcat/os_log/
  server-compact), and browser console/perf layers.
- `builder.rs` — `XmtpLoggingBuilder::from_config(cfg).install()` installs the
  global subscriber.
- `handle.rs` — `LoggingHandle` with reloadable level / file / telemetry slots.
- test subscriber (the obsolete `tracing-forest` tree logger and
  `TestLogReplace` / `REPLACE_IDS` are removed — OTEL spans replace them).

## Consumer changes

- **node** — fully migrated: `init_logging` builds a `LoggingConfig` from
  `LogOptions` and installs once; `flush_telemetry` delegates to the handle.
  Drops the direct `tracing-subscriber` dep.
- **mobile / wasm / xmtp_debug** — repointed `filter_directive` to `xmtp_logging`
  and dropped the `xmtp_common` `logging` feature. These keep their own
  subscribers for now; fully migrating them to the builder is a follow-up.

## Verification

Native + wasm build, clippy, rustfmt, taplo, hakari (regenerated — drops
`tracing-log`, the removed transitive), and unit tests
(`xmtp_logging` + `xmtp_common`) all pass locally.

## Follow-ups (separate PRs)

- Migrate wasm + mobile bindings onto the builder/handle (not just
  `filter_directive`).
- Migrate apps.

> [!NOTE]
> ### Extract logging and telemetry into a dedicated `xmtp_logging` crate as sole owner of tracing-subscriber and OpenTelemetry
> - Creates a new `xmtp_logging` crate with a fluent builder API (`XmtpLoggingBuilder`) that installs a global tracing subscriber with reloadable log level, optional rolling-file output, and optional OTLP telemetry export.
> - Migrates all logging/telemetry code out of `xmtp_common`, removing the `logging` feature flag from that crate; all consumers now depend on `xmtp_logging` directly.
> - Updates the Node binding's `init_logging` to use `xmtp_logging` via a process-global `LOGGING_HANDLE`, and changes `flush_telemetry` to call `handle.flush()` instead of shutting down the exporter.
> - Removes `TestLogReplace` and per-instance log-ID replacement from `xmtp_common::test` and `Tester`; test harnesses use `xmtp_logging::test_logging` instead.
> - Risk: `TestLogReplace` log-ID substitution is permanently removed from test output; tests that relied on named replacements in log lines will no longer see them.
>
> 
> #### Changes since #3708 opened
>
> - Replaced `once_cell` dependency with standard library synchronization primitives [6d0ab95]
> - Standardized workspace dependency notation across Cargo.toml files [6d0ab95]
> - Updated Cargo.lock [6d0ab95]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b45b18f.</sup>
> 
>
>


crates/xmtp_common/src/test/logger.rs
```diff
@@ -1,200 +0,0 @@
-// copy-paste of https://docs.rs/tracing-forest/latest/src/tracing_forest/printer/pretty.rs.html#62
```
insipx, line 1: just going to remove tracing-forest. don't really use it anymore as `lnav` can give span-contextual logs, or `jaeger` can be added to the docker setup to collect traces from each test w/o adding libxmtp dependencies

nix/package/mls_validation_service.nix
```diff
@@ -38,6 +38,10 @@ let
       # Actual source for crates needed to compile mls-validation-service
       (commonCargoSources (root + /crates/xmtp-workspace-hack))
       (commonCargoSources (root + /crates/xmtp_common))
+      # xmtp_common references xmtp_logging in its manifest (test-utils edge), so
```
insipx: these comments are unnecessary

crates/xmtp_logging/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+mod filter;
+mod handle;
+mod layers;
+pub use builder::{XmtpLogging, XmtpLoggingBuilder};
+pub use config::*;
+pub use error::Error;
+pub use filter::filter_directive;
+pub use handle::LoggingHandle;
+
+#[cfg(not(target_arch = "wasm32"))]
```
insipx, line 26: try to separate wasm and non-wasm stuff into different mods if possible to avoid using `#[cfg()]` everywhere

crates/xmtp_common/src/test/logger.rs
insipx: Done — `test/logger.rs` is deleted in this PR (the tracing-forest `Contextual` tree logger + `REPLACE_IDS`/`TestLogReplace` are gone). The test subscriber now lives in `xmtp_logging::test_logging` as a plain fmt logger; spans/OTEL cover the contextual case.

nix/package/mls_validation_service.nix
insipx: Removed.

crates/xmtp_logging/src/lib.rs
insipx: Done in d950ce2. `handle.rs` and `builder.rs` are now thin modules that gate `{native,wasm}` submodules once — the native handle (file/telemetry slots + worker guards) and browser handle (level filter only), and each `install()` body, live in their own files. Removed the scattered per-item `#[cfg(not(target_arch))]`; only the module declarations and a couple of native-only re-exports in `lib.rs` remain gated.

crates/xmtp_logging/src/builder/native.rs
```diff
@@ -0,0 +1,90 @@
...
+            && t.endpoint.is_some()
+        {
+            handle.enable_telemetry(t)?;
+        }
+        if let Some(f) = cfg.file {
+            handle.enable_file(f)?;
+        }
+
+        Ok(handle)
+    }
```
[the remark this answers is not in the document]
insipx: Good catch. Fixed by building the fallible dynamic layers (file writer + OTLP exporter) **before** `try_init()` and seeding the reload slots with them, instead of installing the global subscriber and then calling `enable_file`/`enable_telemetry` (which could fail post-install). Now a bad file path or telemetry endpoint returns `Err` *before* the irreversible `set_global_default`, so the caller can retry with a fixed config. Extracted `build_file_layer`/`build_telemetry_layer` helpers shared between `install` and the runtime `enable_*` methods so there's no duplicated construction logic.

crates/xmtp_logging/src/handle/native.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /// Flush pending telemetry spans (best-effort). File writer lines flush as the
+    /// worker drains and on drop; this primarily forces the OTel exporter to push
+    /// queued spans, e.g. before process exit.
+    pub fn flush(&self) {
+        if let Some(t) = self.guards.lock().telemetry.as_ref() {
+            t.shutdown();
+        }
+    }
```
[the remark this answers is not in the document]
insipx: Fixed in b45b18f. `flush()` was wrongly calling `TelemetryGuard::shutdown()` (terminal). Added a non-destructive `TelemetryGuard::force_flush()` (`SdkTracerProvider::force_flush` — pushes queued spans, exporter stays live) and `flush()` now calls that, so spans created after a flush still export. `shutdown()` stays for the terminal flush-and-stop path (the `Drop` impl, and any explicit shutdown). This also makes node's `flushTelemetry()` safe to call repeatedly.

apps/xmtp_debug/Cargo.toml
```diff
@@ -70,11 +70,12 @@ url.workspace = true
 valuable = { version = "0.1", features = ["derive"] }
 xmtp_api_d14n.workspace = true
 xmtp_api_grpc.workspace = true
-xmtp_common = { workspace = true, features = ["logging"] }
+xmtp_common = { workspace = true }
```
tylerhawkes: this should be `xmtp_common.workspace = true`

bindings/mobile/Cargo.toml
```diff
@@ -35,12 +35,13 @@ uniffi = { version = "0.31.0", default-features = false, features = ["tokio"] }
 xmtp_api.workspace = true
 xmtp_api_d14n.workspace = true
 xmtp_api_grpc.workspace = true
-xmtp_common = { workspace = true, features = ["logging"] }
+xmtp_common = { workspace = true }
```
tylerhawkes: Same here.

bindings/mobile/Cargo.toml
```diff
@@ -87,6 +88,7 @@ xmtp_archive = { workspace = true, features = ["test-utils"] }
 xmtp_common = { workspace = true, features = ["test-utils"] }
 xmtp_configuration = { workspace = true, features = ["test-utils"] }
 xmtp_db = { workspace = true, features = ["test-utils"] }
+xmtp_logging = { workspace = true, features = ["test-utils"] }
```
tylerhawkes, line 91: I wonder if these could be simplified to `xmtp_logging.features = ["test-utils"]` since the workspace is already set to true in the regular dep.

bindings/node/Cargo.toml
```diff
@@ -31,22 +31,17 @@ prost.workspace = true
...
-  "fmt",
-  "json",
-  "chrono",
-] }
 xmtp-workspace-hack.workspace = true
 xmtp_api.workspace = true
 xmtp_api_d14n.workspace = true
 xmtp_api_grpc = { workspace = true, optional = true }
-xmtp_common = { workspace = true, features = ["logging"] }
+xmtp_common = { workspace = true }
```
tylerhawkes: Same here.

bindings/wasm/Cargo.toml
```diff
@@ -16,7 +16,7 @@ crate-type = ["cdylib", "rlib"]
 
 [dependencies]
 xmtp-workspace-hack.workspace = true
-xmtp_common = { workspace = true, features = ["logging"] }
+xmtp_common = { workspace = true }
```
tylerhawkes: And here

bindings/wasm/Cargo.toml
```diff
@@ -50,12 +50,13 @@ web-sys = { workspace = true, features = [
 ] }
 xmtp_api.workspace = true
 xmtp_api_d14n.workspace = true
-xmtp_common = { workspace = true, features = ["logging"] }
+xmtp_common = { workspace = true }
```
tylerhawkes: This is already always a dependency on line 19. Several of these can be moved to be unconditional dependencies since they don't do anything special for wasm.

crates/xmtp_common/Cargo.toml
```diff
@@ -18,13 +18,13 @@ tokio.workspace = true
 tracing.workspace = true
 web-time.workspace = true
 
-# optional/test-utils/dev-deps
+# optional/test-utils/dev-deps. tracing-subscriber survives here only for the
+# bench subscriber + the `traced_test` test helper; all production logging now
+# lives in xmtp_logging.
 once_cell = { workspace = true, optional = true }
```
tylerhawkes: orthogonal, but we don't need once_cell anywhere anymore.

tylerhawkes (approved):
Just some nits on cargo formatting. Looks good otherwise.

apps/xmtp_debug/Cargo.toml
insipx: Done — `xmtp_common.workspace = true`.

bindings/mobile/Cargo.toml
insipx: Done — switched to `xmtp_common.workspace = true`.

bindings/mobile/Cargo.toml
insipx: Tried it, but `xmtp_logging.features = ["test-utils"]` on its own in `[dev-dependencies]` fails to parse: `dependency (xmtp_logging) specified without providing a local path, Git repository, version, or workspace dependency to use`. Cargo doesn't inherit `workspace = true` across dependency tables, so the dev-dep needs its own source. Kept the full `{ workspace = true, features = ["test-utils"] }` form.

bindings/node/Cargo.toml
insipx: Done — `xmtp_common.workspace = true`.

bindings/wasm/Cargo.toml
insipx: Done — `xmtp_common.workspace = true`.

bindings/wasm/Cargo.toml
insipx: Removed the duplicate — `xmtp_common` is already an unconditional dependency on line 19, so the wasm-target copy was redundant. Dropped it.

crates/xmtp_common/Cargo.toml
insipx: Done — dropped `once_cell` entirely. The only remaining use was the bench `LOGGER` static in `bench.rs`, migrated to `std::sync::OnceLock`; removed the dep + its `test-utils`/`bench` feature refs.

PR #70, opened 2023-04-21
Move xmtpv3 into a separate workspace

richardhuaaa:
We are currently using this repo for two separate purposes:
- Core encryption/networking libraries for iOS
- The next XMTP protocol version built on vodozemac (xmtpv3)

They don't share an API, have no code in common, and have conflicting dependencies. Let's move xmtpv3 into a separate workspace for now.

PR #3715, opened 2026-06-03
Make gRPC keepalive overridable via env vars (keep fast defaults)

tylerhawkes:
## What
The native gRPC HTTP/2 keepalive parameters are now overridable via environment variables, read once per process in `apply_channel_options`. **Defaults are unchanged** (fast detection).

## Why
herald-lite #70 (`UNAVAILABLE` / status 14 disconnects on herald dev) was root-caused to libxmtp's own client-side h2 keepalive: ping every 16s, demand a PONG within 10s, even idle. On a cross-cloud Fly→AWS path under load the PONG starves past the 10s deadline, hyper drops the connection, tonic maps it to `UNAVAILABLE`, and the retry layer masks it.

A **global** relax is wrong: mobile wants *fast* dead-connection detection, herald wants *forgiving*. (A global 30/30 broke `client::tests::should_reconnect`, which deliberately relies on ~26s keepalive detection — its comment cites "16s interval + 10s timeout".) So the defaults stay fast and a high-latency deployment opts out via env.

## Env vars (all optional; unset = default)
| env var | overrides | default |
|---|---|---|
| `XMTP_GRPC_KEEPALIVE_INTERVAL_SECS` | `http2_keep_alive_interval` | 16 |
| `XMTP_GRPC_KEEPALIVE_TIMEOUT_SECS` | `keep_alive_timeout` | 10 |
| `XMTP_GRPC_TCP_KEEPALIVE_SECS` | `tcp_keepalive` (0 disables) | 16 |
| `XMTP_GRPC_KEEPALIVE_WHILE_IDLE` | `keep_alive_while_idle` | true |

Read once via a `LazyLock` (servers set them before start), so every connection after the first is a struct copy, not an env read. Native-only (wasm fetch has no h2 knobs). herald rides `@xmtp/node-sdk`, so env vars need **no** binding/SDK change — it just sets them in its Fly deploy.

## Testing
- 4 unit tests for the env parsing (defaults, overrides, `0`-disables-tcp, garbage-falls-back)
- `should_reconnect` passes (~28s) with the fast defaults restored
- `just lint-rust` green

Credit to the herald service owner for the root-cause.

> [!NOTE]
> ### Make gRPC keepalive settings overridable via environment variables
> - Adds a `KeepaliveConfig` struct in [native.rs](https://github.com/xmtp/libxmtp/pull/3715/files#diff-d8ababd4174607b4463cfa01411d0034229c7d7bfccd1aecb5066b2386fb6d6a) that holds interval, timeout, TCP keepalive, and while-idle settings, with defaults matching the previous hardcoded values.
> - Configuration is read once per process via a `LazyLock<KeepaliveConfig>` populated from environment variables (`XMTP_GRPC_TCP_KEEPALIVE_SECS`, etc.).
> - Setting `XMTP_GRPC_TCP_KEEPALIVE_SECS=0` disables TCP keepalive entirely; invalid values fall back to defaults silently.
> - Behavioral Change: `apply_channel_options` now reads from the global config instead of hardcoded constants, so environment variables set before process start affect all channel setups.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a1516db.</sup>
> 
>

### DIFF
--- a/crates/xmtp_api_grpc/src/grpc_client/native.rs
+++ b/crates/xmtp_api_grpc/src/grpc_client/native.rs
@@ -10,6 +10,74 @@ use url::Url;
 
 use std::task::{Context, Poll};
 
+/// HTTP/2 / TCP keep-alive parameters for the gRPC channel.
+///
+/// Defaults favour *fast* detection of a dead connection (~26s = interval + timeout),
+/// which suits mobile clients and the default native path. A deployment on a high-latency
+/// link can relax any of them via environment variables, without affecting other
+/// consumers:
+///
+/// | env var                             | field                        | default |
+/// |-------------------------------------|------------------------------|---------|
+/// | `XMTP_GRPC_KEEPALIVE_INTERVAL_SECS` | `http2_keep_alive_interval`  | 16      |
+/// | `XMTP_GRPC_KEEPALIVE_TIMEOUT_SECS`  | `keep_alive_timeout`         | 10      |
+/// | `XMTP_GRPC_TCP_KEEPALIVE_SECS`      | `tcp_keepalive` (0 disables) | 16      |
+/// | `XMTP_GRPC_KEEPALIVE_WHILE_IDLE`    | `keep_alive_while_idle`      | true    |
+///
+/// Read once per process (servers set these before start), so it is not part of the TLS
+/// endpoint cache key.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct KeepaliveConfig {
+    interval: Duration,
+    timeout: Duration,
+    tcp_keepalive: Option<Duration>,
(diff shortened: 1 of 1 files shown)


insipx (approved):

neekolas:
@tylerhawkes are you sure we want to do this globally, rather than making it configurable?  
  
The reason we had short timeouts before was because on mobile we often rely on the keepalive to let you know that you are offline and to surface errors. Not sure there is a one-size-fits-all solution.

tylerhawkes:
I'll see what it takes to make it configurable.

neekolas (approved):
I like this. Avoids bloating the mobile API surface for a thing that probably only needs to be overwritten by something weird like Herald.

PR #3718, opened 2026-06-04
Add client-side stream-liveness watchdog with reconnect-on-stale

tylerhawkes:
## What
A client-side stream-liveness floor: a `WatchdogStream` combinator that wraps each subscription with an idle timer (reset on every item). If nothing arrives within the idle timeout (default 300s, **env-overridable**), it yields a single `SubscribeError::StreamStale` and terminates; the consume loops treat that as a signal to **re-subscribe from the persisted cursor** instead of hanging forever. A silently-dead stream becomes a recoverable reconnect.

## Why
The v3 path has no server-side keepalive, so an idle subscription whose transport wedges open — e.g. an L7 proxy keeps answering HTTP/2 pings while the backend stream is gone — delivers neither an error nor a close. The consumer just hangs. This is the client-side liveness floor; it's layered with the keepalive work (#3715) and, longer term, a server-side heartbeat (XIP-83) that lets the idle timeout drop well below 300s.

## How
- `subscriptions/watchdog.rs` — the `WatchdogStream<S>` combinator (lazy-armed idle timer via an injectable factory over `BoxDynFuture`, cross-platform native/wasm; implements `FusedStream` once tripped) + a test-only idle-timeout override.
- Timing is read **once at startup** into a process-wide `WatchdogConfig` (`LazyLock`), each knob env-overridable — see **Config** below. Reconnect is **throttled, not delayed**: it measures from when each attempt began, so a stream that was healthy for the whole idle timeout reconnects **immediately** (a couple seconds of backoff on top of a multi-minute idle is pointless), while a *tight* loop — a fast creation failure or a stream that trips right after creation — is paced to ≥ one attempt per floor. The floor uses a **real random jitter** draw (not wall-clock low bits) to de-sync clients tight-looping against the same upstream.
- One **generic runner**, `watchdog::spawn_watchdog_stream`, owns the whole reconnect lifecycle (subscribe → readiness → watchdog-wrap → consume → reconnect-on-stale → `on_close`). `stream_all_messages_with_callback`, `stream_conversations_with_callback`, and the group `stream_messages_with_callback` are now thin wrappers that pass a `subscribe` closure + callback — the triple-duplicated double-loop is gone, so the three can't drift apart. (It's generic over the concrete stream type via the `*_owned` `'static` constructors — no boxing.)
- `SubscribeError::StreamStale` (retryable) drives reconnect; cancellation-token aware throughout. A clean server close or cancellation stops the loop; only `StreamStale` reconnects. A creation failure **propagates as a startup error** before the first successful subscription, and is **retried with the throttle** after it — a transient reconnect-time hiccup no longer silently terminates the subscription (review finding #1).
- On a stale trip the runner **re-subscribes before the reconnect wait**, while the old stream is still in scope — so the new subscription (and, for conversations, its `LocalEvents` broadcast receiver) is already buffering during the pause and events arriving mid-reconnect aren't dropped (addresses the review's resubscribe-before-sleep point; the only residual gap is the brief moment the new subscription is being established).

## Config (env vars, captured once at startup; safe defaults)
- **`XMTP_STREAM_WATCHDOG_ENABLED` (default `false`) — master switch.** The watchdog is **opt-in**: with no env set, streams behave exactly as before this PR (no idle-trip, no reconnect). Only a deployment that wants the liveness floor — herald, long-lived agents — turns it on. Conservative for now; can flip to on-by-default later. (Values: `1`/`true`/`yes`/`on` vs `0`/`false`/`no`/`off`.)
- `XMTP_STREAM_WATCHDOG_IDLE_TIMEOUT_SECS` (default `300`, max `86400`) — idle timeout before a stale trip. A forgiving server-side proxy can raise it; a latency-sensitive mobile client can lower it.
- `XMTP_STREAM_WATCHDOG_RECONNECT_BASE_SECS` (default `1`, max `3600`) — minimum interval between reconnect attempts (the throttle floor). A normal long-idle trip reconnects immediately; this only paces a tight loop.
- `XMTP_STREAM_WATCHDOG_RECONNECT_JITTER_MS` (default `1000`, max `3600000`) — random jitter added to the throttle floor, to de-sync clients tight-looping against the same upstream (`0` disables).

## Tests (all pass; no flakiness across repeated runs)
- Combinator state machine, now **synchronous** — the stream is polled by hand via `futures-test`, fully deterministic with no real clock: pass-through-then-trip-once, clean-end-is-not-a-trip, no-trip-until-the-timer-fires.
- A **proptest** over arbitrary item sequences: every item is yielded in order, then exactly one `StreamStale`, then terminated — fused forever after.
- `new_uses_a_real_timer` — smoke test that the production constructor wires a real wall-clock timer.
- `WatchdogConfig` env parsing (present / explicit-zero / garbage / missing) and jitter-stays-in-bounds.
- `watchdog_trips_on_idle_real_stream` — a real idle v3 stream goes silent and the watchdog trips + terminates (validates the no-server-keepalive premise against the live transport).
- `watchdog_reconnect_keeps_stream_alive` — a message sent *after* a stale-trip still arrives, proving the consume loop re-subscribed (confirmed via logs).

`just lint-rust` green; native + wasm compile.

## Notes / caveats
- The watchdog wraps the **outer** (consume-loop) stream, not the inner network stream. A later phase with a server heartbeat relocates it below the heartbeat filter.
- Only `StreamStale` triggers reconnect. Transport errors still terminate as before (pre-existing behavior, unchanged).
- A healthy *dormant* stream reconnects ~every idle-timeout (a catch-up query + a `StreamClosed`/`StreamOpened` event pair). The server heartbeat removes this.
- Conversation stream: a re-subscribe drops/recreates the local `LocalEvents` broadcast receiver, which has no replay. Network welcomes are caught up from the cursor, but a *locally* created group during the reconnect window is not re-emitted (the caller already holds that group). Documented in code; removed by the later refactor that wraps the network sub below the broadcast.

Merge gate: full CI streaming suite green.

> [!NOTE]
> ### Add client-side stream-liveness watchdog with reconnect-on-stale for message and conversation subscriptions
> - Introduces a new [`watchdog`](https://github.com/xmtp/libxmtp/pull/3718/files#diff-63f916d351d8a9bc3a55b194055219f7089f64c690658c56dd2e3ce2a1550c9c) module with `WatchdogStream`, a stream combinator that detects inactivity and emits `StreamStale` after a configurable idle timeout, then terminates.
> - Adds `spawn_watchdog_stream`, a unified helper used by `stream_messages_with_callback`, `stream_conversations_with_callback`, and `stream_all_messages_with_callback` to transparently re-subscribe after a stale trip or transient failure.
> - Reconnect behavior is env-driven with safe defaults: idle timeout, base reconnect delay, and random jitter are all configurable. A test-only thread-local override is available for unit tests.
> - Adds `SubscribeError::StreamStale`, marked retryable and not requiring a DB reconnect.
> - Behavioral Change: all three subscription callbacks now auto-reconnect on idle instead of silently hanging or terminating; existing callbacks receive no signal when a reconnect occurs.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e1fd7f0.</sup>
> 
>
>

### DIFF
--- a/crates/xmtp_mls/src/groups/subscriptions.rs
+++ b/crates/xmtp_mls/src/groups/subscriptions.rs
@@ -7,11 +7,11 @@ use crate::{
         d14n_compat::{V3OrD14n, decode_group_message},
         process_message::{ProcessFutureFactory, ProcessMessageFuture},
         stream_messages::StreamGroupMessages,
+        watchdog::spawn_watchdog_stream,
     },
 };
 use futures::{FutureExt, Stream, StreamExt, TryStreamExt, future, stream as future_stream};
 use itertools::Itertools;
-use tokio::sync::oneshot;
 use xmtp_api_d14n::{
     protocol::{
         CursorStore, EnvelopeCollection, EnvelopeError, GroupMessageExtractor,
@@ -143,34 +143,29 @@ where
 pub(crate) fn stream_messages_with_callback<Context>(
     context: Context,
     active_conversations: impl Iterator<Item = GroupId> + MaybeSend + 'static,
-    mut callback: impl FnMut(Result<StoredGroupMessage>) + MaybeSend + 'static,
+    callback: impl FnMut(Result<StoredGroupMessage>) + MaybeSend + 'static,
     on_close: impl FnOnce() + MaybeSend + 'static,
 ) -> impl StreamHandle<StreamOutput = Result<()>>
 where
     Context: XmtpSharedContext + 'static,
     Context::ApiClient: XmtpMlsStreams + 'static,
+    Context::Db: 'static,
 {
-    let (tx, rx) = oneshot::channel();
-
-    xmtp_common::spawn(Some(rx), async move {
-        let stream = match StreamGroupMessages::new(&context, active_conversations.collect()).await
-        {
-            Ok(stream) => stream,
-            Err(e) => {
-                tracing::warn!("Failed to create group message stream, closing: {}", e);
-                on_close();
-                return Ok::<_, SubscribeError>(());
-            }
-        };
-        futures::pin_mut!(stream);
-        let _ = tx.send(());
-        while let Some(message) = stream.next().await {
-            callback(message)
-        }
-        tracing::debug!("`stream_messages` stream ended, dropping stream");
-        on_close();
-        Ok::<_, SubscribeError>(())
-    })
+    let cancel = context.cancellation_token().clone();
+    let groups: Vec<GroupId> = active_conversations.collect();
+    // Each `new_owned` reads the last cursor per group, so a re-subscribe resumes where the
+    // dropped stream left off.
+    spawn_watchdog_stream(
+        cancel,
+        "stream_messages",
+        move || {
+            let context = context.clone();
+            let groups = groups.clone();
+            async move { StreamGroupMessages::new_owned(context, groups).await }
+        },
+        callback,
+        on_close,
+    )
 }
 
 #[cfg(test)]
--- a/crates/xmtp_mls/src/subscriptions/watchdog.rs
+++ b/crates/xmtp_mls/src/subscriptions/watchdog.rs
@@ -1,0 +1,731 @@
+//! Stream-liveness floor: a watchdog that detects a silently-dead subscription.
+//!
+//! A long-lived subscription whose transport wedges open (e.g. an L7 proxy keeps
+//! answering HTTP/2 keepalive pings while the backend subscription is gone) delivers
+//! neither an error nor a stream close — the consumer simply hangs forever. There is no
+//! server-side keepalive on the v3 path today, so the client cannot distinguish a
+//! healthy-but-idle stream from a dead one.
+//!
+//! [`WatchdogStream`] wraps an inner subscription and arms an idle timer that resets on
+//! every item. If nothing arrives within the idle timeout (see [`Watchdog::idle_timeout`]),
+//! it yields a single [`SubscribeError::StreamStale`] and then terminates. The consume
+//! loops treat that error as a signal to tear down and reconnect (resuming from the
+//! persisted cursor), turning a silent hang into a recoverable reconnect.
+//!
+//! Because there is no keepalive yet, the timeout is deliberately long: a healthy dormant
+//! stream WILL trip and reconnect periodically. That cost is accepted for the floor and
+//! shrinks once a server heartbeat exists (see XIP-83). This combinator is
+//! endpoint-agnostic and is reused by later phases.
+//!
+//! **Opt-in.** The watchdog is *disabled by default* — with no env vars set, a stream never
+//! trips and behaves exactly as it did before the watchdog existed. A deployment enables it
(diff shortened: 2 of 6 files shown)


crates/xmtp_mls/src/groups/subscriptions.rs
```diff
@@ -153,19 +154,41 @@ where
...
+            let mut stale = false;
+            while let Some(message) = stream.next().await {
+                if matches!(&message, Err(SubscribeError::StreamStale)) {
+                    stale = true;
+                    break;
+                }
+                callback(message);
+            }
+
+            if !stale {
```
insipx: shouldn't this be `if stale`

crates/xmtp_mls/src/subscriptions/watchdog.rs
```diff
@@ -0,0 +1,230 @@
...
+#[pin_project]
+pub(crate) struct WatchdogStream<S> {
+    #[pin]
+    inner: S,
+    idle_timeout: Duration,
+    /// Armed lazily while the inner stream is idle; dropped (reset) on every item so the
+    /// deadline is always measured from the most recent activity.
+    timer: Option<BoxDynFuture<'static, ()>>,
+    /// Once tripped, the stream is finished and must not be polled again.
+    tripped: bool,
```
insipx, line 256: this lends itself well to implement the `FusedStream` trait

crates/xmtp_mls/src/subscriptions/watchdog.rs
```diff
@@ -0,0 +1,230 @@
...
+    use futures::{StreamExt, stream};
+
+    fn ok(n: u8) -> Result<u8, SubscribeError> {
+        Ok(n)
+    }
+
+    /// An idle stream trips after the timeout: one `StreamStale`, then termination.
+    #[xmtp_common::test(unwrap_try = true)]
+    async fn trips_on_idle() {
+        let inner = stream::pending::<Result<u8, SubscribeError>>();
```
insipx: with isolated stream impls like this it can be useful to make the tests synchronous and instead call the streams `poll()` impl manually. [futures-test](https://docs.rs/futures-test/latest/futures_test/) includes some nice test utilities for this as well. sync fns would also open a possibility of running the test against proptest more easily

mostly personal preference tho

insipx (approved):

crates/xmtp_mls/src/subscriptions/watchdog.rs
tylerhawkes: Done — pulled the idle timer behind an injectable factory so the combinator polls synchronously. Rewrote the unit tests to drive `poll_next` by hand via `futures-test`, and added a proptest over arbitrary item sequences (every item in order → one `StreamStale` → fused). Real-clock coverage stays in one smoke test + the integration tests.

crates/xmtp_mls/src/subscriptions/watchdog.rs
tylerhawkes: Done — implements `FusedStream` now; `is_terminated()` flips once tripped, whether by a stale trip or the inner stream ending.

PR #3728, opened 2026-06-05
feat(keepalive-probe): XMTP gRPC connection-health probe + sidecar

mkysel:
## What

A new `apps/keepalive-probe` binary: a probe for XMTP gRPC **connection health**. It holds connections to an XMTP gRPC endpoint with a configurable keepalive (defaults mirror libxmtp's `apply_channel_options`) and reports how long each survives and why it dropped.

Two modes:
- **idle** — *N* bare HTTP/2 connections, to isolate pure transport keepalive behavior.
- **subscribe** (`--subscribe-group <hex>`) — *N* real `MlsApi/SubscribeGroupMessages` streams, logging every payload and disconnect (the herald-shaped test; the V3 backend doesn't gate subscribe).

With `--continual` it becomes a never-exiting daemon (retries through everything, drains on SIGINT/SIGTERM) for running as a long-lived, **non-essential sidecar** next to a real client.

## Why

Built to diagnose [`herald-lite#70`](https://github.com/xmtplabs/convos-assistants/issues/2388) — herald regularly logs `hyper::Error(Http2, KeepAliveTimedOut)` → gRPC `UNAVAILABLE`, retried. Running this probe (raw tonic, no libxmtp) against the same backend from a different host already isolated the failure as **client/host-side, not the V3 backend**: 500 streams held for days with zero disconnects while herald bursted. Productizing it (image + CI) so it can run as a durable sidecar that survives client redeploys.

## CI / build (mirrors `xdbg`, which is also an `apps/` binary excluded from `default-members`)

- `apps/keepalive-probe/docker/Dockerfile` — multi-stage (`rust:1-bookworm` → `debian:trixie-slim` + `tini` for clean SIGTERM).
- `push-keepalive-probe.yml` — builds + pushes `ghcr.io/xmtp/keepalive-probe` (sha + `latest`).
- per-PR clippy gate: `nix/package/keepalive-probe-check.nix` + `ci-checks.nix` + `test-keepalive-probe.yml`, wired into `test.yml` path-detection.
- excluded from hakari (`workspace-hack`) like the other apps.

Passes `clippy -Dwarnings`, `fmt`, `taplo`, `hakari`, and `cargo test` locally. No changes to any existing crate.

> [!NOTE]
> ### Add `keepalive-probe` binary and sidecar for monitoring XMTP gRPC connection health
> - Introduces a new `keepalive-probe` Rust binary crate under [apps/keepalive-probe/](https://github.com/xmtp/libxmtp/pull/3728/files#diff-2ef1e92100a8ad08a977468548118f827a1ef384532a4dbfbd134c169047248f) that spawns N concurrent connections (idle HTTP/2 or subscribe via gRPC streaming) and monitors their lifetimes under configurable keepalive settings.
> - Idle mode establishes a raw TLS+HTTP/2 connection via hyper/rustls and holds it open; subscribe mode uses tonic to stream `SubscribeGroupMessages` from an XMTP node endpoint.
> - On completion or shutdown (SIGINT/SIGTERM), the tool prints a summary with lifetime percentiles, error classification buckets, and subscribe counters.
> - Adds a multi-stage [Dockerfile](https://github.com/xmtp/libxmtp/pull/3728/files#diff-bb3f0d03e50364bb8290474119b27ce2eee13596da42eb68fb99c94ed23fe722), a GitHub Actions push workflow to publish `ghcr.io/xmtp/keepalive-probe`, and Nix/CI clippy checks for the new crate.
>
> 
> #### Changes since #3728 opened
>
> - Replaced single-step group creation and inspection instructions with a three-step sequence for seeding identities, generating groups, and exporting groups to obtain hex group IDs [fcd4b38]
> - Renumbered subsequent setup steps to accommodate the expanded group setup sequence [fcd4b38]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4cfc2b1.</sup>
> 
>
>

### DIFF
--- a/.config/hakari.toml
+++ b/.config/hakari.toml
@@ -37,6 +37,7 @@ workspace-members = [
   "error_glossary",
   "xnet-cli",
   "xnet-lib",
+  "keepalive-probe",
 ]
 third-party = [
   { name = "tonic" },
@@ -59,6 +60,7 @@ workspace-members = [
   "error_glossary",
   "xnet-cli",
   "xnet-lib",
+  "keepalive-probe",
 ]
 third-party = [
   { name = "tonic" },
--- a/.github/workflows/push-keepalive-probe.yml
+++ b/.github/workflows/push-keepalive-probe.yml
@@ -1,0 +1,68 @@
+name: Push keepalive-probe Image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'apps/keepalive-probe/**'
+      - 'crates/xmtp_proto/**'
+      - 'crates/xmtp_common/**'
+      - 'crates/xmtp_configuration/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: xmtp/keepalive-probe
+
+jobs:
+  build:
+    name: Build and push keepalive-probe image
+    runs-on: ubuntu-latest
+    outputs:
+      digest: ${{ steps.push.outputs.digest }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Capture commit metadata
+        id: vars
+        run: echo "short_sha=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Log in to GHCR
+        uses: docker/login-action@v4
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels)
+        id: meta
+        uses: docker/metadata-action@v6
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=raw,value=${{ steps.vars.outputs.short_sha }}
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
+
+      - name: Build and push image
+        id: push
+        uses: docker/build-push-action@v7
+        with:
+          context: .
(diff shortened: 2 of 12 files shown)


mkysel:
Thanks for the reviews — addressed in 781530024.

**Macroscope (inline):**
- ✅ `round_secs` truncation → now rounds to nearest second (+ regression test). It was under-reporting exactly the sub-second timing this probe exists to measure.
- ✅ README `cargo run -p xmtp_debug` → `cargo run -p xdbg` (real package name).
- ✅ Shutdown `established` over-count → now derived from connections that actually reported, with aborted-on-shutdown workers surfaced as a separate `interrupted` line.
- ✅ `--count 0` + `--continual` exiting immediately → now rejected at startup.
- ⏭️ Path filter `crates/xmtp_configuration/**` flagged as a non-dependency: it's a *transitive* dep via `xmtp_proto` (`cargo tree -p keepalive-probe` shows `xmtp_common`, `xmtp_configuration`, `xmtp_cryptography`, `xmtp_macro`), so the trigger is correct. Left as-is.

**Claude review:**
- ✅ Added the subscribe-mode auth warning and the `SSLKEYLOGFILE` note to the README.
- ⏭️ Stagger overflow: `checked_mul` already saturates to `ZERO` with no panic, and the overflow threshold (`stagger × count`) is astronomically beyond any real invocation; left as-is.
- ⏭️ Exponential/jittered reconnect backoff: real reconnects are already spread by per-stream stagger and independent disconnect timing; a fixed 1s backoff is fine for a diagnostic. Noted for later if we see thundering-herd in practice.
- ⏭️ `tonic_mock` smoke test: deferred — the protobuf/channel path is validated by running it against real V3, which is the tool's purpose.

The Macroscope "needs human review" verdict is just code-owner coverage (`@xmtp/protocol-sdk`, `@xmtp/documentation`), not a correctness block.

insipx (approved):

mkysel:
Good catch — `generate` needs `--entity`/`--amount` flags (not a positional `group`), and creating a group needs identities first. Fixed the README to the real `xdbg` sequence in fcd4b38a1: `generate --entity identity --amount 10` → `generate --entity group --amount 1` → `export --entity group --out /tmp/groups.json` to read the id back.

PR #3741, opened 2026-06-07
Return proper status code

### LINKED ISSUE #3130: MLS validation service always return InvalidArgument gRPC code on error

Jr-kenny:
Closes #3130

**What was happening?**
Every error was coming as InvalidArguement, even when the error was just a server side disconnect making users having difficulty understand if the error was permanent or retryable

**What I changed?**
Errors are now sorted by whether they’re temporary, so retryable and server disconnect ones returns unavailable, and the error's short name now gets passed along so users see exactly which error happened.

**How I tested**
Added two unit tests, a bad-input error stays InvalidArgument, a retryable verifier error now returns Unavailable, both carry the label. `cargo test`, `cargo +nightly fmt --check`, and `cargo clippy -- -D warnings` pass.

> [!NOTE]
> ### Return `Unavailable` vs `InvalidArgument` gRPC status codes based on error retryability in MLS validation service
> - Implements `RetryableError` for `GrpcServerError` in [handlers.rs](https://github.com/xmtp/libxmtp/pull/3741/files#diff-ad601199ee09300a54b4df32946614a4e73d4a5872ce1517258590380e5cdd04), classifying `Signature` and retryable `Association` errors as transient.
> - Updates `From<GrpcServerError> for Status` to return `Code::Unavailable` for retryable errors and `Code::InvalidArgument` for non-retryable ones, replacing the previous single-code behavior.
> - Attaches a stable `"error-code"` metadata entry to every returned `Status` using `ErrorCode::error_code()`.
> - Behavioral Change: callers that previously received a uniform error code will now receive different codes depending on whether the error is transient.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1626702.</sup>
> 
>

### DIFF
--- a/apps/mls_validation_service/src/handlers.rs
+++ b/apps/mls_validation_service/src/handlers.rs
@@ -2,8 +2,9 @@ use futures::future::{join_all, try_join_all};
 use openmls::framing::ContentType;
 use openmls::prelude::{MlsMessageIn, ProtocolMessage, tls_codec::Deserialize};
 use openmls_rust_crypto::RustCrypto;
-use tonic::{Request, Response, Status};
+use tonic::{Code, Request, Response, Status, metadata::MetadataValue};
 
+use xmtp_common::{ErrorCode, RetryableError};
 use xmtp_id::key_package::{KeyPackageVerificationError, VerifiedKeyPackageV2};
 use xmtp_id::{
     associations::{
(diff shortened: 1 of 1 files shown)


insipx:
thanks for the contribution!

insipx (approved):

PR #3756, opened 2026-06-10
feat(messages): expose caller-settable idempotency_key across the FFI boundary

### LINKED ISSUE #3750: Application messages cannot be made idempotent: `idempotency_key` is hardcoded to a timestamp at every send entry point
## Summary

The MLS application-message id is derived from a per-message `idempotency_key`, and the wire format (`PlaintextEnvelope.V1`) carries that key so every receiver recomputes the same id. This is exactly the primitive needed to make a **resend idempotent** — but no caller can set it. Every send entry point hardcodes `idempotency_key = now_ns()`, so an application-level retry of the same content always produces a **new** message id, which defeats the content-id dedup that libxmtp (and downstream clients) already implement.

Request: make `idempotency_key` caller-settable on the send API.

## The mechanism is already there and correct

- The id is deterministic given the key: `calculate_message_id(group_id, content, idempotency_key)` — `crates/xmtp_mls/src/utils/mod.rs:34`.
- The key rides the wire as a first-class field — `PlaintextEnvelope.V1.idempotency_key` (tag 2). Its own doc comment: *"A unique value that can be used to ensure that the same content can produce different hashes. May be the sender timestamp."*
- The **receiver recomputes the id from the wire key**, not its own clock: `crates/xmtp_mls/src/groups/mls_sync.rs:1401`–`1405` (`calculate_message_id(self.group_id, &content, &idempotency_key)`).
- Receive-side dedup already collapses equal ids: `group_messages` PK + `store_or_ignore` (`crates/xmtp_db/src/encrypted_store/group_message.rs:434`).

So a stable caller key would produce an identical id at sender, on the wire, and at every receiver — and the existing dedup would make a resend invisible. The plumbing is complete except for one input.

## The problem: the key is hardcoded at all three send paths

None of the send entry points accept the key; each mints `now_ns()` internally:

- `send_message` — `crates/xmtp_mls/src/groups/mod.rs:1149` (`prepare_message(.., |now| into_envelope(message, now))`)
- `send_message_optimistic` — `mod.rs:1220` (same closure)
- `prepare_message_for_later_publish` — `mod.rs:1239` (`let now = now_ns();` → `mod.rs:1242`)
- `into_envelope(encoded_msg, idempotency_key: i64)` — `mod.rs:1455` (typed `i64`; the wire field is already `string`)
- `SendMessageOpts { should_push }` — `crates/xmtp_mls/src/groups/send_message_opts.rs` carries no key field.

The timestamp default is deliberately **anti-idempotent** (per the proto comment) — correct for normal chat, where sending "ok" twice should be two messages. But it leaves no way to opt into idempotency for use cases that need it.

## Impact / use case

Durable agent runtimes (e.g. an XMTP agent that persists its reply and retries delivery after a crash) cannot achieve at-least-once-with-dedup. On recovery the retry re-sends identical bytes but gets a fresh id, so recipients see a duplicate message. The redelivery dedup everyone relies on (same id arriving via stream + sync) does **not** help here, because an app-level resend is a genuinely new id. There is no escape without a caller-settable key.

## Proposed fix

Thread an optional caller key through, defaulting to `now_ns()` to preserve current behavior:

1. Add `idempotency_key: Option<String>` to `SendMessageOpts`.
2. At the send sites, use `opts.idempotency_key.unwrap_or_else(|| now_ns().to_string())` for **both** `into_envelope(..)` and `calculate_message_id(..)` so the local id, the wire field, and the receiver's recomputed id all match. (Widen `into_envelope`'s `i64` param to `&str`/`String`; the wire field is already `string`.)
3. Surface it on the bindings, each of which re-declares its own opts mirror:
   - `bindings/node` (`conversation/content_types.rs` — `send_text`/`send_reply`/`send_reaction`/…) and the `@xmtp/node-sdk` wrapper
   - `bindings/mobile` (`FfiSendMessageOpts`, + uniffi regen)
   - `bindings/wasm` (`SendMessageOpts` mirror + `From` impl)

## Note: a partial idempotent path already exists

`prepare_message_for_later_publish` (`mod.rs:1234`) + `publish_stored_message` (`mod.rs:1279`) is already retryable — the id is fixed at prepare time and persisted, `publish_stored_message` rebuilds the envelope from the **stored** `sent_at_ns` (`mod.rs:1299`) and is a no-op if already `Published` (`mod.rs:1293`). But it forces callers to round-trip and persist libxmtp's returned id between two calls; a caller-settable key is the cleaner primitive and also covers the fused `send_message` path.

---

Examined at `af418bd72` (local checkout HEAD).

mkysel:
## Summary

Application messages could not be made idempotent: every send entry point hardcoded `idempotency_key = now_ns()`, so an application-level retry of identical content always produced a **new** message id. This exposes an optional, caller-settable `idempotency_key`.

The message id is derived from this key (`calculate_message_id(group_id, content, key)`) and the key rides the wire via `PlaintextEnvelope.V1.idempotency_key`, so receivers recompute the same id and deduplicate via the existing `group_messages` PK constraint — enabling **at-least-once-with-dedup** semantics for durable agent runtimes.

Closes #3750

## Changes

**Core (`xmtp_mls`)**
- `SendMessageOpts.idempotency_key: Option<String>` — defaults to `now_ns().to_string()`, preserving current (always-unique) behavior.
- Threaded through `send_message`, `send_message_optimistic`, `prepare_message_for_later_publish`, and `publish_stored_message` (which now reuses the stored key so the wire id matches). `into_envelope` takes `&str`.

**Storage (`xmtp_db`)**
- New migration adds a non-null `idempotency_key` column to `group_messages` (`NOT NULL DEFAULT ''` + backfill `= CAST(sent_at_ns AS TEXT)`; the `''` default never persists — new inserts always supply a value). No table rebuild.
- `StoredGroupMessage.idempotency_key: String` — every construction site persists the exact key its id derives from (incoming messages use the wire-envelope key; group-updates/welcomes use their respective keys; device-sync restore falls back to the timestamp).

**Bindings**
- `idempotencyKey` surfaced on `SendMessageOpts`/`FfiSendMessageOpts` and the `prepareMessage` entry point in **node**, **wasm**, and **mobile**.

## Tests
- New: `test_explicit_idempotency_key_produces_deterministic_id`, `test_default_idempotency_key_is_unique_per_send`, and an end-to-end `test_idempotency_key_crosses_the_wire` (proves the key rides the wire and the receiver derives the same id).
- Verified locally: `xmtp_mls` prepare-message suite (8) + regression sweep (delete/welcome/cleanup/message_list/subscriptions, 73), `xmtp_db` (56), `xmtp_archive` (5). `clippy`/`fmt` clean on touched crates.

## Notes / follow-ups
- **wasm** was not compiled locally (no Nix in this environment); it mirrors the node change and needs `just wasm check` in CI.
- **Downstream deferred** (out of scope here): `@xmtp/node-sdk` re-exports `SendMessageOpts` from the bindings, so `idempotencyKey` flows through once a new `@xmtp/node-bindings` is published; convos-cli then passes `{ idempotencyKey }` at its retry site.
- **Known limitation:** the device-sync `GroupMessageSave` proto carries no idempotency key, so restored messages fall back to `sent_at_ns.to_string()` (harmless — already published). A proto field is a possible follow-up.

> [!NOTE]
> ### Expose caller-settable `idempotency_key` for message deduplication across FFI bindings
> - Adds an optional `idempotency_key` field to `SendMessageOpts` in the core Rust library and propagates it through all FFI layers (mobile, WASM, Node, iOS, Android).
> - When a key is provided, `prepare_message_for_later_publish` derives the `message_id` deterministically from `(group_id, content, idempotency_key)`; repeated calls with the same key return the existing message ID without creating duplicates.
> - When no key is provided, a unique timestamp-based key is generated per send, preserving existing behavior.
> - Adds a new `idempotency_key` column to the `group_messages` DB table via migration; existing rows are backfilled from `sent_at_ns`.
> - `publish_stored_message` now passes the stored `idempotency_key` in the wire envelope so receivers derive the same `message_id` as the sender.
> - Risk: the DB migration adds a `NOT NULL` column with a default backfill — schema change requires migration on all existing databases.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 75e982a.</sup>
> 
>
>

### DIFF
--- a/crates/xmtp_mls/src/groups/mod.rs
+++ b/crates/xmtp_mls/src/groups/mod.rs
@@ -1146,7 +1146,7 @@ where
         self.commit_pending_proposals_if_any().await?;
 
         let message_id =
-            self.prepare_message(message, opts, |now| Self::into_envelope(message, now))?;
+            self.prepare_message(message, opts, |key| Self::into_envelope(message, key))?;
 
         self.sync_until_last_intent_resolved().await?;
 
@@ -1217,7 +1217,7 @@ where
         opts: send_message_opts::SendMessageOpts,
     ) -> Result<Vec<u8>, GroupError> {
         let message_id =
-            self.prepare_message(message, opts, |now| Self::into_envelope(message, now))?;
+            self.prepare_message(message, opts, |key| Self::into_envelope(message, key))?;
         Ok(message_id)
     }
 
@@ -1229,17 +1229,23 @@ where
     /// # Arguments
     /// * `message` - The message content bytes
     /// * `should_push` - Whether to send a push notification when publishing
+    /// * `idempotency_key` - Optional caller-supplied key the message id is
+    ///   derived from. Defaults to the send timestamp when `None`.
     ///
     /// Returns the message ID.
     pub fn prepare_message_for_later_publish(
         &self,
         message: &[u8],
         should_push: bool,
+        idempotency_key: Option<String>,
     ) -> Result<Vec<u8>, GroupError> {
         let now = now_ns();
+        // Resolve the key once: a caller-supplied key makes the resulting id
+        // deterministic; otherwise we fall back to the timestamp (always unique).
+        let idempotency_key = idempotency_key.unwrap_or_else(|| now.to_string());
         let queryable_content_fields = Self::extract_queryable_content_fields(message);
 
-        let message_id = calculate_message_id(self.group_id, message, &now.to_string());
+        let message_id = calculate_message_id(self.group_id, message, &idempotency_key);
         let group_message = StoredGroupMessage {
             id: message_id.clone(),
             group_id: self.group_id,
@@ -1259,6 +1265,7 @@ where
             expire_at_ns: None,
             inserted_at_ns: 0,
             should_push,
+            idempotency_key,
         };
         group_message.store(&self.context.db())?;
 
@@ -1294,9 +1301,10 @@ where
             return Ok(());
         }
 
-        // Create envelope from stored message
+        // Create envelope from stored message, reusing the idempotency key the
+        // message id was derived from so receivers recompute the same id.
         let plain_envelope =
-            Self::into_envelope(&message.decrypted_message_bytes, message.sent_at_ns);
+            Self::into_envelope(&message.decrypted_message_bytes, &message.idempotency_key);
         let mut encoded_envelope = vec![];
         plain_envelope.encode(&mut encoded_envelope)?;
 
@@ -1425,20 +1433,24 @@ where
         envelope: F,
     ) -> Result<Vec<u8>, GroupError>
     where
-        F: FnOnce(i64) -> PlaintextEnvelope,
+        F: FnOnce(&str) -> PlaintextEnvelope,
     {
         // Store the message locally first (with should_push preference)
-        let message_id = self.prepare_message_for_later_publish(message, opts.should_push)?;
+        let message_id = self.prepare_message_for_later_publish(
+            message,
+            opts.should_push,
+            opts.idempotency_key,
+        )?;
 
-        // Fetch the stored message to get the sent_at_ns timestamp
+        // Fetch the stored message to get the resolved idempotency key
         let stored_message = self
             .context
             .db()
             .get_group_message(&message_id)?
             .ok_or_else(|| GroupError::NotFound(NotFound::MessageById(message_id.clone())))?;
 
-        // Create envelope using the stored timestamp for consistency
-        let plain_envelope = envelope(stored_message.sent_at_ns);
+        // Create envelope using the stored idempotency key so the id stays consistent
+        let plain_envelope = envelope(&stored_message.idempotency_key);
         let mut encoded_envelope = vec![];
         plain_envelope.encode(&mut encoded_envelope)?;
 
@@ -1452,7 +1464,7 @@ where
         Ok(message_id)
     }
 
-    fn into_envelope(encoded_msg: &[u8], idempotency_key: i64) -> PlaintextEnvelope {
+    fn into_envelope(encoded_msg: &[u8], idempotency_key: &str) -> PlaintextEnvelope {
         PlaintextEnvelope {
             content: Some(Content::V1(V1 {
                 content: encoded_msg.to_vec(),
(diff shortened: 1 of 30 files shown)


crates/xmtp_mls/src/groups/mod.rs
```diff
@@ -1452,7 +1472,7 @@ where
         Ok(message_id)
     }
 
-    fn into_envelope(encoded_msg: &[u8], idempotency_key: i64) -> PlaintextEnvelope {
+    fn into_envelope(encoded_msg: &[u8], idempotency_key: &str) -> PlaintextEnvelope {
```
insipx, line 1475: mostly curious about datatype choice here, why `String`/`str` rather than keeping `i64` and just letting callers choose their own `i64` or even `Vec<u8>` / fixed array?

neekolas (approved):

crates/xmtp_mls/src/groups/mod.rs
mkysel: Good question — the driving constraint is the wire format. The id is recomputed by *receivers* from `PlaintextEnvelope.V1.idempotency_key`, which is (and has always been) a proto3 `string`/UTF-8 field already populated by every client with `now_ns().to_string()`. So whatever we expose at the API has to land as a valid UTF-8 string on the wire.

- **`i64`**: strictly less general. The point of the feature is letting callers use their own keys — UUIDs, request ids, content hashes — none of which are integers. `String` is a superset (you can still pass `"12345"`).
- **`Vec<u8>`/fixed array**: arbitrary bytes aren't valid UTF-8, so they can't go into the proto `string` field without hex/base64-encoding first — i.e. a leaky abstraction that still encodes at the boundary. Doing it *honestly* would mean changing the proto field `string` → `bytes`, which is a breaking protocol change (the field is already in use; older/other clients parse it as a string), so it'd break dedup against existing peers. Out of scope here.
- **`String`**: matches the existing wire field exactly, `calculate_message_id` already takes `&str`, it's UTF-8 by construction (always a valid proto `string`, no validation/encoding needed), and it maps cleanly to JS/Kotlin/Swift `String` across the three bindings rather than `Uint8Array`/`ByteArray`/`Data`. A caller with binary data can hex/base64 it into a string (which they'd have to do anyway for the UTF-8 wire field).

So `String` isn't arbitrary — it's the most general option that fits the locked `string` wire contract. `Vec<u8>` only wins if we migrate the proto to `bytes`, which isn't worth a breaking change for this.

insipx (approved):

PR #3762, opened 2026-06-11
feat(messages): expose idempotency_key on top-level send helpers

mkysel:
## Summary

Follow-up to #3756 (closes the convenience-helper gap from #3750).

#3756 added a caller-settable `idempotency_key` but only on the low-level generic `send(content, opts)` / `prepareMessage`. Every top-level convenience helper (`send_text`, `send_reply`, `send_remote_attachment`, …) still hardcoded `idempotency_key: None`, so a caller wanting idempotent (crash-safe / dedup-able) resends had to drop to the generic `send` and hand-encode the content. This exposes the key on all of them so it's a one-liner.

Because the message id is `calculate_message_id(group_id, content, key)` and the key rides the wire (`PlaintextEnvelope.V1.idempotency_key`), a stable caller key makes a resend of identical content produce the **same** id, which every consumer already dedups (libxmtp + receive-side PK upsert). Downstream agents (herald-lite#106, convos-assistants#2067) need this for crash-safe resends.

## Changes

**Node** (`bindings/node/src/conversation/`)
- New `#[napi(object)] SendOpts { optimistic?, idempotencyKey? }` bag.
- All 12 `send_*` helpers now take `opts?: SendOpts` (replacing the trailing positional `optimistic` bool), building full opts via a shared `SendMessageOpts::from_send_opts(codec_should_push, opts)`.

**Wasm** (`bindings/wasm/src/conversation.rs`)
- Mirror `Tsify SendOpts` + the same 12-helper rewrite.

**iOS** (`sdks/ios/Sources/XMTPiOS/`) / **Android** (`sdks/android/.../library/`)
- Added `idempotencyKey` to `SendOptions` **and** `MessageVisibilityOptions` (→ `toFfi()`), threaded through `encodeContent` and the `prepareMessage` `noSend` branch (was hardcoded `nil`/`null`). No uniffi regen — `FfiSendMessageOpts` already carries the field from #3756.

**Tests**
- Same-key→same-id / different-key→different-id determinism tests in both node and wasm suites.
- Updated the one positional caller `sendText('gm', true)` → `sendText('gm', { optimistic: true })`.

## Notes

- ⚠️ **Breaking** for direct binding consumers passing `optimistic` positionally to a `send_*` helper — now `{ optimistic }` in the opts bag. Chosen over an appended positional param to keep the signature extensible (per ticket design preference).
- Default behavior unchanged when no key is passed (core still falls back to `now_ns()`).
[...]

PR #3720, opened 2026-06-04
Migrate browser-sdk + node-sdk from xmtp-js into libxmtp

insipx:
## Summary

Migrates `@xmtp/browser-sdk` and `@xmtp/node-sdk` out of `xmtp/xmtp-js` and into this monorepo at `sdks/js/`, alongside the Rust bindings they consume. The SDKs now source the locally-built bindings via Yarn Berry `portal:` instead of published npm pins, with sharded CI that reuses the nix-cached binding builds.

- **History preserved**: `git filter-repo` extracted both folders (415 commits, 17 authors, dates intact) and `jj git fetch` merged them in. The import is the merge commit's second parent.
- **Workspace**: new Yarn Berry root at `sdks/js/` (node-modules linker). SDKs reference bindings via `portal:../../../bindings/{node,wasm}`; `sdks/js/dev/bindings` nix-builds the test-utils binding variants and stages their `dist/` so `portal:` resolves real packages.
- **just**: `mod js 'sdks/js/js.just'` with `install`/`bindings`/`check`/`build`/`lint`/`test`/`test-{node,browser}-sdk-ci` recipes.
- **CI**: `test-{node,browser}-sdk.yml` reusable workflows run sharded vitest (node x2, browser x4). Wired into `test.yml`, gated on `test-node`/`test-wasm` (`!cancelled() && result != 'failure'`) so they run after the binding jobs (nix cache hit, no Rust rebuild) and proceed unless the binding job failed. `lint-js.yml` runs eslint + tsc.
- **Formatting unified**: `sdks/js` is formatted by treefmt prettier (`nix/fmt.nix`), so it's in the same `nix fmt` / flake-check / `lint-config` gate as everything else. The SDK eslint config drops the prettier plugin to avoid double-enforcement.
- **Release**: `browser-sdk`/`node-sdk` registered in `dev/release-tools` (`versionTrack: independent`); `release-{browser,node}-sdk.yml` build the SDK bundle against the nix-built bindings and publish via `npm-publish.yml`.

## Verification

- node-sdk: **179/179** tests pass
- browser-sdk: **195/208** pass on shard 1/4 (Playwright + portal-linked wasm working)
- release-tools: **163/163** tests pass
- `just lint-rust` (clippy + fmt + hakari), `just lint-config` (treefmt incl. JS), `just lint-markdown`, eslint + tsc: all clean

[...]

PR #3764, opened 2026-06-12
feat(js-sdk): expose idempotencyKey on top-level send helpers

mkysel:
## Summary

Follow-up to #3762, which exposed `idempotency_key` on the **binding** `send_*` helpers (node + wasm). The JS **SDKs** (`@xmtp/node-sdk`, `@xmtp/browser-sdk`, now in this repo per #3720) still only forwarded `optimistic`, so SDK callers wanting idempotent (crash-safe / dedup-able) resends couldn't reach the key exposed one layer below. This closes that gap.

Because the message id is `calculate_message_id(group_id, content, key)` and the key rides the wire, a stable caller key makes a resend of identical content produce the **same** id, which every consumer already dedups (libxmtp + receive-side PK upsert).

## Changes

All 12 `send*` convenience helpers now take the binding's `SendOpts` bag (`{ optimistic?, idempotencyKey? }`) in place of the trailing `isOptimistic?: boolean`, mirroring #3762.

**node-sdk**
- `Conversation.ts` — 12 helpers pass `SendOpts` straight through (inherited by `Group`/`Dm`).
- Re-export `SendOpts` from `index.ts`.

**browser-sdk** (threaded across all 4 layers of the worker boundary)
- `Conversation.ts` — 12 public helpers put `opts` on the worker action payload.
- `types/actions/conversation.ts` — 12 action `data` payloads carry `opts?: SendOpts`.
- `workers/client.ts` — 12 dispatch cases forward `data.opts`.
- `WorkerConversation.ts` — 12 helpers pass `opts` to the wasm binding.
- Re-export `SendOpts` from `index.ts`.

**Docs**
- Fixed stale `send()` JSDoc (`isOptimistic` → `optimistic`, the actual `SendMessageOpts` field) and documented `idempotencyKey`.

**Tests**
- Same-key→same-id / different-key→different-id determinism tests in both SDKs' `Group.test.ts`.
- Updated positional callers `sendText(x, true)` → `sendText(x, { optimistic: true })` (node + browser, Group + Dm).

## Notes

- ⚠️ **Breaking** for SDK callers passing `optimistic` positionally to a `send*` helper — now `{ optimistic }` in the opts bag. Chosen over an appended positional param for extensibility, consistent with the binding change in #3762.
- Default behavior unchanged when no key is passed (core falls back to `now_ns()`).
- The portal-linked bindings (`@xmtp/node-bindings`, `@xmtp/wasm-bindings`) are gitignored build artifacts rebuilt by CI via Nix; local typecheck wasn't possible without Nix (same constraint as #3762). Prettier verified clean locally. **Verify in CI:** `just node test`, `just wasm test`, `just lint`.

> [!NOTE]
> ### Replace `isOptimistic` boolean with `SendOpts` object on all `Conversation` send helpers
> - All typed send methods (`sendText`, `sendMarkdown`, `sendReaction`, `sendReply`, etc.) in both the browser and node SDKs now accept a `SendOpts` object instead of a bare `boolean` for the optimistic flag.
> - `SendOpts` carries both `optimistic` and `idempotencyKey` fields, enabling callers to supply a deterministic key for deduplication.
> - `SendOpts` is re-exported from both `@xmtp/browser-sdk` and `@xmtp/node-sdk` so consumers can import it directly.
> - Tests are updated to pass `{ optimistic: true }` and new tests verify that repeated sends with the same `idempotencyKey` produce identical message IDs.
> - Behavioral Change: the second parameter of all typed send helpers is now an options object; passing a raw boolean will break at the TypeScript level.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0b871b7.</sup>
> 
>

### DIFF
--- a/sdks/js/browser-sdk/src/Conversation.ts
+++ b/sdks/js/browser-sdk/src/Conversation.ts
@@ -10,6 +10,7 @@ import {
   type RemoteAttachment,
   type Reply,
   type SendMessageOpts,
+  type SendOpts,
   type TransactionReference,
   type WalletSendCalls,
   type DecodedMessage as XmtpDecodedMessage,
@@ -160,8 +161,10 @@ export class Conversation<ContentTypes = unknown> {
    * @param options - Optional send options
    * @param options.shouldPush - Indicates whether this message should be
    * included in push notifications
-   * @param options.isOptimistic - Indicates whether this message should be
+   * @param options.optimistic - Indicates whether this message should be
    * sent optimistically and published later via `publishMessages`
+   * @param options.idempotencyKey - Optional idempotency key; re-sending
+   * identical content with the same key produces the same deduplicated message id
    * @returns Promise that resolves with the message ID after it has been sent
    */
   async send(content: EncodedContent, options?: SendMessageOpts) {
@@ -176,14 +179,14 @@ export class Conversation<ContentTypes = unknown> {
    * Sends a text message
    *
    * @param text - The text to send
-   * @param isOptimistic - Whether to send the message optimistically
+   * @param opts - Send options (optimistic delivery, idempotency key)
    * @returns Promise that resolves with the message ID after it has been sent
    */
-  async sendText(text: string, isOptimistic?: boolean) {
+  async sendText(text: string, opts?: SendOpts) {
     return this.#worker.action("conversation.sendText", {
       id: this.#id,
       text,
-      isOptimistic,
+      opts,
     });
   }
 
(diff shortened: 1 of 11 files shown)


insipx (approved):

sdks/js/browser-sdk/src/Conversation.ts
```diff
@@ -176,189 +179,186 @@ export class Conversation<ContentTypes = unknown> {
...
 
   /**
    * Sends a read receipt message
    *
-   * @param isOptimistic - Whether to send the message optimistically
+   * @param opts - Send options (optimistic delivery, idempotency key)
    * @returns Promise that resolves with the message ID after it has been sent
    */
-  async sendReadReceipt(isOptimistic?: boolean) {
+  async sendReadReceipt(opts?: SendOpts) {
```
neekolas, line 229: This is objectively a better API with more flexibility. Does make it a breaking change. Could type all these APIs `opts?: SendOpts | boolean`and have a helper that converts the value at runtime to `sendOpts` so we don't accidentally break anything.

neekolas (approved):

sdks/js/browser-sdk/src/Conversation.ts
mkysel: I am cleaning up all the callers as we upgrade. Is there anything else other than `convos-cli` and `herald-lite`?  We will have to call it out if we release a next version for public adoption.

PR #3767, opened 2026-06-12
fix(xmtp_db): make TestInstrumentation database-lock panic opt-in

### LINKED ISSUE #3765: Flaky CI Failure: node-sdk shard 2 — database locked panics crash Vitest workers
## Flaky CI Failure: node-sdk shard 2 — database locked panics crash Vitest workers

**Workflow:** Test
**Failed run:** https://github.com/xmtp/libxmtp/actions/runs/27422730244
**Commit:** 3f65d257abe5932d4595a1733c62bc25812f1e9f
**Failed jobs:** test-node-sdk / Test (node-sdk) shard 2

### Summary

The `test-node-sdk` shard 2 job failed because multiple Rust threads panicked with `database locked` from `crates/xmtp_db/src/encrypted_store/database/instrumentation.rs:36`. These panics crashed 6 Vitest fork worker processes, leaving the corresponding test files unable to complete. The remaining 86 tests passed (183 total tracked), but the 6 unhandled `Worker exited unexpectedly` errors caused Vitest to exit with code 1.

### Error Details

```
thread 'r2d2-worker-2' (7740) panicked at crates/xmtp_db/src/encrypted_store/database/instrumentation.rs:36:29:
database locked
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'r2d2-worker-2' (7778) panicked at crates/xmtp_db/src/encrypted_store/database/instrumentation.rs:36:29:
database locked

thread 'r2d2-worker-2' (7721) panicked at crates/xmtp_db/src/encrypted_store/database/instrumentation.rs:36:29:
database locked

thread 'r2d2-worker-0' (7773) panicked at crates/xmtp_db/src/encrypted_store/database/instrumentation.rs:36:29:
database locked

thread 'tokio-rt-worker' (7534) panicked at crates/xmtp_db/src/encrypted_store/database/instrumentation.rs:36:29:
database locked

thread 'r2d2-worker-2' (7797) panicked at crates/xmtp_db/src/encrypted_store/database/instrumentation.rs:36:29:
database locked
```

Resulting in 6 unhandled Vitest errors:
```
Error: [vitest-pool]: Worker forks emitted error.
Caused by: Error: Worker exited unexpectedly
```

Timed-out test files (workers could not be terminated gracefully):
```
[vitest-pool]: Timeout terminating forks worker for test files .../test/DeviceSync.test.ts.
[vitest-pool]: Timeout terminating forks worker for test files .../test/permissions.test.ts.
[vitest-pool]: Timeout terminating forks worker for test files .../test/Conversations.test.ts.
[vitest-pool]: Timeout terminating forks worker for test files .../test/Group.test.ts.
[vitest-pool]: Timeout terminating forks worker for test files .../test/Preferences.test.ts.
[vitest-pool]: Timeout terminating forks worker for test files .../test/contentTypes.test.ts.
```

Final summary: `9 passed (15), 86 tests passed (183), 6 errors`

### Analysis

The `r2d2` thread pool workers and a `tokio-rt-worker` are all hitting an SQLite `database locked` error, which `instrumentation.rs:36` treats as a panic rather than a recoverable error. This suggests multiple concurrent node-sdk test processes are trying to open/write to the same SQLite database file simultaneously without proper serialization.

The crash happens during the `test/inboxId.test.ts` → `test/libxmtpErrors.test.ts` transition window (around 14:44:44 UTC), when several test workers start up in parallel after early fast tests finish — the most likely moment for a write-write lock contention if multiple processes share or race on the same database path.

This is a **test/code failure**, not an infrastructure failure. Root cause hypothesis: test isolation is insufficient in shard 2 — concurrent Vitest workers may share a database file (or derive the same path from a shared fixture), causing SQLite WAL/exclusive-lock conflicts that the instrumentation layer converts into panics rather than retryable errors.

---

xmtp-coder-agent:
Resolves https://github.com/xmtp/libxmtp/issues/3765

## Problem

`TestInstrumentation` (`crates/xmtp_db/src/encrypted_store/database/instrumentation.rs`) called `panic!("database locked")` on **any** query that failed with `database is locked`. This instrumentation is installed on every SQLite connection whenever the `test` cfg **or** the `test-utils` feature is active — and the node-sdk Vitest suite (`bindings_node`) enables `test-utils`.

Under concurrent test load (CI node-sdk shard 2), Vitest fork workers transiently contend on SQLite locks. Each `database is locked` error panicked on an `r2d2-worker`/`tokio-rt-worker` thread, aborting the worker process and surfacing as flaky `Worker exited unexpectedly` failures with a non-zero Vitest exit code.

`database is locked` is **not** fatal in this codebase: SQLite is configured with `PRAGMA busy_timeout = 5000` and the error is explicitly classified as retryable (`PlatformStorageError::DatabaseLocked => true`; `ClientError` retries on `"database is locked"`). The panic short-circuited that designed retry/backoff machinery, turning a recoverable transient error into a hard crash.

## Fix

Gate the panic behind an opt-in `XMTP_PANIC_ON_DB_LOCK` environment variable (accepts `1`/`true`, matching the existing `SQLCIPHER_LOG` convention):

- **Default (unset):** log the query error at error level and return normally — the existing busy_timeout/retry machinery recovers. No worker crashes.
- **`XMTP_PANIC_ON_DB_LOCK=1`:** preserve the legacy panic for developers explicitly debugging lock contention.

The env read is split from a pure decision helper (`panic_on_db_lock_from_env`) so the opt-in logic is unit-tested without mutating the process environment (Rust 2024 `set_var` is unsafe/racy).

## Unchanged behavior
- Query-error logging (message/table/column/hint) is preserved.
- `DatabaseLocked` remains classified as retryable.
- Begin/Commit/Rollback transaction tracing is unchanged.

## Testing
- New unit tests in `instrumentation.rs` cover: panic disabled by default (unset env), opt-in values (`1`/`true`), and that other values (`0`, `false`, empty, `TRUE`, etc.) do not enable it.
- `cargo test -p xmtp_db --lib instrumentation` → 3 passed.
- `cargo clippy -p xmtp_db --all-features --tests` → clean.
- `cargo fmt -p xmtp_db --check` → clean.

> [!NOTE]
> ### Make `TestInstrumentation` database-lock panic opt-in via `XMTP_PANIC_ON_DB_LOCK` env var
> - Panicking on SQLite "database is locked" errors in `TestInstrumentation` now requires `XMTP_PANIC_ON_DB_LOCK=1` or `XMTP_PANIC_ON_DB_LOCK=true`; previously any occurrence would panic unconditionally.
> - Adds `panic_on_db_lock_enabled` and `panic_on_db_lock_from_env` helpers in [instrumentation.rs](https://github.com/xmtp/libxmtp/pull/3767/files#diff-3f2f010aa407d2e385927c144accbd35ac3b8eb13e576e302dcedbb71aec28d8) to read and evaluate the env var.
> - Sets `XMTP_PANIC_ON_DB_LOCK=true` in the `justfile` test recipes and the [nextest.nix](https://github.com/xmtp/libxmtp/pull/3767/files#diff-aa45367a8dcf70261a484bcc1f097c399301d2c6819f730c07781fb8936eed8a) package so CI and local test runs still opt in.
> - Fixes a flaky test in [stream_messages.rs](https://github.com/xmtp/libxmtp/pull/3767/files#diff-cffe5828e751e2a749629b6acdb0a9a09b407756b9b89f3042a1ecf28055aeb6) by filtering the message stream to application messages only, avoiding races with membership-change commits.
> - Behavioral Change: the default behavior no longer panics on lock contention; environments outside the configured test runners must explicitly set the env var to preserve the old behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 36e833a.</sup>
> 
>

### DIFF
--- a/crates/xmtp_db/src/encrypted_store/database/instrumentation.rs
+++ b/crates/xmtp_db/src/encrypted_store/database/instrumentation.rs
@@ -3,11 +3,32 @@ use diesel::connection::InstrumentationEvent;
 use diesel::result::Error as DieselError;
 use std::fmt::Write;
 
-// Test instrumentatiomn
+// Test instrumentation
 // prints out query on error
-// panics on database lock
+// panics on database lock ONLY when explicitly opted in via the
+// `XMTP_PANIC_ON_DB_LOCK` env var (accepts "1" or "true").
+//
+// `database is locked` is a transient, retryable condition in this codebase
+// (SQLite is configured with `PRAGMA busy_timeout` and the error is classified
+// as retryable). Panicking on it by default aborts the worker thread and crashes
+// downstream test harnesses (e.g. the node-sdk Vitest fork workers), so the panic
+// is opt-in for developers actively debugging lock contention.
 #[allow(unused)]
 pub struct TestInstrumentation;
+
+/// Returns true when the caller has explicitly opted in to panicking on
+/// `database is locked` via the `XMTP_PANIC_ON_DB_LOCK` env var.
+fn panic_on_db_lock_enabled() -> bool {
+    panic_on_db_lock_from_env(std::env::var("XMTP_PANIC_ON_DB_LOCK"))
+}
+
+/// Pure decision helper: the opt-in is enabled only when the env var resolves to
+/// exactly `"1"` or `"true"`. Anything else (unset, empty, other values) keeps the
+/// default non-panicking behavior. Kept separate from the env read so it can be
+/// unit-tested without mutating the process environment.
+fn panic_on_db_lock_from_env(var: Result<String, std::env::VarError>) -> bool {
+    matches!(var, Ok(s) if s == "1" || s == "true")
+}
 
 impl Instrumentation for TestInstrumentation {
     fn on_connection_event(&mut self, event: InstrumentationEvent<'_>) {
@@ -32,7 +53,7 @@ impl Instrumentation for TestInstrumentation {
                             tracing::error!("details: {},", details);
                         }
                         tracing::error!("{}", s);
-                        if s.contains("database is locked") {
+                        if s.contains("database is locked") && panic_on_db_lock_enabled() {
                             panic!("database locked");
                         }
                     }
@@ -49,6 +70,35 @@ impl Instrumentation for TestInstrumentation {
             }
 
             _ => (),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::panic_on_db_lock_from_env;
+    use std::env::VarError;
+
+    #[test]
+    fn db_lock_panic_disabled_by_default() {
+        // Unset env var (the default in CI / node-sdk Vitest workers) must NOT
+        // opt in to panicking, so a `database is locked` error stays recoverable.
+        assert!(!panic_on_db_lock_from_env(Err(VarError::NotPresent)));
+    }
+
+    #[test]
+    fn db_lock_panic_opt_in_values() {
+        assert!(panic_on_db_lock_from_env(Ok("1".to_string())));
+        assert!(panic_on_db_lock_from_env(Ok("true".to_string())));
+    }
+
+    #[test]
+    fn db_lock_panic_ignores_other_values() {
+        for v in ["0", "false", "", "yes", "TRUE", "True"] {
+            assert!(
+                !panic_on_db_lock_from_env(Ok(v.to_string())),
+                "value {v:?} should not enable the panic"
+            );
         }
     }
 }
--- a/crates/xmtp_mls/src/subscriptions/stream_messages.rs
+++ b/crates/xmtp_mls/src/subscriptions/stream_messages.rs
@@ -596,6 +596,7 @@ pub mod tests {
     use crate::tester;
     use futures::stream::StreamExt;
     use rstest::*;
+    use xmtp_db::group_message::GroupMessageKind;
 
     #[xmtp_common::timeout(std::time::Duration::from_secs(30))]
     #[rstest]
@@ -613,15 +614,24 @@ pub mod tests {
         let bob_group = bob_groups.first().unwrap();
         alice_group.sync().await.unwrap();
 
-        let stream = alice_group.stream().await.unwrap();
+        // The `group_updated` membership-change commit from `add_members` above can
+        // race into the stream ahead of the application messages, which made this
+        // test flaky (see xmtp/libxmtp#3765). Filter to application messages so the
+        // assertions below are deterministic regardless of whether the commit
+        // message is delivered through the stream.
+        let stream = alice_group.stream().await.unwrap().filter(|msg| {
+            futures::future::ready(
+                msg.as_ref()
+                    .map(|m| m.kind == GroupMessageKind::Application)
+                    .unwrap_or(true),
+            )
+        });
         futures::pin_mut!(stream);
         bob_group
             .send_message(b"hello", SendMessageOpts::default())
             .await
             .unwrap();
 
-        // group updated msg/bob is added
-        // assert_msg_exists!(stream);
         assert_msg!(stream, "hello");
 
         bob_group
--- a/justfile
+++ b/justfile
@@ -85,11 +85,11 @@ _test-all *args="": (_test-v3 args) (_test-d14n args)
 
 [private]
 _test-v3 *args="":
-    {{ cargo_test }} --profile ci {{ args }}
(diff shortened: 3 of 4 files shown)


xmtp-coder-agent:
## CI fix: deflaked `test_stream_messages`

The Rust Workspace CI run surfaced a failure in `subscriptions::stream_messages::tests::test_stream_messages` — **pre-existing flakiness unrelated to the instrumentation change** (this PR's only other diff is in `xmtp_db`).

**Root cause:** the test opened a message stream after `add_members(bob)` and asserted the first streamed message equaled `"hello"`. The `group_updated` membership-change commit can race into the stream ahead of the application message, so the assertion intermittently saw the commit message (`left: group_updated…`, `right: "hello"`). The test even had a commented-out `assert_msg_exists!` acknowledging this commit message.

**Fix:** filter the stream to `GroupMessageKind::Application` messages so the assertions are deterministic regardless of whether the commit is delivered through the stream (errors pass through unchanged).

**Verification:** `test_stream_messages` passed **10/10** consecutive runs against a live node; `cargo check --tests`, `clippy --tests`, and `fmt --check` all clean for `xmtp_mls`.

This is in-scope with the PR's flaky-CI theme.

crates/xmtp_db/src/encrypted_store/database/instrumentation.rs
```diff
@@ -3,12 +3,33 @@ use diesel::connection::InstrumentationEvent;
...
+// as retryable). Panicking on it by default aborts the worker thread and crashes
+// downstream test harnesses (e.g. the node-sdk Vitest fork workers), so the panic
+// is opt-in for developers actively debugging lock contention.
 #[allow(unused)]
 pub struct TestInstrumentation;
 
+/// Returns true when the caller has explicitly opted in to panicking on
+/// `database is locked` via the `XMTP_PANIC_ON_DB_LOCK` env var.
+fn panic_on_db_lock_enabled() -> bool {
+    panic_on_db_lock_from_env(std::env::var("XMTP_PANIC_ON_DB_LOCK"))
```
insipx, line 22: this makes it default off, so libxmtp workspace test and others which used to have panic on lock now have it off by default. i still want those to panic. can you turn it on for CI please? @xmtp-coder-agent it can stay off for nodesdk tests ONLY

insipx:
@xmtp-coder-agent maybe instead of this env var, node sdk should use an in memory database for tests wherever possible? is the node sdk currently reusing the same db for each of its tests rather than creating a new one? then we can keep test using the panic. the panic is very useful for debuggin race conditions and should be default on.

crates/xmtp_db/src/encrypted_store/database/instrumentation.rs
xmtp-coder-agent: Done — turned the panic back on for the Rust workspace tests while keeping it off for node-sdk only.

Since both the workspace tests and node-sdk compile `xmtp_db` with `test-utils`, the split is made at the test-runner environment rather than at compile time:

- **CI workspace tests:** set `XMTP_PANIC_ON_DB_LOCK=true` in `nix/package/nextest.nix` (covers the v3 + d14n `test-workspace` job). Verified with `nix eval .#nextest.x86_64-linux.v3.XMTP_PANIC_ON_DB_LOCK` → `true`, so it reaches the derivation build env.
- **Local `just test`:** set it in the `_test-v3` / `_test-d14n` recipes (the `ci` / `ci-d14n` nextest profiles), so local workspace runs panic too.
- **node-sdk / node-bindings:** run via `just node test*` (Vitest) and never set the var → stay off by default, so no more crashed fork workers.

Commit: 36e833ae1. Let me know if you'd also like it on for any other CI job (e.g. test-node).

insipx:
Superseded by #3785 — same root cause, inverted default.

Instead of gating the panic behind an opt-**in** `XMTP_PANIC_ON_DB_LOCK` (default off, set on in `justfile` + `nextest.nix`), #3785 keeps the db-lock panic **on by default** and lets consumers driving the full client opt **out** via `XMTP_NO_PANIC_ON_DB_LOCK`, set in the node-sdk / bindings-node `vitest.config` `test.env` (which reaches the fork workers; `globalSetup` does not).

Rationale: a held lock in a pure Rust test is a real bug, so `cargo test` / `nextest` should panic with zero config. One opt-out at the single broken context beats opt-in env vars scattered across runner configs that get lost. The `stream_messages` application-message filter from here is carried over.

Closing in favor of #3785.

PR #3772, opened 2026-06-12
feat(xmtp_api_d14n): BidiConnection actor — own both wire halves, internal liveness

tylerhawkes:
## What

A bidirectional subscription **connection** actor (XIP-83), living at the transport layer in `xmtp_api_d14n` next to the `subscribe_bidi` transport. It owns a single bidi stream end-to-end and is the **sole writer** of the request half.

This supersedes the original per-subscription actor that lived in `xmtp_mls` (`BidiSubscription`): that design held an outbound sender purely to auto-pong, which decoupled the request half's lifetime from the stream's and let a post-close `mutate`/`ping` succeed silently (Macroscope flagged exactly this). Pushing the actor down to where the wire lives fixes it **by ownership** rather than with a side-signal.

## Design

- **Sole owner of both halves.** Callers submit `Command::{Mutate, Probe}` over a channel; the actor multiplexes them with auto-pongs onto the one outbound FIFO. No second outbound sender exists anywhere.
- **Liveness is internal.** Server `Ping` → auto-`Pong`, never surfaced. `probe()` sends a client `Ping` and the actor correlates the returning `Pong` via a oneshot. `BidiEvent` has no `Ping`/`Pong` arm — consumers deal with zero keepalive.
- **Teardown by ownership.** Inbound end/error → the actor drops `wire_out` (request half closes, HTTP/2 torn down both ways), the command receiver (→ `mutate`/`probe` return `Closed`), the event sender (→ `next()` ends), and any pending probe acks. One place, every teardown — the "enqueue into a dead stream" gap can't exist.
- **Bounded probe.** `probe()` self-bounds at `N × keepalive_interval_ms` (N=3; 30s fallback until `Started`, whose cadence the actor captures). `probe_within(Duration)` lets a latency-sensitive caller — e.g. a push-notification handler — bound to seconds. A timeout returns the distinct `ProbeTimedOut` → drop + re-open; tight bounds are safe because resume-from-cursors dedupes.

## Notes

- Native-only (`#[cfg(not(target_arch = "wasm32"))]`), generic over `XmtpMlsBidiStreams` — mock-tested, no deployed node required.
- This is the Phase-1 primitive the herald connection-multiplexer (Phase 2) drives per shard; durability / reconnection stays at the mux / consumer layer, not coupled into the actor.
- Production `Topic` parsing uses the validating `TryFrom` (not the test-only `from_bytes`); malformed `TopicsLive` topics are skipped with a warn.
- 11 unit tests: ownership-teardown, auto-pong-not-surfaced, probe round-trip / timeout / keepalive-derived default, wire-order, unknown-version tolerance.

> [!NOTE]
> ### Add `BidiConnection` actor for bidirectional subscription streams in `xmtp_api_d14n`
> - Adds [connection.rs](https://github.com/xmtp/libxmtp/pull/3772/files#diff-11d1b237ec81cff558f3bbfc0b5f8477bd670265b0fcbe1cd6b700abcfddf1f3), a native-only module exposing `BidiConnection`, a handle that owns both inbound and outbound halves of a gRPC bidi stream.
> - The actor sends an initial `Mutate` first, auto-responds to server `Ping` frames, and correlates client `probe()`/`probe_within()` calls to `Pong` responses for liveness checking.
> - Outbound frames are buffered in a `VecDeque` when the transport is not draining; if the backlog exceeds `MAX_PENDING_FRAMES`, the actor tears down the connection.
> - `BidiEvent` surfaces structured events (`Started`, `CatchUpComplete`, `TopicsLive`, `GroupMessages`, `WelcomeMessages`) while hiding keepalive frames from callers.
> - Dropping `BidiConnection` aborts the actor and closes the stream; the module is excluded from wasm32 targets.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 893718b.</sup>
> 
>

### DIFF
--- a/crates/xmtp_api_d14n/src/queries/v3/connection.rs
+++ b/crates/xmtp_api_d14n/src/queries/v3/connection.rs
@@ -1,0 +1,1044 @@
+//! XIP-83 bidirectional subscription *connection* (native-only).
+//!
+//! This sits one layer below the application: it owns a single bidi stream
+//! end-to-end and is the **sole writer** of the request half. It auto-answers
+//! server `Ping`s, correlates client liveness probes, and forwards only real
+//! subscription events upward — keepalive never reaches the consumer.
+//!
+//! Owning *both* halves is the point. The actor holds the wire-outbound sender
+//! and the inbound stream; when inbound dies it drops both, so the request half
+//! tears down with it and any later `mutate`/`probe` fails with `Closed` by
+//! channel ownership — never by silently enqueueing into a stream nothing reads.
+
+use std::collections::{HashMap, VecDeque};
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU32, AtomicU64, Ordering};
+use std::time::Duration;
+
+use futures::StreamExt;
+use tokio::sync::{mpsc, oneshot};
+use xmtp_common::{AbortHandle, StreamHandle};
+use xmtp_proto::api_client::XmtpMlsBidiStreams;
+use xmtp_proto::mls_v1::subscribe_request::v1::Mutate;
+use xmtp_proto::mls_v1::{
+    GroupMessage, Ping, Pong, SubscribeRequest, SubscribeResponse, WelcomeMessage,
+    subscribe_request, subscribe_response,
+};
+use xmtp_proto::types::Topic;
+
+/// Wire-outbound depth. The actor is the sole writer; if the transport stalls,
+/// `send().await` here backpressures the actor, which backpressures the command
+/// queue — exactly the signal `probe` turns into a fast `Closed`.
+const WIRE_BUFFER: usize = 64;
+/// Caller→actor command depth.
+const COMMAND_BUFFER: usize = 64;
+/// Actor→caller event depth; large enough that a brief consumer stall doesn't
+/// stall wire reads (and thus pong liveness).
+const EVENT_BUFFER: usize = 1024;
+/// XIP-83 client req 2 fallback keepalive, used until the server's `Started`
+/// frame advertises its own cadence.
+const DEFAULT_KEEPALIVE_MS: u32 = 30_000;
+/// `N` from XIP-83 client req 2 (recommended 2–3): the *default* probe deadline
+/// is this many keepalive intervals — generous enough not to false-positive a
+/// slow-but-live link. Latency-sensitive callers (e.g. a notification handler)
+/// pass a much smaller bound to [`BidiConnection::probe_within`].
+const PROBE_TIMEOUT_MULTIPLIER: u32 = 3;
+/// Hard cap on the outbound backlog. Past this, the wire has been wedged long
+/// enough that buffering more is pointless — the transport isn't draining the
+/// request half, so the link is effectively dead — and the actor gives up,
+/// tearing down so the consumer re-opens from cursors on a fresh stream. Sits
+/// well above the command-gated backlog (`WIRE_BUFFER`), so only a sustained
+/// stall (pongs piling up on a half-open link) ever reaches it.
+const MAX_PENDING_FRAMES: usize = WIRE_BUFFER * 2;
+
+/// Events surfaced to the consumer, in wire order. `Ping`/`Pong` never appear —
+/// liveness lives entirely inside the actor.
+#[derive(Debug, Clone, PartialEq)]
+pub enum BidiEvent {
+    /// First frame on every stream; carries the server's keepalive cadence.
+    Started {
+        keepalive_interval_ms: u32,
+        capabilities: Vec<i32>,
+    },
+    /// A `Mutate`'s adds are fully caught up; echoes the Mutate's `mutate_id`.
+    CatchUpComplete {
+        mutate_id: u64,
+    },
+    /// These topics just crossed from catch-up to live.
+    TopicsLive {
+        topics: Vec<Topic>,
+    },
+    GroupMessages(Vec<GroupMessage>),
+    WelcomeMessages(Vec<WelcomeMessage>),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum BidiError {
+    #[error("the bidi connection is closed; re-open and resume from durable cursors")]
+    Closed,
+    #[error("liveness probe timed out; treat the link as dead, drop it, and re-open")]
+    ProbeTimedOut,
+}
+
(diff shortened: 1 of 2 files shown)


insipx:
one thing we'll want to figure out with BiDi is how we handle any potential difference in `V3`/`d14n` APIs while also handling MLS logic. Mostly `v3` and `d14n` differ in how they handle their proto types/what they do with those types handled by the `Extractor` system in d14n. That's where we define common Stream types for each version. 

It could make the delineation between MLS and Bidi stream handling easier than if raw Bidi stream handling is implemented in `xmtp_api_d14n` or even its own stream crate and `xmtp_mls` is given only a small interface surface to interact with it, then we can test both the mls and stream parts separately, and also compose the interface `xmtp_mls` receives with the `xmtp_proto::types` unified `GroupMessage`/`Group` etc types to avoid dealing with raw protobuf types in the more easily complicated MLS code, and also avoid worrying about manually reconciling cursor schemes or other differences between v3/d14n (which would then be handled by the extractors)

crates/xmtp_mls/src/subscriptions/bidi.rs
```diff
@@ -0,0 +1,671 @@
...
+            Some(Response::TopicsLive(live)) => {
+                let event = BidiEvent::TopicsLive {
+                    topics: live.topics,
+                };
+                if emit(&events, event).await {
+                    break;
+                }
+            }
+            Some(Response::Messages(messages)) => {
+                if !messages.group_messages.is_empty()
```
insipx: these message types are where we will need to be careful about the v3/d14n differences. based on client type we can re-use the d14n or v3 extractor schemes here. namely, d14n message return from the `D14nClientStreams` include logic for the iceboxing & dependency resolution based on `depends_on` that v3 does not include. If we are going to use this bidi stream across many clients, we will need a new cursor store implementation that is decoupled from MLS (and potentially includes its own on-disk storage solution or sqlite db?) in order to keep the icebox consistent and avoid depending on any single client store


but maybe i'm speaking too soon and need more info/design on how individual clients will interact with the bidi actor, maybe we can pass along the proto and do the per-version processing later on?

crates/xmtp_mls/src/subscriptions/bidi.rs
```diff
@@ -0,0 +1,671 @@
...
+        capabilities: Vec<i32>,
+    },
+    /// A Mutate's added subscriptions have all caught up to live (one per
+    /// adding Mutate); `mutate_id` echoes the Mutate that started the wave,
+    /// so completions stay attributable when waves overlap.
+    CatchUpComplete { mutate_id: u64 },
+    /// These kind-prefixed topics just crossed from catch-up to live: every
+    /// later frame for a listed topic is live tail (e.g. notify for it,
+    /// don't backfill it).
+    TopicsLive { topics: Vec<Vec<u8>> },
```
insipx: `Vec<Topic>` should work

tylerhawkes: Done — `TopicsLive` now carries `Vec<Topic>`, parsed from the wire bytes via `Topic::from_bytes`.

crates/xmtp_mls/src/subscriptions/bidi.rs
tylerhawkes: Agree — taking your "pass the proto along, process per-version later" path, with the extractor scheme + MLS-decoupled cursor store designed into the d14n follow-up after this lands.

tylerhawkes:
@insipx +1 — pushing bidi handling down behind a small unified-types interface to `xmtp_mls` is the right shape, and I'll do it as part of the d14n lane rather than this PR.

crates/xmtp_mls/src/subscriptions/bidi.rs
```diff
@@ -0,0 +1,671 @@
...
+
+    async fn send(
+        &self,
+        request: subscribe_request::v1::Request,
+    ) -> Result<(), BidiSubscriptionError> {
+        self.outbound
+            .send(request_frame(request))
+            .await
+            .map_err(|_| BidiSubscriptionError::Closed)
+    }
```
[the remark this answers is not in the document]
tylerhawkes: Fixed: `ping()` now uses `try_send`, so a full outbound buffer (a stalled transport) returns `Closed` immediately instead of parking — `mutate()` stays on the awaiting `send()` for backpressure. Added a regression test (`ping_fails_fast_when_outbound_buffer_full`) that fills the buffer via a non-draining transport and asserts `Closed` rather than a hang.

crates/xmtp_api_d14n/src/queries/v3/connection.rs
```diff
@@ -0,0 +1,1044 @@
...
+    // reads and delay auto-pongs (getting us reaped on an otherwise healthy link).
+    let mut pending: VecDeque<SubscribeRequest> = VecDeque::new();
+
+    loop {
+        tokio::select! {
+            // Hand one queued frame to the wire the moment it has capacity,
+            // concurrently with everything else. This is what keeps a busy wire
+            // from blocking the loop. Gated on a non-empty queue so we only
+            // contend for a permit when there's something to flush.
+            permit = wire_out.reserve(), if !pending.is_empty() => {
```
insipx, line 324: TIL abt the `reserve()` api

crates/xmtp_api_d14n/src/queries/v3/connection.rs
```diff
@@ -0,0 +1,1044 @@
...
+        // Wrap the receiver as a `Stream` without pulling in `tokio-stream`:
+        // `poll_recv` is exactly the poll fn a `Stream` needs.
+        let outbound = futures::stream::poll_fn(move |cx| wire_out_rx.poll_recv(cx));
+        let inbound = api.subscribe_bidi(Box::pin(outbound)).await?;
+
+        let keepalive_ms = Arc::new(AtomicU32::new(0));
+        let actor = xmtp_common::spawn(
+            None,
+            run_actor(
+                Box::pin(inbound),
```
insipx, line 140: small nit: shouldn't "inbound" be "wire_in" for consistency -- afaiu the `inbound` = messages we're getting and `wire_out` is messages we're sending off into bidi, where `wire` terminology is differentiating it from the in-mem message channels

insipx (approved):

PR #3788, opened 2026-06-26
feat(xmtp_mls): KeyPackageCleaner scheduled-wake worker (kill the 5s poll)

insipx:
## Problem

`KeyPackagesCleanerWorker` polls the DB every **5 seconds** and opens a `worker_turn` span on every tick. The real work is **~monthly** (30-day key-package rotation + a deletion ~1 day after a rotation marks the old KP). So >99.999% of ticks are no-ops, with two costs:

1. **Telemetry:** ~**82M** empty `worker_turn` spans/day (99.9994% of all `worker_turn` volume, measured in Datadog).
2. **Fleet DB load:** two queries/tick × thousands of clients/process on shared encrypted SQLite ≈ **~2,000 queries/sec** of pure no-op polling at 5,000 clients, scaling linearly.

## Approach

Replace the poll with a **scheduled-wake** worker:

- A coarse **1-hour** fallback interval (configurable per-kind via `WorkerConfig`; backstops the 30-day rotation + 1-day deletion).
- A capacity-1 **re-arm channel** nudged via a context `wake_key_package_worker()` facade, fired after `queue_key_rotation` lowers the rotation deadline to `now+5s`.
- On a nudge the worker waits out the **5s debounce** (a deliberate behavior from #2044 — welcomes *queue* a rotation 5s out to collapse bursts and avoid HPKE errors; rotation is **not** done inline) then runs `maintain()`; on an interval tick it runs `maintain()` directly.
- `maintain()` reads two cheap guards, returns early with **no span** when idle, else opens exactly **one** `worker_turn` span wrapping the delete+rotate work (so failures are still recorded).

A parked client now issues ~no maintenance queries and emits no idle spans; `worker_turn` fires only on real work (~2/month/client). The welcome-queued rotation still fires within ~5-10s (rides the wake channel, independent of the interval — so an operator can set a 24h interval on server infra without slowing rotation).

## Notes

- `WorkerKind::KeyPackageCleaner` and the worker registration are **kept** — no bindings break; the supervisor still restarts the worker on error.
- Extracts a shared `RearmChannel` (`DisappearingChannels` becomes an alias).
- An MLS-invariant audit confirmed the coarse interval is safe: XMTP uses last-resort (reusable) key packages, on-network expiry (~84d) is far beyond the 30-day rotation, and deletion is local-only with a 1-day grace where late deletion is strictly safe.
- Test interval override (`WorkerConfig::for_testing()`, 100ms KP) keeps timing-dependent rotation tests fast; new behavior test exercises the queue→wake→rotate pipeline.

> [!NOTE]
> ### Replace 5s poll with wake-channel event loop in `KeyPackageCleaner` worker
> - Replaces the previous 5-second polling interval in `KeyPackageCleaner` with a `RearmChannel` (capacity-1 lossy mpsc) that lets callers explicitly wake the worker; the fallback idle interval is now 1 hour.
> - `queue_key_rotation` and `process_new_welcome` now call `wake_key_package_worker()` after queuing a rotation, triggering a 5-second debounce window before the worker runs maintenance.
> - Introduces `RearmChannel` as a shared primitive in [`worker/rearm_channel.rs`](https://github.com/xmtp/libxmtp/pull/3788/files#diff-9d19a401ec11f9d747ea657bcbe4a74d7a9d161cc4adb9ab04c942e0864f2ed8), replacing the bespoke `DisappearingChannels` struct with a type alias to it.
> - Test builds use a `WorkerConfig::for_testing()` helper that overrides the `KeyPackageCleaner` interval to 100ms for timing-sensitive tests.
> - Behavioral Change: key-package maintenance no longer runs every 5 seconds by default; it runs within ~5 seconds of an explicit wake signal or after 1 hour if idle.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 53e8377.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/worker/disappearing_messages.rs
+++ b/crates/xmtp_mls/src/worker/disappearing_messages.rs
@@ -2,7 +2,6 @@ use crate::context::XmtpSharedContext;
 use crate::worker::{BoxedWorker, NeedsDbReconnect, Worker, WorkerFactory};
 use crate::worker::{WorkerKind, WorkerResult};
 use futures::TryFutureExt;
-use std::sync::Arc;
 use std::time::Duration;
 use thiserror::Error;
 use xmtp_common::time::now_ns;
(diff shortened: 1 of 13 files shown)


crates/xmtp_mls/src/worker/disappearing_messages.rs
```diff
@@ -15,43 +14,9 @@ use xmtp_db::{StorageError, prelude::*};
...
-    /// non-blocking: if a nudge is already queued (slot full) or no worker is
-    /// consuming, the send is dropped — the worker recomputes from the DB on its
-    /// next wake regardless, so a dropped duplicate nudge changes nothing.
-    pub fn rearm(&self) {
-        let _ = self.sender.try_send(());
-    }
-}
+/// Wake channel for the disappearing-messages worker. Alias of the shared
+/// [`RearmChannel`](crate::worker::rearm_channel::RearmChannel).
+pub type DisappearingChannels = crate::worker::rearm_channel::RearmChannel;
```
tylerhawkes, line 19: This can all be done on the existing worker with a specific time out in the future instead of regular polling. If they expire after 90 days, we update the task to roll it to 30 days out every time we create one that way as long as the app wakes up some time in the 60 day window it gets rolled.

tylerhawkes (changes requested):

PR #3791, opened 2026-06-27
feat(xmtp_mls): KeyPackageCleaner sleeps to the exact rotation deadline (no poll)

insipx:
**Draft** — alternative to #3788 implementing @tylerhawkes's review feedback. Opening for review, not yet ready to merge.

## What

Replaces KeyPackageCleaner's **5-second poll** with a worker that **sleeps until the exact next rotation deadline** (`next_key_package_rotation_ns`) — no fixed-interval poll, so sleeping apps aren't woken for work whose time we already know.

This is the approach from the #3788 review:
> *"This can all be done on the existing worker with a specific timeout in the future instead of regular polling. If they expire after 90 days, we update the task to roll it to 30 days out every time we create one — as long as the app wakes up sometime in the 60-day window it gets rolled."*

## Why

The 5s poll cost ~**82M** empty `worker_turn` spans/day (99.9994% of `worker_turn` volume) + ~**2,000 no-op DB queries/sec** at 5,000 clients/process. Real work is ~monthly (30-day rotation; deletion ~1 day after a KP is superseded).

## How

- **`xmtp_common::time::sleep` is now wasm-safe** — it chunks long durations internally so a multi-day sleep doesn't overflow `gloo-timers` (which casts ms to `i32` for JS `setTimeout`; a 30-day sleep would otherwise wrap and fire immediately). Benefits every long sleep in the codebase.
- **The worker sleeps to `next_key_package_rotation_ns`.** A pure `WakePlan { RunNow, SleepUntil(deadline) }` decides; on `RunNow` (NULL/past deadline) it runs maintenance, on `SleepUntil` it `select!`s a re-arm-channel recompute vs `sleep(deadline - now) → maintain()`. Existing code already rolls the rotation deadline 30 days forward after each rotation, so the worker self-perpetuates. Deletion of expired KPs is opportunistic in the work pass (latency-tolerant; local-only with a grace).
- **The ~5s welcome-queued rotation** still fires promptly: `Client::queue_key_rotation` (welcome + public) lowers the deadline to `now+5s` and nudges the worker via a `wake_key_package_worker()` facade → the parked worker recomputes and rotates ~5s later.
- **Span-gated** `maintain()` — one `worker_turn` span only when there's real work, wrapping it so failures are recorded.

`WorkerKind::KeyPackageCleaner` and registration are kept (no bindings break); supervisor restart is the failure backstop.

## Relationship to #3788

#3788 is the 1-hour-fallback version of this worker; this is the exact-deadline version (no fallback poll). If #3788 lands first, this becomes a small follow-up swapping the interval for the deadline sleep.

## MLS safety

Last-resort reusable KPs; on-network expiry ~90d vs 30d rotation = 60-day window; deletion local-only where late deletion is strictly safe. A worker waking at the rotation deadline (or any time in the window) is safe.

> [!NOTE]
> ### Replace polling loop in `KeyPackagesCleanerWorker` with deadline-driven sleep and re-arm channel
> - Replaces the periodic 5-second jittered poll in `KeyPackagesCleanerWorker.run` with an event-driven loop that sleeps until the exact next rotation or deletion deadline, waking early via a new `RearmChannel` when work is queued.
> - Adds `RearmChannel`, a capacity-1 lossy wake channel, replacing the duplicate implementation previously used by `DisappearingMessages`.
> - Adds `next_key_package_rotation_ns` and `min_key_package_delete_at_ns` DB queries to `QueryIdentity` and `QueryKeyPackageHistory` traits so the worker can determine precise sleep durations.
> - `queue_key_rotation` and `WelcomeSync::process_new_welcome` now call `wake_key_package_worker()` to immediately signal the cleaner when rotation is queued.
> - Fixes a wasm `setTimeout` i32 overflow for sleeps longer than ~24.8 days by chunking into 1-day intervals in `xmtp_common::time::sleep`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1e40d81.</sup>
> 
>

### DIFF
--- a/crates/xmtp_common/src/time.rs
+++ b/crates/xmtp_common/src/time.rs
@@ -84,7 +84,10 @@ where
 /// JS `setTimeout` (via gloo-timers) casts the millisecond value to `i32`,
 /// so any duration > ~24.8 days overflows and fires immediately. Chunking into
 /// at-most-1-day pieces lets callers sleep the full requested duration.
-#[allow(dead_code)] // only called from wasm arm; also exercised by unit tests
+///
+/// Only the wasm `sleep` path needs this; compiled for wasm (real use) and under
+/// `test` (native coverage of the pure arithmetic), so it is never dead code.
+#[cfg(any(all(target_family = "wasm", target_os = "unknown"), test))]
 fn sleep_chunks(duration: Duration, max: Duration) -> (u64, Duration) {
     let max_ns = max.as_nanos().max(1);
     let full = (duration.as_nanos() / max_ns) as u64;
--- a/crates/xmtp_common/src/time.rs
+++ b/crates/xmtp_common/src/time.rs
@@ -78,11 +78,34 @@ where
     }
 }
 
+/// Split `duration` into whole `max`-sized chunks plus a remainder.
+/// Returns `(number_of_full_chunks, remainder)`. Pure; used for wasm-safe sleeping.
+///
+/// JS `setTimeout` (via gloo-timers) casts the millisecond value to `i32`,
+/// so any duration > ~24.8 days overflows and fires immediately. Chunking into
+/// at-most-1-day pieces lets callers sleep the full requested duration.
+#[allow(dead_code)] // only called from wasm arm; also exercised by unit tests
+fn sleep_chunks(duration: Duration, max: Duration) -> (u64, Duration) {
+    let max_ns = max.as_nanos().max(1);
+    let full = (duration.as_nanos() / max_ns) as u64;
+    let rem = Duration::from_nanos((duration.as_nanos() % max_ns) as u64);
+    (full, rem)
+}
+
 #[doc(hidden)]
 pub async fn sleep(duration: Duration) {
     wasm_or_native! {
         native => {tokio::time::sleep(duration).await},
-        wasm => {gloo_timers::future::sleep(duration).await},
+        wasm => {
+            // JS setTimeout (via gloo-timers) caps near i32::MAX ms; chunk so a
+            // multi-day sleep elapses fully instead of wrapping to ~0.
+            const MAX_CHUNK: Duration = Duration::from_secs(60 * 60 * 24); // 1 day
+            let (full, rem) = sleep_chunks(duration, MAX_CHUNK);
+            for _ in 0..full {
+                gloo_timers::future::sleep(MAX_CHUNK).await;
+            }
+            gloo_timers::future::sleep(rem).await;
+        }
     }
 }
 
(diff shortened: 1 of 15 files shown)


crates/xmtp_common/src/time.rs
```diff
@@ -78,11 +78,34 @@ where
     }
 }
 
+/// Split `duration` into whole `max`-sized chunks plus a remainder.
+/// Returns `(number_of_full_chunks, remainder)`. Pure; used for wasm-safe sleeping.
+///
+/// JS `setTimeout` (via gloo-timers) casts the millisecond value to `i32`,
+/// so any duration > ~24.8 days overflows and fires immediately. Chunking into
+/// at-most-1-day pieces lets callers sleep the full requested duration.
+#[allow(dead_code)] // only called from wasm arm; also exercised by unit tests
```
tylerhawkes: This should be gated for wasm if that's all it's used for

crates/xmtp_mls/src/worker/key_package_cleaner.rs
```diff
@@ -116,21 +137,83 @@ where
...
+                        // `None` means every sender was dropped (context torn down) —
+                        // stop rather than busy-spin on a closed channel.
+                        msg = receiver.recv() => { if msg.is_none() { return Ok(()); } }
+                        // Deadline elapsed: time to do maintenance.
+                        () = xmtp_common::time::sleep(dur) => {
+                            self.maintain(next_rotation).await?;
+                        }
+                    }
+                }
+            }
```
tylerhawkes, line 175: What's preventing this whole worker from being dropped and getting implemented as a generic task? Anytime the key package gets rotated it could get rescheduled and then when it actually kicks off it would end up rescheduling itself.

insipx: I have an implementation here that does this: https://github.com/xmtp/libxmtp/pull/3795 but it touches more code, and there are some tricky edge cases with getting the worker to wake within X sec. time of getting a welcome, while also juggling deadlines for key package rotation and deletion

insipx: it's also entirely possible I need to rethink this KP stuff entirely. it seems we actually want two isolated worker tasks, key package cleaning, and key package rotation they don't necessarily have to be coupled as they are now. maybe that gives way to some other race conditions though but i'll investigate.

crates/xmtp_common/src/time.rs
insipx: Done in 1e40d818 — replaced `#[allow(dead_code)]` with `#[cfg(any(all(target_family = "wasm", target_os = "unknown"), test))]`, so `sleep_chunks` is compiled only on wasm (real use) and under `test` (keeps native coverage of the pure arithmetic). Native non-test builds no longer compile it — no dead code, no allow. Verified: native build/test + `wasm32-unknown-unknown` check all clean.

PR #3792, opened 2026-06-29
feat(proto): add KpMaintenance task variant + regen xmtp_proto

insipx:
Base of a 2-PR stack. Adds an empty `KpMaintenance` variant (field `=4`) to the `Task` oneof and regenerates the vendored `xmtp_proto`. No behavior change on its own.

The impl PR (stacked on this) uses this variant to run key-package maintenance as a recurring TaskRunner task, replacing the polling `KeyPackagesCleanerWorker` (~82M `worker_turn` spans/24h).

Note: `proto/mls/database/task.proto` must also land on `xmtp/proto` and `proto_version` be bumped for a stock regen; this PR vendors the generated `.rs` so the impl PR compiles.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SZTNgPeSH4WtQfN5qB33QF

> [!NOTE]
> ### Add `KpMaintenance` task variant to the MLS database proto
> Adds a new `KpMaintenance` oneof variant (tag 4) to the `Task` message in the MLS database proto, along with serde serialization support. The new message type is empty and follows the same pattern as existing task variants. Generated files in [xmtp.mls.database.rs](https://github.com/xmtp/libxmtp/pull/3792/files#diff-1b48cc83eb618a15345516b946b4138dae9c2be24155cc8d9a950038f118096c) and [xmtp.mls.database.serde.rs](https://github.com/xmtp/libxmtp/pull/3792/files#diff-fc113d60966783e19f7aee5738ce2e99a205b0ba65cf81541386c8e27c12e682) are regenerated from the updated proto definition.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 09e8ff8.</sup>
> 
>

PR #3793, opened 2026-06-29
feat: run key-package maintenance on the TaskRunner (kill the poll loop)

insipx:
**Depends on #3792** (proto). Until #3792 merges, this PR's diff includes the proto-regen commit at its base; review only the `xmtp_db`/`xmtp_mls` commits here.

Moves key-package maintenance off the fixed-interval `KeyPackagesCleanerWorker` (~82M `worker_turn` spans/24h, 99.9994% of all worker spans) onto the existing TaskRunner as a single recurring, deadline-driven task. ≤1-day fallback wake; prompt (~5s) rotation after welcome acceptance preserved.

## What changes
- One recurring `KpMaintenance` TaskRunner task: deletes expired key packages, rotates when due, re-arms to the next rotation/deletion deadline.
- `reschedule_kp_task`: atomic re-arm that re-reads the live KP deadline inside the same txn as the write (closes a nudge race that could otherwise delay a queued rotation up to 30 days).
- Recurring task is never reaped (capped backoff + reaper renews-in-place), self-healing singleton seed, backoff-safe 5s nudge on `queue_key_rotation`.
- Old `KeyPackagesCleanerWorker` poll loop removed; `WorkerKind::KeyPackageCleaner` kept as the enable gate (KP-enabled implies TaskRunner registered).

## Tests
- `xmtp_db`: readers, singleton seed, atomic floor-re-read (3 cases), backoff-safe nudge.
- `xmtp_mls`: recurring-never-reaped, KpMaintenance arm, wasm-safe rotation behavior test; deletion-timing guard preserved.
- wasm test-compile + clippy `-D warnings` clean. Live-node tests run in CI.

https://claude.ai/code/session_01SZTNgPeSH4WtQfN5qB33QF

> [!NOTE]
> ### Replace key-package cleaner poll loop with TaskRunner-driven maintenance
> - Removes the standalone `KeyPackagesCleanerWorker` polling loop; key-package deletion and rotation are now handled by the `TaskRunner` via a new `KpMaintenance` task type in [key_package_maintenance.rs](https://github.com/xmtp/libxmtp/pull/3793/files#diff-1d4be0f0b8283620c52870358d20c933cc8086d19b8c804906ce1fc6d99af0eb) and [tasks.rs](https://github.com/xmtp/libxmtp/pull/3793/files#diff-b3d01dc512a077cf4b625f516f8c36ea61978f070bcce720ed8d522065a268b6).
> - Introduces a singleton `KpMaintenance` DB task that is seeded on client build (when the KeyPackageCleaner gate is enabled) and is never reaped — expired or exhausted rows are atomically replaced and re-armed.
> - `queue_key_rotation` and welcome-envelope processing now call `bump_kp_maintenance_task` and wake the TaskRunner so maintenance runs within a debounce window rather than waiting for a distant scheduled deadline.
> - Adds `next_key_package_rotation_ns` and `min_key_package_delete_at_ns` DB queries used to compute the next maintenance deadline atomically inside `reschedule_kp_task`.
> - Behavioral Change: clients with the KeyPackageCleaner gate enabled register the TaskRunner instead of a separate worker; key-package maintenance is now event-driven with deadline flooring rather than fixed-interval polling.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cd35452.</sup>
> 
>

### DIFF
--- a/crates/xmtp_db/src/encrypted_store/identity.rs
+++ b/crates/xmtp_db/src/encrypted_store/identity.rs
@@ -174,7 +174,10 @@ pub(crate) mod tests {
                 .unwrap();
 
             // Reader returns the currently-scheduled deadline.
-            assert_eq!(conn.next_key_package_rotation_ns().unwrap(), Some(far_future));
+            assert_eq!(
+                conn.next_key_package_rotation_ns().unwrap(),
+                Some(far_future)
+            );
 
             // Queuing a rotation pulls the deadline earlier (~5s out).
             conn.queue_key_package_rotation().unwrap();
--- a/crates/xmtp_db/src/encrypted_store/identity.rs
+++ b/crates/xmtp_db/src/encrypted_store/identity.rs
@@ -52,6 +52,8 @@ pub trait QueryIdentity {
         rotation_interval_ns: i64,
     ) -> Result<(), StorageError>;
     fn is_identity_needs_rotation(&self) -> Result<bool, StorageError>;
+    /// Absolute ns of the next scheduled key-package rotation, or None if not queued.
+    fn next_key_package_rotation_ns(&self) -> Result<Option<i64>, StorageError>;
 }
 
 impl<T> QueryIdentity for &T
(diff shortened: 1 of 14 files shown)


insipx:
Reopening on origin branches for a proper stacked PR (base = the proto PR's branch).

PR #3802, opened 2026-06-29
fix(db): fix batch_read_from_cache pairwise filter broken into cross product

### LINKED ISSUE #3800: bug(db): batch_read_from_cache returns wrong results pairwise filter broken into cross-product
In crates/xmtp_db/src/encrypted_store/association_state.rs:127-135, the batch_read_from_cache function builds a SQL query that uses IN lists independently instead of pairwise comparison:

    rust
    let (inbox_ids, sequence_ids): (Vec<String>, Vec<i64>) = identifiers.into_iter().unzip();
    let query = dsl::association_state
        .select((dsl::inbox_id, dsl::sequence_id, dsl::state))
        .filter(
            dsl::inbox_id
                .eq_any(inbox_ids)
                .and(dsl::sequence_id.eq_any(sequence_ids)),
        );

    The query translates to (inbox_id IN list1) AND (sequence_id IN list2), which is a cross-product match, not a pairwise match. If the caller requests [(inbox_a, 1), (inbox_b, 2)], this query will also return (inbox_a, 2) and (inbox_b, 1) if those rows exist returning wrong cached association states.

    Impact: Any caller that uses batch_read_from_cache with multiple (inbox_id, sequence_id) pairs can receive association state entries that do not correspond to the requested pair. This can cause identity verification logic to read stale or incorrect association state data.

    Suggested fix: Rewrite the query to use chained OR conditions that pair each inbox_id with its corresponding sequence_id, for example using a FIND_IN_SET approach or constructing pairwise (inbox_id = X AND sequence_id = Y) OR (...) conditions.

Sertug17:
Closes #3800

    Replace the unzip + eq_any approach with into_boxed() + or_filter()
    to ensure each (inbox_id, sequence_id) pair is matched individually
    instead of producing a cross-product (inbox_id IN list) AND (sequence_id IN list).

> [!NOTE]
> ### Fix `batch_read_from_cache` returning cross-product matches instead of exact pairs
> The previous implementation filtered by `inbox_id IN (...)` AND `sequence_id IN (...)` separately, causing a cross-product of all input inbox IDs and sequence IDs to be returned instead of only the exact `(inbox_id, sequence_id)` pairs requested.
>
> The fix in [`association_state.rs`](https://github.com/xmtp/libxmtp/pull/3802/files#diff-581d80a6e674bed695f641d46fdc06c7a810f0f66a113d191daec50a5fab2a60) replaces the two-vector approach with iterative `or_filter` calls, each matching one exact `(inbox_id, sequence_id)` pair using a boxed query.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8e1128c.</sup>
> 
>

### DIFF
--- a/crates/xmtp_db/src/encrypted_store/association_state.rs
+++ b/crates/xmtp_db/src/encrypted_store/association_state.rs
@@ -124,15 +124,16 @@ impl<C: ConnectionExt> QueryAssociationStateCache for DbConnection<C> {
             return Ok(vec![]);
         }
 
-        let (inbox_ids, sequence_ids): (Vec<String>, Vec<i64>) = identifiers.into_iter().unzip();
-
-        let query = dsl::association_state
+        let mut query = dsl::association_state
             .select((dsl::inbox_id, dsl::sequence_id, dsl::state))
-            .filter(
-                dsl::inbox_id
-                    .eq_any(inbox_ids)
-                    .and(dsl::sequence_id.eq_any(sequence_ids)),
-            );
+            .into_boxed();
+
+        for (inbox_id, sequence_id) in &identifiers {
+            let predicate = dsl::inbox_id
+                .eq(inbox_id.clone())
+                .and(dsl::sequence_id.eq(*sequence_id));
+            query = query.or_filter(predicate);
+        }
 
         let association_states =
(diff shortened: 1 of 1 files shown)


insipx (approved):

insipx:
thanks!

PR #3823, opened 2026-07-06
refactor(xmtp_mls): extract shared message-processing pipeline

tylerhawkes:
## Summary

XIP-83 client-integration **Phase A**: extract the message/welcome decode pipeline out of the live-stream poll state-machines into standalone, reusable seams — then route the live streams through those same seams so there is exactly one implementation.

Two commits, best reviewed in order:

1. **`extract reusable message-processing pipeline from StreamGroupMessages`** — adds the seams:
   - `process_one(factory, msg) -> Processed`: DB fast-path (already-stored → forward without re-decrypting), else the full decrypt/store pass (incl. recovery sync). Dedup and cursor bookkeeping deliberately stay **caller-side** — the live stream tracks them per-stream (`GroupList`), while the upcoming XIP-83 multiplexing layer tracks them centrally (per-topic high-water marks).
   - `process_welcome_one(...) -> WelcomeOutcome` + `ProcessWelcomeResult::into_outcome()`: the welcome analog, lifted from `StreamConversations::filter_welcome`.
2. **`route live streams through the shared processing seams`** — proves the extraction is behavior-preserving:
   - `StreamGroupMessages::on_waiting`/`resolve_futures` now share the pipeline via sync `prepare`/`finish` halves (the poll state-machine can't `.await`), instead of duplicating the logic. Also drops a per-message clone on the decrypt path.
   - `filter_welcome` now interprets results via `into_outcome()`; ignore-path debug logs preserved.
   - `Prepared::Ready` carries the stored message **by value**, so a "fast-path hit without a message" state is unrepresentable.

Note for reviewers: `process_one` / `process_welcome_one` have **no non-test callers yet by design** — they are the seams the XIP-83 stream router (next phase) consumes. The live streams already exercise their internals via `prepare`/`finish`/`into_outcome`, so the shared halves are covered by every existing streaming test.

## Test plan

- New unit tests: `process_one` DB fast-path (asserts `create()` is never reached) and fast-path-miss pipeline run, via a `StubFactory`; `into_outcome` ignore/ignore-with-id mappings.
- Existing streaming integration tests cover the rewired live paths.
- `cargo clippy -p xmtp_mls --tests` clean, fmt clean.

> [!NOTE]
> ### Extract shared message and welcome processing pipeline for XIP-83 Phase A
> - Adds `process_one` in [process_message.rs](https://github.com/xmtp/libxmtp/pull/3823/files#diff-c533e00eddfcecd8eb7fb1e7b5c8984143d191b48d28c44b8fea4dfbb06528d2) — a single-message async API that checks a DB fast-path via `prepare` and falls back to decrypt/store, returning a normalized `Processed` result.
> - Adds `process_welcome_one` and `WelcomeOutcome` in [process_welcome.rs](https://github.com/xmtp/libxmtp/pull/3823/files#diff-6a8fc2f2e683ba89ea325dab846f3caec443d5bbd3bd0b96f5e259300ff94296), giving callers a consistent welcome-processing result outside of stream poll machinery.
> - Refactors `StreamConversations.filter_welcome` and `MessageStream` in the stream files to delegate to the new shared `prepare`/`finish`/`into_outcome` helpers, removing duplicated inline logic.
> - Changes `group_id` field type from `Vec<u8>` to typed `GroupId` across `ProcessedMessage` and related structures.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized aaec92a.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/subscriptions/process_message.rs
+++ b/crates/xmtp_mls/src/subscriptions/process_message.rs
@@ -94,6 +94,64 @@ where
     }
 }
 
+/// Outcome of running a single raw group message through the processing pipeline,
+/// independent of any stream/poll machinery.
+///
+/// Dedup (skipping already-seen cursors) and cursor bookkeeping are intentionally
+/// **not** handled here — they are the caller's responsibility, because different
+/// consumers track them differently. The live [`super::stream_messages`] stream uses
+/// a per-stream `GroupList`; the XIP-83 bidi multiplexing manager keeps central
+/// per-topic high-water marks plus per-subscriber in-flight cursors. This seam is
+/// only the decode half (DB fast-path, then decrypt/store + recovery-sync), so both
+/// consumers decode identically.
+#[derive(Debug)]
+pub struct Processed {
+    /// The decrypted message to deliver, if this envelope produced one. `None` means
+    /// the envelope was processed but surfaced nothing (e.g. a commit); the caller
+    /// should still advance its cursor to `next_cursor`.
+    pub message: Option<StoredGroupMessage>,
+    /// The group this envelope belongs to.
+    pub group_id: Vec<u8>,
+    /// The cursor the caller should advance this group to after handling the result.
+    pub next_cursor: Cursor,
+    /// The cursor of the envelope we attempted to process (for tracking/logging).
+    pub tried: Cursor,
+}
+
+/// Run a single raw group message through the shared processing pipeline.
+///
+/// 1. **Fast path:** if the message is already stored locally (e.g. a replayed
+///    envelope after a cursor'd re-add — see XIP-83 client integration), return it
+///    without decrypting.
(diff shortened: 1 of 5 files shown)


insipx (approved):
this is all good, but I will mention that with bidi and reliable streams we can probably do better

All this factory boilerplate is mostly a side-effect of having to keep our own cursor book-keeping. since bidi is allowed to advance the cursor, we can put more effort into making the `process` in `mls_sync` itself idempotent rather than layer complexity into the streams. previously, calling process was a dead end, since advancing the cursor was disallowed we needed a heuristic to tell whether we could call `process()` or needed to separately fork out to `sync()` (boils down to is this new message a commit that requires advancing the cursor in `RefreshState` or not). On top of that were some extra optimizations to reduce to a single db call (what `process()` should already do). if advancing the cursor is allowed, then calling process with allow_cursor_increment should be enough to remove much of the boilerplate

PR #3825, opened 2026-07-06
feat(xmtp_api_d14n): BidiTransport — process-level topic ledger + envelope demux

tylerhawkes:
## Summary

XIP-83 client integration **Phase B1**: the process-level transport layer. `BidiTransport<B>` owns THE bidi `Connection` for a process and demuxes raw wire deliveries by topic to refcounted `TopicLease` holders. It speaks only transport vocabulary — topics, envelopes, leases, waves — and knows nothing about MLS, databases, decryption, or streams. It holds **no cursors**: it forwards the resume positions a lease hands it, and the durable/application cursors stay with the client-level consumer (the stream router, next phase).

Independent of #3823 (the `xmtp_mls` pipeline seams) — the two merge in either order; the stream router stacks on both.

### Semantics

- **Every `lease()` is a cursored `Mutate` wave**, sent even when another lease already holds the topic: per XIP-83, the cursored re-add IS the catch-up/replay mechanism. Duplicates that replay causes for other leases are the consumer's dedup job (it owns the delivery high-water marks).
- **Removes are refcounted**: a topic leaves the wire only when its last lease derefs (`Drop` = deref). The wire opens lazily on the first lease and closes gracefully (half-close + bounded drain) after the last.
- **One backpressure policy at every layer**: a lease that stops draining is dropped at its channel bound — never blocking the wire or sibling leases — and recovers by re-leasing from durable cursors.
- **Endpoint keying is the caller's obligation** (documented): a transport has no idea what its opener dials; nothing here is process-global, so multiple transports (one per endpoint/env) compose naturally.

### Deadlock discipline (found in review, designed out)

The ledger task is the sole drainer of the wire's events, so it must never await the wire's bounded command channel — a delivery flood plus a drop-and-re-lease stampede could otherwise park the ledger and the connection actor on each other's full channels (a reachable AB-BA deadlock). All outbound waves therefore go through a new non-blocking `Connection::try_mutate` (returns the wave on `Full`/`Closed`) plus an ordered ledger-side outbox, flushed opportunistically with a timer backstop. The graceful-close budget now covers `finish()` itself, not just the drain, so a parked actor can't leak the close task.

### Deferred to later phases (documented in-module)

Transparent reconnect (today a dead wire closes every lease; consumers re-lease, which re-opens), half-open-wire liveness probing (reconnect phase owns probe + re-open; consumers keep the existing watchdog floor until then), and `Started` capability gating.

## Test plan

12 unit tests over a scripted multi-session mock wire: lazy open with cursored adds; cursored re-add on a shared topic; per-topic demux (incl. an unroutable message skipped without sinking its batch); shared-topic fan-out; `CatchUpComplete`/`TopicsLive` routing to their owners; refcounted removes + graceful half-close (asserted distinguishable from a hard abort) + reopen; slow-lease drop without blocking siblings; wire-death lease closure + reopen; empty-lease rejection; retryable open failure. Plus 2 core tests for `try_mutate` (`Full` under a parked actor with recovery after drain; `Closed` after finish). Crate suite 189/189, clippy clean, wasm target still compiles (module is native-gated).

> [!NOTE]
> ### Add process-level topic ledger and envelope demux via `BidiTransport` in `xmtp_api_d14n`
> - Introduces [`bidi_transport.rs`](https://github.com/xmtp/libxmtp/pull/3825/files#diff-d4217bbd27403da346b2451dab35cab51079f0ccb533baa2a2248a7fa3dbe458) with a `BidiTransport` struct that manages a single bidi `Connection`, supports multiple ref-counted topic leases, and demultiplexes wire deliveries to per-lease channels.
> - Adds a `TransportBinding` trait for building mutate waves and extracting topics; implements it for `V3Binding` in [`connection.rs`](https://github.com/xmtp/libxmtp/pull/3825/files#diff-11d1b237ec81cff558f3bbfc0b5f8477bd670265b0fcbe1cd6b700abcfddf1f3).
> - Extends `TopicExtractor` in [`topics.rs`](https://github.com/xmtp/libxmtp/pull/3825/files#diff-8e8ab12f4b3a817cd6b9c5704fa5bed892794e9d1952f8cc06267276b94bdc18) to extract topics from v3 group message, welcome message, and welcome pointer responses.
> - Adds `Connection::try_mutate` in [`bidi.rs`](https://github.com/xmtp/libxmtp/pull/3825/files#diff-c49e1661d673a984b586cc256686366332160bc09d8324aff059ff2d9753879f) for non-blocking mutation enqueue, returning `TryMutateError::Full` or `TryMutateError::Closed` instead of awaiting.
> - Risk: the ledger run loop drops wedged leases under backpressure and uses an `Outbox` to avoid deadlocks; edge cases around lease teardown and graceful close have timeout-based fallbacks.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c9d257b.</sup>
> 
>

### DIFF
--- a/crates/xmtp_api_d14n/src/queries/v3/connection.rs
+++ b/crates/xmtp_api_d14n/src/queries/v3/connection.rs
@@ -5,13 +5,17 @@
 //! `GroupMessage`/`WelcomeMessage` (v3 carries a single id cursor and the
 //! consumer decodes), via the [`BidiBinding`] trait. Native-only.
 
+use crate::protocol::Envelope;
 use crate::queries::bidi::{BidiBinding, Connection, Event, Inbound, parse_topics};
+use crate::queries::bidi_transport::TransportBinding;
 use xmtp_proto::api_client::XmtpMlsBidiStreams;
 use xmtp_proto::mls_v1::subscribe_request::v1::Mutate;
+use xmtp_proto::mls_v1::subscribe_request::v1::mutate::Subscription;
 use xmtp_proto::mls_v1::{
     GroupMessage, Ping, Pong, SubscribeRequest, SubscribeResponse, WelcomeMessage,
     subscribe_request, subscribe_response,
 };
+use xmtp_proto::types::Topic;
 
 /// The v3 (MLS API) wire binding for a bidi subscription.
 pub struct V3Binding;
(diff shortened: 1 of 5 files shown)


crates/xmtp_api_d14n/src/queries/v3/connection.rs
```diff
@@ -80,6 +83,48 @@ impl BidiBinding for V3Binding {
...
+        }
+    }
+
+    fn group_topic(msg: &GroupMessage) -> Option<Topic> {
+        let group_message::Version::V1(v1) = msg.version.as_ref()?;
+        Some(Topic::new_group_message(&v1.group_id))
+    }
+
+    fn welcome_topic(msg: &WelcomeMessage) -> Option<Topic> {
+        // Both shapes carry the topic identifier: a full welcome and a pointer
```
insipx: `msg.topic().ok()` from [Envelope](https://xmtp.github.io/libxmtp/xmtp_api_d14n/protocol/traits/trait.Envelope.html#tymethod.topic) trait should work for both `GroupMessage` and `WelcomeMessage` types as well

insipx (approved):

PR #3826, opened 2026-07-06
feat(xmtp_mls): StreamRouter — per-stream decode, dedup, and delivery over the bidi transport

tylerhawkes:
## Summary

The client-level **`StreamRouter`** (XIP-83 client integration, phase B2): one router per client, sitting on the process-level `BidiTransport` (#3825 — this PR is stacked on it). Consumers get decoded, deduplicated streams (`stream_messages`, `stream_conversations`) whose subscribe **is** the catch-up: each stream leases its topics with durable resume cursors and the server replays from there.

### Design

- **One task per stream, nothing shared.** Each stream owns its lease, its dedup state, and its bounded delivery channel. A stall in one (e.g. a multi-second recovery sync in the pipeline) backs up only that stream's own lease — siblings are never delayed. The router task only seeds, leases, and spawns.
- **Decode via the shared pipeline seams** (`process_one` / `process_welcome_one` from #3823). Decrypt/store happens once per unique message; replay copies resolve from the DB fast-path.
- **Dedup**: steady state is a per-topic `GlobalCursor` high-water mark. During a stream's catch-up window (subscribe → its `CatchUpComplete`), wire order is legitimately not cursor-monotonic (the transport's documented replay semantics), so the window instead checks a frozen subscribe-time floor + an exact-identity seen-set, pre-seeded with locally-stored messages above the durable cursor (the streaming pipeline stores **without** advancing `refresh_state`, so the durable cursor alone is not a delivery floor).
- **Welcome streams** keep a per-stream known-welcome set (streams with different filters must never suppress each other) and seed the wire from the durable `EntityKind::Welcome` cursor.
- Streams outlive router handles; dropping a `RouterStream` is the unsubscribe (lease deref → refcounted wire removal).

### Adversarial review (2 independent passes) — all findings fixed in this PR

Cross-stream welcome suppression via shared dedup state; router-lifetime contract violation (handles dropping killed live streams); catch-up window missing its seed floor (sibling replay leaked pre-subscribe history); recovery-sync fallback double-delivery (window now records the delivered identity, not just the envelope's); welcome wire seed read from `group_cursors` max instead of the durable Welcome cursor (replay storms + mixed-originator mis-seeding); re-subscribe re-delivery of everything streamed since the last full sync (`messages_newer_than` seeding restored, mirroring the legacy stream); inline decode starving sibling streams (solved structurally by per-stream tasks).

## Test plan

- Live integration (local node): live delivery; catch-up replay from durable cursors; **re-subscribe does not re-deliver** (pins the stored-message seeding); **two sibling conversation streams both receive the same welcome** (pins per-stream dedup).
- Unit: catch-up window semantics (frozen-seed floor, identity dedup that doesn't swallow leapfrogged replays, delivered-identity recording, window→live handoff).
- clippy clean, fmt clean; full suite unaffected.

> [!NOTE]
> ### Add `StreamRouter` for per-stream decode, dedup, and delivery over the bidi transport
> - Introduces [`StreamRouter`](https://github.com/xmtp/libxmtp/pull/3826/files#diff-f9a7fe85e5e034811e7a4454198333ec3adcdabd79c68f1c3a952279950de15c) (native-only) with `stream_messages` and `stream_conversations` APIs that lease topics via a shared `BidiTransport`, spawn per-stream consumer tasks, and return bounded `RouterStream`s.
> - Adds `StreamDedup` with `Syncing`/`Live` states to correctly deduplicate messages during the catch-up window (using frozen seeds + exact-identity sets) before switching to high-water-mark dedup in steady state.
> - Updates `QueryGroupMessage::messages_newer_than` to return `Vec<(GroupId, Cursor)>` so cursors are attributed to their originating group, which is required for correct per-topic position seeding on resubscribe.
> - `fold_stored` seeds per-topic resume positions and the exact-identity seen set from durable storage before leasing live topics, enabling no-redelivery resubscribe.
> - Adds live integration tests in [`stream_router_tests.rs`](https://github.com/xmtp/libxmtp/pull/3826/files#diff-94a932ab54db31971beb99d604b2c29cceb4236704ba8ee89cc57491201914f8) covering live delivery, cursor catch-up, resubscribe idempotency, and sibling welcome stream delivery.
> - Behavioral Change: `messages_newer_than` signature change is a breaking API update affecting all callers, including [`stream_messages.rs`](https://github.com/xmtp/libxmtp/pull/3826/files#diff-cffe5828e751e2a749629b6acdb0a9a09b407756b9b89f3042a1ecf28055aeb6) and the mock in [`mock.rs`](https://github.com/xmtp/libxmtp/pull/3826/files#diff-561d136919d4d75db380706389549bc558fbd0bcdfefb1f77a821bcbcf5cdf20).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3dc778d.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/subscriptions/stream_router.rs
+++ b/crates/xmtp_mls/src/subscriptions/stream_router.rs
@@ -1,0 +1,804 @@
+//! XIP-83 client-level stream router.
+//!
+//! One [`StreamRouter`] per client. Each consumer stream leases its topics
+//! from the process-level [`BidiTransport`], decodes raw wire deliveries
+//! through the shared pipeline seams ([`process_one`] /
+//! [`process_welcome_one`]), and delivers decoded results on a bounded
+//! channel. Everything here speaks *stream* vocabulary — streams, messages,
+//! conversations, durable cursors — the transport's wire/topic/envelope world
+//! stays below the lease boundary.
+//!
+//! ## Shape: one task per stream, nothing shared
+//!
+//! Every stream is its own task owning its lease, its dedup state, and its
+//! delivery channel. Streams share no mutable state, so a stall in one — say
+//! a multi-second recovery sync inside the pipeline — backs up only that
+//! stream's own lease channel (where the transport's drop policy applies to
+//! it alone) and never delays a sibling. The router task itself only seeds,
+//! leases, and spawns; it holds no delivery path.
+//!
+//! ## Cursors and dedup
+//!
+//! Streams are seeded from the client's **durable** application cursors
+//! (`refresh_state`), never wire cursors: the cursored lease is the one
+//! catch-up mechanism (a re-add below the server's floor replays history).
+//! Steady-state dedup is a per-topic [`GlobalCursor`] position — the
+//! transport fans a shared topic to every lease, and each stream
+//! independently skips what it has already delivered, while decrypt/store
+//! still happens once per unique message (the pipeline's DB fast-path serves
+//! every later copy from storage).
+//!
+//! During a stream's **catch-up window** (subscribe until its
+//! `CatchUpComplete`) wire order is not cursor-monotonic, so the position is
+//! not consulted; the window dedups against the frozen subscribe-time seed
+//! (drop everything at-or-below it — including a sibling's replay of older
+//! history) plus an exact-identity seen-set (see [`StreamDedup`]).
+//!
+//! Welcome dedup is a per-stream known-welcome set, consulted and updated
+//! sequentially inside that stream's task — exactly the serialization
+//! [`process_welcome_one`]'s contract asks for. Exact-identity sets are
+//! immune to the ordering caveat, so welcome streams need no window.
+//!
+//! ## Backpressure (same policy as every layer below)
+//!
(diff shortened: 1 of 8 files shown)


crates/xmtp_mls/src/subscriptions/stream_router.rs
```diff
@@ -0,0 +1,804 @@
...
+/// requested (see the transport's delivery-order contract) — so the window
+/// checks the **frozen subscribe-time seed** (anything at-or-below the
+/// durable resume point is never for this stream, including a sibling
+/// lease's replay of older history) plus an exact-identity seen-set. The set
+/// is bounded by the window: it drops at `CatchUpComplete`, when the live
+/// positions (which kept advancing) take over.
+enum StreamDedup {
+    Syncing {
+        /// Per-topic durable positions frozen at subscribe.
+        seeds: HashMap<Topic, GlobalCursor>,
```
insipx, line 217: this is the same as https://xmtp.github.io/libxmtp/xmtp_proto/types/struct.TopicCursor.html which derefs to a hashmap

insipx (approved):

PR #3828, opened 2026-07-08
feat(xmtp_api_d14n): transport delivery guarantee + transparent reconnect

tylerhawkes:
Phase E of the XIP-83 client integration, part 1 of 2: the `BidiTransport` becomes the single reconnect authority, and gains a delivery guarantee strong enough that consumers never observe a wire flap.

## Commit 1 — per-lease delivery positions (transport-level dedup guarantee)

`TransportBinding` learns per-message cursor extraction (`group_cursor`/`welcome_cursor`, via the shared `CursorExtractor`) plus `advance`/`covers`/`meet` ops (v3: `max`/`<=`/`min`). Each lease tracks per-topic state:

- **Catch-up window** (before its `CatchUpComplete`): admit anything above the topic's frozen add-cursor **floor**. Filtering by the advancing position here would be wrong — wire order is not cursor-monotonic during catch-up, so an early live delivery would swallow the replay behind it (the "leapfrog trap"). The advancing position is still recorded.
- **Caught up**: admit only strictly past the advancing position, check-then-advance.

Net guarantee: a caught-up lease receives each trackable message **at most once**, and a lease in its window never sees a sibling's deeper replay below its own floor. Untrackable cursors fail open (the consumer's DB dedup absorbs, a silent drop would lose data).

## Commit 2 — transparent reconnect

Leases now **survive wire death**. The ledger keeps an in-memory, advance-only wire-position map (`last_seen: Topic → Cursor`) as it demuxes — wire bookkeeping, not durable-cursor authority. On `Wire(None)`:

- Reconnect on capped exponential backoff (100ms → 30s), forever — an offline process resumes when the network returns.
- One **resume wave** re-adds every leased topic from `last_seen`, held down to the floor of any lease still owed its catch-up replay; that wave's `CatchUpComplete` resolves every such lease (their windows are still open, so the replay overlap is absorbed by commit 1's guarantee).
- A lease taken while the wire is down registers and rides the next resume open.

`TopicLease::next() → None` now means only: this consumer fell behind and was dropped, or the transport shut down.

## Tests

Six new mock-wire tests: sibling-replay dedup for caught-up leases, window admits the replay behind an early live delivery, below-floor history filtered, reconnect resumes from last-seen (not the floor), interrupted catch-up re-owed by the resume wave, dead-wire lease rides the resume open. 6/6 new + all existing transport/bidi tests green; `just lint` clean.

Follow-up in this stack: silence watchdog + `suspend()`/`resume()` (next PR). A later XIP-83 revision (server-tagged deliveries) will delete most of commit 1's filtering machinery; until servers tag, this is the required client-side reconstruction.

> [!NOTE]
> ### Add transparent reconnect and delivery guarantees to the d14n bidi transport
> - Implements automatic reconnect in `run_ledger` with capped exponential backoff (100ms–30s); active leases are no longer closed on wire death and resume after reconnect via `Ledger.resume_adds`.
> - Adds per-lease delivery filtering in `Ledger.demux` and `LeaseState.admit`: during catch-up, only messages above the frozen floor are admitted; after catch-up, only messages strictly beyond the tracked position are admitted (at-most-once semantics).
> - `Ledger` now tracks per-topic `last_seen` positions and `WaveOwner` to correctly resolve `CatchUpComplete` across multiple leases after a reconnect.
> - `V3Binding` in [connection.rs](https://github.com/xmtp/libxmtp/pull/3828/files#diff-11d1b237ec81cff558f3bbfc0b5f8477bd670265b0fcbe1cd6b700abcfddf1f3) implements the new `TransportBinding` cursor APIs (`group_cursor`, `welcome_cursor`, `advance`, `covers`, `meet`) using `sequence_id` as the wire position.
> - Behavioral Change: a wire flap no longer tears down leases; callers that previously relied on lease closure to detect disconnects will no longer receive that signal.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 988cf08.</sup>
> 
>

### DIFF
--- a/crates/xmtp_api_d14n/src/queries/bidi_transport.rs
+++ b/crates/xmtp_api_d14n/src/queries/bidi_transport.rs
@@ -1007,13 +1007,22 @@ where
 {
     xmtp_common::spawn(None, async move {
         let mut wire = wire;
-        let _ = tokio::time::timeout(GRACEFUL_CLOSE_BUDGET, async {
+        let drained = tokio::time::timeout(GRACEFUL_CLOSE_BUDGET, async {
             if wire.finish().await.is_err() {
                 return; // actor already gone — nothing to drain
             }
             while wire.next().await.is_some() {}
         })
         .await;
+        if drained.is_err() {
+            // Expected-degraded, not an error: a long server-side catch-up
+            // tail (or a process thaw after suspension) can outlive the
+            // budget, and dropping the wire here is the designed ending.
+            tracing::debug!(
+                budget_ms = GRACEFUL_CLOSE_BUDGET.as_millis() as u64,
+                "bidi transport: graceful-close drain budget expired; dropping the wire"
+            );
+        }
     });
 }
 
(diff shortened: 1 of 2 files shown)


crates/xmtp_api_d14n/src/queries/bidi_transport.rs
```diff
@@ -20,8 +20,19 @@
...
 //!   catch-up/replay mechanism (XIP-83), and the server replays from the given
-//!   cursor. Duplicates that replay causes for other leases are dedup'd by the
-//!   consumer, which tracks delivery high-water marks; the transport does not.
+//!   cursor.
+//! - **The delivery guarantee.** The ledger keeps per-lease per-topic delivery
+//!   positions (seeded at the lease's add cursors). During a lease's catch-up
+//!   window (subscribe → its `CatchUpComplete`) it admits anything above the
+//!   topic's *frozen floor* — never filtering by the advancing position, or an
+//!   early live delivery would swallow the replay behind it. Once caught up, a
+//!   lease receives each trackable message **at most once**: a sibling's
```
insipx, line 29: what happens when a delivered message fails processing on the mls side? How does tracking deal with that -- MLS has no internal cache of delivered messages, in current streams if delivery fails it expects to re-query at some point, which i guess historically happened outside the stream during a `sync()`

insipx (approved):

PR #3832, opened 2026-07-09
feat(xmtp_mls,bindings): route callback streams over bidi behind XMTP_BIDI_STREAMS_ENABLED

tylerhawkes:
Phase C of the XIP-83 client integration — **stacked on #3829**. The client callback surface (mobile + node) learns to route its streams over the bidi transport, dark behind `XMTP_BIDI_STREAMS_ENABLED`.

## Commit 1 — pull-to-push adapter (`xmtp_mls`)

`router_callbacks.rs` adapts the `StreamRouter`'s pull streams to the callback shape the FFI bindings consume: spawn, pump, callback per item, `on_close` when the stream ends. The router rides a **process-global shared transport** — one wire per process regardless of client count — and every entry point gates on `bidi_streams_enabled()` (the env var, read per call), so the whole path stays dark until explicitly enabled.

## Commit 2 — mobile dispatch

`FfiConversations::stream` / `stream_all_messages` branch on the gate; the legacy path is untouched. The mobile client holds its API behind a trait object, so `XmtpMlsBidiStreams` is plumbed through the dyn-API surface (`full_api`, boxed streams, combined, d14n) to make the bound satisfiable there.

## Commit 3 — crate-level suspend/resume helpers

`suspend_bidi_streams()` / `resume_bidi_streams()` wrap the shared transport's app-lifecycle pair (#3829) — resume resolves at catch-up, the background-fetch "catch up, then done" primitive — and no-op when bidi was never active. **Crate-level only for now**: exposing them over the FFI boundary is deliberate follow-on work.

## Commit 4 — single-conversation streams

`FfiConversation::stream` was the one callback stream still pinned to the legacy path — a conversation object holds only the context, while the client's cached router hangs off the `Client`. Rather than thread client handles through every conversation construction site, the bidi arm builds an **ephemeral router scoped to the one conversation**: router state is per-stream and its task exits with its last stream, while dedup correctness already lives in the transport's per-lease positions — so a second router beside the client's cached one is safe by design.

## Commit 5 — node dispatch

Mirrors mobile: `conversations.stream`, `conversations.streamAllMessages`, and `conversation.stream` take the bidi path when the env var is truthy. Consent/preference/deletion streams ride the sync topics and stay legacy by design.

## Semantics & rollout notes

- **`on_close` semantics differ from legacy for a wedged consumer** (macroscope flagged): a bidi stream whose callback stops draining is dropped by the transport (protecting the shared wire) and fires `on_close` — the caller re-subscribes from durable cursors. Legacy instead stalls its own dedicated stream indefinitely. Wire flaps do NOT end bidi streams (leases survive reconnects). Whether the adapters should internally re-subscribe on that drop instead of surfacing `on_close` is an open design call — flagged for review.

- **Dark by default.** Nothing changes unless `XMTP_BIDI_STREAMS_ENABLED` is set — intended for dev/agent deploys until the fallback latch lands. There is **no automatic legacy fallback yet** (that latch is follow-on work), so the gate must only be enabled against v3 backends — a `MigrationClient`/`D14nClient` refuses `subscribe_bidi` at runtime.
- **Interim scope (messages)**: bidi `stream_all_messages` covers the groups known at subscribe time; the welcome auto-subscribe reflex is the next phase. Don't flip the env var for always-on agents until that lands.
- **Interim scope (conversations)**: the bidi conversations stream speaks the welcome topic only — a group/DM *created locally by the same client* does not fire its own `conversations.stream` callback (legacy multiplexes `LocalEvents::NewGroup` for that). Documented at the `WelcomeConsumer`; parity arrives with the same reflex work, which fans local new-group events into both stream kinds.

## Review-round fixes

- **Sync-group interception restored** (macroscope, High): the bidi `stream_all_messages` arm subscribed sync groups (same query as legacy) but forwarded their messages to the callback — surfacing internal device-sync payloads to apps. It now mirrors legacy exactly: sync-group traffic nudges the device-sync worker (`NewSyncGroupMsg`) and never reaches the callback. (Macroscope's suggested fix — dropping sync groups from the subscription — would have silently lost that worker nudge.) Pinned by a live test with a real sync group.
- **Empty subscription stays open** (macroscope): `stream_all_messages` on an account with no matching conversations used to hit the transport's empty-lease refusal and error-close immediately; it now stays open with nothing subscribed, matching legacy. Pinned by a live test.
- The false "falls back to legacy streams" comment on the `MigrationClient`/`D14nClient` refusal is corrected — the latch is follow-on work, v3 backends only (also disclosed above).
- The single-auth-identity assumption of the process-shared transport is documented at the module doc, and transport initialization logs at info.
- Interim-scope doc comments added at both bidi entry points.

## Tests

Eight live tests in `router_callbacks_tests.rs`: callback stream delivers live messages, new-conversation surfacing, sibling clients share the process transport, single-conversation scoping (two groups — only the streamed one delivers), suspend → miss → resume → replay E2E across **two** background cycles, lifecycle helpers as no-ops before any transport exists, empty-subscription-stays-open, and sync-group-messages-intercepted-not-delivered. The node vitest stream suites (`Conversations.test.ts` / `Conversation.test.ts`) pass identically with the env var off (8/8) and on (8/8 + 1/1), including conversation-type filters and the new-installation DM case.

> [!NOTE]
> ### Route MLS callback streams over bidi transport behind `XMTP_BIDI_STREAMS_ENABLED` flag
> - Adds `bidi_streams_enabled()` gate in [router_callbacks.rs](https://github.com/xmtp/libxmtp/pull/3832/files#diff-0c27658258694997d993ae3e2e444204d0c98a70f368560fd306949e8849549d) that reads the `XMTP_BIDI_STREAMS_ENABLED` env var to select between the new bidi path and the legacy streaming path.
> - Implements `stream_all_messages_with_callback_bidi` and `stream_conversations_with_callback_bidi` on `Client`, backed by a process-wide shared `BidiTransport` cached in a `OnceLock`; a free function handles per-conversation streams via an ephemeral router.
> - Updates all mobile and Node binding stream entry points (`stream`, `stream_groups`, `stream_dms`, `stream_messages`) to branch on `bidi_streams_enabled()` and call the appropriate path.
> - Adds `MaybeBidiStreams` trait to `FullXmtpApiT` bounds and stubs out `XmtpMlsBidiStreams` for `BoxedStreamsClient`, `MigrationClient`, and `D14nClient` (the latter two return unretryable errors at runtime).
> - Adds `suspend_bidi_streams()` / `resume_bidi_streams()` lifecycle helpers and integration tests covering delivery, suspend/resume replay, and sync-group interception.
> - Behavioral Change: when `XMTP_BIDI_STREAMS_ENABLED` is set, all subscription callbacks run over the bidi transport; the legacy path is unchanged when the flag is unset.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized afc96f0.</sup>
> 
>

### DIFF
--- a/bindings/mobile/src/mls.rs
+++ b/bindings/mobile/src/mls.rs
@@ -77,6 +77,7 @@ use xmtp_mls::mls_common::group::GroupMetadataOptions;
 use xmtp_mls::mls_common::group_metadata::GroupMetadata;
 use xmtp_mls::mls_common::group_mutable_metadata::MessageDisappearingSettings;
 use xmtp_mls::mls_common::group_mutable_metadata::MetadataField;
+use xmtp_mls::subscriptions::router_callbacks::bidi_streams_enabled;
 use xmtp_mls::{
     client::Client as MlsClient,
     groups::{
@@ -1615,6 +1616,45 @@ impl From<&FfiMetadataField> for MetadataField {
     }
 }
 
+impl FfiConversations {
+    /// One seam for every conversation stream: route over the shared bidi
+    /// wire when `XMTP_BIDI_STREAMS_ENABLED` is set, legacy otherwise. Lives
+    /// outside the exported impl (uniffi must not see the rust-only types).
+    fn stream_conversations_dispatch(
+        &self,
+        conversation_type: Option<ConversationType>,
+        callback: Arc<dyn FfiConversationCallback>,
+    ) -> FfiStreamCloser {
+        let client = self.inner_client.clone();
+        let close_cb = callback.clone();
+        if bidi_streams_enabled() {
+            // Interim bidi scope — locally-created conversations do not
+            // surface on this stream yet: see the entry point's docs.
+            FfiStreamCloser::new(RustXmtpClient::stream_conversations_with_callback_bidi(
+                client,
+                conversation_type,
+                false,
+                move |convo| match convo {
+                    Ok(c) => callback.on_conversation(Arc::new(c.into())),
+                    Err(e) => callback.on_error(e.into()),
+                },
+                move || close_cb.on_close(),
+            ))
+        } else {
+            FfiStreamCloser::new(RustXmtpClient::stream_conversations_with_callback(
+                client,
+                conversation_type,
+                move |convo| match convo {
+                    Ok(c) => callback.on_conversation(Arc::new(c.into())),
+                    Err(e) => callback.on_error(e.into()),
+                },
+                move || close_cb.on_close(),
+                false,
+            ))
+        }
+    }
+}
+
 #[uniffi::export(async_runtime = "tokio")]
 impl FfiConversations {
     #[tracing::instrument(level = "debug", skip_all)]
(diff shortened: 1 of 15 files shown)


cameronvoell (approved):
no issues found, LGTM!

PR #3847, opened 2026-07-13
fix: install rustls crypto provider explicitly so device sync doesn't panic on Apple

### LINKED ISSUE #3846: device sync panics on startup with ctor
original: https://github.com/xmtplabs/convos-ios/pull/1169

What

Paired devices get conversations via welcomes but never old messages ("Earlier messages are hidden for privacy"). That's because message history is not carried by XMTP device sync automatically — an installation only receives pre-existing messages when it explicitly asks its peers for a history archive, and nothing in the app ever did.

    XMTPClientProvider.requestHistorySync() — wraps the SDK's sendSyncRequest() (history-server URL resolves per environment; the device-sync worker imports the answering archive automatically).
    MessagingServiceProtocol.requestHistorySync() — waits for inbox ready, calls through on the real client.
    SessionManager.refreshAfterPairingCompleted() now fires a best-effort history sync request on the adopted identity's fresh client. Pairing is the one moment the peer (initiator) is guaranteed online to answer. Emits pairing.history_sync_requested (QAEvent) on success; failure is logged and dropped.
    Corrected the SessionStateMachine.clientOptions comment that claimed device sync carries message history across devices automatically (it only carries consent/preference records).
    QA test 44 extended: Device A seeds a "History probe 44" message pre-pair; new history_sync_delivers criterion asserts pairing.history_sync_requested fires on the joiner and the probe arrives.
    New HistorySyncIntegrationTests (2 tests, currently .disabled — see blocker) covering request round-trip and end-to-end archive delivery between two installations of one inbox against the local node + history server.

The upstream blocker

With current libxmtp nightlies (4.11.0-nightly, and likely everything since the reqwest 0.13 upgrade), any history-archive transfer aborts the app on Apple platforms — and the responder crash-loops afterwards:

    libxmtp installs its rustls crypto provider in xmtp_cryptography/src/lib.rs via #[ctor::ctor] (ctor 0.12): rustls::crypto::ring::default_provider().install_default().
    ctor 0.12 registers constructor records in a _CTOR0_ISIZE_FN section and relies on a dispatcher being registered as a real Mach-O initializer. When libxmtpv3.a is linked into a Swift binary, that dispatcher never registers — ctor's own source notes its Apple link section (__DATA,__mod_init_func) is "no longer supported by Apple". Verified by dumping __init_offsets of both a macOS swift-test binary and the built Convos.debug.dylib: the CTOR record is present but orphaned; no ctor initializer exists (the app's only initializers are Sentry's).
    So the rustls process-default provider is never installed, and reqwest 0.13 (built with rustls-no-provider) hits panic!("No provider set") (async_impl/client.rs, default_rustls_crypto_provider) the first time the device-sync worker builds its HTTP client for an archive upload/download.
    The xcframework builds with panic = 'abort' → the responding app aborts. And because the sync request stays pending in the sync group, the responder dies again ~30s after every relaunch (confirmed on-sim: two identical SIGABRT crash reports) until the request is purged.

Consent/preference sync and all gRPC node traffic are unaffected (no reqwest); only the history-server HTTP path is broken. This also means sendSyncArchive/processSyncArchive (the PIN archive path) are equally unusable.

Suggested upstream fix: install the provider explicitly instead of via ctor — e.g. rustls::crypto::ring::default_provider().install_default() under a std::sync::Once in the mobile bindings' client-create path.

insipx:
Resolves https://github.com/xmtp/libxmtp/issues/3846

## Problem

`xmtp_cryptography` installs the process-default rustls crypto provider **only** through a constructor:

```rust
#[cfg(not(target_arch = "wasm32"))]
#[ctor::ctor]
fn install_rustls_crypto_provider() {
    let _ = rustls::crypto::ring::default_provider().install_default();
}
```

`ctor` 0.12 relies on a Mach-O initializer registered via the `__DATA,__mod_init_func` link section, which **Apple no longer supports**. When `libxmtpv3.a` is statically linked into a Swift binary the constructor record is orphaned and never fires, so the provider is never installed. The first time the device-sync worker builds an HTTP client for a history-archive upload/download, `reqwest` 0.13 (built with `rustls-no-provider`) hits `panic!("No provider set")`. Because the xcframework is built with `panic = 'abort'`, the responding app aborts and then crash-loops (~30s after each relaunch) until the pending sync request is purged.

Only the history-server **HTTP** path is affected; gRPC node traffic and consent/preference sync are unaffected.

## Fix

- `xmtp_cryptography` now exposes a public, idempotent `install_crypto_provider()` guarded by `std::sync::Once`.
- The existing `#[ctor::ctor]` delegates to it (unchanged behavior on non-Apple native platforms).
- The mobile bindings call `xmtp_cryptography::install_crypto_provider()` explicitly at the earliest entry points that build network/HTTP clients — `connect_to_backend` and `create_client` — so the provider is installed regardless of whether the constructor fires.

The explicit call is idempotent and cheap, so it is a no-op on platforms where the constructor already ran.

## Validation

- `cargo check -p xmtp_cryptography -p xmtpv3` — passes.
- `cargo clippy -p xmtp_cryptography -p xmtpv3 --no-deps -- -Dwarnings` — passes (no new warnings from the changed crates).
- `cargo fmt --check` — clean.

Behavior on non-Apple native and WASM targets is unchanged.

> [!NOTE]
> ### Fix rustls crypto provider panic in device sync on Apple by installing it explicitly
> - Adds a public `install_crypto_provider()` function in [`xmtp_cryptography`](https://github.com/xmtp/libxmtp/pull/3847/files#diff-a075887b8d87bf780924d062938b4ae965cef3f87b354fd35e3179f5f59b60b5) that idempotently installs the rustls `ring` default provider using `Once`, so repeated calls are safe.
> - Calls `install_crypto_provider()` at the entry point of `create_client` and `connect_to_backend` in both the [mobile](https://github.com/xmtp/libxmtp/pull/3847/files#diff-ce381c5891b3b5967aca582076bc5241ba04fc310d39949fed1859da13d876ec) and [Node](https://github.com/xmtp/libxmtp/pull/3847/files#diff-86fd7673bc2918a6066d44922085f9ea6fd5930c8fa9cc643363d2807fedc30e) bindings to ensure the provider is registered before any TLS/HTTP usage by device-sync workers.
> - The existing `install_rustls_crypto_provider` constructor is updated to delegate to the new helper rather than calling `install_default()` directly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 683d66d.</sup>
> 
>

### DIFF
--- a/bindings/mobile/src/mls.rs
+++ b/bindings/mobile/src/mls.rs
@@ -149,6 +149,11 @@ pub async fn connect_to_backend(
     auth_handle: Option<Arc<gateway_auth::FfiAuthHandle>>,
 ) -> Result<Arc<XmtpApiClient>, FfiError> {
     init_logger();
+    // Install the rustls crypto provider explicitly. On Apple platforms the `#[ctor::ctor]`
+    // in `xmtp_cryptography` never fires (the constructor link section is unsupported), so
+    // relying on it would leave reqwest without a provider and panic on the history-sync HTTP
+    // path. Idempotent, so it is safe to call on every entry point. See issue #3846.
+    xmtp_cryptography::install_crypto_provider();
 
     let client_mode = client_mode.unwrap_or_default();
 
@@ -374,6 +379,9 @@ pub async fn create_client(
 ) -> Result<Arc<FfiXmtpClient>, FfiError> {
     let ident = account_identifier.clone();
     init_logger();
+    // See `connect_to_backend` — ensure the rustls provider is installed before the
+    // device-sync worker builds its history-server HTTP client. Idempotent. See issue #3846.
+    xmtp_cryptography::install_crypto_provider();
 
     let DbOptions {
         db,
--- a/bindings/node/src/client/backend.rs
+++ b/bindings/node/src/client/backend.rs
@@ -44,6 +44,10 @@ impl BackendBuilder {
   #[napi]
   #[xmtp_common::err_span]
   pub async fn build(&self) -> Result<Backend> {
+    // Ensure the rustls crypto provider is installed before any TLS/HTTP client is built,
+    // independent of whether the `#[ctor::ctor]` in `xmtp_cryptography` fired. Idempotent.
+    // See issue #3846.
+    xmtp_cryptography::install_crypto_provider();
     {
       let mut consumed = self
         .consumed
--- a/bindings/node/src/client/create_client.rs
+++ b/bindings/node/src/client/create_client.rs
@@ -226,6 +226,12 @@ async fn create_client_inner(
   app_version: Option<String>,
   nonce: u64,
 ) -> Result<Client> {
+  // Install the rustls crypto provider explicitly rather than relying solely on the
+  // `#[ctor::ctor]` in `xmtp_cryptography`, whose constructor link section does not run on
+  // some platforms (notably Apple). Without it, the device-sync worker's history-server HTTP
+  // client panics with "No provider set". Idempotent. See issue #3846.
+  xmtp_cryptography::install_crypto_provider();
+
   let root_identifier = account_identifier.clone();
   let internal_account_identifier = account_identifier.try_into()?;
   let identity_strategy = IdentityStrategy::new(inbox_id, internal_account_identifier, nonce, None);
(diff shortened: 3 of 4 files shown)


bindings/mobile/src/mls.rs
```diff
@@ -149,6 +149,11 @@ pub async fn connect_to_backend(
     auth_handle: Option<Arc<gateway_auth::FfiAuthHandle>>,
 ) -> Result<Arc<XmtpApiClient>, FfiError> {
     init_logger();
+    // Install the rustls crypto provider explicitly. On Apple platforms the `#[ctor::ctor]`
+    // in `xmtp_cryptography` never fires (the constructor link section is unsupported), so
+    // relying on it would leave reqwest without a provider and panic on the history-sync HTTP
+    // path. Idempotent, so it is safe to call on every entry point. See issue #3846.
+    xmtp_cryptography::install_crypto_provider();
```
insipx, line 156: make sure this happens for node as well

tylerhawkes (approved):
Why don't the ctor functions get called on Apple?

PR #3866, opened 2026-07-17
feat(xmtp_mls): same-commit registry lookahead — register-then-write in one commit

tylerhawkes:
## Problem

Receivers dispatched and policy-checked every `AppDataUpdate` proposal against the **pre-commit** `COMPONENT_REGISTRY` snapshot. A commit that registers a new component and writes to it in the same commit was therefore rejected by every receiver — the write's type dispatch found no registry entry (`UnknownComponent`) and its policy lookup hit deny-by-default (`NoRegistryEntry`). "Register in an earlier commit" was documented as a convention (`registry_table.rs`) but enforced by nothing on the send side: any future client that batched the two would produce commits every fielded receiver rejects. Since v1.11 is the first stable release carrying the app-data validators, this acceptance-widening is only shippable now.

## Change

New `RegistryOverlay` (`groups/app_data/mod.rs`): tracks the registry snapshot as a commit's proposals are walked in queue order, folding each `COMPONENT_REGISTRY` mutation so subsequent proposals see it. **Both receive stages fold through this one type, so their semantics cannot drift:**

- Apply path (`accumulate_app_data_updates`): the in-batch chaining already computes the post-proposal registry snapshot; the overlay adopts it (`set_snapshot`) instead of re-applying the delta. Folding here is optimistic — results are discarded unless commit validation later passes.
- Validate path (`validate_app_data_update_proposals_in_commit`): the overlay folds a registry mutation only **after** that proposal passes its own hardcoded super-admin-only policy check, so overlaid entries carry exactly the trust of entries committed earlier. Trusting them same-commit is equivalent to trusting them one commit later — no new abuse surface.

Queue order is fixed by the commit, so every honest receiver folds identically. Write-before-register in queue order still deterministically fails. `Remove` of the whole registry component clears the overlay to deny-all (that operation is itself rejected at validation — see #3863). The change-control docs in `registry_table.rs` and the stale corner-case comment in `accumulate_app_data_updates` now state the queue-order rule instead of the old trap.

`validate_one_app_data_update` becomes `pub(super)` (mirroring its `_with_old_value` sibling) for direct test coverage.

## Testing

- `test_accumulate_registry_lookahead_register_then_write`: registry-insert + write batched in queue order applies; the write dispatches through the just-registered String type. **Mutation-verified**: with the overlay's snapshot adoption neutralized, the test fails; restored, it passes.
- `test_accumulate_registry_lookahead_write_before_register_fails`: reversed queue order still fails `UnknownComponent` — the overlay looks back, never ahead.
- `test_validate_registry_lookahead_policy_sees_folded_entry`: drives `validate_one_app_data_update` + `RegistryOverlay::fold` directly — the same pair the commit-validation loop walks: deny-by-default before the fold, allow-all entry admits the write after it.
- Existing intra-batch chaining test untouched and green. Full `xmtp_mls` + `xmtp_mls_common` suites against the local node: 1502/1502 (known flake `should_reconnect` retried). Changed files clean under clippy 1.97 (remaining flags on this machine are pre-existing trunk debt in identity.rs / mls_sync.rs / bidi_fuzz_tests.rs).
- Not covered (noted honestly): an end-to-end staged-commit test through the full commit-validation loop — no production sender currently batches register+write in one commit, and hand-building the staged commit needs a harness this PR doesn't add. The loop-level fold is exercised via the direct-drive test above.

Composes with #3863: entry-level validation there means only structurally valid, super-admin-authored registry mutations ever reach the overlay.

Part of the pre-v1.11 app-data hardening plan (P0.6) — the last of the Phase-0 protocol-acceptance changes. Siblings: #3863, #3864, #3865.

https://claude.ai/code/session_01PkZpajnDV829SVuW9danrW

> [!NOTE]
> ### Add same-commit registry lookahead so register-then-write succeeds in one commit
> - Introduces `RegistryOverlay` in [mod.rs](https://github.com/xmtp/libxmtp/pull/3866/files#diff-f0cc46ebb4d5cc3d18a7be0c2cab862e1d491117402b03370c19b1ca6dadcf3d) to maintain an in-batch view of the component registry while processing `AppDataUpdate` proposals, updating the overlay as each registry mutation is applied in queue order.
> - Updates `accumulate_app_data_updates` and `validate_app_data_update_proposals_in_commit` in [validated_commit.rs](https://github.com/xmtp/libxmtp/pull/3866/files#diff-233994c0ec18d23d0636e849849a3500e44611a4f64c6fffb1d0931aa1461ed6) to use the overlay so later proposals in the same commit see components registered earlier in the batch.
> - A write queued before its registration in the same commit still fails (`UnknownComponent`); removing the registry mid-batch yields a deny-all view for subsequent proposals.
> - Adds tests covering register-then-write success, write-before-register failure, and direct overlay validation behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ec6ef73.</sup>
> 
>
>

### DIFF
--- a/crates/xmtp_mls/src/groups/app_data/mod.rs
+++ b/crates/xmtp_mls/src/groups/app_data/mod.rs
@@ -41,6 +41,109 @@ use self::component_source::{
     ComponentSourceError, apply_app_data_update_payload, read_from_app_data_dict,
 };
 
+/// Same-commit registry lookahead: tracks the `COMPONENT_REGISTRY`
+/// snapshot as a commit's `AppDataUpdate` proposals are walked in queue
+/// order, so a component *registered* by an earlier proposal in the
+/// same commit dispatches and policy-checks exactly like one registered
+/// in an earlier commit. Without this, register-then-write batched into
+/// one commit is rejected by every receiver (the write's type dispatch
+/// and deny-by-default policy lookup would only see the pre-commit
+/// snapshot) — a permanent trap enforced by nothing on the send side.
+///
+/// Both receive stages fold through this one type so their semantics
+/// cannot drift:
+/// - the apply path ([`accumulate_app_data_updates`]) folds
+///   optimistically — its results are discarded unless commit
+///   validation later passes;
+/// - the validate path
+///   (`validate_app_data_update_proposals_in_commit`) folds only after
+///   the registry proposal itself has passed its hardcoded
+///   super-admin-only policy check, so the overlaid entries carry the
+///   same trust as ones committed earlier.
+///
+/// Queue order is fixed by the commit, so every honest receiver folds
+/// identically; write-before-register in queue order deterministically
+/// still fails.
+pub(crate) struct RegistryOverlay {
+    registry: ComponentRegistry,
+    snapshot_bytes: Option<Vec<u8>>,
+}
+
+impl RegistryOverlay {
+    /// Start from the pre-commit dict state of `mls_group`.
+    pub(crate) fn from_group(mls_group: &OpenMlsGroup) -> Result<Self, ComponentSourceError> {
+        Ok(Self {
+            registry: load_component_registry(mls_group)?,
+            snapshot_bytes: read_from_app_data_dict(ComponentId::COMPONENT_REGISTRY, mls_group),
+        })
+    }
+
+    /// Start from an already-loaded registry (the validator pre-loads
+    /// one per commit) plus the matching pre-commit dict bytes.
+    pub(crate) fn new(registry: ComponentRegistry, snapshot_bytes: Option<Vec<u8>>) -> Self {
+        Self {
+            registry,
+            snapshot_bytes,
+        }
+    }
+
+    /// The registry as of the last folded proposal (initially the
+    /// pre-commit snapshot). Hand this to type dispatch and policy
+    /// lookups for each subsequent proposal.
+    pub(crate) fn registry(&self) -> &ComponentRegistry {
+        &self.registry
+    }
+
+    /// Fold one `COMPONENT_REGISTRY` `AppDataUpdate` operation. Callers
+    /// invoke this only for proposals targeting the registry component,
+    /// in queue order.
+    pub(crate) fn fold(
+        &mut self,
+        operation: &AppDataUpdateOperation,
+    ) -> Result<(), ComponentSourceError> {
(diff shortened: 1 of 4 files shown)


tylerhawkes:
## Re-review (D9): BLOCK — recommend deferring this loosening out of v1.11

Adversarial re-review + an independent verifier found two critical issues, both rooted in the same-commit register-then-write **loosening** this PR introduces:

1. **The central invariant is false.** The PR's premise is "both receive stages fold registry updates identically in queue order." They don't: the apply stage walks `committed_proposals()` in raw commit-body order, while the validate stage walks `staged_commit.app_data_update_proposals()`, which the openmls fork **sorts** by `(component_id, removes-first)` (`proposal_store.rs:505`). Because `COMPONENT_REGISTRY` (0x8000) is the minimum id, registry mutations always sort to the front of validation — so validation is order-insensitive while apply enforces body order. No same-version fork is reachable today (apply gates first, deterministically), but every "the two stages can't drift" comment shipped here is wrong, one shape (`[register X, write X, Remove registry]`) both-accepts yet silently de-migrates, and any future refactor flips acceptance silently.

2. **It arms an honest-sender fork.** The loosened `accumulate_app_data_updates` (with the overlay) runs at *sender* stage time too, so an honest sender can build+publish a register+write batch that every receiver rejects at the standalone-proposal arrival gate (which validates against the committed registry, with no overlay) → the sender forks itself off the group.

**Recommendation: defer.** libxmtp doesn't currently produce same-commit register-then-write (registration and component-writes each go through their own propose-then-commit cycle), so this is speculative future-proofing that arms two fork risks while falsifying its own invariant. Keeping the strict pre-PR behavior (write-to-unregistered-component rejected) is the safe choice for the permanent-floor v1.11 contract. Same-commit register-then-write can be added later, correctly, gated behind a `PROPOSALS_MIN_PROTOCOL_VERSION` floor bump — loosening acceptance post-cut needs a floor bump anyway, so nothing is lost by deferring.

If we do want it in v1.11, the reconciliation is non-trivial: validation must iterate the same source/order as apply (not the fork's sorted iterator), the standalone-proposal arrival path needs consistent overlay semantics, and the contradictory change-control docs at both validator entry points need fixing — see `app-data-v1.11-hardening-plan.md` "D9 re-review results".

insipx (approved):

tylerhawkes:
This currently goes against XIP-81 spec, so will be closing for now.

PR #3867, opened 2026-07-17
fix(bindings,sdks): gate enable_proposals behind a per-language unstable namespace

tylerhawkes:
## Problem

`enableProposals()` is a one-way, irreversible per-group migration: the moment any app calls it, that group's cohort is permanently on the app-data format and every pre-1.11 member is paused until upgrade. It must not sit on the **stable** API surface for the v1.11 cut until the app-data hardening work (#3863–#3865 plus the remaining product fixes) has landed and been re-reviewed.

But first-party apps (convos-ios / convos-android) still need a way to call it — so hiding it outright is wrong. The first revision of this PR did exactly that on mobile (`internal` / non-`public`), which **removes** the capability from every downstream consumer, and merely **renamed** it on JS (`unstableEnableProposals`), which gives no compile-time signal when it graduates.

## Change

One consistent model across all four bindings: an **`unstable` namespace** that groups pre-release functions under a single opt-in. The function keeps its **real, final name**; promotion to stable is a **compile-time break** for callers of the gated form (the function moves out of the namespace). New unstable functions — the custom-component-id internals landing next — accrete onto the same namespace with **no per-function annotation**.

| Binding | Gate (applied once) | Call site |
| --- | --- | --- |
| node | `conversation.unstable` namespace object (`UnstableConversation`, `#[napi(getter)]`) | `conversation.unstable.enableProposals(opts)` |
| wasm | same (`#[wasm_bindgen(getter)]`) | `conversation.unstable.enableProposals(opts)` |
| iOS | `@_spi(Unstable) public var unstable: UnstableGroup` | `@_spi(Unstable) import XMTPiOS` → `group.unstable.enableProposals(...)` |
| Android | `@UnstableApi val unstable: UnstableGroup` (`@RequiresOptIn` marker, mirrors existing `DelicateApi`) | `@OptIn(UnstableApi::class)` → `group.unstable.enableProposals(...)` |

- **node**: new `bindings/node/src/conversation/unstable.rs` holding `UnstableConversation` + the `EnableProposalsOptions` it consumes, plus a `#[napi(getter)] unstable` on `Conversation`.
- **wasm**: `UnstableConversation` + `#[wasm_bindgen(getter)] unstable`, inline in `conversation.rs` (matching that binding's single-file layout).
- **iOS**: new `UnstableGroup.swift`; a single `@_spi(Unstable)` on the `Group.unstable` accessor gates the whole type — adding functions needs no per-function attribute.
- **Android**: new `UnstableGroup.kt` + `UnstableApi.kt` (`@RequiresOptIn(level = ERROR)` marker); the marker is applied **once** on the `Group.unstable` accessor, not per function.
- Doc links that referenced `Group.enableProposals` now point at `UnstableGroup.enableProposals`.
- Core `MlsGroup::enable_proposals`, the mobile FFI, and the read surfaces (`proposalsEnabled`, `membershipCapabilities`) are unchanged.

**Promotion, when the hardening lands:** move `enableProposals` off the `unstable` namespace onto the primary type (all four). Each move is a compile-time break for existing opt-in callers.

## Testing

- `cargo check -p bindings_node` ✅, `cargo check -p bindings_wasm --target wasm32-unknown-unknown` ✅, clippy `-Dwarnings` ✅ on both (native + wasm32), fmt clean.
- No in-repo references to the old `unstableEnableProposals` name; no bare `Group.enableProposals` references remain.
- iOS/Android are visibility/namespace-only changes that mirror existing patterns; **not compiled locally** (no Xcode/Gradle+nix in this env) — CI validates the mobile lanes. No Swift/Kotlin code call sites reference the moved symbol (only doc links, now updated).

Part of the pre-v1.11 app-data hardening plan (P1.1 / D5). Supersedes the rename/hide approach in the first revision of this PR.

> [!NOTE]
> ### Gate `enableProposals` behind a per-language unstable namespace across Node, WASM, Android, and iOS SDKs
> - Moves `enableProposals` off the stable `Conversation`/`Group` surface and onto a new `unstable` accessor (`conversation.unstable.enableProposals` / `group.unstable.enableProposals`) in all SDK bindings.
> - Introduces `UnstableConversation` (Node, WASM) and `UnstableGroup` (Android, iOS) types that hold the actual implementation.
> - Each platform uses its idiomatic opt-in mechanism: `@_spi(Unstable)` on iOS, `@RequiresOptIn` (`@UnstableApi`) on Android, and a plain getter on Node/WASM.
> - Behavioral Change: callers using `enableProposals` directly on `Conversation` or `Group` will have a compile/runtime break and must migrate to the `unstable` accessor.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dfd88ec.</sup>
> 
>

### DIFF
--- a/bindings/node/src/conversation/metadata.rs
+++ b/bindings/node/src/conversation/metadata.rs
@@ -3,28 +3,6 @@ use napi::bindgen_prelude::Result;
 use napi_derive::napi;
 use xmtp_mls::mls_common::group_metadata::GroupMetadata as XmtpGroupMetadata;
 
-/// Options for [`Conversation::enableProposals`]. Mirrors
-/// [`xmtp_mls::groups::EnableProposalsOptions`].
-#[napi(object)]
-pub struct EnableProposalsOptions {
-  /// Skip the pre-flight key-package capability check. Post-d14n
-  /// every client supports proposals by version floor alone; set
-  /// `true` to bypass the per-member scan in that environment.
-  pub force: Option<bool>,
-  /// Override the `MIN_SUPPORTED_PROTOCOL_VERSION` floor. `None`
-  /// defaults to `xmtp_configuration::PROPOSALS_MIN_PROTOCOL_VERSION`.
-  pub min_version: Option<String>,
-}
-
-impl From<EnableProposalsOptions> for xmtp_mls::groups::EnableProposalsOptions {
-  fn from(opts: EnableProposalsOptions) -> Self {
-    xmtp_mls::groups::EnableProposalsOptions {
-      force: opts.force.unwrap_or(false),
-      min_version: opts.min_version,
-    }
-  }
-}
-
 #[napi]
 pub struct GroupMetadata {
   metadata: XmtpGroupMetadata,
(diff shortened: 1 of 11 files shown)


tylerhawkes:
Follow-up gap (does not block this PR, which is approved): the SDK wrappers gate `enableProposals`, but the raw mobile uniffi symbol `FfiConversation.enableProposals` remains public and callable by apps directly (the reviewer's MAJOR point). The clean fix is to rename the FFI method to `unstable_enable_proposals` (matching the wasm/node `unstableEnableProposals`) and regenerate `xmtpv3.swift` + the two SDK wrapper call sites. The rename compiles clean, but it needs a working `sdks/ios/dev/bindings` regen (the generated binding's FFI checksums can't be hand-edited safely). Tracked in the app-data hardening plan as a pre/post-cut follow-up.

insipx (approved):

PR #3870, opened 2026-07-17
feat(bindings): unify the app-data FFI surface across mobile, wasm, and node

tylerhawkes:
## Problem

Three inconsistencies in the app-data FFI surface, from the pre-v1.11 review — each one bakes divergent app expectations across platforms the moment the first stable release ships:

1. **Migration introspection was split disjointly**: wasm/node exposed `proposalsEnabled()` but not `membershipCapabilities()`; mobile exposed the capabilities snapshot but no boolean — and its docs told apps to detect migration by scanning `contextExtensions` for `AppDataDictionary`, coupling iOS/Android apps to an internal MLS extension identity.
2. **`updateAppData` took a bare `String`** on all three bindings — no way to ever add a parameter without a breaking signature change.
3. **The `MetadataField` enums disagreed**: mobile's `FfiMetadataField` was missing the `MessageExpirationFromNs`/`InNs` variants wasm/node have.

## Changes

- Mobile gains `proposals_enabled()` (same core call as wasm/node); the `membershipCapabilities` docs now steer apps to the semantic bool for the "is this group migrated?" question and reserve the capabilities snapshot for "which members block migration".
- `updateAppData` takes an options record on all three bindings (`FfiUpdateAppDataOptions` / `UpdateAppDataOptions { value }`), the same growable pattern as `EnableProposalsOptions`. The **public Swift/Kotlin SDK signatures are unchanged** (`updateAppData(appData: String)` — the wrappers construct the options internally), so no app-facing break anywhere.
- `FfiMetadataField` gains the two `MessageExpiration*` variants, mapped to the core `MessageDisappear*` fields — all three bindings now expose the same field set.
- Checked-in iOS uniffi bindings (`xmtpv3.swift`) regenerated via `sdks/ios/dev/bindings` (nix). Android Kotlin bindings are generated at build time and need no checked-in update.

## Testing

`xmtpv3` + `bindings_node` suites: 153/153. `just wasm check` and `cargo check` green across all three bindings. In-repo mobile tests updated to the options struct (no external consumers exist — verified in the P1.1 sweep).

Part of the pre-v1.11 app-data hardening plan (P1.4 + P1.5 + P1.6). Siblings: #3863–#3869.

https://claude.ai/code/session_01PkZpajnDV829SVuW9danrW

> [!NOTE]
> ### Unify `updateAppData` FFI surface across mobile, WASM, and Node bindings
> - Replaces the bare `String` parameter in `updateAppData` with an `UpdateAppDataOptions`/`FfiUpdateAppDataOptions` options object across all bindings (mobile, WASM, Node). Public SDK APIs for iOS, Android, and JS are unchanged.
> - Adds a `proposalsEnabled()` method to the mobile FFI, iOS SDK, and Android SDK to expose whether a group uses AppData-proposal-based metadata updates.
> - Adds `messageExpirationFromNs` and `messageExpirationInNs` variants to `FfiMetadataField` in the mobile bindings, mapping to the underlying disappear-from/in fields.
> - Behavioral Change: callers using the raw FFI or native addon directly (not via the SDK wrappers) must update `updateAppData` call sites to pass an options object.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 74bc001.</sup>
> 
>

### DIFF
--- a/bindings/mobile/src/mls.rs
+++ b/bindings/mobile/src/mls.rs
@@ -1688,12 +1688,32 @@ impl TryFrom<FfiPermissionPolicySet> for PolicySet {
     }
 }
 
+/// Options for [`FfiConversation::update_app_data`]. A record (rather
+/// than a bare `String` parameter) so future knobs can be added
+/// without breaking compiled apps — same pattern as
+/// [`FfiEnableProposalsOptions`].
+///
+/// WARNING: uniffi Records get NO default field values unless the field
+/// carries `#[uniffi(default = ...)]`. Any field added later MUST carry
+/// a uniffi default (and a serde/napi default on the wasm/node
+/// `UpdateAppDataOptions`), or the generated Swift/Kotlin constructors
+/// change and the addition breaks compiled apps.
+#[derive(uniffi::Record, Clone, Default, Debug)]
+pub struct FfiUpdateAppDataOptions {
+    /// The new value for the group's opaque `APP_DATA` string slot.
+    pub value: String,
+}
+
 #[derive(uniffi::Enum, Debug)]
 pub enum FfiMetadataField {
     GroupName,
     Description,
     ImageUrlSquare,
     AppData,
+    // Present on the wasm and node bindings from the start; added here
+    // so the three bindings expose the same field set.
+    MessageExpirationFromNs,
+    MessageExpirationInNs,
 }
 
 impl From<&FfiMetadataField> for MetadataField {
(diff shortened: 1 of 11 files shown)


tylerhawkes:
Note for reviewers: the regenerated `sdks/ios/.../xmtpv3.swift` in this PR sweeps in ~290 lines of PR #3857's public API surface (`catchUpToLive`, `FfiCatchUpSummary`, `suspendStreams`, `resumeStreams`) — #3857 landed those Rust methods without regenerating the checked-in bindings, so the first regen after it (this PR) picks them up. The swept-in content is correct and required for the file to match head Rust; it is not authored here. A follow-up should add a CI check that the checked-in bindings are current so this stops riding on unrelated PRs.

PR #3874, opened 2026-07-17
feat(xmtp_api_d14n): bound the bidi Mutate frame by chunking add-waves

tylerhawkes:
Bounds the XIP-83 bidi `Mutate` frame so a single add-wave can never approach the transport's 25 MiB encode limit — the last enablement-blocker from the bidi release audit (item 6). Independent of #3873 (different regions of `bidi_transport.rs`).

## Problem
An add-wave rode a single `Mutate` frame carrying every topic. A lease over very many groups — or a reconnect resuming an agent's whole subscription set — could approach the 25 MiB `max_encoding_message_size` (~650k topics), and silently assumed the server's inbound limit matched ours. Agents (many groups, never suspended) are the growth case.

## Change
Every add-set is split into ≤`MAX_MUTATE_TOPICS` (1000) topics per frame, each its own wave, at all three emit sites: cold/warm `lease()`, the reconnect resume wave, and each reconnect reissue (helper `chunk_waves`).

Chunking a lease into several same-owner waves required **one** change to the completion invariant: `complete()` now also defers a lease that still owns another pending wave, so a chunked lease isn't caught up until its **last** chunk resolves — reusing the existing `holds`/deferred-completion machinery. This new arm is:
- a **monotone tightening** of the existing blocker predicate — it can only add deferral, never remove it;
- a **no-op for an unchunked lease** (the single wave is removed before `complete()` runs), so single-frame behavior is byte-for-byte unchanged.

The per-frame cap is a **per-transport field** (`Ledger::chunk_cap`) defaulting to `MAX_MUTATE_TOPICS`; a test-only constructor (`new_with_chunk_cap`) shrinks it so the property test can fuzz chunked waves without leasing thousands of topics — no shared global, so concurrent tests never race on it.

## Testing
- **58 existing** bidi transport/connection tests pass unchanged (single-chunk path behavior-identical).
- **2 scripted tests**: a >1000-topic lease splits into bounded frames while staying deferred until its last chunk completes; a >1000-topic reconnect resume chunks the same way.
- **Property fuzz**: the model-based test gains a chunked variant (cap 1–2) exercising the multi-wave-per-lease completion accounting under randomized interleaved delivery, yank overlap, and dedup. The model's "exactly one `CatchUpComplete` per lease" assertion is the tripwire. **Passed 400 cases**; the default variant passed 256.
- `cargo clippy --all-features --all-targets` clean on both changed files.

## Review
Reviewed by two independent adversarial passes before opening:
- **Completion accounting** (5 scenarios: no-op-for-unchunked, termination/exactly-once under any CatchUpComplete order, sibling-gap composition, deref-mid-catch-up, `reconnect_plan` under k-waves-per-lease) — all CONFIRMED-OK.
- **`lease()`/`reopen()` refactor** (10 items incl. the suspend-cold-open peeled-chunk survival and the reopen initial-frame no-double-send) — all OK; the empty-lease path is strictly safer than before.

Both independently noted one pre-existing (non-regression) behavior: a lease registered while the wire is down and deref'd before reopen leaves never-sent waves dangling until the reopen drain — identical to the single-wave path, self-heals at reopen.

https://claude.ai/code/session_017AVuJ7VAgen2fFhtJQ6zid

> [!NOTE]
> ### Bound bidi Mutate frames by chunking large add/remove waves in `xmtp_api_d14n`
> - Adds `MAX_MUTATE_TOPICS` and `MAX_MUTATE_BYTES` constants plus a `chunk_by_budget` helper to split topic sets into size-bounded chunks before sending Mutate frames.
> - `Ledger::chunk_waves` and `chunk_mutate_adds` apply these limits when building add-waves for new leases, reconnect resumes, and bounded catch-up in `catch_up.rs`.
> - Lease drops now also chunk remove topics into multiple Mutate frames, each with a non-zero wave id for ack correlation.
> - Blocker selection in `LedgerTask::register` is fixed so a multi-chunk lease is not marked caught up until its final wave completes.
> - Risk: leases and reconnects that previously sent a single large Mutate frame now generate multiple frames; completion tracking covers all outstanding wave ids.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a6f6d37.</sup>
> 
>

### DIFF
--- a/crates/xmtp_api_d14n/src/queries/bidi_transport.rs
+++ b/crates/xmtp_api_d14n/src/queries/bidi_transport.rs
@@ -177,6 +177,65 @@ const OUTBOX_RETRY_INTERVAL: std::time::Duration = std::time::Duration::from_mil
 /// converts unbounded memory growth into the channel-wedge recovery: the
 /// lease is dropped and its consumer re-leases from durable cursors.
 const WITHHELD_FRAMES_CAP: usize = 512;
+/// The most topics packed into one `Mutate` frame. A larger add-wave — a
+/// lease over very many groups, or a reconnect resuming an agent's whole
+/// subscription set — is split into this many topics per frame. A predictable
+/// per-frame count (bounding memory and latency per wave) that works alongside
+/// the byte budget below; whichever is reached first closes the frame. Chunks
+/// of one lease share its wave owner, so the lease stays catching-up until its
+/// last chunk resolves ([`LedgerTask::complete`]). Held as a field
+/// ([`Ledger::chunk_cap`]) so tests shrink it per transport without a shared
+/// global; production is always this value.
+pub(crate) const MAX_MUTATE_TOPICS: usize = 1000;
+/// Byte ceiling for one `Mutate` frame's topic payload — well under the
+/// transport's 25 MiB encode limit, leaving headroom for the frame's own
+/// framing. Bounding by *bytes*, not just count, keeps a frame under the limit
+/// regardless of topic identifier length: [`MAX_MUTATE_TOPICS`] alone would not,
+/// since a [`Topic`] is a `Vec<u8>` with no type-level size bound (today every
+/// topic is a 17–33 B kind+id, but the guarantee should not rest on that).
+pub(crate) const MAX_MUTATE_BYTES: usize = 16 * 1024 * 1024;
+/// Upper bound on one entry's encoded overhead in a `Mutate` beyond its topic
+/// bytes: the `id_cursor` varint (≤10 B) plus protobuf field tags and length
+/// prefixes for the `Subscription`, generous enough to stay an upper bound
+/// across both bindings' cursor shapes (a removes entry has no cursor, so this
+/// over-counts it — safe, it only makes chunks smaller).
+const PER_ENTRY_OVERHEAD: usize = 64;
+
+/// Upper bound on the bytes one topic contributes to a `Mutate` frame: the
+/// full topic bytes (kind byte + identifier) plus [`PER_ENTRY_OVERHEAD`]. An
+/// over-estimate, so a byte-budgeted chunk never exceeds the budget.
+fn topic_wire_cost(topic: &Topic) -> usize {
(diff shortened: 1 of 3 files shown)


tylerhawkes:
Addressed in 970ee7f5.

`MAX_MUTATE_TOPICS` is now a *secondary* bound. The primary is a byte budget (`MAX_MUTATE_BYTES` = 16 MiB), enforced per frame by a shared `chunk_by_budget` helper that closes a chunk when either the count cap **or** the byte budget would be exceeded. Per-entry cost comes from `topic_wire_cost` = full topic bytes (kind + identifier) + a generous `PER_ENTRY_OVERHEAD` (64 B for the cursor varint and protobuf framing), an upper bound — so a frame stays under the 25 MiB encode limit **regardless of topic identifier length**. The `Vec<u8>`-backed `Topic` no longer has to be assumed small for the guarantee to hold.

I also extended the same bound to the removes wave in `retire` (a mass unsubscribe of a large multi-topic lease), which was the other unchunked `Mutate` producer — the two `build_mutate` call sites are now the only ones, and both go through the budget.

New tests: `a_lease_over_the_byte_budget_splits_into_bounded_frames` (byte-driven split with the count cap effectively off) and `a_mass_unsubscribe_chunks_the_removes_wave`.

tylerhawkes:
Added a second commit completing the frame-bounding. The bounded catch-up path (`catch_up_to_live` → `catch_up_bidi`) opens its own `BidiConnection` with a hand-built `history_only` Mutate naming every topic the client owns — a Mutate producer outside the transport that the wave chunking in the first commit doesn't reach. A client priming a large account would open catch-up with one oversized frame.

The initial subscription set is now chunked to the same per-`Mutate` limit through a new `chunk_mutate_adds` helper, so the bound has a single source of truth: the first chunk seeds the open, the rest ride the existing follow-up queue as their own waves, each completed like a discovery add. Small accounts are byte-identical — empty overflow, unchanged bookkeeping. The per-group discovery follow-up already carries a single topic, so it's untouched.

insipx (approved):

PR #3894, opened 2026-07-23
test(sdks/js): fix two residual delivery races in the JS test poll conversions

tylerhawkes:
Follow-up to #3891 fixing two residual races its poll conversions left behind, both caught by CI runs today (#3872's rerun and this PR's own first run).

## streamConsent: batch-count equality can overshoot

`should stream consent updates` waits on `expect(batches.length).toBe(n)` — exact equality on a counter that only grows. Background workers can emit their own consent batches at any time (the failing run logged worker sync activity mid-test), so the count can skip past `n` between polls; once it does, the equality can never match again and the `waitFor` times out. Observed as `expected 5 to be 3` after 31s on node-sdk shard 1.

Fixed by making the waits content-based:

- After each issued update, poll until that update's `(entityType, entity, state)` appears anywhere in the stream — extra worker batches can't break it.
- The final two-entry `setConsentStates` polls for a **single batch** containing both entries, preserving the batched-delivery semantic the old `batches[2].length === 2` assertion was checking.
- The exact-index assertions (`batches[0][0]…`) are gone: batch indices aren't stable under worker traffic, and everything they verified is covered by the content waits.

## sendSyncRequest: the one-shot fix #3891 missed

#3891 established that device-sync messages are one-shot: published before the peer installation has joined the sync group, they are undecryptable for it forever. It moved `sendSyncArchive` and the consent update inside the retry loops — but left `sendSyncRequest` issued once, outside the poll. Lose that join race and client1's worker never sees the request, no archive ever comes back, and the round-trip poll times out no matter how long it retries. Observed as `expected undefined to be truthy` at 30s on **both** node-sdk shards in this PR's first run.

Fixed by re-issuing the request on each poll attempt, same pattern as the archive test. Multiple requests are harmless: the poll exits on the first successful import, and the message-identity assertion was already conditional on counts matching.

Both fixes applied identically to node-sdk and browser-sdk.

## Verification

- node-sdk `Preferences.test.ts`: 3 consecutive runs against the local backend, 5/5 each.
- node-sdk `DeviceSync.test.ts`: 3 consecutive runs, 3/3 each.
- `yarn eslint` and `tsc` clean on all four files; `just lint-treefmt` reports 0 changes.

> [!NOTE]
> ### Fix delivery races in JS SDK device sync and consent streaming tests
> - Replaces single-shot `sendSyncRequest` calls with retry-inside-poll logic in `DeviceSync` tests for both browser and node SDKs, re-issuing the request on each `vi.waitFor` attempt to handle async group joins.
> - Adds a `ROUND_TRIP_WAIT` constant (90s timeout, 3s interval) to account for the two-hop background-worker round trip in device sync tests.
> - Replaces batch-count and index-based assertions in `Preferences` streaming tests with content-based checks using an `observed()` helper, tolerating unstable batch boundaries from background workers.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4ae5065.</sup>
> 
>

### DIFF
--- a/sdks/js/browser-sdk/test/DeviceSync.test.ts
+++ b/sdks/js/browser-sdk/test/DeviceSync.test.ts
@@ -11,6 +11,12 @@ import { createRegisteredClient, createSigner } from "@test/helpers";
 // on loaded CI runners.
 const WAIT = { timeout: 30_000, interval: 1000 };
 
+// The request→archive→import round trip crosses two background-worker hops
+// (one worker builds and uploads the archive, the other downloads and
+// imports it), so give it a much larger budget than the single-hop waits
+// and pace the retries so the workers aren't flooded with requests.
+const ROUND_TRIP_WAIT = { timeout: 90_000, interval: 3000 };
+
 describe("DeviceSync", () => {
   it("should sync consent across installations", async () => {
     const { signer: boSigner } = createSigner();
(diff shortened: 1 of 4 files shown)


tylerhawkes:
Pushed a follow-up: the sendSyncRequest round trip still timed out on shard 2 with the re-issue fix in place — it crosses two background-worker hops (archive build+upload, then download+import), and 30s isn't enough budget on a loaded runner. The round-trip poll now gets 90s with a 3s re-issue cadence (`ROUND_TRIP_WAIT`), well under the 120s vitest ceiling. Single-hop waits keep the 30s budget.

insipx (approved):

PR #3897, opened 2026-07-24
fix(xmtp_mls,xmtp_proto): durable AddMissingInstallations reconciliation on the TaskRunner

insipx:
## Problem

The node-sdk `DeviceSync > should sync messages across installations using sendSyncRequest and syncAllDeviceSyncGroups` test has been failing most CI runs since Jul 23 (10 of 14 node-sdk jobs across main and several branches), always with the same signature: the group never appears on the second installation for the full 90s poll, while the same code sometimes converges in ~3s.

Root cause: when a new installation registers, the DeviceSync worker's `NewSyncGroupFromWelcome` handler ran `add_new_installation_to_groups()` **inline, exactly once**:

- per-group failures inside `queue_for_each` were warn-logged and swallowed (`groups/intents/queue.rs`), so a single lost race (e.g. `MissingSequenceId` when the membership add races identity propagation — see the diagnosis on `tyler/seqid-retryable`) silently skipped the add;
- the triggering event is never re-fired and nothing else in the test scenario sends a message that would run the add-missing-installations-on-send fallback, so the miss is **permanent** — 90s of re-issued sync requests can't recover it (with `elements: []` the archive round trip is empty, so the welcome path is the only delivery path).

Retryability classification alone (`tyler/seqid-retryable`) can't fix this: the queue-time errors are dropped regardless of classification, and an abandoned publish leaves the intent dormant with nothing to re-trigger it.

## Fix

Make the reconciliation durable: the welcome handler now enqueues one `AddMissingInstallations { group_id }` task per eligible group on the TaskRunner (`create_or_ignore_task`, deduped by payload hash, then a wake). The TaskRunner arm loads the group and runs the existing idempotent `MlsGroup::add_missing_installations()`; failures ride the standard task backoff (2s ×1.5 capped at 60s, 20 attempts, 3-day expiry), missing/malformed groups delete the task.

- `xmtp_proto`: new `Task` oneof variant `AddMissingInstallations` (tag 7) hand-added to the checked-in generated files ahead of the upstream xmtp/proto PR (same pattern as the earlier KpMaintenance work). Wire encoding pinned in `data_hash_encoding_is_pinned`.
- `xmtp_mls`: `schedule_add_installations_to_groups()` replaces the inline bulk add; new TaskRunner arm; `queue_for_each` (now dead, and the error-swallowing culprit) removed.
- tests: enqueue+dedup unit test, missing-group task-deletion unit test, live end-to-end integration test (welcome → task → commit → welcome). `test_only_added_to_correct_groups` polls for convergence (the metric now fires at schedule time), and `test_has_same_sync_group_as` converges on epoch too — the post-join KeyUpdate commit riding the first HMAC-cycle message landed deterministically inside its old assert window.

## Verification

- `xmtp_mls` 707/707, `xmtp_db` + `xmtp_proto` 322/322 (`cargo nextest`, ci profile)
- `just lint-rust` clean
- node-sdk `DeviceSync.test.ts` (including the exact failing CI test) 3/3 green against freshly built bindings

## Follow-ups

- Upstream xmtp/proto PR adding the variant (message + oneof field 7 — field numbers are wire-pinned by persisted task rows), then `dev/gen_protos.sh` regen + `proto_version` bump.
- Independent JS-side deflake: the DeviceSync tests request `elements: []`, which produces an empty archive; requesting `Messages` would make each retry attempt self-sufficient.

> [!NOTE]
> ### Add durable `AddMissingInstallations` task reconciliation to the TaskRunner
> - Introduces a new `AddMissingInstallations` proto task variant and wires it into the `TaskWorker` dispatch loop, replacing inline membership adds with retryable, deduplicated durable tasks.
> - When `add_missing_installations()` fails during sync group creation in `DeviceSyncClient`, a durable task is scheduled instead of silently dropping the add.
> - `SyncWorker.evt_new_sync_group_from_welcome` now schedules tasks via `schedule_add_installations_to_groups` rather than executing inline adds, and recovers from broadcast receiver lag by re-scheduling reconciliation.
> - The `TaskWorker` deletes `AddMissingInstallations` tasks permanently on malformed group ID or group not found, and retries on transient errors.
> - Behavioral Change: `QueueIntentBuilder::queue_for_each` is removed; callers must use the new task-based scheduling path.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 761b197.</sup>
> 
>


insipx:
**Blocked on xmtp/proto#341** — this PR's generated-code changes for the `AddMissingInstallations` variant must land upstream first. The gen files here are now verbatim `dev/gen_protos.sh` output from that proto branch (including `proto_descriptor.bin`); after xmtp/proto#341 merges, the only remaining step is a `proto_version` repin + regen commit (expected no-op diff). Marked draft until then.

tylerhawkes:
@insipx three clarifications from the diagnosis on `tyler/seqid-retryable`, since this PR lands the durable half of it:

1. `get_sync_group`'s creation branch still has the one-shot shape this PR removes elsewhere: `create_and_insert` persists the row, then `add_missing_installations().await?` runs inline exactly once — if that fails, the `Some(...)` branch never re-attempts and the sync group stays permanently under-membered. Enqueueing the same `AddMissingInstallations` task for the just-created group (instead of, or as a safety net behind, the inline add) closes it with machinery this PR already builds.

2. The schedule step itself is still at-most-once from the event edge: `run_internal`'s `while let Ok(event) = receiver.recv().await` exits on a broadcast `Lagged`, the supervisor restarts the loop on the same receiver, and the lagged events — including the `NewSyncGroupFromWelcome` that would have scheduled the tasks — are gone with no task row ever created. `schedule_add_installations_to_groups()` is cheap and dedup'd, so running it once on worker (re)start or on `Lagged` recovery makes the scheduling level-triggered instead of edge-triggered.

3. For the writeup's why-now: the onset matches the Blacksmith migration (#3883, merged Jul 22 17:40 UTC) — zero failures of this test in the 40 node-sdk runs before it, ~85% of shard-instances during US-night hours after, identical trees passing and failing on either side of the boundary. The race predates Jul 23; the fleet change flipped the odds. That also explains the 3s-vs-never bimodality in the Problem section.

#3895 stays complementary: each task attempt's `add_missing_installations` re-runs the identity fetch under the intent machinery's retry loop, which the retryable classification re-arms.

tylerhawkes (approved):

PR #3931, opened 2026-08-02
fix: preserve ordered welcome cursor retries

Jr-kenny:
Fixes #3396

When welcome messages are queried in cursor order, a retryable failure can be followed by another welcome. The later welcome was still processed and advanced the durable welcome cursor, which made the failed welcome disappear from the next sync.

I reproduced this locally with two real MLS welcomes. The first hit the retryable `LockUnavailable` path and the second was valid. The old implementation failed the regression with `later welcome advanced the durable cursor past a retryable failure`. The fix stops the batch at the first retryable error after the configured retries, so the failed cursor stays available for the next sync.

Non-retryable welcomes still get skipped so one malformed or permanently invalid welcome does not block the rest of the batch.

Checks run on macOS:

- `cargo fmt --all -- --check`
- `cargo clippy -p xmtp_mls --all-targets --all-features -- -D warnings`
- `cargo test -p xmtp_mls --lib groups::welcome_sync::tests`, 14 passed
- `RUST_LOG=off cargo test -p xmtp_mls --lib only_test_sync_welcomes -- --nocapture` against the local Docker backend, 1 passed
- `RUST_LOG=off cargo test -p xmtp_mls --lib -- --test-threads=1`, 705 passed, 6 unrelated timing or shared-backend failures, 4 ignored

The full suite failures were outside `welcome_sync.rs`. The changed welcome-sync tests and the service-backed welcome sync test pass.

> [!NOTE]
> ### Fix welcome cursor to stop processing on retryable failures in `WelcomeService`
> - Replaces the stream-based welcome processing in `sync_welcomes` with a new sequential `process_welcomes_with` helper in [welcome_sync.rs](https://github.com/xmtp/libxmtp/pull/3931/files#diff-5ad5d6ca4506c4960727c32b22b9168a128f14af4d61497cd168ba5c80fb43f5).
> - On a retryable error, processing halts immediately so later welcomes cannot advance the cursor past the failure point; non-retryable errors are skipped and processing continues.
> - Behavioral Change: `sync_welcomes` previously continued past all failures; it now stops on the first retryable failure, meaning welcomes after that point are excluded from the returned groups until the next sync.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7be9540.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/groups/welcome_sync.rs
+++ b/crates/xmtp_mls/src/groups/welcome_sync.rs
@@ -509,7 +509,6 @@ mod tests {
     use openmls::prelude::MlsMessageOut;
     use prost::Message;
     use rstest::*;
-    use std::sync::Mutex;
     use tls_codec::Serialize;
     use xmtp_common::Generate;
     use xmtp_configuration::Originators;
@@ -818,7 +817,7 @@ mod tests {
         );
 
         let (context, validator) = TestWelcomeSetup::builder()
-            .validator(NoopValidator)
+            .validator(SequenceValidator { fail: false })
             .context(context)
             .nested_transaction_calls(|db: &mut MockDbQuery| {
                 db.expect_find_group().returning(|_id| Ok(None));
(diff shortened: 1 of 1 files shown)


insipx (approved):

insipx:
looks reasonable to me

PR #3933, opened 2026-08-03
fix(xmtp_api_grpc): relax h2 keepalive defaults to stop bidi stream churn

tylerhawkes:
## Why

Production incident (P1 on the network `mls` DB, Jul 28–Aug 3): the ~6 bidi-enabled endpoints in production generated ~1,450 stream reconnects/hour, and every reconnect replayed a server-side catch-up wave against the database. The reconnect cadence matched libxmtp's h2 transport keepalive almost exactly: prod heralds cycled roughly once per keepalive interval, and the single bidi-enabled mobile device churned at ~10s intervals. Mechanism: with a 16s ping / 10s ack deadline, one straggling PING ack (cross-cloud RTT spike, radio wake latency) kills the entire HTTP/2 connection and every stream on it — the same failure previously root-caused as herald-lite #70. The env-var escape hatch added for #70 (#3715) was never wired into any deployment, so the aggressive defaults were live everywhere.

## What

Defaults move from **16s interval / 10s ack timeout / 16s TCP keepalive** to **45s / 20s / 45s** (still env-overridable per deployment via `XMTP_GRPC_KEEPALIVE_*`):

- ~65s worst-case dead-path detection (interval + timeout) instead of ~26s — deliberately traded for connections that survive transient stalls instead of tearing down healthy long-lived streams
- 45s cadence stays inside common 60s middlebox idle timers (a 60s cadence would race them)
- 20s ack window doubles tolerance for the stalls that actually killed connections
- `keep_alive_while_idle` unchanged (true) — idle bidi streams still get checked, just at the relaxed cadence

The doc comment states the honest contract: where the XIP-83 Subscribe ping/pong or the stream watchdog is enabled, the app layer detects dead streams itself; for default consumers (mobile FFI, node bindings with neither enabled) this transport keepalive remains the **only** dead-path detector — just a slower, less trigger-happy one. `apps/keepalive-probe` defaults and README move in lockstep so a bare probe run keeps mirroring a real client.

Historical release notes documenting the old defaults are intentionally untouched; the next release's notes must document the new values.

## Adversarial review

Reviewed adversarially before opening. Findings, all addressed: the original draft claimed app-level liveness owns detection (false for default consumers — both mechanisms are opt-in; comment rewritten and the detection regression shrunk from ~90s to ~65s worst-case), a 60s cadence sat exactly on common idle-timeout boundaries (now 45s), and the keepalive-probe tool still hardcoded the old defaults while claiming to mirror libxmtp (updated). The review could not refute: no code anywhere derives timing from the old 16/10/26 numbers (watchdog, bidi backoff, and request timeouts are independent; worst-case keepalive failure still lands under the 120s request timeout), the process-wide `LazyLock` config remains coherent with the TLS endpoint cache, and wasm has no transport keepalive to diverge.

## Verification

- `cargo test -p xmtp_api_grpc --lib` — 25 passed (keepalive default + override + parsing tests updated meaningfully)
- `cargo check -p keepalive-probe`, `cargo clippy -p xmtp_api_grpc` — clean

## Companions

- xmtp-node-go#565: catch-up wave scans rewritten (prune + rotation + LATERAL probes) so reconnects that do happen cost the DB almost nothing
- convos-assistants: herald env overrides set generously so the XIP-83 ping/pong owns liveness on the herald fleet

> [!NOTE]
> ### Relax HTTP/2 keepalive defaults in `xmtp_api_grpc` to reduce bidi stream churn
> - Increases `DEFAULT_INTERVAL` from 16s to 45s, `DEFAULT_TIMEOUT` from 10s to 20s, and `DEFAULT_TCP_KEEPALIVE` from 16s to 45s in [`native.rs`](https://github.com/xmtp/libxmtp/pull/3933/files#diff-d8ababd4174607b4463cfa01411d0034229c7d7bfccd1aecb5066b2386fb6d6a) to reduce unnecessary ping-pong on long-lived bidirectional streams.
> - Updates the `keepalive-probe` tool defaults and README to match the new libxmtp values (45s interval, 20s timeout).
> - Test files pin faster keepalive values via environment variables (`XMTP_GRPC_KEEPALIVE_INTERVAL_SECS`, `XMTP_GRPC_KEEPALIVE_TIMEOUT_SECS`) so existing reconnect and conversation tests remain fast and deterministic.
> - Behavioral Change: Any deployment relying on the old 16s/10s defaults will now use 45s/20s unless overridden by environment variables.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d2139cc.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/client.rs
+++ b/crates/xmtp_mls/src/client.rs
@@ -2068,12 +2068,21 @@ pub(crate) mod tests {
         assert_eq!(item[0].state, ConsentState::Allowed);
     }
 
-    #[xmtp_common::timeout(Duration::from_secs(50))]
+    #[xmtp_common::timeout(Duration::from_secs(100))]
     #[rstest::rstest]
     #[xmtp_common::test(unwrap_try = true)]
-    // Set to 50 seconds to safely account for the 16 second keepalive interval and 10 second timeout
+    // Detection of the black-holed connection comes from the h2 transport keepalive.
+    // Pin it fast (5s ping / 5s ack) so the failure lands in seconds under nextest,
+    // whose process-per-test isolation guarantees the pin is read before the
+    // process-wide config latches. Under a plain `cargo test`, a sibling test may
+    // latch the library defaults (45s/20s) first and the pin becomes a no-op, so
+    // the timeout budget also covers their ~65s worst-case detection.
     #[cfg_attr(any(target_arch = "wasm32"), ignore)]
     async fn should_reconnect() {
+        unsafe {
+            std::env::set_var("XMTP_GRPC_KEEPALIVE_INTERVAL_SECS", "5");
+            std::env::set_var("XMTP_GRPC_KEEPALIVE_TIMEOUT_SECS", "5");
+        }
         toxiproxy_test(async || {
             let alix = Tester::builder().proxy().build().await;
             let bo = Tester::builder().build().await;
--- a/apps/keepalive-probe/README.md
+++ b/apps/keepalive-probe/README.md
@@ -20,8 +20,8 @@ The defaults mirror libxmtp's real channel config in
 
 | flag | default | libxmtp source |
 |------|---------|----------------|
-| `--ka-interval` | `16s` | `http2_keep_alive_interval` |
-| `--ka-timeout` | `10s` | `keep_alive_timeout` ← the #70 knob |
+| `--ka-interval` | `45s` | `http2_keep_alive_interval` |
+| `--ka-timeout` | `20s` | `keep_alive_timeout` ← the #70 knob |
 | `--ka-while-idle` | `true` | `keep_alive_while_idle` |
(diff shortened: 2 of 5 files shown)


insipx (approved):

crates/xmtp_mls/src/client.rs
```diff
@@ -2068,12 +2068,21 @@ pub(crate) mod tests {
         assert_eq!(item[0].state, ConsentState::Allowed);
     }
 
-    #[xmtp_common::timeout(Duration::from_secs(50))]
+    #[xmtp_common::timeout(Duration::from_secs(100))]
```
tylerhawkes: Addressed the harness-dependence finding: the fast keepalive pin stays (seconds-fast under nextest, the repo standard), and the test timeout now also budgets for the library defaults' ~65s worst-case detection, so the test passes under a plain parallel `cargo test` where a sibling may have latched the process-wide config first.

PR #3946, opened 2026-08-07
fix(xmtp_mls): skip phantom members on partial key package failure (#3945)

### LINKED ISSUE #3945: add_members can write phantom members to group membership when key package fetch partially fails
### Describe the bug

When `add_members` / `UpdateGroupMembership` runs for multiple inbox IDs and key package fetch succeeds for some installations but fails for others, libxmtp can still record the failed inbox in the group membership payload (legacy GCE extension or migrated AppData `GROUP_MEMBERSHIP` dict) even though no MLS `Add` proposal was emitted for that inbox.

That leaves membership metadata and the MLS ratchet tree out of sync: the dict says the inbox is a member, but there is no corresponding leaf.

### Expected behavior

Only inbox IDs that received at least one verified key package and an `Add` proposal should be written into the membership update. Inboxes whose installations all failed KP fetch/verify should be omitted from the membership payload (while still being reported via `failed_installations` / error paths as today).

This should match the behavior already implemented on the `ProposeMemberUpdate` path in `mls_sync.rs`, which filters adds before building the AppData payload because an inbox without an MLS leaf must not appear in membership metadata.

### Steps to reproduce

1. Create a group and call `add_members` with two inbox IDs: **A** (valid published key packages) and **B** (no verifiable key packages).
2. If A succeeds but B's installations all fail, the commit may proceed (`new_installations` non-empty).
3. Membership extension / AppData dict can still include **B** with no corresponding `Add` proposal.

### Affected code

- `crates/xmtp_mls/src/groups/mls_sync/update_group_membership.rs` — `apply_update_group_membership_intent`
- `crates/xmtp_mls/src/groups/mls_sync.rs` — `get_membership_update_intent`

### Proposed fix

Filter net-new adds to inboxes that actually received a verified key package, mirroring the existing `ProposeMemberUpdate` path. Preserve sequence_id bumps for existing members.

crazywriter1:
Fixes #3945

## Summary

- Filter net-new inbox adds out of `UpdateGroupMembership` when no installation for that inbox received a verified key package.
- Apply the same filter in `get_membership_update_intent` so queued intents match what the commit will actually add.
- Reuse a shared `inbox_ids_from_new_key_packages` helper (also used by the existing `ProposeMemberUpdate` path).
- Add unit tests for the filter helpers.

## Problem

On partial `add_members` success (inbox A gets KPs, inbox B does not), the `UpdateGroupMembership` path could write B into the membership extension / AppData `GROUP_MEMBERSHIP` dict without an MLS `Add` proposal — dict and ratchet tree diverge.

## Test plan

- [x] `cargo test -p xmtp_mls strip_unverified filter_unverified --lib`
- [x] `cargo fmt --check` on changed files
- [ ] CI

> [!NOTE]
> ### Skip phantom members on partial key package failure in MLS group membership updates
> - Adds `strip_unverified_new_adds` in [update_group_membership.rs](https://github.com/xmtp/libxmtp/pull/3946/files#diff-56361bf0fcb1e9400991ea5996f3629427f28874ca78d913d3fd695be7c736f4) to remove net-new inboxes from `GroupMembership` when no corresponding key packages were provided, preventing divergence between the membership dict and the MLS tree state.
> - Adds `inbox_ids_from_new_key_packages` helper to centralize extraction of inbox IDs from key packages, emitting tracing warnings on credential decode or parse failures instead of silently skipping.
> - `apply_update_group_membership_intent` now calls `strip_unverified_new_adds` before computing the membership diff, and returns `None` (no commit/proposal) when the only remaining changes were unverified adds.
> - Behavioral Change: group sync may now produce no commit where it previously would have committed a membership update that included phantom members.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized be67792.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/groups/mls_sync/update_group_membership.rs
+++ b/crates/xmtp_mls/src/groups/mls_sync/update_group_membership.rs
@@ -193,15 +193,11 @@ pub(crate) async fn apply_update_group_membership_intent(
         return Ok(None);
     }
 
-    let failed_installations_changed =
-        new_group_membership.failed_installations != old_group_membership.failed_installations;
-
     if leaf_nodes_to_remove.is_empty()
         && changes_with_kps.new_key_packages.is_empty()
         && membership_diff.updated_inboxes.is_empty()
         && membership_diff.added_inboxes.is_empty()
         && membership_diff.removed_inboxes.is_empty()
-        && !failed_installations_changed
     {
         return Ok(None);
     }
--- a/crates/xmtp_mls/src/groups/mls_sync/update_group_membership.rs
+++ b/crates/xmtp_mls/src/groups/mls_sync/update_group_membership.rs
@@ -35,8 +35,24 @@ pub(crate) fn inbox_ids_from_new_key_packages(
     new_key_packages
         .iter()
         .filter_map(|kp| {
-            let credential = BasicCredential::try_from(kp.leaf_node().credential().clone()).ok()?;
-            parse_credential(credential.identity()).ok()
+            let credential = match BasicCredential::try_from(kp.leaf_node().credential().clone()) {
+                Ok(credential) => credential,
+                Err(e) => {
+                    tracing::warn!(?e, "failed to decode key package leaf credential");
+                    return None;
+                }
+            };
+            match parse_credential(credential.identity()) {
+                Ok(inbox_id) => Some(inbox_id),
+                Err(e) => {
+                    tracing::warn!(
+                        ?e,
+                        "failed to parse inbox id from key package credential; \
+                         skipping inbox for phantom-member verification"
+                    );
+                    None
+                }
+            }
         })
         .collect()
 }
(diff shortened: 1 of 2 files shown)


insipx:
great pr @crazywriter1 ! just some lints to resolve. thanks for the contribution

crates/xmtp_mls/src/groups/mls_sync/update_group_membership.rs
```diff
@@ -25,6 +28,56 @@ use xmtp_mls_common::{
...
 
+/// Inbox ids that received at least one Add proposal in this commit.
+pub(crate) fn inbox_ids_from_new_key_packages(
+    new_key_packages: &[KeyPackage],
+) -> std::collections::HashSet<String> {
+    new_key_packages
+        .iter()
+        .filter_map(|kp| {
+            let credential = BasicCredential::try_from(kp.leaf_node().credential().clone()).ok()?;
+            parse_credential(credential.identity()).ok()
```
insipx: this ok is potentially dangerous, and we should since it will drop the inbox id from `verified_adds` which strips it from `new_membership` resulting in a commit-creation failure. in practice it may never occur since credential should always parse an inbox id but it could be good to `trace::warn` on credential parse failure here.

insipx (approved):

PR #3940, opened 2026-08-05
fix(xmtp_common): stop Android device-sync abort by using webpki roots (the gRPC path already does this)

cameronvoell:
### Problem
  
  Device sync crash-loops on Android — the process aborts ~3s after every launch:

  ```
  Abort message: 'Expect rustls-platform-verifier to be initialized'
    reqwest::connect  ->  WorkerRunner::spawn (DeviceSync)
  ```

  reqwest 0.13 constructs `rustls_platform_verifier::Verifier` whenever no explicit roots are set. On Android that verifier calls into the JVM and
  needs the **host app** to initialize it over JNI with a `Context` *and* ship the crate's Kotlin component. libxmtp ships as a plain `.so` through
  uniffi, so neither happens. Only the archive HTTP path is affected; gRPC is fine.

  ### When this regressed

  `199e5034c` (#3306, 2026-03-10) bumped `reqwest 0.12.12 "rustls-tls"` → `0.13.2 "rustls-no-provider"`. 0.12 bundled the Mozilla root store; 0.13
  removed that option and made the platform verifier unconditional.

  | | status |
  |---|---|
  | libxmtp v1.10.0 / android-4.10.0 | clean |
  | libxmtp v1.11.0 / android-4.11.0 | **affected** |

  Same bump caused the iOS crash fixed in #3847. This is the Android half.

  ### Fix

  New `xmtp_common::http::client()` hands reqwest a rustls config using bundled webpki roots on Android, so the platform verifier is never constructed.
  Mirrors `xmtp_api_grpc`, which already forces `tls-webpki-roots` on Android/iOS, and restores pre-#3306 behavior. Applied to both archive call sites
  — `device_sync/worker.rs` (download, the crashing path) and `xmtp_archive/exporter.rs` (upload).

  **Only Android changes.** The branch is behind `#[cfg(target_os = "android")]`:

  ```rust
  // Android
  reqwest::Client::builder().use_preconfigured_tls(bundled_roots_tls_config())
  // iOS, node, wasm, desktop
  reqwest::Client::builder()
  ```
  
  The only delta elsewhere: `Client::new()`'s internal `.expect()` becomes an `Err` via `?`. No signature or FFI changes, and `Cargo.lock` gains no new
  packages — webpki-roots was already in the tree via tonic.

  ### Verification

  New `HistorySyncTlsTest` on a Pixel 10 Pro — `tests=1 failures=0 skipped=0`, 15.5s. Both TLS legs confirmed against
[...]

PR #3947, opened 2026-08-07
fix(xmtp_mls): treat membership sequence_id 0 as unset when resolving association state

insipx:
## Problem

`create_group` seeds the membership extension with the creator at `sequence_id: 0` as a placeholder (`build_starting_group_membership_extension(creator_inbox_id, 0)`). Nothing replaces it with the creator's real sequence id until the first `UpdateGroupMembership` intent runs.

`0` is a sentinel meaning "unset", and `get_installation_diff` honors it on the **old** membership side:

```rust
let starting_sequence_id = match old_group_membership.get(inbox_id) {
    Some(0) => None,
    Some(i) => Some(*i as i64),
    None => None,
};
```

The reads on the **new** membership side did not. The actor credential check in `ValidatedCommit::from_staged_commit` resolved the committer's association state at the dict's literal value, and since no identity update can have `sequence_id <= 0`, `get_association_state_with_verifier` returns `MissingIdentityUpdate`. That error is explicitly non-retryable, so the intent is moved to `Error` and never recovers.

Any commit authored between group creation and the first membership update hits this. Apps that create a group and immediately write metadata hit it every time.

## Evidence

Seen in production logs from three independent clients. The failing commit's membership dict is the only one out of 824 logged that carries a `0`:

```
Group created. {group_id: "bdc50949"}
MetadataUpdate intent routing
Commit published successfully {group_id: "bdc50949", intent_id: 20148, MetadataUpdate}
Group context extensions proposal found:
  GroupMembership { members: {"d52b1ff1…": 0}, failed_installations: [] }
ERROR Error validating commit for own message. Intent ID [20148]:
  InstallationDiff(Client(Association(MissingIdentityUpdate)))
```

122ms from group creation to a permanently errored intent.

## Fix

Apply the same `Some(0) => None` guard at the two new-membership read sites:

- `groups/validated_commit.rs` — the actor/proposer credential check
- `identity_updates.rs` — `ending_sequence_id` in `get_installation_diff`

Fixing it at the read sites (rather than resolving the real sequence id at creation time) also repairs groups already in the wild that carry a `0`.

### Note on the loosened check

Treating `Some(0)` as "use latest" means the actor check resolves against the newest known state instead of pinning to an agreed sequence id. The exposure is nil in practice: `0` only ever exists on a group whose sole member is its creator — it is replaced the moment any `UpdateGroupMembership` runs, and adding members necessarily bumps it — so no remote member ever validates a commit against a `0` dict.

## Tests

New `groups/tests/test_starting_membership_sequence_id.rs` covers both routes to the bad read:
- `metadata_update_on_freshly_created_group_succeeds` — GCE proposal carries `{creator: 0}`
- `key_update_on_freshly_created_group_succeeds` — no GCE proposal, so `get_latest_group_membership` falls back to the staged commit's group context, which still holds `{creator: 0}`

Both fail deterministically before the fix (reproducing the production error exactly) and pass after.

`test_update_policies_empty_group` had codified the buggy behavior via `expect_err("Should fail to update group name before first sync")`. That assertion is flipped to expect success.

## Test plan

- [x] `just test crate xmtp_mls` — 712/712 passed
- [x] `cargo fmt -p xmtp_mls --check` clean
- [ ] CI

> [!NOTE]
> ### Fix membership `sequence_id` 0 treated as unset when resolving association state
> - A freshly created group has `sequence_id: 0` in its membership, which was previously interpreted as a real sequence ID, causing non-retryable `MissingIdentityUpdate` failures on commits made before the first membership sync.
> - In [`validated_commit.rs`](https://github.com/xmtp/libxmtp/pull/3947/files#diff-233994c0ec18d23d0636e849849a3500e44611a4f64c6fffb1d0931aa1461ed6), credential verification now maps `sequence_id: 0` to `None`, falling back to the latest known association state instead of resolving at sequence 0.
> - In [`identity_updates.rs`](https://github.com/xmtp/libxmtp/pull/3947/files#diff-93f82a1bc3c5238249de128f10cdaca01b706f8b455fde435ca1184739600fde), `get_installation_diff` applies the same `Some(0) | None => None` mapping for the ending sequence ID.
> - Two regression tests in [`test_starting_membership_sequence_id.rs`](https://github.com/xmtp/libxmtp/pull/3947/files#diff-76a0f94c30fa6a91f00976e7331bbaac4bfc2e501a7e6b729aa1fd0478da80a9) confirm that metadata updates and key updates succeed on a freshly created, unsynced group.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7aa50c9.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/groups/tests/mod.rs
+++ b/crates/xmtp_mls/src/groups/tests/mod.rs
@@ -14,6 +14,7 @@ mod test_network;
 mod test_prepare_message_for_later_publish;
 mod test_proposals;
 mod test_send_message_opts;
+mod test_starting_membership_sequence_id;
 mod test_validate_app_data_update;
 mod test_welcome_pointers;
 mod test_welcomes;
@@ -2590,16 +2591,10 @@ async fn test_update_policies_empty_group() {
     let policy_set_2 = Some(PreconfiguredPolicies::AdminsOnly.to_policy_set());
     let amal_group_2 = amal.create_group(policy_set_2, None).unwrap();
 
-    // Verify empty group fails to update metadata before syncing
-    amal_group_2
-        .update_group_name("New Group Name 2".to_string())
-        .await
-        .expect_err("Should fail to update group name before first sync");
-
-    // Sync the group
-    amal_group_2.sync().await.unwrap();
-
-    //Verify we can now update the group name
+    // A solo group still carries the creation-time `sequence_id: 0` placeholder
+    // in its membership extension until something bumps it, and that must not
+    // block a metadata update. This previously failed with a non-retryable
+    // `MissingIdentityUpdate` — see `test_starting_membership_sequence_id`.
     amal_group_2
         .update_group_name("New Group Name 2".to_string())
         .await
--- a/crates/xmtp_mls/src/groups/tests/test_starting_membership_sequence_id.rs
+++ b/crates/xmtp_mls/src/groups/tests/test_starting_membership_sequence_id.rs
@@ -1,0 +1,45 @@
+//! Regression coverage for the `sequence_id: 0` placeholder that
+//! `build_starting_group_membership_extension` writes at group creation.
+//!
+//! `create_group` is purely local and seeds the membership extension with the
+//! creator at sequence_id `0`. Nothing replaces it with the creator's real
+//! sequence id until an `UpdateGroupMembership` intent runs, and neither the
+//! metadata-update nor the key-update path triggers one.
+//!
+//! `0` is a sentinel meaning "unset" — `get_installation_diff` honors it on the
+//! *old* membership side (`Some(0) => None`), but the reads on the *new* side
+//! treat it as a literal sequence id. The actor credential check in
+//! `ValidatedCommit::from_staged_commit` then resolves the committer's
+//! association state at sequence id `0`, and since no identity update can have
+//! `sequence_id <= 0`, it fails with `MissingIdentityUpdate` — which is not
+//! retryable, so the intent lands in `Error` permanently.
+
+use crate::tester;
+
+/// A metadata update issued before anything has bumped the creator's
+/// placeholder sequence id must still commit successfully.
+#[xmtp_common::test(unwrap_try = true)]
+async fn metadata_update_on_freshly_created_group_succeeds() {
+    tester!(alix);
+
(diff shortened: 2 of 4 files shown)


insipx:
### CI triage

Both required checks (`Test`, `Lint`) are green. Two jobs went red; neither is caused by this PR, and no code changes were needed.

**`cross-version (nightly)`** — now passing after re-run. Original failure was infra, not a test failure:

> The self-hosted runner lost communication with the server.

No job log was even uploaded (log blob 404s), consistent with the runner dying mid-step.

**`test-android / Integration Tests (Android)`** — still red, but pre-existing and repo-wide, not from this change. Two re-runs failed with two *different* environmental causes:

1. Attempt 1 — hit the 20-minute step timeout at **199/202 tests completed, 0 failed**. Progress stalls right after test #194, then trickles at 60–90s per test.
2. Attempt 2 — never reached the tests at all; the Docker backend failed to come up during `Start backend`: `container libxmtp-node-1 exited (1)`.

The attempt-1 signature reproduces identically on three unrelated branches — stall after #194, then timeout:

| run | branch | tail |
|---|---|---|
| [31123923861](https://github.com/xmtp/libxmtp/actions/runs/31123923861) | `fix/issue-3438` | 201/202, timeout |
| [31052878001](https://github.com/xmtp/libxmtp/actions/runs/31052878001) | `fix/node-surface-failed-installations-3763` | 201/202, timeout |
| [31209286315](https://github.com/xmtp/libxmtp/actions/runs/31209286315) | `fix/phantom-member-partial-add` | 199/202, timeout |

For contrast, [run 31047907702](https://github.com/xmtp/libxmtp/actions/runs/31047907702) (Aug 5) ran all 202 in **3m26s**, going 183 → 199 in ten seconds. I checked whether #3940 was the trigger — it is not; that Aug-5 passing run already contained it.

`test-android` is explicitly excluded from the required `Test` gate ([`test.yml#L253-254`](https://github.com/xmtp/libxmtp/blob/main/.github/workflows/test.yml#L253-L254)), so this does not block merge — but the tail-stall looks worth its own issue.

Only remaining blocker is the required approving review.

insipx:
Superseded by #3948, which is the same commit (`7aa50c92`) opened from a branch on `xmtp/libxmtp` instead of a fork. No code change.

CI was already green on both required gates here (`Test` and `Lint`); it re-runs on #3948.

PR #3954, opened 2026-08-07
feat(xmtp_mls): verify our own key package is live on the network

insipx:
## The problem

Key-package rotation is driven **purely by a local deadline column**
(`identity.next_key_package_rotation_ns`). `rotate_if_needed` reads
`is_identity_needs_rotation()`, and if the column says "not due" it returns
`Ok(false)` and does nothing. Nothing in the client ever verifies that a usable key
package is actually published on the network.

When that column is wrong, the failure is silent, total, and self-sustaining:

1. The client stops rotating. Its key package passes `not_after` and becomes unusable.
2. `fetch_key_packages` for that installation returns nothing.
3. Anyone adding it to a group records it in `failed_installations` instead of emitting
   an `Add` proposal. It is listed as a group member **with no MLS leaf**.
4. With no leaf it never receives a welcome, so the existing welcome-driven rotation
   nudge (`queue_key_rotation`) never fires either.
5. The client observes **no local error at all**. It syncs, sends messages, and creates
   groups against the groups where it still has valid leaves.

Closed loop, no exit, no logs.

A dev-network installation reached exactly this state and stayed unreachable for weeks.
Across seven client builds in a 31-hour window, with rotation roughly two months overdue,
the `"Start rotating keys and uploading the new key package"` INFO line never appeared
once. **Its absence was the only externally visible symptom.** A comparison client on the
same build and network rotated normally, so the rotation machinery itself was fine.

The systemic gap: local rotation state is never reconciled against network reality.

## What this adds

`KpLiveness` — a recurring TaskRunner singleton that probes the network for this
installation's own key package on a ~24h throttle, and queues a rotation when it is
absent, unverifiable, or close to expiry.

```text
KpLiveness dispatch
  ├─ throttled (checked within the interval)  -> reschedule, no network
  └─ probe fetch_key_packages([my_installation_id])
        ├─ valid, ours, comfortable lifetime  -> INFO, stamp, +24h
        ├─ absent | unverifiable | expiring   -> WARN, queue_key_rotation(), +24h
        └─ inconclusive (offline, 5xx, defect) -> WARN, do NOT stamp, +1h
```

It is seeded next to `KpRotation`/`KpDeletion`. `create_or_ignore_task` keeps an existing
row's deadline, but a *fresh* seed is due immediately — so **the first client build after
upgrade checks at once**, which is what rescues installations that are already broken.

Repair reuses `queue_key_rotation`, the same already-tested path the welcome nudge uses.
Liveness never rotates inline.

## Why a separate task, not a branch inside `KpRotation`

Folding the probe into the existing `KpRotation` arm was evaluated and rejected:

- **Schedule.** Liveness needs ~24h; `KpRotation`'s deadline follows the rotation column
  (~30 days). Sharing a row means multiplexing two schedules onto one column — and
  deriving the watchdog's cadence from the very column whose corruption it detects.
- **Retry budget.** A network probe fails routinely on offline clients. A shared row means
  a failing probe drives *rotation* into exponential backoff, and vice versa.
- **Nudges.** `PullInDeadline` targets a task by `data_hash`. With one row there is no way
  to say "verify now" as distinct from "rotate now".
- It would need the same DB migration anyway, so it avoids the proto work, not schema work.

A watchdog that shares a schedule and a retry budget with its subject fails with its subject.

## Details worth reviewing

Several of these came out of an adversarial review pass and are the non-obvious parts:

- **"Absent" has two wire shapes.** A short response is `MismatchedKeyPackages`, but the
  network actually returns a **present-but-empty payload** — exactly what
  `xdbg query fetch-key-packages` showed for the affected installation. Both mean absent.
  This is why the probe calls the API directly instead of `MlsStore`, which collapses an
  empty payload into a generic verification error.
- **The API matches responses to requests by position.** A backend/cache/ordering defect
  can return a perfectly valid key package belonging to *someone else*. It verifies, it has
  a healthy lifetime, and we would record ourselves as reachable while staying unaddable —
  rebuilding the exact loop this exists to break. The probe compares the leaf signature key.
  A mismatch is `Inconclusive`, **not** unhealthy: an upload cannot repair a mapping defect,
  and rotating would make one backend incident churn the whole fleet every interval.
- **Only "asked for 1, got 0" proves absence.** Any other count mismatch is a backend
  defect, classified `Inconclusive` for the same reason.
- **The throttle column is never trusted blindly.** A stamp in the *future* — clock skew at
  write time, or corruption — would suppress the watchdog for months. That is byte-for-byte
  the failure this check exists to catch, so the fix must not reproduce it.
- **Clamping the stamp is not enough.** The same clock jump also writes a future deadline
  onto the *task row*, and the dispatcher skips future-dated tasks — so the handler's clamp
  becomes unreachable because the handler never runs. Startup reconciliation therefore
  enqueues the liveness pull-in **unconditionally**.
- **Probe failures return a deadline, never `Err`**, so an offline client cannot drive this
  task into the TaskRunner's backoff. Genuine storage errors still propagate for the
  DB-reconnect contract.
- **Observability.** `rotate_if_needed` now logs at INFO on *both* branches. The silent
  "not due" branch is what hid the incident.

A commit-driven trigger (our own installation appearing in `failed_installations`) was
built and then **removed**: it wrote durable task rows from unvalidated remote input,
before `expected_diff_matches_commit` and the credential checks — and it is near-unreachable
anyway, since an installation in `failed_installations` has no leaf and cannot decrypt the
commit naming it.

Known limitations, including the pre-existing unclaimed-TaskRunner race and its real
destructive worst case, are documented in §5a of the design doc rather than glossed.

## Testing

`crates/xmtp_mls/src/worker/key_package_maintenance.rs`. Classification is extracted into
**pure functions** (`now_ns` is a parameter, not a hidden clock) and asserted directly over
every wire shape — no network, no override — so they cannot go false-green if the probe body
changes. A live-backend test additionally pins real backend behavior; that is the test that
found the empty-payload shape.

Covers: healthy → no rotation; absent → rotation queued **and** nudged; throttle blocks a
repeat probe; nudges coalesce to one pull-in; future stamp does not disable the check;
a clock-skew-stranded row is confirmed non-dispatchable then rescued by reconciliation;
every `next_liveness_deadline` boundary; every response shape including a valid key package
for a *different* installation.

- `just lint-rust` — clean
- `cargo nextest run -p xmtp_mls -p xmtp_db` — **936 passed** (`should_reconnect` is
  known-flaky and retried green)
- `just wasm check` — passes

## Dependency

Requires xmtp/proto#342, **already merged** as `f53e21af`.
`crates/xmtp_proto/proto_version` pins that rev and the generated Rust was regenerated
through `dev/gen_protos.sh` — not hand-edited. The regenerated diff covers only the new
`KpLiveness` message and its oneof arm.

## Design doc

`docs/superpowers/specs/2026-08-07-kp-liveness-check-design.md`

> [!NOTE]
> ### Add key package liveness check to verify own key package is live on the network
> - Introduces a new `KpLiveness` task that periodically probes the network to verify the installation's own key package is present, valid, and not near expiry.
> - Probes classify the network result into `Healthy`, `Absent`, `Unverifiable`, `ExpiringSoon`, or `Inconclusive`; actionable outcomes (`Absent`, `Unverifiable`, `ExpiringSoon`) queue a key package rotation.
> - Throttling is managed via a new `key_package_liveness_checked_at_ns` column in the identity table, with constants for check interval, retry interval, and minimum remaining lifetime.
> - The liveness task is seeded and reconciled at startup; `KpRotation` dispatches now also nudge the liveness task unconditionally.
> - Adds a new `KpLiveness` proto variant and DB migration for the liveness timestamp column.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8cf9617.</sup>
> 
>


insipx:
### CI fix pushed (`f33e167` → `8cf9617`)

`test-workspace` and `test-android` failed on the first push. Root cause was a **test-config value**, not the feature logic — but the failing test was right and I was wrong.

**What broke.** I had given `KEY_PACKAGE_LIVENESS_INTERVAL_NS` a 3-second override in `test/mls.rs` so throttle tests could sleep past it. That made *any* client built more than 3s after registration probe the network on startup. `bindings/mobile`'s `create_client_does_not_hit_network` asserts that building a client performs no network I/O, and it correctly caught this. The Android job showed the same cause from a different angle: it reached 200/202 with **0 failures** and then timed out, with the final tests crawling at ~90s each — extra probes across 202 client builds.

**The fix.** All three liveness constants now live in `common/mls.rs` with **no prod/test split**, so the interval is a realistic 1 day everywhere. Tests that need the check to run write the throttle column directly via `set_key_package_liveness_checked_at_ns` instead of sleeping. That is both correct and faster — it removed ~8s of sleeping from the suite.

The short interval was the bug. The assertion was load-bearing, and shortening a production interval purely to make a test convenient is what broke it.

**Verified locally, matching what CI runs:**

- `cargo nextest run --profile ci` (v3) — 2478/2478 pass
- `cargo nextest run --features d14n --profile ci-d14n -E 'package(xmtp_mls)' -E 'rdeps(xmtp_mls)'` — all pass
- `create_client_does_not_hit_network` under d14n — passes (previously failed 4/4)
- `cargo clippy --locked --all-features --all-targets -- -Dwarnings` and `cargo fmt --check` — clean

I had only run the v3 profile before the first push; CI runs v3 **and** d14n, which is how this reached CI at all. My mistake.

Also worth noting for reviewers: this incidentally resolves the review bot's suggestion about co-locating `KEY_PACKAGE_LIVENESS_MIN_REMAINING_LIFETIME_NS` — all three liveness constants are now together in `common/mls.rs`. Section 4 of the design doc records why there is deliberately no test override.

insipx:
### CI status on `8cf9617`: the real failure is fixed; the one remaining failure is a known repo-wide flake

**Fixed and green:** `test-workspace / Test (Rust Workspace)` now passes. That was the genuine regression from the 3s test-interval override, described in the previous comment. Everything else is green — WASM, Node, browser-sdk (all 4 shards), iOS, Android *unit* tests, cross-version, cross-talk, lint, cargo-deny.

**Remaining failure:** `test-android / Integration Tests (Android)` — this is [#3953](https://github.com/xmtp/libxmtp/issues/3953), not this PR.

Evidence it is pre-existing and unrelated:

- It fails on **`main`** in 4 of the last 5 failing `main` runs, including [`31223948308`](https://github.com/xmtp/libxmtp/actions/runs/31223948308) and [`31221165776`](https://github.com/xmtp/libxmtp/actions/runs/31221165776) from earlier today, both predating this branch.
- The signature matches #3953 exactly: progress is brisk to 194/202, then the last few tests crawl, the job reports **`0 failed`**, and the step times out. No test asserts anything wrong; the step budget is consumed elsewhere.
- I initially suspected my own change caused it, since the first run on this branch did add extra per-client network I/O. **That turned out to be wrong.** Removing those probes did not fix it — the second run failed identically (201/202 vs 200/202, i.e. noise). I have corrected my earlier comment on #3953 accordingly rather than leave a bad lead there.

One detail I contributed back to #3953 that may help whoever picks it up: the tail crawl has an **exact 90-second cadence** (six consecutive intervals at 90s to the second), which looks like a fixed timeout being hit rather than gradual resource saturation.

Not merging or force-passing anything — flagging it so the red check is not mistaken for a defect in this PR. Happy to rebase or re-run if you'd like a fresh roll of that job.

PR #3955, opened 2026-08-08
fix(xmtp_mls): stop one unresolvable member from jamming a group

### LINKED ISSUE #3952: Unresolvable inbox in GroupMembership permanently jams a group (MissingSequenceId retried forever, member cannot be removed)
### Describe the bug

A group can reach a state where its `GroupMembership` dict contains an inbox id that has **zero identity updates on that network**. Once that happens the group is permanently jammed: it can no longer add members, remove members, update installations, or send messages — and the bad member cannot be removed, because removal goes through the very code that fails.

Observed in production (placeholders used throughout):

```
WARN  Could not find existing sequence ID for inbox <INBOX_X>
WARN  Intent failed to sync and will be retried.
      {group_id: "<GROUP_G>", intent_id: <N>, intent_kind: UpdateGroupMembership, state: ToPublish}
```

`<INBOX_X>` has **0 identity updates** on the network the client is connected to. The same inbox id resolves fine on a *different* network. Inbox ids are deterministic from `address + nonce`, so the same id is computable on every network; only the identity log is network-scoped.

### Why the group jams

`MlsGroup::get_membership_update_intent` (`crates/xmtp_mls/src/groups/mls_sync.rs:4129`) builds its work list from **every** inbox already in the membership dict, plus any explicit adds:

```rust
let mut inbox_ids = existing_group_membership.inbox_ids();   // mls_sync.rs:4137
inbox_ids.extend_from_slice(inbox_ids_to_add);               // mls_sync.rs:4138
load_identity_updates(self.context.api(), &conn, &inbox_ids).await?;   // :4141
let latest_sequence_id_map = conn.get_latest_sequence_id(&inbox_ids)?; // :4143
```

`QueryIdentityUpdates::get_latest_sequence_id` (`crates/xmtp_db/src/encrypted_store/identity_update.rs:182-205`) simply **omits** inboxes with no rows, so an unresolvable inbox yields `None` and lands in the catch-all arm of the `try_fold`:

```rust
(_, _) => {
    tracing::warn!("Could not find existing sequence ID for inbox {}", inbox_id);
    return Err(GroupError::MissingSequenceId);   // mls_sync.rs:4166-4172
}
```

`try_fold` aborts the whole computation, so **one** unresolvable member fails the intent for the **entire** group. Everything that funnels through this function dies with it:

> ⚠️ **The warning text is ambiguous.** That catch-all arm covers two very different situations and logs the identical message for both:
> - `(None, None)` — a **net-new add** of an inbox with no identity log. No dict entry is created; `add_members` just fails.
> - `(None, Some(_))` — an inbox **already in the committed dict** with no identity log. This is the permanent jam.
>
> Because the log line does not say which, the production log alone cannot prove the inbox is in the dict. The message should name the case (and ideally the group id).

| Caller | file:line |
| --- | --- |
| `add_members` | `crates/xmtp_mls/src/groups/mod.rs:1910` |
| `remove_members` | `crates/xmtp_mls/src/groups/mod.rs:1985` |
| `add_missing_installations` | `crates/xmtp_mls/src/groups/mls_sync.rs:4100` |
| `maybe_update_installations` → `send_message` | `mls_sync.rs:4079`, `mod.rs:1448` |
| `maybe_update_installations` → `publish_messages` | `mod.rs:1497` |
| `maybe_update_installations` → `update_installations` | `mod.rs:1512` |
| `maybe_update_installations` → `MlsGroup::sync` | `mls_sync.rs:477`, `mls_sync.rs:481` |
| durable `AddMissingInstallations` task | `crates/xmtp_mls/src/worker/tasks.rs:506-536` |

**There is no recovery path.** `remove_members` (`mod.rs:1983-1993`) calls `get_membership_update_intent(&[], inbox_ids)`, and the fold at `mls_sync.rs:4146` iterates the *existing* membership without excluding `inbox_ids_to_remove` (there is even a `TODO:nm` acknowledging this at `mls_sync.rs:4136`). So the removal of the bad member fails on the bad member. Even if that were fixed, `IdentityUpdates::get_installation_diff` (`crates/xmtp_mls/src/identity_updates.rs:554-566`) resolves an association state for every `removed_inboxes` entry and would fail the same way.

Note that the bad entry is never re-validated once it is in: `get_installation_diff` (`identity_updates.rs:497-552`) only resolves inboxes in `added_inboxes` / `updated_inboxes` / `removed_inboxes`. An entry whose `sequence_id` never changes is silently carried through every subsequent commit, so the group state stays "valid" while being permanently unusable.

### Error classification is wrong for this case

`GroupError::MissingSequenceId` is classified retryable (`crates/xmtp_mls/src/groups/error.rs:616-622`), justified by a comment about identity registration "not having propagated to reads yet", with a regression test locking it in (`error.rs:668-680`).

That reasoning holds only for the **net-new add** case. For an inbox that is *already committed into the group* and has *no identity log at all* on this network, the condition is permanent — a fresh `load_identity_updates` at `mls_sync.rs:4141` has already re-read from the network and come back empty. Retrying re-runs the same failing read forever.

The two cases are distinguishable in the existing match:

- `(None, None)` — inbox is a **net-new add**; registration may genuinely not have propagated yet. Transient is a defensible default (bounded).
- `(None, Some(_))` — inbox is **already in the committed membership dict**, so it must have been resolvable when the commit was accepted. If a fresh network read now returns nothing, this is permanent, not lag.

### How does an unresolvable inbox get into the dict? (not fully settled)

I traced every path that writes a `GroupMembership` into local state. Under current `main` they all appear to enforce resolvability:

- **Local add path** — `get_membership_update_intent` rejects `(None, None)` (`mls_sync.rs:4166`), so `add_members` cannot introduce one.
- **Welcome** — `InitialMembershipValidator::check_initial_membership` (`crates/xmtp_mls/src/groups/welcomes/validated_membership.rs:47-54`) resolves an association state for **every** member in the dict; an unresolvable one surfaces `AssociationError::MissingIdentityUpdate` from `get_association_state_with_verifier` (`crates/xmtp_mls/src/identity_updates.rs:92-101`) and the welcome is rejected.
- **Remote commit** — `ExtractExpectedDiff` → `get_installation_diff` (`crates/xmtp_mls/src/groups/validated_commit.rs:1171-1180` → `identity_updates.rs:536-543`) resolves every added/updated inbox, so a commit adding an unresolvable inbox is rejected.
- **Bootstrap / AppData migration commit** — membership-neutral and inbox-set-equality checked (`crates/xmtp_mls/src/groups/app_data/bootstrap_validator.rs:669+`), so it cannot introduce a new inbox.

So the entry requires either (a) a client running an older/different build, or (b) a client whose **local** `identity_updates` table could resolve the inbox even though the network it was publishing to could not.

Worth noting how easy it is for a caller to *produce* a network-unverified inbox id in the first place: every binding exports a purely deterministic, network-independent derivation — `generate_inbox_id(identifier, nonce)` at `bindings/mobile/src/identity.rs:38`, `bindings/node/src/inbox_id.rs:35`, `bindings/wasm/src/inbox_id.rs:31`, re-exported as `generateInboxId` in the JS SDKs (`sdks/js/browser-sdk/src/utils/inboxId.ts:15`) — right next to `add_members(inbox_ids)` (`bindings/mobile/src/mls.rs:2909`, `bindings/wasm/src/conversation.rs:662`), which takes raw inbox ids with no registration check. An app backend that derives an inbox id from a wallet address (or copies one across environments) and passes it to `addMembers` gets exactly the `(None, None)` warning above. Today libxmtp does reject that — but with a retryable, unnamed `MissingSequenceId` that is indistinguishable from the jam case. The network-checked alternative (`get_inbox_id_for_identifier`, `bindings/mobile/src/mls.rs:521`) exists but nothing steers callers to it.

For completeness, the API layer itself does not fabricate ids: `get_inbox_ids` in `crates/xmtp_api_d14n/src/queries/d14n/identity.rs:110-146` is a real server query, not a client-side derivation.

(b) is structurally possible today. The `identity_updates` table is keyed only by `(inbox_id, sequence_id)` (`crates/xmtp_db/src/encrypted_store/schema_gen.rs:135-142`) — **there is no network/environment column**, and `get_membership_update_intent` trusts the local DB (`mls_sync.rs:4143`). Rows only enter via `load_identity_updates` (`identity_updates.rs:594-632`), so any client that ever pointed the *same database file* at a different network keeps those rows and will happily resolve a foreign-network inbox id afterwards. The mobile SDKs namespace the DB by env (e.g. `sdks/ios/Sources/XMTPiOS/Client.swift:494`, `xmtp-<env>-<inboxId>.db3`), but `bindings/node/src/client/create_client.rs:119` takes a raw caller-supplied `db_path` with no env scoping — a backend/bot reusing one path across envs is not protected.

If such a client authored the add, the partial-failure guard at `mls_sync.rs:4197-4202` would not stop it: it only errors when `inbox_ids_to_add` is non-empty **and** `new_installations` is empty, so a mixed add (one good member + one cross-network member) passes. The resulting commit is then **non-retryably rejected by every other member** — `CommitValidationError::InstallationDiff` → `ClientError::Association(MissingIdentityUpdate)` falls through `ClientError::is_retryable`'s `_ => false` (`crates/xmtp_mls/src/client.rs:214-231`), `CommitValidationError::is_retryable` (`validated_commit.rs:137-144`), and `mls_sync.rs:1240-1252` then advances the cursor and skips the commit. That forks the group in addition to jamming it.

**What I could not settle from code alone**, and the evidence that would settle it:

0. **Which arm actually fired** — `(None, None)` (a failing `addMembers` call, no dict entry) or `(None, Some(_))` (a real dict entry). Adding the case to the log message, or dumping the dict, settles this first. Everything below assumes the latter.
1. The group's actual membership dict — does it contain `<INBOX_X>`, and at what `sequence_id`? If the stored `sequence_id` matches the value the inbox has on the *other* network, cross-network local-DB resolution is proven.
2. The commit that introduced the entry: author inbox/installation, epoch, timestamp, and the libxmtp version that authored it. (`xdbg query batch-query-commit-log` on the group, or a `ValidatedCommit`/transcript dump from any member.)
3. Whether other members of the group are also stuck, or whether they rejected that commit and forked (distinguishes "everyone accepted it" from "only the author accepted it").
4. Which SDK/binding and which `db_path` scheme the authoring client used.

### Proposed fix

**1. Split the two cases in `get_membership_update_intent` (`mls_sync.rs:4146-4176`) instead of failing the whole fold.**

- `(None, None)` where the inbox **is** in `inbox_ids_to_add` → return a distinct, **non-retryable** error naming the inbox (e.g. `GroupError::InboxNotRegistered { inbox_id }`) so `add_members` fails fast and the host app can say "this user is not registered on this network" rather than retrying forever.
- `(None, Some(_))` where the inbox is an **existing** dict member → do **not** abort. `warn!` once, skip that inbox, and keep building the intent from the members that do resolve. This alone unjams every group already in this state and restores `send_message`.

**2. Make removal possible.** Exclude `inbox_ids_to_remove` from `inbox_ids` before the fold (the existing `TODO:nm` at `mls_sync.rs:4136`), and make `get_installation_diff` tolerate an unresolvable inbox in `removed_inboxes` (`identity_updates.rs:554-566`) — such an inbox has no leaf in the tree, so "no installations to remove" is the correct answer, not an error.

**3. Validate at the add boundary, not at the sync boundary.** The check belongs in `get_membership_update_intent` at the point where a net-new inbox is first admitted (`mls_sync.rs:4162-4165`, the `(Some(latest), None)` arm) — i.e. reject an inbox whose association state resolves to zero installations, and reject when *all* of a net-new inbox's key packages failed. Today `mls_sync.rs:4197-4202` only guards the all-adds-failed case; a mixed batch slips through. (Overlaps with #3945, which covers the phantom-member half of this.)

**4. Reclassify.** Keep a retryable variant for the genuine propagation race on net-new adds, but bound it; introduce a distinct non-retryable variant for "existing dict member with no identity log on this network", and update the regression test at `error.rs:668-680` to assert both.

**5. Consider making local identity-update rows network-scoped**, or at minimum scope the DB path by env in the non-mobile bindings (`bindings/node/src/client/create_client.rs:119`), so a client can never resolve an inbox that the network it is publishing to cannot.

### Affected code

- `crates/xmtp_mls/src/groups/mls_sync.rs:4129-4214` — `get_membership_update_intent`
- `crates/xmtp_mls/src/groups/error.rs:96`, `:616-622`, `:668-680` — `MissingSequenceId` + retryability
- `crates/xmtp_mls/src/identity_updates.rs:489-566` — `get_installation_diff`
- `crates/xmtp_mls/src/groups/mod.rs:1899-1940`, `:1983-1993` — `add_members` / `remove_members`
- `crates/xmtp_db/src/encrypted_store/identity_update.rs:182-205` — `get_latest_sequence_id`

### Related

- #3945 — `add_members` can write phantom members to group membership when key package fetch partially fails. Same failure surface (dict entry with no MLS leaf), different trigger (KP fetch failure vs. no identity log at all).

insipx:
Fixes #3952.

## The problem

A group's `GroupMembership` dictionary can end up holding an inbox id that has
zero identity updates on the network the client is connected to. Once that
happens the group is permanently unusable, and the entry cannot be removed.

`MlsGroup::get_membership_update_intent` builds its work list from **every**
inbox in the dictionary and then runs a `try_fold`. `get_latest_sequence_id`
omits inboxes with no rows, so the unresolvable inbox lands in the catch-all
arm, which returns `GroupError::MissingSequenceId` and aborts the whole
computation. One member kills the intent for the entire group.

### Blast radius

Everything membership-related funnels through that one function, so the group
cannot send messages either:

| Caller | Path |
| --- | --- |
| `add_members` | `groups/mod.rs` |
| `remove_members` | `groups/mod.rs` |
| `add_missing_installations` | `groups/mls_sync.rs` |
| `send_message` | via `maybe_update_installations` |
| `publish_messages` | via `maybe_update_installations` |
| `update_installations` | via `maybe_update_installations` |
| `MlsGroup::sync` | via `maybe_update_installations` |
| durable `AddMissingInstallations` task | `worker/tasks.rs` |

`MissingSequenceId` is classified retryable, so the intent retries the same
failing read forever.

### No recovery path

`remove_members` calls `get_membership_update_intent(&[], inbox_ids)` and the
fold walked the existing membership **without** excluding the inboxes being
removed — there was a `TODO:nm` acknowledging exactly this. Even past that,
`get_installation_diff` resolved an association state for every removed inbox
and failed the same way. The bad member was unremovable by construction.

## The fix

The user-facing goal is recovery: a group already in this state must become
usable again, and the bad entry must be removable.

**1. `get_membership_update_intent` — skip instead of abort.**
The old catch-all arm covered two different situations and logged the same
message for both. It is now split:

- `(None, Some(_))` — the group already committed this inbox and the network
  has no identity updates for it. Warn, skip the member, and **leave its
  dictionary entry untouched**. Dropping the entry would itself be a membership
  change that other members do not expect, so the entry stays and the rest of
  the membership still updates. This alone restores `send_message` and every
  other caller in the table above.
- `(None, None)` — a net-new add of an inbox with no identity log. Still an
  error. Registration may genuinely not be visible to reads yet.

Members named in `inbox_ids_to_remove` are now excluded from the fold (the
`TODO:nm`). They stay in the `load_identity_updates` refresh on purpose:
`get_installation_diff` still has to resolve them to know which leaves the
commit drops, and dropping the refresh would make a stale local log look
identical to an absent one.

**2. `get_installation_diff` — a removed inbox with no identity log removes no
installations.**
Such an inbox has no leaf in the tree, so "nothing to remove" is the correct
answer, not an error. The check is deliberately narrow: it only skips when the
local log is **empty** after a refresh. A short-but-present log still fails,
because `get_association_state` needs the exact sequence id and a partial log
means "stale", not "absent". Removed inboxes were added to the refresh filter
so the emptiness test is authoritative.

**3. `expected_diff_matches_commit` — expected removals are restricted to
installations that are in the tree.**
A commit can only remove a leaf that exists. Without this, a member that *can*
resolve the entry (the one plausible way it got written in the first place)
computes a non-empty expected removal set, sees a commit that removed no
leaves, and rejects it — trading the jam for a fork. This tightens the
comparison onto the domain the commit can actually act on; it does not let a
commit skip removing a leaf that is present.

A fourth change was needed on the original base: the "nothing to do" early
return in `apply_update_group_membership_intent` ignored `removed_inboxes`, so
a dictionary-only removal produced no commit at all. #3946 landed the same
condition (plus `added_inboxes`) while this branch was open, so after the
rebase only a comment remains there.

### Composition with #3949

#3949 unions the publish-time key-package failures into
`new_group_membership.failed_installations`, which `expected_diff_matches_commit`
reads as `failed_installation_ids`. That predicate and the new
`existing_installation_ids` predicate are ANDed, so both only ever shrink
`filtered_expected` and the result is order-independent.

The two are in fact disjoint by construction. A `failed_installations` entry is
an installation id, and you can only derive an installation id for an inbox
whose association state resolves — so an inbox with **no** identity log, the
case change 2 skips, can never contribute one. `test_remove_inbox_with_bad_installation_from_group`
(the test #3949's comment calls out as sensitive to this field) and both of
#3949's new tests pass alongside the five added here.

## Retryability

Defect 3 in the issue — `MissingSequenceId` retried forever for a permanent
condition — is resolved by change 1 rather than by a new error variant. After
the split, the permanent case produces no error at all, and `MissingSequenceId`
is reachable **only** from a net-new add of an unregistered inbox. That case is
plausibly transient, so retryable stays correct and the existing regression
lock in `groups/error.rs` stays valid. The two other emitters of
`MissingSequenceId` (`CommitPendingProposals`, `ProposeMemberUpdate`) are both
add-only paths, so the narrowed contract holds crate-wide. Adding a
non-retryable variant would have left it dead. The variant docs, the
`is_retryable` comment and the regression test comment were updated to state
the narrowed contract, and the two warnings now name which case fired and the
group id — the ambiguity the issue called out.

## Commit validation

The skip is chosen so that the unresolvable entry never appears in the
membership diff: it is not added, not updated, and not removed. It therefore
never reaches `get_installation_diff`, and `validate_membership_diff` (which
only inspects `updated_inboxes`) is unaffected. Every member computes the same
diff from the same dictionary, so the commit is accepted by all of them.

## Tests

New module `crates/xmtp_mls/src/groups/tests/test_unresolvable_member.rs`. Each
test builds the jammed state by writing a dictionary entry with no leaf via a
local-only group-context-extensions commit — the commit never reaches the
network, so it reproduces a group that some other client already jammed.

- a group with one unresolvable member can still send a message, and the entry
  is still there afterwards
- a group with one unresolvable member can still add and remove other members
- the unresolvable member can be removed, and the group is fully usable
  afterwards: a new member joins and exchanges a message
- a dictionary entry with no leaf whose inbox **is** resolvable can also be
  removed (covers change 3)
- `MissingSequenceId` is still raised, and still retryable, for a net-new add
  of an unregistered inbox — and is not raised for an already-committed member

All five fail on `main` and pass with this change, re-verified after every
rebase (#3946, #3948, #3949). Each change was also reverted individually
against the new tests to confirm every one of them is load-bearing.

### Known boundary

A group still in this state cannot onboard a **new** installation, because
`InitialMembershipValidator::check_initial_membership` resolves every inbox in
the dictionary before accepting a welcome. That is deliberately left alone
here: relaxing welcome validation has its own security surface, and it is not
needed for recovery, since removing the bad entry restores joining as well.
The third test above proves that ordering end to end.

## Verification

- `just test crate xmtp_mls` — 722 passed, 0 failed (1 flaky)
- `just test d14n` (`--features d14n --profile ci-d14n`) — 803 passed, 0 failed
  (27 flaky; high flake count comes from a shared local backend)
- `just lint-rust` — clean, re-run after every rebase

> [!NOTE]
> ### Fix unresolvable group members blocking MLS group commits
> - In `get_membership_update_intent`, members already in the group but lacking identity updates are now skipped (with a warning) instead of returning `MissingSequenceId`, which previously jammed the entire group.
> - `MissingSequenceId` is now only raised for attempted additions of inboxes with no identity updates, where it remains retryable.
> - Commit validation in `expected_diff_matches_commit` no longer rejects commits that attempt to remove installations without a leaf in the current tree.
> - Installation diff calculation tolerates removed inboxes with no identity updates by skipping their installation removals rather than erroring.
> - A new test suite in [test_unresolvable_member.rs](https://github.com/xmtp/libxmtp/pull/3955/files#diff-1c0e61a8ce5b992846428be96b3ddcef992bf0c9505026bd509f486635d0f012) verifies that unresolvable members do not block sends, adds, removes, or other group operations.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f771d35.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/groups/error.rs
+++ b/crates/xmtp_mls/src/groups/error.rs
@@ -101,8 +101,10 @@ pub enum GroupError {
     UserLimitExceeded,
     /// Sequence ID not found.
     ///
-    /// No sequence ID for an inbox after an identity-update refresh —
-    /// its registration hasn't propagated yet. Retryable.
+    /// An inbox that the caller wants to add has no sequence id after an
+    /// identity-update refresh. Its registration is not visible yet. Retryable.
+    /// A member that the group already committed never raises this error. The
+    /// membership builders skip that member and keep its dictionary entry.
     #[error("SequenceId not found in local db")]
     MissingSequenceId,
     /// Addresses not found.
@@ -624,12 +626,10 @@ impl RetryableError for GroupError {
             Self::DeleteMessage(e) => e.is_retryable(),
             Self::DeviceSync(e) => e.is_retryable(),
             Self::MergePendingCommit(e) => e.is_retryable(),
-            // Only emitted when a fresh `load_identity_updates` network
-            // refresh still has no sequence id for an inbox — i.e. its
-            // registration hasn't propagated to reads yet. Retrying re-runs
-            // the fetch, so the miss is transient; classifying it
-            // non-retryable permanently failed sync-group membership adds
-            // that raced a new installation's identity propagation.
+            // Only an inbox that the caller wants to add raises this error. Its
+            // registration is not visible to reads yet, so a retry can succeed. An
+            // already committed member cannot reach here: the membership builders
+            // skip it. See `get_membership_update_intent`.
             Self::MissingSequenceId => true,
             Self::NotFound(_)
             | Self::UserLimitExceeded
@@ -682,10 +682,10 @@ mod tests {
 
     #[xmtp_common::test]
     fn missing_sequence_id_is_retryable() {
-        // Regression lock: while non-retryable, a membership add racing a
-        // new installation's identity propagation burned all publish
-        // attempts instantly and permanently failed the sync-group add
-        // (surfaced as the DeviceSync sendSyncRequest CI failures).
+        // Regression lock: while non-retryable, a membership add that raced a new
+        // installation's identity propagation burned all publish attempts at once.
+        // The add then failed for good. Only unregistered adds reach this error, so
+        // a retry is correct. See `get_membership_update_intent`.
         assert!(GroupError::MissingSequenceId.is_retryable());
     }
 
--- a/crates/xmtp_mls/src/groups/mls_sync.rs
+++ b/crates/xmtp_mls/src/groups/mls_sync.rs
@@ -4128,47 +4128,62 @@ where
     ) -> Result<UpdateGroupMembershipIntentData, GroupError> {
         self.load_mls_group_with_lock_async(async |mls_group| {
             let existing_group_membership = extract_group_membership(mls_group.extensions())?;
-            // TODO:nm prevent querying for updates on members who are being removed
             let mut inbox_ids = existing_group_membership.inbox_ids();
             inbox_ids.extend_from_slice(inbox_ids_to_add);
             let conn = self.context.db();
-            // Load any missing updates from the network
+            // Members that leave keep their refresh. `get_installation_diff` must
+            // resolve them to know which leaves the commit removes.
             load_identity_updates(self.context.api(), &conn, &inbox_ids).await?;
 
             let latest_sequence_id_map = conn.get_latest_sequence_id(&inbox_ids as &[&str])?;
 
             // Get a list of all inbox IDs that have increased sequence_id for the group
-            let changed_inbox_ids =
-                inbox_ids
-                    .iter()
-                    .try_fold(HashMap::new(), |mut updates, inbox_id| {
-                        match (
-                            latest_sequence_id_map.get(inbox_id as &str),
-                            existing_group_membership.get(inbox_id),
-                        ) {
-                            // This is an update. We have a new sequence ID and an existing one
-                            (Some(latest_sequence_id), Some(current_sequence_id)) => {
-                                let latest_sequence_id_u64 = *latest_sequence_id as u64;
-                                if latest_sequence_id_u64.gt(current_sequence_id) {
-                                    updates.insert(inbox_id.to_string(), latest_sequence_id_u64);
-                                }
-                            }
-                            // This is for new additions to the group
-                            (Some(latest_sequence_id), None) => {
-                                // This is the case for net new members to the group
-                                updates.insert(inbox_id.to_string(), *latest_sequence_id as u64);
-                            }
-                            (_, _) => {
-                                tracing::warn!(
-                                    "Could not find existing sequence ID for inbox {}",
-                                    inbox_id
-                                );
-                                return Err(GroupError::MissingSequenceId);
+            let changed_inbox_ids = inbox_ids
+                .iter()
+                .filter(|inbox_id| !inbox_ids_to_remove.contains(*inbox_id))
+                .try_fold(HashMap::new(), |mut updates, inbox_id| {
+                    match (
+                        latest_sequence_id_map.get(inbox_id as &str),
+                        existing_group_membership.get(inbox_id),
+                    ) {
+                        // This is an update. We have a new sequence ID and an existing one
+                        (Some(latest_sequence_id), Some(current_sequence_id)) => {
+                            let latest_sequence_id_u64 = *latest_sequence_id as u64;
+                            if latest_sequence_id_u64.gt(current_sequence_id) {
+                                updates.insert(inbox_id.to_string(), latest_sequence_id_u64);
                             }
                         }
+                        // This is for new additions to the group
+                        (Some(latest_sequence_id), None) => {
+                            // This is the case for net new members to the group
+                            updates.insert(inbox_id.to_string(), *latest_sequence_id as u64);
+                        }
+                        // The group already committed this inbox. The network has no
+                        // identity updates for it. Skip the member. Keep its dictionary
+                        // entry. If we stop here, one member blocks every commit in the
+                        // group. If we drop the entry, the commit makes a membership
+                        // change that the other members reject.
+                        (None, Some(_)) => {
+                            tracing::warn!(
+                                group_id = %self.group_id,
+                                %inbox_id,
+                                "Group member has no identity updates. Skipped the member and kept its entry."
+                            );
+                        }
+                        // The caller wants to add an inbox that has no identity updates.
+                        // The registration may not be visible yet. This error retries.
+                        (None, None) => {
+                            tracing::warn!(
+                                group_id = %self.group_id,
+                                %inbox_id,
+                                "Cannot add an inbox that has no identity updates."
+                            );
+                            return Err(GroupError::MissingSequenceId);
+                        }
+                    }
 
-                        Ok(updates)
-                    })?;
+                    Ok(updates)
+                })?;
             let extensions = mls_group.extensions().clone();
             let old_group_membership = extract_group_membership(&extensions)?;
             let mut new_membership = old_group_membership.clone();
--- a/crates/xmtp_mls/src/groups/mls_sync/update_group_membership.rs
+++ b/crates/xmtp_mls/src/groups/mls_sync/update_group_membership.rs
@@ -196,6 +196,8 @@ pub(crate) async fn apply_update_group_membership_intent(
 
     // Run this guard before the writeback below. A change that only moves
     // `failed_installations` must not make an empty commit.
+    // A dictionary entry can exist with no leaf in the tree. Adding or removing
+    // such an entry moves no leaf. The commit must still carry the change.
     if leaf_nodes_to_remove.is_empty()
         && changes_with_kps.new_key_packages.is_empty()
         && membership_diff.updated_inboxes.is_empty()
--- a/crates/xmtp_mls/src/groups/tests/mod.rs
+++ b/crates/xmtp_mls/src/groups/tests/mod.rs
@@ -17,6 +17,7 @@ mod test_prepare_message_for_later_publish;
 mod test_proposals;
 mod test_send_message_opts;
 mod test_starting_membership_sequence_id;
+mod test_unresolvable_member;
 mod test_validate_app_data_update;
 mod test_welcome_pointers;
 mod test_welcomes;
--- a/crates/xmtp_mls/src/groups/tests/test_unresolvable_member.rs
+++ b/crates/xmtp_mls/src/groups/tests/test_unresolvable_member.rs
@@ -1,0 +1,152 @@
+//! An inbox in the group membership dictionary can have no identity updates on
+//! this network. The group must stay usable, and the member must be removable.
+//! See xmtp/libxmtp#3952.
+//!
+//! A new installation still cannot join such a group. `check_initial_membership`
+//! resolves every inbox in the dictionary before it accepts a welcome. Remove the
+//! entry first. `test_unresolvable_member_can_be_removed` proves that order.
+
+use crate::context::XmtpSharedContext;
+use crate::groups::group_membership::GroupMembership;
+use crate::groups::send_message_opts::SendMessageOpts;
+use crate::groups::validated_commit::extract_group_membership;
+use crate::groups::{GroupError, build_group_membership_extension};
+use crate::identity_updates::load_identity_updates;
+use crate::tester;
+use crate::utils::TestMlsGroup;
+use crate::utils::test::FullXmtpClient;
+use xmtp_common::RetryableError;
+use xmtp_db::prelude::*;
+
+/// Add an inbox to the group membership dictionary with no leaf in the tree.
+/// The commit stays local, so no other member has to accept it. This copies a
+/// group that a different client already jammed.
+fn add_dict_entry(client: &FullXmtpClient, group: &TestMlsGroup, inbox_id: &str, sequence_id: u64) {
+    let provider = client.context.mls_provider();
+
+    group
+        .load_mls_group_with_lock(client.context.mls_storage(), |mut mls_group| {
+            let mut extensions = mls_group.extensions().clone();
+            let mut membership = extract_group_membership(&extensions)?;
+            membership.add(inbox_id.to_string(), sequence_id);
+            extensions.add_or_replace(build_group_membership_extension(&membership))?;
+
+            mls_group
+                .update_group_context_extensions(
+                    &provider,
+                    extensions,
+                    &client.identity().installation_keys,
+                )
+                .unwrap();
+            mls_group.merge_pending_commit(&provider).unwrap();
+            Ok(())
+        })
+        .unwrap();
+}
+
+/// Add an inbox that has no identity updates on this network.
+fn jam_group(client: &FullXmtpClient, group: &TestMlsGroup) -> String {
+    let unresolvable = hex::encode(xmtp_common::rand_array::<32>());
+    add_dict_entry(client, group, &unresolvable, 1);
+    unresolvable
+}
+
+fn membership(client: &FullXmtpClient, group: &TestMlsGroup) -> GroupMembership {
+    group
+        .load_mls_group_with_lock(client.context.mls_storage(), |mls_group| {
+            Ok(extract_group_membership(mls_group.extensions())?)
+        })
+        .unwrap()
+}
+
+#[xmtp_common::test(unwrap_try = true)]
+async fn test_unresolvable_member_does_not_block_messages() {
+    tester!(alix);
+    let group = alix.create_group(None, None)?;
+    let unresolvable = jam_group(&alix, &group);
+
+    // This refresh reads every member of the dictionary. The jam stops it.
+    group.update_installations().await?;
+    group
+        .send_message(b"hello", SendMessageOpts::default())
+        .await?;
+
+    // The skip keeps the entry. A drop is itself a membership change.
+    assert!(membership(&alix, &group).get(&unresolvable).is_some());
+}
+
+#[xmtp_common::test(unwrap_try = true)]
+async fn test_unresolvable_member_does_not_block_other_members() {
+    tester!(alix);
+    tester!(bo);
+    tester!(caro);
+    let group = alix.create_group(None, None)?;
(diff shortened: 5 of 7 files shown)


insipx:
Thanks — went through the four suggestions. Three applied in `c03fc36`, one declined with reasoning.

**1. Log message clarity — applied.** The warning now reads `"Group member has no identity updates. Skipped the member and kept its entry."` The "kept its entry" half is the part an operator cannot infer, so it belongs in the message. I left the inbox id as a structured field rather than interpolating it into the text, so it stays greppable and does not duplicate.

**2. Why the empty-vs-short distinction matters — applied.** The comment now states the semantics, not just the mechanism:

```
// forever. An empty log after the refresh above means the inbox has no
// identity here. A short log means stale local data, so it still fails:
// `get_association_state` needs the exact sequence id.
```

**3. Filter order in `expected_diff_matches_commit` — declined.** Both `failed_installation_ids` and `existing_installation_ids` are `HashSet<Vec<u8>>`, so each `contains` is O(1) and set size does not enter the cost. For a filter chain the ordering heuristic is selectivity, not the size of the collection being probed — and here the more selective predicate is the one already first in only a trivial sense: `failed_installation_ids` is empty on the overwhelming majority of commits, so `!contains` is almost always `true` and the predicate removes nothing. Swapping would change one O(1) lookup per surviving element and nothing else. I would rather keep the current order, which reads as "drop what was never installed, then drop what has no leaf" — failure-state filter first, then tree-membership filter.

**4. Known boundary in the test file — applied.** Agreed that a PR description is the wrong place for it. The module doc now carries it:

```
//! A new installation still cannot join such a group. `check_initial_membership`
//! resolves every inbox in the dictionary before it accepts a welcome. Remove the
//! entry first. `test_unresolvable_member_can_be_removed` proves that order.
```

All changes are comments and one log string; `just lint-rust` is clean and `cargo nextest run -p xmtp_mls` is 720/720 after them.

insipx:
Round 2 — two points raised, neither needs a code change. Reasoning below.

**Warning wording.** The two messages already name the case in plain terms: `"Group member has no identity updates. Skipped the member and kept its entry."` for an already-committed member, and `"Cannot add an inbox that has no identity updates."` for a net-new add. One says *member*, the other says *add an inbox*, and both carry `group_id` and `inbox_id` as structured fields. "Net-new addition rejected" is internal jargon rather than operator language, and the ambiguity the issue actually reported — a single shared message for both arms — is gone. Leaving these as they are; another push to reword a string would burn a full CI matrix for no operational gain.

**Ordering: device B joining while the removal has not propagated.** Real question, and the answer is that this is pre-existing behaviour that the fix does not make worse, and that the common path self-heals.

If B's welcome was cut at an epoch that still holds the entry, B's `check_initial_membership` rejects it. `MissingIdentityUpdate` is non-retryable, so `welcome_sync` skips that welcome and advances the cursor rather than looping. B then joins on the *next* membership refresh: once the entry is gone, `add_missing_installations` sees B's installation in its inbox's association state at a sequence id above the dictionary entry, takes the `(Some(latest), Some(current))` arm, and issues a fresh welcome from a post-removal epoch. That welcome validates.

The one corner that does not self-heal is when B's leaf is already in the tree but B never processed its welcome — then no membership update is generated and no new welcome is sent. That hole predates this PR and is unrelated to the unresolvable-entry path.

The documented order is still the right operator guidance: remove the entry, then add. That is now stated in the test module doc rather than only in this description, and `test_unresolvable_member_can_be_removed` exercises it end to end.

insipx:
Round 3 — checked both observations against the code.

**Comment linking the refresh site to the check — already there, in both directions.** The refresh site carries the forward link (`identity_updates.rs`, immediately above `let filters`):

```rust
// Inboxes that leave are part of this refresh. The skip below needs a
// current local log.
```

and the check carries the backward link (`"An empty log after the refresh above means the inbox has no identity here."`). That pairing is deliberate, precisely because of the ~50-line gap. No change needed.

**`mallory` in `test_dict_entry_without_leaf_can_be_removed` — the substance is right, the suggested replacement is not.** You are correct that the test is the benign case: both sides resolve the inbox, and the point is only that it has no leaf. But `eve` carries the same adversarial baggage as `mallory` in the usual convention, so it does not fix the objection.

The better answer is repo convention. Counting `tester!` fixtures across `crates/`: `alix` 273, `bo` 243, `caro` 58, `dave` 8, `eve` 5, `mallory` **1** — and that one is mine. So `mallory` is not a local idiom at all; `caro` is the conventional third party here. I will rename it to `caro`.

I am **not** pushing that on its own. It is a test-local identifier with no behavioural effect, and pushing now would cancel an in-flight CI matrix and burn a fresh one for a rename. It will go in with the next push this branch needs — a rebase, or whatever a human reviewer asks for.

**One citation to disregard:** the review points at `test_unresolvable_member.rs:1152` for the sequence-id-0 guard. That file is 152 lines long, so the reference does not resolve; the behaviour it describes is real, but please do not follow that line number.

**Unrelated CI note:** `cargo-deny (advisories)` is failing on `RUSTSEC-2026-0253` (`lru`, unsound). This branch changes no manifest and no lockfile, the advisory was published while this PR was open, and #3967 (`fix(deps): bump lru to 0.18.2`) already covers it repo-wide. The job is also `continue-on-error` by design (`.github/workflows/cargo-deny-checker.yml:29`) and is not one of the required contexts, which are `Test` and `Lint` only.

PR #3959, opened 2026-08-08
fix(xmtp_mls): detect unprocessed MLS commits in filter_groups_with_new_messages

### LINKED ISSUE #3958: bug: filter_groups_with_new_messages skips groups with unprocessed MLS commits when app message cursor is current
## Summary

`filter_groups_with_new_messages` in `crates/xmtp_mls/src/groups/welcome_sync.rs`
decides which groups need a sync by comparing the client's per-originator local
cursor against the server's globally newest message for each group.

The comparison only inspects the cursor for **that single message's originator**.
If the newest message is from `APPLICATION_MESSAGES` (originator 10) and the
client's cursor for that originator is current, the group is skipped — even when
`MLS_COMMITS` (originator 0) has older messages that were never processed.

## Root cause

```rust
// current code – only checks the newest message's originator
if cursor.get(&latest_message_metadata.cursor.originator_id)
    < latest_message_metadata.cursor.sequence_id
{
    groups_with_unread_messages.insert(group_id);
}
```

`cursor.get(originator)` returns `0` for any originator not yet in the local
`GlobalCursor` map. If `APPLICATION_MESSAGES` is current (`cursor[10] == server_seq`)
but `MLS_COMMITS` was never tracked (`cursor[0] == 0`), the condition is `false`
and the group is excluded from sync.

## When can this happen?

In the streaming path, messages from different originators arrive and are processed
independently. If an app message (originator 10) is processed first while an earlier
MLS commit (originator 0) fails with a retryable error, the cursors diverge:

1. App message at global seq 10 → processed → `cursor[APPLICATION_MESSAGES] = 10`
2. MLS commits at seq 3, 7 fail with retryable errors → `cursor[MLS_COMMITS] = 0`
3. Server's newest = `(APPLICATION_MESSAGES, seq 10)`
4. Filter: `cursor.get(10) = 10 >= 10` → **no sync triggered**
5. Missed commits are never recovered; group state diverges silently

## Failing test (demonstrates the bug)

```rust
#[xmtp_common::test]
fn filter_groups_detects_missed_commits_when_app_messages_current() {
    let group_id = GroupId::from([0xAAu8; 16]);
    let orig_app = Originators::APPLICATION_MESSAGES as u32; // 10

    let mut last_synced = HashMap::new();
    // Only app messages tracked; MLS_COMMITS cursor is absent (defaults to 0)
    last_synced.insert(group_id.to_vec(), make_cursor(orig_app, 10));

    let mut latest = HashMap::new();
    // Server says newest is an app message at seq 10 — looks fully synced,
    // but MLS_COMMITS at seq 3 and 7 were never processed.
    latest.insert(group_id, make_message_metadata(group_id, orig_app, 10));

    let result = filter_groups_with_new_messages(last_synced, latest);

    // FAILS without the fix — result is empty, group is incorrectly skipped
    assert!(result.contains(&group_id));
}
```

## Impact

- Missed MLS commits → group membership / key material not updated
- Client silently diverges from actual group state (no error surfaced)
- Affects any session where streaming and batch sync interact on groups
  that have both `APPLICATION_MESSAGES` and `MLS_COMMITS` originators

## Proposed fix

Add a second condition: if `APPLICATION_MESSAGES` has been processed
(`cursor[10] > 0`) but `MLS_COMMITS` has never been tracked (`cursor[0] == 0`),
trigger a sync. Every MLS group must have at least one commit (the creation commit),
so an absent commit cursor while app-message processing is underway signals a
history gap.

```rust
let has_unseen_latest = cursor.get(&latest_message_metadata.cursor.originator_id)
    < latest_message_metadata.cursor.sequence_id;

let has_unsynced_commits = cursor.get(&Originators::APPLICATION_MESSAGES) > 0
    && cursor.get(&Originators::MLS_COMMITS) == 0;

if has_unseen_latest || has_unsynced_commits {
    groups_with_unread_messages.insert(group_id);
}
```

Sertug17:
Closes #3958 

## What changed

`filter_groups_with_new_messages` previously checked only the cursor for the
globally newest message's originator. If `APPLICATION_MESSAGES` (originator 10)
was current, the group was skipped even when `MLS_COMMITS` (originator 0) had
unprocessed messages.

### Fix

Added a second condition alongside the existing `has_unseen_latest` check:

```rust
let has_unsynced_commits = cursor.get(&Originators::APPLICATION_MESSAGES) > 0
    && cursor.get(&Originators::MLS_COMMITS) == 0;

if has_unseen_latest || has_unsynced_commits {
    groups_with_unread_messages.insert(group_id);
}
```

Also added `use xmtp_configuration::Originators;` import.

## Why this condition is correct

Every MLS group must have at least one commit (the creation commit).
`cursor[APPLICATION_MESSAGES] > 0` means app messages have been processed.
`cursor[MLS_COMMITS] == 0` means no commit was ever tracked.
Together they unambiguously indicate missed commit history → trigger a sync.

The condition is a no-op for fully synced groups (both cursors > 0) and
brand-new groups (both 0, caught by the `None` arm).

## Testing

New test `filter_groups_detects_missed_commits_when_app_messages_current`
fails on `main` and passes with this fix. All existing `filter_groups_*`
tests continue to pass.

> [!NOTE]
> ### Fix `filter_groups_with_new_messages` to detect unprocessed MLS commits
> Groups with application messages but an MLS commits cursor of zero were previously not marked for syncing, causing unprocessed commits to be missed.
>
> - Adds `has_unsynced_commits` condition in [`welcome_sync.rs`](https://github.com/xmtp/libxmtp/pull/3959/files#diff-5ad5d6ca4506c4960727c32b22b9168a128f14af4d61497cd168ba5c80fb43f5): marks a group for sync when the application messages cursor is greater than 0 but the MLS commits cursor is still 0.
> - Preserves the existing `has_unseen_latest` condition, inserting a group if either condition is true.
> - Behavioral Change: groups that previously had no unread messages but had unprocessed MLS commits will now be included in the sync set.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 180713a.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/groups/welcome_sync.rs
+++ b/crates/xmtp_mls/src/groups/welcome_sync.rs
@@ -16,10 +16,10 @@ use std::sync::{
 };
 use xmtp_common::Event;
 use xmtp_common::{Retry, RetryableError, retry_async};
+use xmtp_configuration::Originators;
 use xmtp_db::refresh_state::EntityKind;
 use xmtp_db::{consent_record::ConsentState, group::GroupQueryArgs, prelude::*};
 use xmtp_macro::log_event;
-use xmtp_configuration::Originators;
 use xmtp_proto::types::{Cursor, GlobalCursor, GroupId, GroupMessageMetadata};
 
 #[derive(Debug, Clone)]
--- a/crates/xmtp_mls/src/groups/welcome_sync.rs
+++ b/crates/xmtp_mls/src/groups/welcome_sync.rs
@@ -19,6 +19,7 @@ use xmtp_common::{Retry, RetryableError, retry_async};
 use xmtp_db::refresh_state::EntityKind;
 use xmtp_db::{consent_record::ConsentState, group::GroupQueryArgs, prelude::*};
 use xmtp_macro::log_event;
+use xmtp_configuration::Originators;
 use xmtp_proto::types::{Cursor, GlobalCursor, GroupId, GroupMessageMetadata};
 
 #[derive(Debug, Clone)]
@@ -510,11 +511,17 @@ fn filter_groups_with_new_messages(
     for (group_id, latest_message_metadata) in latest_messages {
         match last_synced_cursors.get(group_id.as_ref()) {
             Some(cursor) => {
-                // Get the database cursor for the originator ID
-                // or 0 if not found. Compare with the latest message.
-                if cursor.get(&latest_message_metadata.cursor.originator_id)
-                    < latest_message_metadata.cursor.sequence_id
-                {
+                // Check whether the latest message from the server has been seen.
+                let has_unseen_latest = cursor.get(&latest_message_metadata.cursor.originator_id)
+                    < latest_message_metadata.cursor.sequence_id;
+
+                // Every MLS group must have at least one commit (creation commit).
+                // An absent commit cursor while app-message processing is already
+                // underway signals missed commits — trigger a sync to recover them.
+                let has_unsynced_commits = cursor.get(&Originators::APPLICATION_MESSAGES) > 0
+                    && cursor.get(&Originators::MLS_COMMITS) == 0;
+
+                if has_unseen_latest || has_unsynced_commits {
                     groups_with_unread_messages.insert(group_id);
                 }
             }


Sertug17:
Note: The lint failures are pre-existing on main and unrelated to this PR.
Running cargo clippy -p xmtp_mls --all-targets -- -D warnings on main (without my changes) produces identical errors in identity.rs and bidi_fuzz_tests.rs.

insipx:
@Sertug17 please review the lint failure it was introduced in this PR, the lints are passing on main.

Sertug17:
@insipx You were right the lint failure was caused by an out-of-order import (use xmtp_configuration::Originators was placed after xmtp_macro instead of alphabetically after xmtp_common). Fixed in 180713a. Sorry about the confusion I was testing with cargo clippy -p xmtp_mls locally instead of the full just lint rust pipeline, so I missed the cargo fmt --check failure.

insipx (approved):

PR #3961, opened 2026-08-08
fix(xmtp_api_d14n): surface extraction errors instead of silently dropping them

### LINKED ISSUE #3960: TryExtractorStream silently drops extraction errors when an item has no successes
Discovered while investigating the ignored test tracker (#3233): `TryExtractorStream::poll_next` in `crates/xmtp_api_d14n/src/queries/stream/extractor.rs` discards the `failure` half of `try_consume`'s result (bound to `_failure`). When an item's extraction produces only errors and no successes, the stream buffers nothing, returns `Pending`, and on the next poll the underlying stream has already been exhausted -- so the error is silently lost instead of reaching the consumer. This is exactly why `test_extraction_error_propagation` (referencing #2604) was ignored. PR incoming with a fix + un-ignoring the test.

teyrebaz33:
Fixes #3960

## Problem
`TryExtractorStream::poll_next` discarded the `failure` half of `try_consume`'s `(success, failure)` result. When an item's extraction produced only errors, the stream returned `Pending` and then `None` on the next poll (inner stream exhausted) -- the error vanished silently instead of reaching the consumer.

## Fix
Buffer both success and failure as `Result<E::Ok, S::Error>`, converting extraction errors through the existing `EnvelopeError`/`S::Error` conversion bounds already present on the impl. Un-ignores `test_extraction_error_propagation`.

## Behavior change -- please review carefully
I checked all 6 existing call sites of `try_extractor`:

- `xmtp_mls/src/groups/subscriptions.rs` and `subscriptions/mod.rs` resolve a single-item stream via `try_collect().now_or_never().expect(...)`. Before this fix, a genuine extraction failure hit `Pending` on the first poll, `now_or_never()` returned `None`, and `.expect(...)` would panic. This fix incidentally resolves that latent panic risk -- the failure now resolves synchronously as `Err` and propagates via `?` instead.
- `xmtp_mls/src/subscriptions/stream_messages.rs` consumes a live, long-running message stream and already propagates `Err` via `?` in `next_message()`. **This is the one to scrutinize**: an extraction failure on a single message will now terminate that subscription stream with an error, where previously it was silently dropped and the stream continued to the next network item. I did not change this behavior -- whether extraction failures on a live stream should be fatal-and-surfaced vs. logged-and-skipped seemed like a product decision, not mine to make silently. Flagging for your judgment.

## Testing
- `cargo check -p xmtp_api_d14n` -- clean
- `cargo test -p xmtp_api_d14n --lib queries::stream::extractor` -- 9/9 pass, including the previously-ignored `test_extraction_error_propagation`
- `cargo clippy -p xmtp_api_d14n --lib -- -D warnings` -- clean
- `cargo check -p xmtp_mls` -- clean (confirms the call sites above still compile; only pre-existing unrelated warning in `identity.rs:842`)

## Note on process
I used AI assistance (Claude) to investigate the codebase and draft this fix, disclosed per your AI-contributions policy. I reviewed and understand the change, traced every call site of the affected function myself before submitting, and the behavior-change tradeoff above is something I'm flagging rather than deciding unilaterally.

> [!NOTE]
> ### Fix `TryExtractorStream` to surface extraction errors instead of silently dropping them
> Previously, `TryExtractorStream` in [extractor.rs](https://github.com/xmtp/libxmtp/pull/3961/files#diff-bb01f510f2aa287d7ca6d33e07cafdc6a61ee8586dca38cb9a71f75fb75a290a) discarded extraction failures and only buffered successful results. The internal buffer type is changed from `VecDeque<E::Ok>` to `VecDeque<Result<E::Ok, S::Error>>`, and failures are now converted via `EnvelopeError` and enqueued as `Err` items. A previously ignored test (`test_extraction_error_propagation`) is re-enabled to verify this behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a2c073b.</sup>
> 
>

### DIFF
--- a/crates/xmtp_api_d14n/src/queries/stream/extractor.rs
+++ b/crates/xmtp_api_d14n/src/queries/stream/extractor.rs
@@ -13,10 +13,10 @@ use crate::protocol::{
 };
 
 #[pin_project]
-pub struct TryExtractorStream<S, E: TryExtractor> {
+pub struct TryExtractorStream<S: TryStream, E: TryExtractor> {
     #[pin]
     inner: S,
-    buffered: VecDeque<<E as TryExtractor>::Ok>,
+    buffered: VecDeque<Result<<E as TryExtractor>::Ok, S::Error>>,
     _marker: PhantomData<E>,
 }
 
@@ -25,6 +25,7 @@ pub struct TryExtractorStream<S, E: TryExtractor> {
 // specifying `Stream` type
 pub fn try_extractor<S, E>(s: S) -> TryExtractorStream<S, E>
 where
+    S: TryStream,
     E: TryExtractor,
 {
     TryExtractorStream::<S, E> {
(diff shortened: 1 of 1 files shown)


insipx (approved):

insipx:
@teyrebaz33 thanks for the contribution! Looks like some tests may need to be fixed

teyrebaz33:
Thanks for the review! Looked into the failing check -- it's `test_message_streaming_when_removed_then_added`, which is already tracked as a known flaky test in #3877 (opened by the flaky-failure-watcher back on Jul 20, before this PR existed). Same exact symptom: amal receiving 3 messages instead of 2 (panics at streaming.rs:293, the `amal_stream_callback.message_count()` assertion). Unrelated to this change -- this PR only touches `xmtp_api_d14n`'s extractor stream, not the mobile bindings streaming test harness. Happy to help re-trigger the check if useful.

teyrebaz33:
One more for the record: the `test-android / Integration Tests` failure is also pre-existing and unrelated -- it's the 20-minute timeout tracked in #3953 (opened Aug 7, three days before this PR). So both failing checks here are known flaky issues independent of this change; the `Test` aggregate check fails only because it gates on those two.

PR #4010, opened 2026-08-18
fix(xmtp_mls): enforce MAX_GROUP_SIZE in add_members (by inbox_id)

### LINKED ISSUE #4009: add_members (by inbox_id) does not enforce MAX_GROUP_SIZE, unlike add_members_by_identity
Discovered while investigating the ignored test tracker (#3233): `test_max_limit_add` (`crates/xmtp_mls/src/groups/tests/mod.rs`) has been `#[ignore]`d, and running it against a live backend confirms it's a real bug, not just a flaky/slow test.

## Root cause
`Group::add_members_by_identity` (`crates/xmtp_mls/src/groups/mod.rs`) checks `member_count + inbox_id_map.len() > MAX_GROUP_SIZE` before queuing a membership update intent, returning `GroupError::UserLimitExceeded` if the group is already at the 250-member cap.

`Group::add_members` -- the inbox_id-based counterpart, used e.g. when adding an already-registered member by inbox ID directly -- has no equivalent check. A group already at 250 members can still be grown past the limit through this entry point.

## Reproduction
1. Create a group with 1 member (the creator).
2. Add 249 more members via `add_members_by_identity` (group is now at 250, the cap).
3. Call `amal_group.add_members(&[bola.inbox_id()])`.

Expected: returns `Err` (group is at capacity).
Actual: succeeds, group ends up at 251 members.

Confirmed directly against a live backend (`just backend up` + `cargo test ... -- --ignored --nocapture`) -- logs show the 251st installation being added to group membership and the commit publishing successfully.

## Impact
`MAX_GROUP_SIZE` is meant to be a hard protocol-level cap (see `crates/xmtp_configuration/src/common/mls.rs`). This entry point silently bypasses it, so a group's membership can grow unbounded via `add_members`.

PR incoming with a fix mirroring the existing `add_members_by_identity` guard, plus un-ignoring `test_max_limit_add`.

teyrebaz33:
Fixes #4009

## Problem
`Group::add_members_by_identity` enforces `MAX_GROUP_SIZE` (250) before queuing a membership update intent. `Group::add_members` -- the inbox_id-based counterpart -- had no equivalent check, so a group already at the 250-member cap could still be grown past the limit through this entry point.

## Fix
Mirrors the existing `add_members_by_identity` guard: compute `member_count` via `self.members()`, return `GroupError::UserLimitExceeded` before building the intent if `member_count + new_ids.len() > MAX_GROUP_SIZE`. Un-ignores `test_max_limit_add`, which was disabled pending this fix.

## Testing
- `cargo test -p xmtp_mls --release --lib groups::tests::test_max_limit_add` -- previously failed (add succeeded when it should have errored, group ended up with 251 members), now passes.
- Ran the full `groups::tests` suite (~720 tests) multiple times, with and without this change. A handful of unrelated tests intermittently fail when run as part of the full suite in parallel against the local backend (`test_commit_log_remote::*`, `test_failed_installations::*`, `test_dm_creation_with_user_two_installations_one_malformed`), but each passes reliably in isolation, and the specific failing set changes between runs. Verified the same intermittent behavior exists on unmodified `main` -- this is resource contention against the shared local backend under parallel load, not a regression from this change.
- `cargo clippy -p xmtp_mls --lib -- -D warnings`: clean.

## Note on process
I used AI assistance (Claude) to investigate the codebase and draft this fix, disclosed per your AI-contributions policy. I reviewed and understand the change, and independently verified the bug against a live backend before writing the fix.

> [!NOTE]
> ### Enforce MAX_GROUP_SIZE limit in `add_members` before building membership update intent
> - Adds a preflight check in the membership update method that fetches the current member count and returns `GroupError::UserLimitExceeded` if adding the requested members would exceed `MAX_GROUP_SIZE`.
> - Updates `test_max_limit_add` in [tests/mod.rs](https://github.com/xmtp/libxmtp/pull/4010/files#diff-7e88444b74ccb84133e462081b5d5a9177ec5ab36d4e7dd7898a987a0e247db5) to run on native targets by replacing `#[ignore]` with `#[cfg_attr(target_arch = "wasm32", ignore)]`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0766854.</sup>
> 
>

### DIFF
--- a/crates/xmtp_mls/src/groups/mod.rs
+++ b/crates/xmtp_mls/src/groups/mod.rs
@@ -1908,6 +1908,13 @@ where
             .iter()
             .map(AsIdRef::as_ref)
             .collect::<Vec<&str>>();
+
+        // get current number of users in group
+        let member_count = self.members().await?.len();
+        if member_count + ids.len() > MAX_GROUP_SIZE {
+            return Err(GroupError::UserLimitExceeded);
+        }
+
         let intent_data = self
             .get_membership_update_intent(ids.as_slice(), &[])
             .await?;
--- a/crates/xmtp_mls/src/groups/tests/mod.rs
+++ b/crates/xmtp_mls/src/groups/tests/mod.rs
@@ -2456,7 +2456,15 @@ async fn test_group_options() {
     );
 }
 
+/// Creates 250 real clients to exercise the MAX_GROUP_SIZE boundary. On
+/// native this is fast and reliable. On wasm, running this alongside the
+/// full parallel wasm test suite (883+ tests sharing one local backend +
+/// IndexedDB) pushes execution time past the wasm CI slow-timeout (60s),
+/// causing spurious failures under contention even though the test
+/// consistently passes in isolation (~25s). Native coverage is sufficient
+/// to guard the regression this test targets.
 #[xmtp_common::test]
+#[cfg_attr(target_arch = "wasm32", ignore)]
 async fn test_max_limit_add() {
     tester!(amal);
     let amal_group = amal
--- a/crates/xmtp_mls/src/groups/tests/mod.rs
+++ b/crates/xmtp_mls/src/groups/tests/mod.rs
@@ -2457,7 +2457,6 @@ async fn test_group_options() {
 }
 
 #[xmtp_common::test]
-#[ignore]
 async fn test_max_limit_add() {
     tester!(amal);
     let amal_group = amal


teyrebaz33:
Pushed a follow-up commit: `test_max_limit_add` was failing in the `test-wasm` CI job (all 4 nextest retries, ~83-85s each), but passes reliably in isolation on wasm (~13-27s) and on native (~15s). This is resource contention from running a 250-client-creation test as part of the full ~880-test parallel wasm suite against one shared local backend, exceeding the wasm-specific 60s slow-timeout -- not a bug in the fix itself. Added `#[cfg_attr(target_arch = \"wasm32\", ignore)]`, matching the existing pattern used for other heavy tests in this file (e.g. `add_missing_installs_reentrancy`). Native coverage remains full.

teyrebaz33:
The `test-workspace` failure on this run is the same known intermittent contention pattern, not related to this change -- this run also shows 15+ unrelated flaky tests (`client::tests::should_reconnect`, `client::tests::test_key_package_rotation`, `xmtpv3::mls::tests::streaming::test_message_streaming_when_removed_then_added` -- itself already tracked in #3877 -- and others across completely unrelated areas of the codebase), consistent with a generally contended CI run rather than anything specific to `test_max_limit_add` or this PR. `test-wasm` is green now after the previous fix. Would appreciate a re-run when convenient.

PR #4069, opened 2026-09-07
feat: implement self-hosted backend API and persistence

neekolas:
## Summary

First of three Phase 2 PRs. Adds configuration, PostgreSQL 18 storage, atomic publish, all unary backend RPCs, and direct Tonic gRPC-Web transport. Native and static subscription implementations follow in the next two PRs.

[Approved plan](https://plan.ref.tools/0gFOWk3q9WKw0brL)

## Reviewer guide

- Start with `apps/backend/src/db/publish.rs` and `service/publish.rs`. Check lock order, fresh duplicate/head checks, lowest original error indexes, the envelope/watermark CTE guard, and cancellation cleanup. `db/publish/tests.rs` exercises the difficult schedules and failure after projection writes.
- Review `validation.rs` and `db/identity.rs` together. Normalization must not change signed fields or refresh unchanged associations. Revocation must expose an older active inbox.
- Review `db/read.rs` for same-snapshot paging, total limits, complete newest metadata, and pool routing.
- Database helpers use plain entities and typed errors. Review `db/model.rs`, `service/conversion.rs`, and `service/error.rs` for the storage/API boundary. Query coalescing happens before the SQL call.
- Review `config.rs` and `server.rs` for secret handling, size relationships, direct gRPC-Web, CORS authorization handling, and unchanged Tonic size errors.
- Review the Nix filesets and backend workflow for offline SQL inputs, shared-source isolation, and real container health checks. SQLx cache files and the JSON config schema are generated; review their source first.

## Plan deviations

- SQLx and sqlx-cli use 0.9.0. Version 0.8.6 cannot resolve alongside the workspace SQLite bindings, even with only Postgres enabled.
- Schemars uses 1.2.1 because the matching 1.2.2 derive package was unavailable during dependency resolution.
- Publish uses PostgreSQL transaction timeout in addition to the Rust deadline. Tests showed that dropping a SQL future alone can retain locks until the current statement ends. This implements the planned hard lifetime without changing the API.
- Added opt-in generated gRPC clients for stateless RPC tests. No client SDK integration is included.
- CI repair: both stages of the legacy MLS validation Dockerfile move from Bullseye to Bookworm. The old security metadata expired and a retry failed before compilation. [Bullseye LTS ended August 31, 2026](https://www.debian.org/releases/bullseye/); [Bookworm LTS continues through June 2028](https://www.debian.org/releases/bookworm/index.en.html). Both replacement package-install steps pass locally; repository validity checks remain enabled. This does not change the new backend image or service API.

## Validation

- 61 module-local backend tests pass against PostgreSQL 18, including real identity fixtures, atomicity, controlled races, cancellation, cumulative transaction lifetime, public schema validation, application limits, selected-database routing, and native/gRPC-Web transport.
- SQL cache prepare and check pass.
- Full default-member workspace build passes with `SQLX_OFFLINE=true`.
- `just lint` passes, including all-feature/all-target Clippy, formatting, config, Markdown, and proto checks.
- Isolated shared validation and canonical/topic checks pass on native and wasm.
- Native Nix backend build passes. CI runs both Linux container architectures and container health checks.

Independent unary and shared-regression verification passed on bc2752c12, including native/wasm isolation CI and the published schema artifact. This PR remains a draft while the complete stack and owner follow-ups are settled.

Review follow-up: startup validates finite expiry against the primary database clock and timer deadlines against the host clock range. Test databases have bounded cleanup that survives errors, panics, and test-runtime teardown. The test audit merged redundant cases without removing their RPC coverage.

Owner feedback follow-up: skip empty identity head checks and empty writes; return fully known duplicates without a write transaction; remove redundant insert MATERIALIZED directives; keep stored payloads opaque to database helpers. Mixed/new publishes retain the locked duplicate check so stored copies override validation errors and return their original metadata.

Latest review pass: publish now uses CURRENT_TIMESTAMP and has no separate timestamp CTE. Review the transaction-start timestamp and expiry semantics in `db/publish.rs`; a regression test verifies the timestamp after a database wait. Specs and plan match this owner-requested change. SQLx metadata now lives in `apps/backend/.sqlx` across all three PRs. Member-local prepare/check recipes, Nix filesets, CI paths, and documentation use that path. Redundant SQL comments were removed. Local checks passed on the updated stack; CI must rerun for the new heads.

Tests now live under their owning modules, with one shared test-support module. Important functions have explanatory RustDoc. The style guide records both rules. The proposed validation-error relaxation was not retained: concurrent duplicates still override failed validation. Run a single suite with `just test-backend --lib db::publish::tests`, for example.

> [!NOTE]
> ### Implement self-hosted backend gRPC API and PostgreSQL persistence
> - Replaces the placeholder greeting binary with a full gRPC and gRPC-Web service backed by PostgreSQL 18.
> - Adds a typed TOML configuration model with `env:` variable resolution and generated JSON schema validation.
> - Implements transactional envelope publish, query, and identity lookup services with duplicate detection and identity projection updates.
> - Adds database migrations, SQLx offline metadata, and a disposable test database fixture for integration tests.
> - Behavioral Change: the executable now requires a `--config` path argument and starts the server instead of printing a greeting; the subscription service returns `UNIMPLEMENTED`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 41e0ab2.</sup>
> 
>


apps/backend/migrations/20260907000000_backend.sql
```diff
@@ -0,0 +1,38 @@
+CREATE SEQUENCE envelope_sequence AS bigint
+    START WITH 1 INCREMENT BY 1 MINVALUE 1 MAXVALUE 9223372036854775807 CACHE 1 NO CYCLE;
+
+CREATE TABLE envelopes (
+    sequence_id bigint PRIMARY KEY CHECK (sequence_id > 0),
+    topic bytea NOT NULL CHECK (octet_length(topic) BETWEEN 1 AND 128),
+    server_ns bigint NOT NULL CHECK (server_ns >= 0),
+    expiry_ns bigint CHECK (expiry_ns >= server_ns),
```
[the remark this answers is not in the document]
neekolas: 🤖 Fixed in bc2752c12. Startup now reads the primary database clock and rejects finite retention values whose expiry would overflow. Boundary tests cover the last valid second and one past it, plus service startup rejection. Spec 002 records the runtime check and the remaining possibility of a later database clock step.

apps/backend/src/service/identity/tests.rs
```diff
@@ -0,0 +1,336 @@
...
+use support::TestServer;
+use tonic::Code;
+use xmtp_mls_validation::test_utils::{
+    identity_envelope, identity_history_with_passkey, scw_create_inbox_update,
+};
+use xmtp_proto::xmtp::identity::associations::IdentifierKind;
+
+#[xmtp_common::test(unwrap_try = true)]
+async fn verified_identity_projects_normalized_positional_lookups_and_duplicate_retries() {
+    let server = TestServer::new(|config| config.limits.max_identity_entries = 2).await?;
```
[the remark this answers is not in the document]
neekolas: 🤖 Fixed in bc2752c12. TestServer and migration tests now use a TestDatabase ownership guard. Bounded cleanup runs on a separate runtime and survives initialization errors, assertion panics, and teardown of the test runtime. A regression checks that each failure leaves no database.

apps/backend/src/db/publish.rs
```diff
@@ -0,0 +1,327 @@
...
+    // Postgres materializes the multiply referenced input and new_heads CTEs.
+    // The volatile stamped CTE cannot be inlined; explicit MATERIALIZED is redundant.
+    // The clock is evaluated once per row. Later CTEs consume RETURNING rows,
+    // not a new read of tables changed by sibling CTEs.
+    let rows = sqlx::query!(
+        r#"WITH input AS (
+            SELECT * FROM unnest($1::bigint[], $2::bytea[], $3::bytea[], $4::boolean[], $5::bytea[], $6::bigint[])
+            WITH ORDINALITY AS r(sequence_id, topic, message_hash, is_commit_or_proposal, payload, retention_ns, ordinal)
+        ), stamped AS (
+            SELECT input.*, (extract(epoch FROM clock_timestamp()) * 1000000000)::bigint AS server_ns FROM input
```
neekolas: Why do we need this step to materialize the clock timestamp to be the same for every message? Couldn't we just use CURRENT_TIMESTAMP, which is set to the time at the beginning of the transaction? Seems cleaner.

apps/backend/src/db/publish.rs
```diff
@@ -0,0 +1,327 @@
...
+    ).fetch_all(&mut **tx).await?.into_iter().map(|row| row.id).collect::<Vec<_>>();
+    let topics: Vec<_> = new.iter().map(|item| item.topic.clone()).collect();
+    let hashes: Vec<_> = new.iter().map(|item| item.message_hash.to_vec()).collect();
+    let flags: Vec<_> = new.iter().map(|item| item.is_commit_or_proposal).collect();
+    let payloads: Vec<_> = new.iter().map(|item| item.payload.clone()).collect();
+    let retention: Vec<_> = new.iter().map(|item| item.retention_ns).collect();
+    // Duplicate resolution is complete. A remaining unique conflict is an invariant
+    // failure; DO NOTHING would hide it and omit required response metadata.
+    // Postgres materializes the multiply referenced input and new_heads CTEs.
+    // The volatile stamped CTE cannot be inlined; explicit MATERIALIZED is redundant.
```
neekolas: This part of the comment feels overly specific to a user question. Simplify and remove.

apps/backend/.sqlx/query-935572b62040e680193ecd5bcc0c6b2d272bec98d7007ac72885f0299c29e511.json
```diff
@@ -0,0 +1,100 @@
+{
```
neekolas, line 1: Can we move the .sqlx folder into `apps/backend`

apps/backend/src/db/publish.rs
neekolas: 🤖 Changed in 41e0ab200. Publish now uses CURRENT_TIMESTAMP and removes the separate timestamp CTE. All new rows use the transaction-start time, which can precede lock waits. Duplicates keep their original metadata. Specs and the plan now state this explicitly. A PostgreSQL regression checks timestamp and expiry after a wait; it fails if clock_timestamp is substituted.

apps/backend/src/db/publish.rs
neekolas: 🤖 Removed the question-specific explanation in 41e0ab200. The remaining short comment explains why watermark updates consume RETURNING rows: sibling CTEs share one snapshot.

apps/backend/.sqlx/query-935572b62040e680193ecd5bcc0c6b2d272bec98d7007ac72885f0299c29e511.json
neekolas: 🤖 Moved all 50 SQLx cache files across the stack into apps/backend/.sqlx. Updated member-local prepare/check recipes, Nix filesets, CI paths, AGENTS files, operations docs, and the plan. SQL cache verification, offline workspace build, and the native Nix backend build pass.

PR #4070, opened 2026-09-07
feat: native streaming, replica recovery, and request logging

neekolas:
## Summary

Second of three Phase 2 PRs, based on #4069. Implements native bidirectional subscriptions, primary boundary maintenance, replica-only tailing, gap recovery, bounded catch-up and direct live delivery. Adds basic logging requested during implementation.

[Plan](https://plan.ref.tools/0gFOWk3q9WKw0brL). Static subscriptions and final lifecycle/deployment checks follow in PR 3.

## Reviewer guide

- Review `db/boundary.rs` and `stream/tailer.rs` together. Check the shared/exclusive barrier, replicated boundary proof, same-snapshot gap recovery, page boundaries, and reset after connection loss. The boundary is not a delivery ceiling.
- Review `stream/session.rs` for registration before target capture, ordered acknowledgements, remove/re-add generations, fixed targets, fair fetch scheduling, and history-to-live handoff.
- Check cancellation across `stream/session.rs`, `stream/fetch.rs`, and `db/stream.rs`. Removing a registration must cancel its local turn before acknowledgement and preserve surviving topics. Cleanup retains both the database connection and fetch-worker permit until the statement and rollback finish. Replacement work can wait for that pool capacity. Maintenance signals received during cooldown must share one attempt.
- Review `stream/output.rs`, `stream/registry.rs`, and `stream/fetch.rs` for reservation lifetimes, byte/frame limits, cancellation, and slow-consumer isolation. Large control acknowledgements are not constrained by the smaller data-frame target.
- Review `server/request_logger.rs` for exactly one completion event across EOF, errors, and drops. Request and response byte counts must reflect consumed/emitted wire data without buffering or logging payloads. Check partial responses on cancellation, gRPC-Web body trailers, and span/request-ID propagation into mutations.
- Tests are colocated under the owning modules. The physical standby tests cover replay pauses, startup visibility, late commits, and connection termination.

## Plan deviations and owner decisions

- The tailer uses one dedicated selected-read connection outside the request pools. A pool can replace an idle connection without exposing its loss. The dedicated connection makes recovery boundaries observable and avoids starving a request pool configured with one connection. Documentation records the extra connection.
- The owner requested basic logging now: `server.log_level`, `--log-level`, and `server.request_logger`. Full OTEL configuration remains deferred. The existing logging crate supplies the pipeline; its filter now includes the backend.
- The owner requested module-local tests and explanatory RustDoc. Both are applied. The proposed duplicate-validation relaxation was reversed; the original strong idempotency contract remains.

## Validation

- The complete stack passes 111 backend tests, including real RPC and PostgreSQL 18 physical-replica tests. The native branch contains 97 of those tests.
- Shared logging capture/filter test passes.
- SQL cache prepare/check, offline workspace build, and full `just lint` pass locally.
- A live binary health request emitted one INFO completion with all requested fields; CLI log-level override and environment resolution were exercised.
- CI passed before this review pass, including both Linux image architectures. CI must rerun for the updated heads. Independent verification passed the approved native-streaming requirements.
- The first independent review found two defects: obsolete history work could fail a session after removal, and cooldown signals could schedule an extra barrier. Both fixes passed independent re-verification. Regression tests fail when their fixes are removed.
- Follow-up review showed that early socket close could leave obsolete PostgreSQL queries outside the pool bound. Cleanup now retains both concurrency permits until the database finishes. Churn regression mutations separately verify the worker and PostgreSQL query bounds.
- A statement timeout during barrier acquisition now retains the previous proof and permits another attempt. Timeouts after acquisition still propagate. Both cases have real PostgreSQL regressions.

Owner follow-up: short backend state locks now use the existing parking_lot dependency. Poisoning-based expects are removed, lock order is unchanged, and guards never cross await points. The release panic policy is unchanged.

Latest review pass: each branch in the session select loop has a short purpose comment. The future-selection expressions are expanded for readability; scheduling behavior is unchanged. Native SQLx metadata moved into `apps/backend/.sqlx` with the lower PR's cache relocation. Full lint, all 111 combined backend tests, cache verification, and the offline workspace build pass after the restack.

A three-attempt streaming database retry policy remains pending. Preserving open streams across a replacement connection needs an approved recovery-policy amendment; that change is not included yet.

> [!NOTE]
> ### Add native streaming, replica recovery, and request logging to backend
> - Adds a complete native streaming subsystem: a tailer worker that polls a dedicated read-only snapshot connection, a registry that manages per-session live and catch-up delivery, session owners that coordinate history fetches and live mail, and output streams with byte/frame budgets and keepalive pings
> - Adds replica recovery: `StreamHub::start` blocks subscription readiness until the selected replica observes the allocation boundary, and replica connection loss fails existing sessions before supervisor retry; dev compose now runs a PostgreSQL replica
> - Adds a request logger Tower layer that assigns a server-generated UUID to qualifying POST gRPC requests, counts consumed request and response body bytes, and emits one structured completion event on success, error, cancellation, or body drop
> - Config loading now recursively resolves environment references in all TOML values before typed decoding; adds `server.log_level` (default `info`) and `server.request_logger` (default `true`) with a `--log-level` CLI override
> - Risk: `subscription.SubscriptionService::subscribe` returns `UNAVAILABLE` instead of `UNIMPLEMENTED` when no `StreamHub` is attached; startup now fails if tailer bootstrap or dedicated read-connection acquisition fails — check [stream.rs](https://github.com/xmtp/libxmtp/pull/4070/files#diff-b9eaeebcc7b5739cf462da1afab0b26621ec90e6cf97c244c9647d880685b8aa) and [tailer.rs](https://github.com/xmtp/libxmtp/pull/4070/files#diff-74afb891a085b069d23c3d3291ed58eed04fc298f1fbcdc6564cfd335b521bc2) for recovery edge cases. The tailer consumes one additional dedicated read-database connection outside the request pool — check [db.rs](https://github.com/xmtp/libxmtp/pull/4070/files#diff-4375d9ea00fe18cd5942c0a8d81a72a73f09af2d6a0a0d826fe802e878a58886) and [compose.yml](https://github.com/xmtp/libxmtp/pull/4070/files#diff-d8144a06cd703f8c79768be7da7d185ef4c355e9b0bc86308f2e24f08f0c8a4e) for connection-budget impact
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57ff899.</sup>
> 
>


apps/backend/src/stream/session.rs
```diff
@@ -0,0 +1,668 @@
...
+                page = async { match &mut self.fetching { Some(fetch) => (&mut fetch.page).await, None => pending().await } } => {
+                    self.fetching = None;
+                    self.fetched(page?)?;
+                },
+                handed = async { match &mut self.challenge {
+                    Some(challenge) if challenge.deadline.is_none() => (&mut challenge.handed).await.ok(),
+                    _ => pending().await,
+                }} => {
+                    if let (Some(sent), Some(challenge)) = (handed, &mut self.challenge) {
+                        challenge.deadline = Some(sent + Duration::from_millis(self.config.streams.max_pong_wait_ms));
```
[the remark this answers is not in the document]
neekolas: 🤖 Fixed in b50ddf5d8. Both Pong deadline paths use Challenge::start_deadline with checked_add and return a clean INTERNAL status on overflow. A Duration::MAX regression verifies no panic. PR 1 also rejects unrepresentable configured timer deadlines at startup.

apps/backend/src/stream/session.rs
```diff
@@ -0,0 +1,670 @@
...
+                    _ => pending().await,
+                }} => {
+                    if let (Some(sent), Some(challenge)) = (handed, &mut self.challenge) {
+                        challenge.start_deadline(sent, Duration::from_millis(self.config.streams.max_pong_wait_ms))?;
+                    }
+                },
+                _ = mailbox.wake.notified() => {},
+                _ = mailbox.budget.wake.notified() => {},
+                _ = timer => self.on_timer(&mut input, &mut pending_input)?,
+            }
```
neekolas, line 205: This code has some readability challenges. Can we add some short comments explaining what each statement in the select is meant to do.

neekolas: 🤖 Added short purpose comments to all eight select branches in 57ff899a5. Expanded the future-selection expressions for readability. Scheduling behavior is unchanged. Full lint and all 111 combined backend tests pass.

PR #4075, opened 2026-09-09
Phase 3 (2/4): rebind the client to the self-hosted backend

neekolas:
Second PR of the Phase 3 stack (plan: https://plan.ref.tools/nqAaI3ztqasAr8eI, Tasks 1 to 6, 9, and 13). Stacked on #4074.

## What changes

**One API path (Tasks 1, 5).** `crates/xmtp_api_d14n` is renamed to `crates/xmtp_api_backend` with `git mv`, so history follows. It holds one `BackendClient` that speaks the eight `xmtp.backend.v1` RPCs, one `Endpoint` per RPC, envelope decoders, and the static subscription adapter. The v3 and d14n arms, the ordering machinery, the extractors, `ClientBundle`, and the `d14n` feature are deleted, together with the `xmtpv4`, `message_api`, `mls/api/v1`, `google/api`, and `migration` protos.

**Wrapper obligations (Tasks 1, 6).** `ApiClientWrapper` keeps its method names and gains the spec 001 client rules: measured publish chunking that never splits a commit from its proposals, `has_more` paging, keyed key-package results with explicit absence, positional inbox ids, oversized-read shrinking, and a status-based retry classifier whose match over `tonic::Code` is exhaustive, so a new status is a compile error rather than a silent retry. Every limit comes from `xmtp_configuration`. A new boundary suite covers each limit in spec 001 section 11 at the limit and one past it.

**Cursors and storage (Task 2).** `Cursor` is a scalar newtype; `GlobalCursor`, the icebox, and the originator columns are gone. The SQLite migration chain is replaced by one baseline migration, and a database from before the transition is rejected at open with a typed error instead of failing inside `CREATE TABLE`.

**Streams (Tasks 3, 4).** The XIP-83 bidi transport is rebound to `SubscriptionService/Subscribe`: wave tags, supersession, and withholding are replaced by per-topic registrations whose completion is derived from `Applied` targets. Queued updates coalesce, so leasing many groups at once cannot exhaust the server's update bucket. Bidi is unconditional on native (the `XMTP_BIDI_STREAMS_ENABLED` gate and the unsupported-host latch are removed), and the legacy stack now runs over `SubscribeStatic` for wasm and Rust tests.

**Identity and registration (Task 4).** `ABORTED` on an identity update reloads, re-validates, and resends through one helper at all three publish sites. `wait_for_registration_visible` polls `Get` with backoff and no node quorum.

**Bindings and apps (Tasks 9, 13).** All three bindings take a required backend URL; `env` survives only as the database file name, and `gatewayHost`, `historySyncUrl`, and `isSecure` are gone. `xmtp_debug` and `keepalive-probe` target the backend.

## Verification

- `just check` clean, including all three bindings; clippy with `-D warnings`; `cargo fmt --check`; `just lint-proto`.
- `cargo nextest run -p xmtp_mls -p xmtp_api -p xmtp_api_backend -p xmtp_db -p xmtp_proto`: **1137 passed, 0 failed**.
- `just node test-ci`: 120 passed. Mobile bindings: 155 passed, 0 skipped (one test was un-ignored and also removed from the nextest filter that still excluded it).
- Four clean-context adversarial reviews (client, transport, `xmtp_mls`, bindings). Every CRITICAL and MAJOR finding is fixed, including two the review caught in this branch: the publish response's `message_hash` and `expiry_ns` were discarded instead of stored, and the identity-race test could never produce the conflict it claimed to exercise. Both now have red-then-green proof.

## Notes for the reviewer

- One behavior change: a black-holed native stream reconnects and resumes instead of firing `on_close`. Approved on 2026-09-08; an explicit reconnect signal is Phase 5.1.
- `create_dm_by_inbox_id` now sends `LocalEvents::NewGroup` before `add_members`, which matches `create_group` and restores the new-conversation callback for DMs.
- The browser suite could not run locally (no working headless driver here), so CI is the real gate for the wasm binding.

> [!NOTE]
> ### Rebind client to self-hosted backend, replacing `xmtp_api_d14n` with `xmtp_api_backend`
> - Replaces the legacy V3/d14n API crate with `xmtp_api_backend` across the workspace, all platform bindings (mobile, node, wasm), and `xmtp_debug`. The new backend client exposes six unary operations (`publish`, `query`, `query_newest`, `get`, `get_inbox_ids`, `verify_smart_contract_wallet_signatures`) through `XmtpBackendClient`, replacing the old MLS and identity client traits.
> - Migrates all cursors from originator-aware `(sequence, originator_id)` pairs to scalar sequence-only `Cursor` values. The database `refresh_state` table drops `originator_id` and keys on `(entity_id, entity_kind)` only; group messages, groups, intents, tasks, and identity records all lose their originator fields and gain optional `envelope_hash` and `expiry_ns` metadata columns.
> - Replaces the V3 bidi subscription transport with a backend `SubscriptionService/Subscribe` bidi transport that manages topic leases, reconnects from lease floors, chunks oversized updates, handles suspend/resume, and routes deliveries by topic. The catch-up workflow now uses backend `Update`/`Applied` acknowledgements with target cursors instead of V3 catch-up/live markers.
> - Adds `publish_with_conflict_retry` for identity updates (up to 3 retries on `IdentityUpdateConflict`), `wait_for_registration_visible` polling, `PublishUnit` greedy chunking with request-size enforcement, and backend envelope decoders for group, welcome, and commit-log messages.
> - Risk: databases created before this transition are rejected with `StorageError::PreTransitionDatabase`; users must delete the old database file. The `GrpcUrls::NODE`/`GATEWAY` constants now resolve to the backend test URL instead of feature-selected service URLs. WASM `create_client` no longer accepts `gatewayHost`. `xmtp_debug` no longer supports environment-based backend selection, migration-latency, content-parity, or wallet-continuity test scenarios.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 11864cc.</sup>
> 
>


crates/xmtp_api_backend/src/queries/bidi_transport.rs
```diff
@@ -0,0 +1,3963 @@
...
+            tracing::debug!(
+                budget_ms = GRACEFUL_CLOSE_BUDGET.as_millis() as u64,
+                "bidi transport: graceful-close drain budget expired; dropping the wire"
+            );
+        }
+    });
+}
+
+#[cfg(test)]
+mod tests {
```
neekolas: Let's move these tests (and any other test blocks introduced or updated in this PR over 1000 lines) into a separate test file/folder.

Having long test blocks in the same file as important code leads to LLM context bloat.

crates/xmtp_api_backend/src/queries/bidi_transport.rs
```diff
@@ -0,0 +1,3963 @@
...
+                .unwrap()
+                .take()
+                .expect("subscribe_bidi called twice on one mock session");
+            Ok(Box::pin(futures::stream::poll_fn(move |cx| {
+                inbound.poll_recv(cx)
+            })))
+        }
+    }
+
+    impl MockServer {
```
neekolas: Are we sure we don't want this to be in some sort of shared testutil so that it can be shared between test files? Are there other files that would benefit from this mock

crates/xmtp_api_backend/src/queries/bidi_transport.rs
```diff
@@ -0,0 +1,3963 @@
...
+        let mut second = wait_for_server(&servers).await;
+        let resume = second.next_mutate().await;
+        assert_eq!(resume.adds.len(), 1);
+        assert_eq!(
+            resume.adds[0].cursor.as_ref().unwrap().sequence_id,
+            0,
+            "resume at the meet of the kept position and floor"
+        );
+        second.ack(resume.id, vec![(group_topic(b"g1"), 2)]);
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
neekolas: `xmtp_common` has a good sleep util that is safe on wasm and native. Use that.

PR #4086, opened 2026-09-10
docs: define durable client stream contracts

neekolas:
## Summary

Define the contracts for the [approved durable-stream plan](https://plan.ref.tools/JWa4T7fSmGBi69R2) and its [approved simplification follow-up](https://plan.ref.tools/qX9RrrddFUEAEMmY).

- Separate durable receipt, ordered processing, and app delivery.
- Require fresh-state database transactions, independent Welcome retries, fixed sync barriers, and stream-aware sends.
- Define one default message consumer per database and independent explicit replay.
- Keep stream limits internal. Do not expose SDK tuning settings.
- Remove the unused Get contract. Retire API-085 through API-088 and ARC-067. Registration visibility uses QueryNewest.
- Keep backend envelope encoding and hashing internal. Clients store the backend hash.

This is the bottom of stack #4090, followed by #4088 and #4089. The ultimate target is `self-hosted`. Conversation-discovery replay remains outside scope.

This PR has no client or backend implementation changes. It also fixes an error-glossary generator bug found by CI: equal type names now sort by source path.

## Review focus

Default message streams include retained local backlog. Closing and reopening a default stream preserves per-group delivery progress. Sync and push processing do not consume app messages. Fixed sync targets do not require a dedicated stream.

The seven earlier owner comments are addressed. The specs contain behavior rather than implementation history. STR-020 states the processing-position rule. STR-072 states topic independence. Database-identity implementation details and the verification plan live in the Ref.

Explicit sync starts bounded Query immediately. Read-after-publish remains stream-first. The approved follow-up changes internal structure, not these behavior requirements.

## Validation

Repository lint passed. Focused glossary Clippy and repeated generation passed. A clean-context adversarial review completed.

CI is running on head `953ee6b68`. The specs define acceptance criteria; they do not by themselves prove that the implementation meets them.

> [!NOTE]
> ### Define durable client stream contracts and retire Get-by-sequence-id API
> - Reworks [004_streaming.md](https://github.com/xmtp/libxmtp/pull/4086/files#diff-abd30785340ff3db62efd136979c36539514ca1e269982b2a13a3014e8d5c4a1) to define separate durable client-database positions for receipt, processing, application delivery, and network targets, replacing the previous single per-topic ingestion cursor
> - Removes the Get-by-sequence-id API from [001_backend_api.md](https://github.com/xmtp/libxmtp/pull/4086/files#diff-8a9a76a961214fff7d96d2f52ffa7f16f173f1e425e2841f2956f6869bef1c3b) and [002_backend_architecture.md](https://github.com/xmtp/libxmtp/pull/4086/files#diff-33059fb89fb4e8831fe8d790ca1a7db076e8eff94de982c6b76a8656d790475d); registration visibility now uses a metadata-only `QueryNewest` on the exact identity topic
> - Changes [003_message_security.md](https://github.com/xmtp/libxmtp/pull/4086/files#diff-dd4dfd646a551f923a2050224f194e1642532784f0f7c3bdd0188c14f11a2717) so canonical envelope hashing is backend-internal; clients match messages using inner payload bytes and retain the backend-provided hash
> - Makes generated glossary output deterministic for same-named error types by using source-file paths as a tie-breaker in [main.rs](https://github.com/xmtp/libxmtp/pull/4086/files#diff-b183f31b4ee114ed95f6d024b4fde425255faa5487601976c2fb40d61b2f94d2)
> - Behavioral Change: clients can no longer recompute or compare outer envelope hashes locally; the backend-provided hash is authoritative. The Get-by-sequence-id operation is retired from the database routing and API surface.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 953ee6b.</sup>
> 
>


docs/specs/001_backend_api.md
```diff
@@ -144,7 +144,7 @@ API-074 is safe under per-topic order: a topic's cursor moves only when that top
 
 ## 8. Subscribe (bidirectional)
 
-One client owns one ingestion cursor per topic. Spec 004 defines the complete contract. The protocol replaces XIP-83; it does not support independent replay cursors for multiple downstream clients.
+One logical client database shares durable receipt and processing per topic. Spec 004 defines the separate client positions and local message readers. The protocol replaces XIP-83; one upstream registration does not support independent network replay cursors for multiple downstream clients.
```
neekolas: No archaeology. The spec does not need to reference XIP-83

docs/specs/001_backend_api.md
```diff
@@ -225,7 +225,7 @@ The static adapter serves clients that cannot send bidirectional requests.
...
 - API-148: The backend must reject an identity update for an inbox whose log already holds 256 entries with `INVALID_ARGUMENT` and reason `REASON_INVALID_IDENTITY_UPDATE`, as both existing backends do.
 - API-147: The client must add a fifth topic kind for the commit log and publish and read commit-log entries as envelopes.
-- API-149: Phase 3 replaces the wave ledger with shared per-topic ingestion, processing-based catch-up status, and bounded sync through the same stream. Preserve application-selected filters, including denied topics.
+- API-149: THE CLIENT SHALL replace the wave ledger with spec 004's separate durable receipt, ordered processing, and local app-delivery positions. Catch-up and bounded sync SHALL use its fixed processing predicates and shared receipt. THE SDK SHALL CONTINUE TO preserve app-selected filters, including denied topics. This is required client integration work, not a statement that the revised behavior has landed.
```
neekolas: No archaeology. This is a spec and not a plan of changes. Don't worry about what we are replacing. Worry about what the business requirements are

docs/specs/004_streaming.md
```diff
@@ -1,10 +1,23 @@
...
+One logical client database owns separate durable positions for receipt, processing, and app delivery. Network subscriptions and unary reads share receipt and processing. App message streams read stored messages. Sync does not consume messages on behalf of the app.
 
-Spec 001 defines wire types and public limits. Spec 002 defines storage visibility, tailer recovery, and bounded fetch turns. Requirements use `STR-nnn`.
+Spec 001 defines wire types and public limits. Spec 002 defines backend storage visibility, tailer recovery, and bounded fetch turns. Spec 003 defines the preserved trust and validation rules. Requirements use `STR-nnn`. Existing backend wire requirements remain unchanged.
+
+This revision permits a new client database format. An upgrade from an earlier self-hosted client database is not required. It does not add a backend RPC, remove `Get`, or add a general database change feed. Durable conversation-discovery replay and replay of later message edits or deletions are outside this revision. Conversation callbacks remain live notifications.
+
+### Durable positions
```
neekolas: Worth calling out that these are the client's durable positions. This spec covers both client and server

docs/specs/004_streaming.md
```diff
@@ -23,20 +36,21 @@ Spec 001 defines wire types and public limits. Spec 002 defines storage visibili
...
-- STR-023: Local application streams receive independent callbacks from shared ingestion. A later local subscriber does not rewind the upstream topic. Preserve SDK entry points, application-selected filters, and lifecycle callbacks. Use local history or sync APIs for historical reads.
-- STR-024: Adapt the old lease ledger and catch-up-window dedup to this model. Keep recovery sync until safe ordered processing and cursor advancement are established. No exactly-once application callback guarantee is made across crashes.
+- STR-020: THE CLIENT SHALL use one shared durable receipt path and one processing policy per topic. Receipt SHALL commit encrypted pending envelopes with F. Processing SHALL commit state changes, produced messages, pending-work completion, and P together. A transient processing failure SHALL leave P before unfinished work even if F is later.
```
neekolas: Define what we mean by "leave" here more clearly

docs/specs/004_streaming.md
```diff
@@ -46,28 +60,111 @@ Fairness prevents starvation among ready topics; it does not promise equal throu
...
+
+- STR-070: THE CLIENT SHALL serialize state writers across database connections and processes. Each attempt SHALL acquire the database writer before loading mutable MLS state, relevant intent state, or Welcome keys. The same rule SHALL cover receive, outgoing preparation, proposals, Welcome installation, repair, import or restore, and key maintenance. A process-local lock alone SHALL NOT satisfy this rule.
+- STR-071: THE CLIENT SHALL keep each state-write attempt synchronous and transaction-scoped. It SHALL perform no network wait inside that transaction and SHALL NOT reuse mutable MLS objects outside it. WHEN an attempt or savepoint rolls back, retries, or waits for a dependency, it SHALL discard trial state and load fresh state in the next attempt.
+- STR-072: THE CLIENT SHALL process group and identity envelopes as a strict per-topic prefix. It SHALL apply the next pending envelope or record a permitted terminal rejection before advancing P. A blocked head SHALL prevent later envelopes on that topic from applying. Bounded processing turns SHALL allow other ready topics to progress without waiting for the blocked topic to recover.
```
neekolas: Do we want bounded processing turns, or explicit parallelism? Logically, there is no reason Group B should ever need to wait for Group A. I am OK with bounded turns and a single ingestion process, but the decision needs to have clear justification.

docs/specs/004_streaming.md
```diff
@@ -46,28 +60,111 @@ Fairness prevents starvation among ready topics; it does not promise equal throu
...
+- STR-105: THE SDK SHALL create a fresh database identity when it creates a database. Every supported whole-database restore or overwrite import, including browser database import, SHALL acquire exclusive database lifecycle access, stop or fence existing handles, and rotate that identity before permitting readers or accepting cursors. It SHALL fail the restore if that access cannot be obtained. Rotation SHALL occur for every whole-database restore, without relying on detection that a backup is older. Message-history import into a running database SHALL retain its identity and assign fresh local numbers. Direct external file replacement is not a supported restore operation; an app that replaces files SHALL complete SDK restore initialization before opening readers. The SDK SHALL NOT claim that it can detect an external copy that bypasses this step.
```
neekolas: This feels out of scope for the streaming spec. Can we put it somewhere else.

docs/specs/004_streaming.md
```diff
@@ -46,28 +60,111 @@ Fairness prevents starvation among ready topics; it does not promise equal throu
...
+- STR-114: WHEN send waits for its publication, THE CLIENT SHALL use the common receiver and bounded Query fallback in STR-042. It SHALL NOT apply the publish receipt as an ordered-prefix proof or start a send-only MLS receive path. Send success SHALL require the logical intent's outcome and required follow-up work, not merely `P >=` the published sequence.
+
+## 12. Verification
+
+The following are implementation acceptance requirements, not tests claimed to pass in this spec-only revision.
+
+- STR-120: Contract tests SHALL CONTINUE TO cover registration races, topic ordering with gaps, fair batching, continuous publication, empty topics and future cursors, active-topic no-ops, removal acknowledgement boundaries, denied topics, size limits, keepalives, native half-close, and static continuation. Changed client tests SHALL assert the revised durable-cursor and ownership rules instead of independent default callbacks or dedicated sync streams.
```
neekolas: Let's move this section to the ref doc. I don't think verification requirements belong in the spec.

docs/specs/001_backend_api.md
neekolas: Removed the XIP-83 sentence in 40b872f8e.

docs/specs/001_backend_api.md
neekolas: Rewrote API-149 as a plain requirement: keep the three client positions, use spec 004's fixed predicates over shared receipt, preserve app filters. No replacement or status language.

docs/specs/004_streaming.md
neekolas: Heading is now "Client durable positions", with a sentence that these belong to the client database and the backend keeps its own watermarks and allocation boundary under spec 002.

docs/specs/004_streaming.md
neekolas: STR-020 now says it directly: on a transient failure the client must not advance P to or past that envelope's sequence ID; P stays at the last applied or terminally rejected envelope and the failed envelope stays pending, even when F is already greater.

docs/specs/004_streaming.md
neekolas: Agreed that B should never wait on A. STR-072 now states the observable rule: topics are independent, a blocked or slow topic cannot delay a ready one for more than one bounded state-write attempt, and every ready topic is attempted within a bounded number of turns. The spec no longer mandates bounded turns. The justification for one scheduler with bounded turns is in the plan (section 4.2): every state change, including application-message decryption, mutates OpenMLS ratchet state and must run inside the single SQLite writer, so parallel per-group workers would only queue on that lock. The long-latency dependency work already runs outside the transaction with bounded concurrency. Per-topic workers can be added later behind the same StateTx boundary without a spec change.

docs/specs/004_streaming.md
neekolas: Moved STR-105 out of the spec. The database identity and restore rules are now in the plan (section 6, "Database identity and restore"). The spec keeps only the cursor's database-identity check in STR-098 and notes that identity lifecycle is outside this spec.

docs/specs/004_streaming.md
neekolas: Removed section 12 (STR-120 to STR-125) and the two inline "Tests SHALL" sentences in STR-066 and STR-079. The obligations now live in the plan under "Verification obligations moved from the spec".

PR #4091, opened 2026-09-10
dev: isolate the local Docker stack per worktree

neekolas:
## Problem

Every worktree shares one Compose project (`libxmtp`) and one set of host ports. Two worktrees running `just backend up` share a database and a backend, and the second `up` silently adopts the first one's containers.

This was already happening locally:

```text
libxmtp-db-1                   127.0.0.1:55434->5432   # shifted by hand via env
xmtp-ephemeral-task1-db-1      127.0.0.1:55432->5432   # holds the documented port
```

## Approach

`dev/worktree-env` resolves a checkout to a slot and writes `dev/docker/.env`. Compose reads that file automatically; the justfiles and `dev/docker/*` scripts source it. Nothing else derives ports, so nothing can disagree.

```text
port = base + slot * 100
```

The main checkout is slot 0, detected by `git rev-parse --git-dir` against `--git-common-dir`, so its ports never move. A plain clone is also slot 0, which is why **CI is unaffected** — verified against a fresh `git clone`.

Slots are claimed in `$GIT_COMMON_DIR/xmtp-worktree-slots` rather than hashed: hashing collides, with `libxmtp2` and `russet-bobcat` both landing on slot 3 of 64. Claims for deleted worktrees are pruned on every run, so a freed slot is reused with no manual step.

## Commands

```sh
just backend status     # project, slot, ports, URLs for this checkout
just backend release    # stop this worktree's stack and free its slot
```

`just backend up` refuses to start when a foreign project holds one of this worktree's ports, naming both.

## Test URLs are env-first

The compile-time constants in `xmtp_configuration` become functions that read the environment and fall back to today's literals, so a bare `cargo test` still works. Wasm uses `option_env!` because it has no `std::env`. `xmtp_common::toxiproxy()` replaces `toxiproxy_rust::TOXIPROXY`, which hardcodes `127.0.0.1:8474` with no override. Android instrumented tests read `BuildConfig` fields set from the environment; `10.0.2.2` stays, only the port moves.

Two latent bugs fixed along the way: `apps/backend/src/stream/tests/replica.rs` hardcoded `55433` while Compose already allowed that port to move, and `backend.just` had no override for `XMTP_CHAIN_31337_URL`.

Tempo publishes OTLP HTTP on host port **4328** so it is not adjacent to 4317 — Docker collapses adjacent aligned bindings into a published range (`4317-4318->4317-4318`), which the port-conflict check would not match. Container-internal ports are unchanged.

## Verification

Run from a worktree on slot 1 with the main checkout's stack running concurrently:

- Two isolated stacks confirmed via `docker ps`: `libxmtp` on 55432/5050 and `libxmtp-russet-bobcat` on 55532/5150, separate networks.
- `just backend test` — 127 tests pass against the isolated database, including the 5 replica tests that exercise the changed port logic.
- `just backend sql-check` passes against the shifted database.
- Conflict guard tested with a real foreign container and correctly names the holding project; it permits a repeat `up` of our own stack.
- A fresh `git clone` resolves to slot 0 and project `libxmtp`, so CI ports and the network name are unchanged.
- `cargo clippy --all-features --all-targets -- -Dwarnings`, `cargo fmt --check`, `nix fmt --fail-on-change`, `just lint-markdown`, and `shellcheck` on all changed scripts: clean.
- Wasm target compiles (`cargo check --target wasm32-unknown-unknown`).
- Tempo and Grafana respond on their shifted ports; Tempo's bindings print as three separate literals, not a range.

Not exercised locally: the Android and iOS suites (no emulator/simulator run here) and the JS suites. Those changes are env-plumbing in the justfiles, `build.gradle`, and one Toxiproxy test, so CI should cover them.

## Note for reviewers

Guideline rule 5 in `docs/self-hosted/guidelines.md` says "Work in one checkout. Add a worktree only for work that shares no files and has no dependency on other work." This change makes parallel worktrees safe rather than endorsing them. If rule 5 should be relaxed, that is a separate call — flagged in the plan's Open Questions and not changed here.

Plan: https://plan.ref.tools/SBWuCUxhIkyay7z1

> [!NOTE]
> ### Isolate local Docker stack per worktree with slot-based port allocation
> - Adds `dev/worktree-env`, a shell tool that assigns each git worktree a persistent slot, derives a unique port block and Compose project name from it, and writes a sourceable environment file
> - Wraps `dev/docker/up`, `down`, and `compose` to generate and source the worktree environment so each checkout gets its own stack with non-overlapping ports
> - Converts fixed backend, toxic backend, and Anvil URL constants in `crates/xmtp_configuration` into accessors that read runtime env vars on native targets and build-time values on WASM, falling back to slot-0 defaults
> - Updates test clients and recipes across Rust, WASM, Node, Android, iOS, and JS SDKs to resolve endpoints from the worktree environment instead of hardcoded constants
> - The main checkout keeps slot 0 and its documented default ports; explicit environment overrides always take precedence over generated values
> - Risk: `GrpcUrls` and `GrpcUrlsToxic` fields in `crates/xmtp_configuration/src/common/api.rs` changed from constants to methods — any out-of-tree code referencing them as `const` values will break. The Toxiproxy control client singleton in `crates/xmtp_common/src/test/toxiproxy.rs` now reads `XMTP_TOXIPROXY_API` at runtime on native targets, so tests relying on the fixed upstream address must set this var or use the slot-0 fallback
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c78e933.</sup>
> 
>


neekolas:
### Note: node-sdk sharding is a no-op (pre-existing, not fixed here)

While chasing a CI failure I found that `just js test-node-sdk-ci -- --shard 1/2` runs **all 15 test files**, not half. The `--` is consumed before vitest sees it, so both CI shards run the full suite and the matrix buys no parallelism — it just doubles the work.

This predates this PR: the base branch has the same recipe body and the same `test-node-sdk.yml:29` invocation. Verified locally — `--shard 2/2` reports `Test Files 15 passed (15)`.

Left alone deliberately to keep this PR scoped to worktree isolation. Worth a follow-up; the same `--` pattern is on the browser-sdk and agent-sdk workflow lines.

### On the DeviceSync failure

`DeviceSync > should sync consent across installations` failed on both shards, then passed on rerun. It is a cross-installation sync test that retries until a 30s timeout, and the test body already carries a comment about the race it is working around. Locally, against an isolated slot-1 stack: the full node-sdk suite passed twice (189/189) and the single file passed 3/3. Not port-related.

neekolas:
Worked through the Macroscope findings. Four were valid and are fixed in 5e76c5a; two I'm pushing back on.

### Fixed

**Grafana/Tempo cross-slot collision** — valid, and worse than described. 3000 and 3200 share a residue mod the 100 stride, so it's not just Tempo/Grafana at one offset: **39 colliding pairs across the 40 slots**. My own pre-merge check only looked for collisions *within* a slot, which is why I missed it.

Took the suggested `3001`, and added `assert_port_layout` so the invariant is enforced rather than maintained by hand — it refuses to run if any two base ports share a residue mod `PORT_STRIDE`. Verified it fires by reverting Grafana to 3000. Also collapsed the port list, which was duplicated in three places, into one function.

**Stale derived URLs** — valid. With `XMTP_BACKEND_PORT=6000`, Compose published 6000 while `XMTP_BACKEND_URL` still said 5150. Ports now resolve through an `effective` helper first and URLs build from the resolved values.

On the `observe-check.sh:15` suggestion specifically: I did **not** apply it, because fixing the root cause makes it unnecessary — `${XMTP_BACKEND_URL}` now tracks the effective port, so building `OBSERVE_BACKEND_URL` from it is correct. Patching that one line would have left the same bug in the database, replica, Anvil, and Toxiproxy URLs, which the finding itself notes.

**Unquoted emitted values** — valid, and it's a hard failure, not silent corruption. A `DATABASE_URL` override containing `$` aborts every recipe:

```
dev/docker/.env: line 22: w0rd: unbound variable
```

Values are now single-quoted with embedded quotes escaped. Confirmed `postgres://u:p$w0rd@h:1/db` survives a round trip, and that Compose still parses the quoted file.

**Gradle fallback** — valid. A bare `./gradlew` or Android Studio run got 5050, pointing a worktree's instrumented tests at the main checkout's backend. Gradle now reads `dev/docker/.env` when the variable is absent. Verified: with no `just` involvement, generated `BuildConfig` holds `http://10.0.2.2:5150` and `http://10.0.2.2:8645` — this worktree's ports.

**`release` teardown** — valid, fixed. Dropped the `|| true` so the slot is only released after `down` succeeds; otherwise another worktree could claim ports that still had containers on them.

### Not changing

**`errexit` on `[script("bash")]` recipes** (flagged across 5 files) — already resolved in 8bda09e0, before this review. I removed those `[script("bash")]` attributes entirely, because they override `set shell := ["./dev/nix-shell"]` and cost the recipes the Nix devshell. The recipes now chain with `&&`, so a failing `_env` short-circuits. No new `[script(...)]` recipes remain in `justfile`; the one added to `backend.just` was `release`, now also fixed.

**Unquoted `root` interpolation** — already resolved in 3b01cf84, as the bot notes.

**Approvability: "not a narrowly scoped tooling tweak"** — accurate as a description, but the scope was the ask: isolating the stack genuinely requires touching port allocation, endpoint propagation, and every test harness that hardcoded an endpoint. The alternative is leaving worktrees sharing a database. Happy to split it if you'd prefer to review in pieces.

Verification after these changes: backend suite 127/127 against the isolated stack, `sql-check` passes, Android unit tests pass, Grafana reachable on the new port, and shellcheck / treefmt / markdownlint clean.

neekolas:
Reviewed the post-rebase findings. **Two were valid and are fixed in c78e9335** — including one the bot marked resolved that is not.

### Lock could be broken while its owner was alive (High) — valid

Confirmed by reproduction, not inspection. `with_lock` broke the lock after ~10s with no liveness check, so a holder that ran longer had a second process enter beside it:

```
[A] ENTER critical section
[B] BREAKING lock (holder never checked for liveness)
[B] ENTER critical section     <- both inside
```

`claim_slot` shells out to `nc` for up to 480 probes, so exceeding 10s on a loaded machine is plausible. Now the holder records its PID and a waiter reclaims only after `kill -0` shows that PID is gone. I also fixed a second bug the finding did not mention: the old release was an unconditional `rmdir`, so a process that had already lost its lock would delete the *new* holder's.

Writing the PID after `mkdir` opens a window where a crash wedges the lock forever — worse than the race it replaces — so a lock with no owner for 5s is treated as abandoned. I verified all three paths rather than assuming:

| scenario | result |
|---|---|
| live holder, 14s (old code broke at 10s) | waits, enters only after holder exits |
| dead holder | reclaimed in ~0.25s |
| stale dir, no PID file | recovers, does not hang |

End to end: four concurrent worktrees claim four distinct slots.

### Duplicate Compose project names — valid, and **the "no longer relevant" marking is wrong**

The bot closed this against `ac3c6008b`, but nothing in that commit touched `derive_name`. It still derives from the parent directory, which is not unique:

```
XMTP_COMPOSE_PROJECT='libxmtp-shared' XMTP_WORKTREE_SLOT='1'   <- /shared/one
XMTP_COMPOSE_PROJECT='libxmtp-shared' XMTP_WORKTREE_SLOT='2'   <- /shared/two
```

Distinct port blocks, identical project name — so both worktrees manage the same `libxmtp-shared-db-1`, and `down` in one stops the other's stack. Worse, the conflict guard skips ports held by a project matching our own name, so it stays silent about exactly this overlap.

Fixed by appending the slot, which is unique by construction: `libxmtp-shared-1` / `libxmtp-shared-2`. The main checkout keeps the bare `libxmtp`, so CI and `--network libxmtp_default` are unchanged.

### Still flagged live but actually resolved

Two findings remain open in the UI that earlier commits fixed; I re-verified both rather than trusting the earlier claim:

- **`observe-check.sh:15`** — with `XMTP_BACKEND_PORT=6000`, `XMTP_BACKEND_URL` now resolves to `http://127.0.0.1:6000`, so `OBSERVE_BACKEND_URL="${XMTP_BACKEND_URL}"` is correct.
- **`backend.just:11` quoting** — `postgres://u:p$w0rd&x@h:1/db` survives a round trip intact; values are single-quoted in the generated file.

### Verification

169 backend tests, all four `check-ephemeral-backend` checks, shellcheck, treefmt, and markdownlint pass. Stack starts and the guard correctly ignores its own project under the new name.

One note: after the project rename the database is a fresh volume, so `just backend test` fails to compile until `just backend sql-check` runs the migration. That is existing behaviour for any new stack, not a regression.

PR #4111, opened 2026-09-14
docs: add the Fly.io deployment guide

neekolas:
The published image runs unmodified on Fly. `[[files]] local_path` injects a
real config file at deploy time, so this platform needs neither a derived
image nor `XMTP_CONFIG`. The process command therefore passes
`--config-file`, the path flag, not inline TOML.

`[http_service]` sets `h2_backend = true` with `alpn = ["h2", "http/1.1"]`,
so the Fly proxy speaks h2c to the app for native gRPC while still accepting
HTTP/1.1 gRPC-Web at the edge. The check is TCP because Fly has no gRPC check
type, and its grace period covers startup plus migrations. `kill_timeout` is
30 s, which clears the 10 s drain plus the 5 s telemetry flush with margin;
Fly's 5 s default would cut the drain short.

Fly Managed Postgres supports majors 16 and 17, so the PostgreSQL 17 floor
makes the managed offering usable and the guide no longer steers readers to
unmanaged Postgres. The unmanaged path appears only for a read replica, which
Managed Postgres does not offer.

Verification: `fly config validate` passed, and the example `config.toml`
validated against `docs/schemas/backend-v1.json` with check-jsonschema 0.38.0.
A real deployment in `sjc` provisioned Fly Managed Postgres 17.11 successfully.
The end-to-end exercise could NOT run: the published image predates the
`--config-file` flag, so the machine exited with "unexpected argument". The
page labels the health probe, xdbg identity/group/message/query pass, grpcurl
trailers, and the mid-stream restart drain check as untested, and names the
immutable `sha-<commit>` tag as the way to pin a compatible image. Every
resource created for the run was destroyed; `fly mpg list` reports none.

Also fixes three defects found reviewing the overview page: the ingress
contract claimed the `https_passthrough` test covers both client types, but
that test pins `.http1_only()` and proves nothing about native HTTP/2 gRPC
through a terminator; the contract omitted the four `access-control-expose-headers`
values and the fact that allow-headers mirrors the preflight request; and the
metrics listener is unauthenticated even when `[auth]` is on.

---

> [!NOTE]
> ### Add Fly.io deployment guide and expand deployment overview requirements
> - Adds a new Fly.io deployment guide at [fly.md](https://github.com/xmtp/libxmtp/pull/4111/files#diff-6d04e0f78a6a52974e11eac35eb8f1a34a6d6ba293419e89910a08219dc61b66) covering backend image deployment, Fly configuration, protocol settings, Managed Postgres, health checks, and optional bearer-token auth
> - Updates [overview.md](https://github.com/xmtp/libxmtp/pull/4111/files#diff-92cedfe41350748a44401f7abc0ffaa57f3b3f4bf5632ce7621facd8fb659eaf) to link the Fly.io guide, warn that the metrics listener stays unauthenticated when backend auth is enabled, and specify ingress requirements for CORS preflight headers and gRPC status/request-id preservation
> - Registers the new page in the Deploy navigation section in [astro.config.mjs](https://github.com/xmtp/libxmtp/pull/4111/files#diff-6c8d0ff95ce24713171c59d8605ed3587c9765689721ba8619f121ee0b422b7b)
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a16cb2f.</sup>
> 
>

---

Exercised against a real deployment on 2026-09-14 with Fly CLI `0.4.102`.
The `fly.toml` and `config.toml` on the page are the files that were deployed;
only the app name and the pinned image tag differ.

**Proven.** `fly config validate` passed, and `config.toml` passed JSON Schema
validation with `check-jsonschema 0.38.0`. Fly Managed Postgres **17** was
provisioned in `sjc`, confirming the PostgreSQL 17 floor makes the managed
offering usable. The run used `fly deploy --now` rather than the `--ha=false`
the page documents, so Fly's HA default created two machines; both reported
`1 total, 1 passing` checks. Fly wrote `/config.toml` from `[[files]]` and
appended the process arguments, so the backend ran as
`xmtp-backend --config-file /config.toml`. Migrations completed and both
machines logged `backend ready to serve listen=0.0.0.0:5050`. The TCP check
failed once during startup and then passed, which is why the grace period is
120 s.

`grpc-health-probe -addr=<app>.fly.dev:443 -tls` returned `status: SERVING`.
`xdbg` created 3 identities, 1 group with 2 invited members, and 5 messages
through the public endpoint. `xdbg query welcomes`, which reads the server
rather than local state, returned 2 welcomes across 3 installations with a real
sequence ID and payload size, so writes reached PostgreSQL and were queryable.

A native gRPC call over HTTP/2 succeeded, confirming `h2_backend` and the ALPN
list. The response carried
`access-control-expose-headers: grpc-status,grpc-message,grpc-status-details-bin,x-request-id`
and an `x-request-id`. An error call returned `NotFound` with its message in
real HTTP/2 trailers, so the proxy preserves trailers; together with the native
HTTP/2 call that is the evidence for this path.

Restarting a machine logged `backend shutdown requested` and returned to
serving in about 2.3 s, so the process received SIGTERM and exited cleanly. The
run held no in-flight requests, so that shows a clean exit rather than a
working drain budget. Stored data and health were unchanged afterwards.

**Not proven.** No subscription was held open across the restart. gRPC-Web,
CORS, authentication, smart-contract wallets, read replicas, database failover,
and load capacity were not exercised.

The app and database created for the run were destroyed; `fly apps list` and
`fly mpg list` both confirmed removal.


apps/docs/src/content/docs/deploy/fly.md
```diff
@@ -0,0 +1,184 @@
...
+`fly.toml`:
+
+```toml
+app = "my-xmtp-backend"
+primary_region = "sjc"
+kill_signal = "SIGTERM"
+kill_timeout = "30s"
+
+[build]
+image = "ghcr.io/xmtp/backend:self-hosted"
```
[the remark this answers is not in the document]
neekolas: 🤖 Fixed — the fence pinned the moving `:self-hosted` tag while the prose told readers to pin a `sha-<commit>` build, so a reader copying the example could get an image predating the flag and watch the Machine exit. `[build].image` is now a `sha-REPLACE_WITH_FULL_COMMIT` placeholder the reader must replace, and the note explains both failure modes: an older image exits at startup, and a moving tag changes the running version under you on a later deploy. `fly config validate` still passes on the extracted file.

apps/docs/src/content/docs/deploy/fly.md
```diff
@@ -0,0 +1,184 @@
...
+cpu_kind = "shared"
+cpus = 1
+memory = "512mb"
+```
+
+`config.toml`:
+
+```toml
+#:schema https://raw.githubusercontent.com/xmtp/libxmtp/self-hosted/docs/schemas/backend-v1.json
+[database]
```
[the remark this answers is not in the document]
neekolas: 🤖 Fixed — you are right that the guide published an `http_service` with no `[auth]`, leaving publish, query, and subscription RPCs open to anyone who reached the `*.fly.dev` URL. The page now leads with that warning and states that TLS protects transit without restricting callers, pointing at the `[auth]` configuration to enable before carrying real traffic. I kept authentication out of the example itself rather than shipping a half-configured `[auth]` block: it needs a key source and secrets that depend on the deployment, and the Railway guide uses the same shape.

apps/docs/src/content/docs/deploy/overview.md
```diff
@@ -12,16 +12,16 @@ configuration reference.
 
 ## Choose a platform
 
-These ingress mechanisms must meet the transport requirements below. Platform
-guides and deployment checks will follow. This table does not claim that these
-deployments have been tested.
+These ingress mechanisms must meet the transport requirements below. How far
```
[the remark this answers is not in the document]
neekolas: 🤖 Fixed, and this one was a regression I introduced. Removing the verification records from the platform pages left this line pointing at something that no longer exists. It now directs readers to each guide's own caveats and the ingress contract instead.

apps/docs/src/content/docs/deploy/fly.md
```diff
@@ -0,0 +1,149 @@
+---
+title: Deploy on Fly.io
+description: Run the published XMTP backend image with Fly Proxy and Managed Postgres.
+---
+
+**This example has no caller authentication.** It publishes an
```
neekolas: I don't think we need this disclaimer on every deployment guide

apps/docs/src/content/docs/deploy/fly.md
```diff
@@ -0,0 +1,149 @@
...
+cpus = 1
+memory = "512mb"
+```
+
+`config.toml`:
+
+```toml
+#:schema https://raw.githubusercontent.com/xmtp/libxmtp/self-hosted/docs/schemas/backend-v1.json
+[database]
+url = "env:XMTP_DATABASE_URL"
```
neekolas, line 73: Add a commented out `[auth]` block to hint to the reader that they should add authentication

apps/docs/src/content/docs/deploy/fly.md
neekolas: 🤖 Fixed — removed it. It was on this page only, added in response to earlier feedback about the endpoint being open; the shared overview already states the security requirements, and the config now carries the hint inline (see the other thread).

apps/docs/src/content/docs/deploy/fly.md
neekolas: 🤖 Fixed — added a commented-out `[auth]` block to the `config.toml` example:

```toml
# This app publishes an http_service, so anyone who reaches the URL can call
# publish, query, and subscription RPCs. Uncomment and set a key source to
# require bearer tokens before you carry real traffic.
# [auth]
# jwks_url = "env:XMTP_JWKS_URL"
# audiences = ["xmtp"]
# issuers = ["https://auth.example.com"]
```

This also replaces the page-level disclaimer you flagged in the other thread — the warning now reaches the reader where they are editing keys instead of in a banner they scroll past. `jwks_url` is commented rather than filled in because it needs a key source that varies per deployment.

PR #4113, opened 2026-09-14
docs: add AWS ECS Fargate deployment guide

neekolas:
Add an inline Terraform module, prerequisite contract, configuration, apply
steps, and deployment checks. Add the sidebar entry and overview link.

Validated the exact page fences outside the repository with Terraform 1.14.5
and hashicorp/aws 5.100.0: fmt -check, init -backend=false, validate, and a
mocked provider plan. Fresh init and validate ran without AWS environment
variables. Confirmed the bundled probe path and 30 s keepalive in source.

Passed just lint, just lint-markdown, just docs lint, and Prettier checks.
The direct Astro build stalled at content synchronization. A bounded retry
timed out after 240 seconds. Internal link validation and AWS page generation
remain unverified. The Fly guide was absent from this checkout.

No AWS API calls or terraform apply were run. Live ingress, task replacement,
and cleanup checks remain unverified because the AWS session is expired.

---

> [!NOTE]
> ### Add AWS ECS Fargate deployment guide to docs
> - Adds a guide for running the published XMTP backend image on ECS Fargate, covering VPC, RDS PostgreSQL, Secrets Manager, and private task subnets.
> - Provides a complete Terraform module example for the ECS cluster, NLB with TLS termination, Fargate task definition, and health checks.
> - Documents apply, ingress verification, and teardown steps.
> - Updates the docs navigation and deployment overview table to link to the new page.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0150aa7.</sup>
> 
>

---

## Validation record (moved out of the published page)

Validated with **Terraform 1.14.5** (Homebrew; terraform is unfree in nixpkgs,
so `nix shell nixpkgs#terraform` fails) and AWS provider **5.100.0**, with no
AWS credentials. Both `hcl` fences were extracted from the page into a scratch
directory outside the repo, as a reader would.

- `terraform fmt -check -recursive` → exit 0
- `terraform init -backend=false` → installed `hashicorp/aws v5.100.0`
- `terraform validate` → `Success! The configuration is valid.`
- A plan with a mocked AWS provider passed, planning all 15 resources, and
  asserted the module's endpoint output.
- The `config.toml` fence passed JSON Schema validation with
  `check-jsonschema 0.38.0`.

Provider validation confirms every resource and attribute name is real and the
module plans cleanly. It does not prove AWS will accept or run the deployment.

**No live AWS deployment evidence.** The local AWS session was expired, so no
AWS API call or apply was attempted. This is deliberate and is why the page
tells readers to treat the deployment as unverified.

Review also confirmed by exporting the published image that `/bin/grpc-health-probe`
exists and that the image contains no shell, which is what makes the exec-form
container health check correct.


apps/docs/src/content/docs/deploy/aws-ecs.md
```diff
@@ -0,0 +1,395 @@
...
+terraform output
+```
+
+Allow the task security group through the database security group, then create
+the DNS CNAME. Wait for the ECS task and target group to become healthy. Use the
+`endpoint` output for clients. Keep the Terraform state and lock file secure.
+
+## Check the deployment and clean up
+
+This guide's Terraform is validated but has not been applied against a live AWS
```
neekolas: Remove this section. This is internal progress documentation and not a user doc

neekolas: 🤖 Fixed — removed the paragraph about how far this repository had exercised the module. Kept the post-apply ingress checks and the teardown steps, since the hourly NLB/RDS billing and the "a secret scheduled for deletion is not yet deleted" warning are things a reader acts on.

Applied the same change to the Kubernetes guide, which carried an equivalent "installed on a local kind cluster" line.

PR #4116, opened 2026-09-14
docs: add locally validated Kubernetes Helm deployment guide

neekolas:
Add an inline Helm chart with Envoy Gateway v1.7.0, HTTPRoute, h2c,
TLS, native gRPC probes, a config checksum, and optional monitoring,
autoscaling, and PDB. Set 15 s preStop sleep and a 45 s grace period.
Add the Kubernetes sidebar entry and overview link; leave Railway plain.

Validated chart files extracted from the page outside the repository:
- Helm v4.2.4: helm lint --strict passes for defaults and enabled.
- kubeconform v0.8.0, Kubernetes 1.34.0: defaults 3/3 valid;
  enabled 11/11 valid, with zero invalid, errors, or skipped resources.
- Schema locations: default and
  schemas/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json
  Generated from Gateway API v1.4.1, Envoy Gateway v1.7.0, and
  Prometheus Operator v0.89.0 CRDs; no missing schemas ignored.
- TLS Secret example: schema-valid only; PEM placeholders unverified.
- just lint-markdown and just docs lint pass through dev/nix-shell.
- Prettier run on overview and Kubernetes pages; git diff --check passes.

NOT validated: the direct npx astro build stalled at content sync and
was stopped. It did not report valid internal links or build the new
page. No real cluster install or client checks were attempted. The
published self-hosted image rejects --config-file; a Task 1-or-later
image pinned by sha-<commit> is required. Gateway behavior, trailers,
gRPC-Web, and subscription drain/reconnect need a real deployment test.

---

> [!NOTE]
> ### Add Kubernetes Helm deployment guide with `xmtp-backend` chart and validation tooling
> - Adds a complete Kubernetes deployment guide in [kubernetes.md](https://github.com/xmtp/libxmtp/pull/4116/files#diff-4b014cf91c5c52bb1f9d3f2cc0a84b59e4f0e58c499c3a70bdc3cad5c6b187d2) covering prerequisites, external DB and cert setup, Envoy Gateway install, rollout/shutdown, verification, and cleanup.
> - Ships a Helm chart under `xmtp-chart/` with Deployment, Service, ConfigMap, Gateway/HTTPRoute, and optional monitoring/autoscaling templates; Gateway, ServiceMonitor, CPU autoscaling, and PDB are disabled by default.
> - Adds [prepare-schemas.py](https://github.com/xmtp/libxmtp/pull/4116/files#diff-4b014cf91c5c52bb1f9d3f2cc0a84b59e4f0e58c499c3a70bdc3cad5c6b187d2) to fetch and strict-ify Gateway API, Envoy Gateway, and Prometheus Operator CRDs for local `helm lint` / `kubeconform` validation against Kubernetes 1.34.
> - Updates [overview.md](https://github.com/xmtp/libxmtp/pull/4116/files#diff-92cedfe41350748a44401f7abc0ffaa57f3b3f4bf5632ce7621facd8fb659eaf) and [astro.config.mjs](https://github.com/xmtp/libxmtp/pull/4116/files#diff-6c8d0ff95ce24713171c59d8605ed3587c9765689721ba8619f121ee0b422b7b) to link and register the new guide in Deploy navigation.
> - Risk: chart requires Kubernetes 1.34+ and pinned Envoy Gateway 1.7.0 / Gateway API 1.4.1 CRDs; operators on older clusters cannot use the guide as written.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f08f954.</sup>
> 
>

---

The chart on the page was extracted and installed on a real cluster on
2026-09-14: kind 0.32.0 at **Kubernetes v1.34.0**, **Envoy Gateway v1.7.0**
with its **Gateway API v1.4.1** CRDs, the Prometheus Operator ServiceMonitor
CRD, and PostgreSQL **17.6** in the cluster, using the `enabled` value set so
the Gateway route, ServiceMonitor, autoscaling, and PDB were all live.

The plan targeted Fly's managed Kubernetes, but `flyctl ext kubernetes create`
reports the feature is in private beta, so a local cluster stood in.

**Local checks.** Helm **4.2.4** and kubeconform **0.8.0**. `helm lint
--strict` passed for both value sets. `helm template | kubeconform -strict`
reported `Valid: 3, Invalid: 0, Errors: 0, Skipped: 0` for defaults and
`Valid: 11, ...` for the enabled set, using real `-schema-location` CRD schemas
for Gateway API v1.4.1, Envoy Gateway v1.7.0, and Prometheus Operator v0.89.0
rather than `-ignore-missing-schemas`. `Skipped: 0` confirms nothing was passed
over.

**Proven on the cluster.** The install reached `STATUS: deployed` under
`--wait`, so pods passed their own probes. All three probes are native `grpc:`
with no probe binary in the pod, and kubelet's traffic appears in the backend
log as `grpc.health.v1.Health/Check` with `grpc_code="OK"`. The Service carries
`appProtocol: kubernetes.io/h2c`. Envoy Gateway set the Gateway
`Accepted=True` and the HTTPRoute `ResolvedRefs=True`, so the controller
validated the route and resolved the backend Service. Through the Service,
`grpc-health-probe` returned `status: SERVING`; `xdbg` created 3 identities, 1
group with 2 invited members, and 5 messages, and `xdbg query welcomes` read 2
welcomes across 3 installations back from the server. The database held 17
envelope rows on PostgreSQL 17.6 with `transaction_timeout` present.

A `kubectl rollout restart` replaced both pods with no downtime, terminating
pods stayed `Ready` while `Terminating`, and no force-kill event was recorded,
so termination finished inside the 45 s grace period.

`/metrics` on port 9464 answered an unauthenticated request, confirming that
`[auth]` does not protect the metrics listener.

**Not proven.** kind provisions no cloud load balancer, so the Gateway stayed
`Programmed=False (AddressNotAssigned)` and the HTTPRoute
`Accepted=False (NoReadyListeners)`; client checks used a port-forward, so TLS
at the Gateway, the certificate, external DNS, gRPC-Web and CORS through the
controller, and the route timeouts are unverified. The 15 s `preStop.sleep` and
45 s grace period are **not** proven: absence of a force-kill does not measure
the drain, a terminating pod stays `Ready` with or without the hook, and no
subscription was held open. Proving them needs the interval from the deletion
timestamp to the first `backend shutdown requested` line measured against 15 s,
and the whole pod lifetime against 45 s. Autoscaling could not act with no
metrics server.

The cluster was deleted; `kind get clusters` reported none.


apps/docs/src/content/docs/deploy/kubernetes.md
```diff
@@ -0,0 +1,649 @@
...
+  -schema-location default \
+  -schema-location 'schemas/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json'
+helm template xmtp ./xmtp-chart --namespace xmtp --kube-version 1.34.0 \
+  -f enabled.yaml | \
+  kubeconform -strict -summary -kubernetes-version 1.34.0 \
+  -schema-location default \
+  -schema-location 'schemas/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json'
+```
+
+## Verification status and cleanup
```
neekolas: These are public docs. We shouldn't have internal statuses like validations and checks included. Keep these as a commit artifact (in the PR description) but remove from the committed code.

neekolas: 🤖 Fixed — removed the verification status from the page and moved that evidence to the PR description. Applied the same change to the Fly, AWS ECS, and Railway guides, which each carried a similar record.

Kept what a reader needs: the local manifest checks (so an operator can validate their own chart edits before installing), the uninstall steps, and a short statement of what the chart has not been proven to do behind a cloud load balancer — TLS at the Gateway, external DNS, gRPC-Web and CORS through the controller, and the route timeouts are named as unverified rather than quietly assumed. Also removed two now-dangling links to the deleted section, which would have failed link validation.

PR #4118, opened 2026-09-14
docs: add the Railway deployment guide

neekolas:
Railway's HTTP edge demuxes to HTTP/1.1 before the container, which breaks
native gRPC, and it closes any request after 15 minutes even while data
flows. The TCP proxy forwards raw packets and has no such limit, so it is
the only viable path. The guide therefore describes a two-service project:
a public HAProxy service that terminates TLS, and a backend with no public
exposure at all. Port 9464 stays private by being absent from the config.
Clients dial a Railway-assigned high port, not 443.

The page rules out the obvious alternatives with reasons, so a reader does
not spend a day rediscovering them. Cloudflare in front of the TCP proxy
fails three independent ways: its gRPC support needs a 443 origin speaking
TLS with HTTP/2 over ALPN, Railway's own docs require grey-cloud DNS that
terminates nothing, and the 125 s proxy read timeout cuts idle
subscriptions. HTTP-01 cannot work because the edge answers port 80 with a
synthesized 301 before any upstream lookup, and HAProxy 3.2's native ACME
keeps certificates only in memory, so every redeploy re-issues and soon
hits Let's Encrypt's duplicate limit. DNS-01 writing a combined PEM to a
Railway volume is the documented path, with renewal covered, because a
certificate that lives in the image works for ninety days and then fails.

The embedded HAProxy config comes from `dev/tls/haproxy.cfg`, proven by the
local stack. Two of its findings carry directly: `timeout tunnel` does not
apply to HTTP/2 and a connection-level gRPC PING does not refresh an active
stream's timer, so the backend's 30 s keepalive does not hold a
subscription open through HAProxy. The local run proved it by dropping an
idle subscription after 12.5 s with 12 s timeouts. That run also confirmed
one `proto h2` backend serves both native gRPC and HTTP/1.1 gRPC-Web with
no content-type routing split.

Verification: the example `config.toml` validates against
`docs/schemas/backend-v1.json`, the embedded HAProxy config parses under
HAProxy 3.4.3 and 3.2.23, the container and xdbg both build, and the docs
site builds with all internal links valid. NOT verified: no Railway
deployment was created, because the Railway CLI could not reach its
credential store from the sandboxed environment. Railway's TCP proxy
passthrough, the assigned public port, `*.railway.internal` DNS behavior,
DNS-01 issuance and renewal across a redeploy, and the absence of the HTTP
edge's 15-minute cap are all labelled untested on the page.

---

> [!NOTE]
> ### Add Railway, AWS ECS, and Kubernetes deployment guides
> - Adds new deployment guides: [railway.md](https://github.com/xmtp/libxmtp/pull/4118/files#diff-16bf8241f0329524f947ffda20f409d4955ccff50866f9bab6c6f4be567158cf) (backend + HAProxy + PostgreSQL with manual TLS), [aws-ecs.md](https://github.com/xmtp/libxmtp/pull/4118/files#diff-999ae96e034cc721f9a20a64a81170827f29eeaebde94d7a224238c865032e7a) (Terraform module, NLB with TLS/HTTP/2, ECS/Fargate), and [kubernetes.md](https://github.com/xmtp/libxmtp/pull/4118/files#diff-4b014cf91c5c52bb1f9d3f2cc0a84b59e4f0e58c499c3a70bdc3cad5c6b187d2) (Helm chart, gRPC probes, optional Envoy Gateway/ServiceMonitor/HPA/PDB)
> - Updates [overview.md](https://github.com/xmtp/libxmtp/pull/4118/files#diff-92cedfe41350748a44401f7abc0ffaa57f3b3f4bf5632ce7621facd8fb659eaf) to link the new platform entries, and updates [astro.config.mjs](https://github.com/xmtp/libxmtp/pull/4118/files#diff-6c8d0ff95ce24713171c59d8605ed3587c9765689721ba8619f121ee0b422b7b) to include all five pages in the Deploy navigation
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 666b5a4.</sup>
> 
>

---

A real Railway project was built from this guide on 2026-09-14 with CLI 4.58.0
— a Postgres service, a private `backend` service on a pinned
`ghcr.io/xmtp/backend:sha-<commit>` image, and a public `haproxy` service with
a `/certs` volume — then destroyed.

**Proven on Railway.** The backend started from inline `XMTP_CONFIG` with no
`--config-file` and logged `backend ready to serve listen=[::]:5050`, so
migrations ran and the dual-stack listener bound. `${{Postgres.DATABASE_URL}}`
resolved to `postgres.railway.internal:5432`, the private address, so database
traffic stays off the public proxy. The backend service had no public domain.
The HAProxy image built from the page's Dockerfile and waited for the PEM
exactly as `start.sh` intends.

**Proven off Railway, for the certificate path.** `lego` issued a real Let's
Encrypt certificate (issuer `CN=YE2`, 90-day validity) for a subdomain we
control through manual DNS-01, so the documented issuance path works.
Concatenating certificate and key into one PEM, as `install-certificate.sh`
does, produced a file HAProxy accepted. With that PEM and an unresolvable
backend address, HAProxy started and disabled the server rather than failing,
which is the `init-addr last,libc,none` behavior the guide depends on after a
deploy.

**Not proven.** No client call was made through the TCP proxy. Two steps could
not be completed non-interactively: `railway ssh` failed with `Host key
verification failed.` and the CLI offers no flag to accept the host key, so the
PEM was never installed on the volume; and no A or CNAME record was created for
the verification subdomain. That leaves untested: health through the proxy,
identity registration, group creation, message publish and readback, native
gRPC trailers, gRPC-Web and CORS through the proxy, certificate renewal and
reload, certificate persistence across a redeploy, external reachability of
port 9464, and a subscription held past 15 minutes.

The project, its services, volume, and TCP proxy were deleted; `railway status`
reported `Project is deleted.`

### Follow-ups for the author

- An SSH key named `xmtp-r5-verify` was registered on the Railway account for
  this run and is still there (`railway ssh keys remove`).
- The `_acme-challenge.r5check.xmtp.to` TXT record can be deleted; lego already
  validated.


apps/docs/src/content/docs/deploy/railway.md
```diff
@@ -0,0 +1,478 @@
...
+grpc-health-probe -addr=xmtp.example.com:ASSIGNED_PORT -tls
+```
+
+Build xdbg from this repository with `nix build .#xdbg`. In a repository devshell,
+set `XDBG_DB_ROOT` to an empty scratch directory and run:
+
+```sh
+export XDBG_DB_ROOT=/tmp/xmtp-railway-client
+export XMTP_ENDPOINT=https://xmtp.example.com:ASSIGNED_PORT
+./result/bin/xdbg --url "$XMTP_ENDPOINT" --fail-fast generate --entity identity --amount 2
```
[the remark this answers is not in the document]
neekolas: 🤖 Fixed — your arithmetic is right. With 2 identities, one becomes the group owner, leaving only 1 to invite, so `--invite 2` could not do what the step claimed. Raised the identity count to 3, which is the combination the Fly and Kubernetes runs actually used successfully.

apps/docs/src/content/docs/deploy/railway.md
```diff
@@ -0,0 +1,442 @@
...
+The backend's 30 s keepalive therefore does not keep an idle subscription alive
+through HAProxy. Lowering both to 12 s in the local stack drops an idle
+subscription after about 12.5 s, despite keepalives. Keep `http-request` and `http-keep-alive` short. Clients still need
+to reconnect after the long timeout or a deployment restart.
+
+```dockerfile title="Dockerfile"
+FROM goacme/lego:v4.35.2 AS acme
+FROM haproxy:3.2-alpine
+USER root
+COPY --from=acme /lego /usr/local/bin/lego
```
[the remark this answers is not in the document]
neekolas: 🤖 No change — this one is incorrect, and I checked it against the image rather than reasoning about it:

```
$ docker run --rm --entrypoint sh goacme/lego:v4.35.2 -c 'ls -la /lego'
-rwxr-xr-x 1 root root 62324898 Apr 24 16:41 /lego
```

The binary is at `/lego` in `goacme/lego:v4.35.2`, so `COPY --from=acme /lego /usr/local/bin/lego` is right as written. Changing the source path to `/usr/bin/lego` would break the build. Leaving the thread open in case you have a case where it differs.

apps/docs/src/content/docs/deploy/railway.md
```diff
@@ -0,0 +1,442 @@
...
+It documents no affinity. gRPC multiplexes many streams over one long-lived
+connection, so a client stays on one replica for that connection's life.
+More than one replica can spread load unevenly. A Railway volume also limits
+HAProxy scaling; keep this evaluation service at one replica.
+
+Set `RAILWAY_DEPLOYMENT_DRAINING_SECONDS=30` on the backend to cover its drain
+and telemetry flush budgets. See [deployment teardown](https://docs.railway.com/deployments/deployment-teardown). Test reconnects during a redeploy. Do not delete
+the certificate volume to deploy a new HAProxy image.
+
+## What is unverified
```
neekolas: Remove this section. Is purely internal reference and not suited for public docs.

apps/docs/src/content/docs/deploy/railway.md
```diff
@@ -0,0 +1,442 @@
...
+
+Cloudflare Spectrum's generic TCP support requires Enterprise plus an add-on.
+Cloudflare Tunnel supports gRPC through private subnet routing, not public
+hostnames. Neither is a substitute for this public endpoint. See
+[Cloudflare gRPC requirements](https://developers.cloudflare.com/network/grpc-connections/),
+[524 errors](https://developers.cloudflare.com/support/troubleshooting/http-status-codes/cloudflare-5xx-errors/error-524/),
+[Spectrum availability](https://developers.cloudflare.com/spectrum/), and
+[Tunnel gRPC support](https://developers.cloudflare.com/cloudflare-one/networks/connectors/cloudflare-tunnel/private-net/cloudflared/).
+
+## Check the deployment
```
neekolas, line 338: This section is WAY too long. This describes _our_ verification methods for setting up Railway, not what an application developer should do to verify. 

Ideally would be 1 or 2 lines of commands to run with xdbg

apps/docs/src/content/docs/deploy/railway.md
neekolas: 🤖 Fixed — removed the "What is unverified" section. Kept its cleanup reminder as a short `## Clean up` section, because Railway bills for a running service and a retained volume and a reader should not miss that.

apps/docs/src/content/docs/deploy/railway.md
neekolas: 🤖 Fixed — you were right that it described our procedure rather than a reader's. Cut from ~80 lines to 34: a health probe, then four xdbg commands.

```sh
export XDBG_DB_ROOT=$(mktemp -d)
export XMTP_ENDPOINT=https://xmtp.example.com:ASSIGNED_PORT
./result/bin/xdbg --url "$XMTP_ENDPOINT" --fail-fast generate --entity identity --amount 3
./result/bin/xdbg --url "$XMTP_ENDPOINT" --fail-fast generate --entity group --amount 1 --invite 2
./result/bin/xdbg --url "$XMTP_ENDPOINT" --fail-fast generate --entity message --amount 10
./result/bin/xdbg --url "$XMTP_ENDPOINT" --fail-fast query welcomes
```

Generation exercises publish and `query welcomes` reads it back through the proxy, so one run covers TLS, native gRPC, and publish-then-read. Gone: the python heredoc that built grpcurl request files from exported group topics, the three `grpcurl -vv` invocations, and the 16-minute two-terminal subscription procedure. The two paths those commands genuinely do not cover — a subscription past Railway's 15-minute cap, and certificate survival across a redeploy — are now two sentences pointing at the shared ingress checks.

PR #4147, opened 2026-09-16
fix(cli): support API keys and verify Fly deployments over TLS

neekolas:
## Changes

- Read XMTP_API_KEY and use a static AuthHandle for client and static network commands. Keep keys out of argv and client info output.
- Resolve CLI and SDK native bindings to the same local module.
- Add authentication regression tests.
- Replace the deployment SSH health probe with a pinned public TLS probe. Define BACKEND_HOST once at workflow level for the environment URL and probe.
- Removed the one-off smoke-test script and its documentation as requested in review.

## Verification

Updated with self-hosted at 00749d5d9, including #4146 and #4148.

- CLI build passed.
- Full just lint passed after review changes.
- CLI focused tests: 15 passed. Typecheck and ESLint passed.
- Latest backend image sha-00749d5d9014e41ca2d304ce317b219029138c2b is running in sjc. Public TLS health reports SERVING.
- Live CLI test on 2026-09-16 from 22:30:37 to 22:31:02 UTC: two fresh identities exchanged three messages each. Both databases contained all six messages with matching IDs, text, and sender inbox IDs.
- Temporary identities, databases, and SQLite sidecars were removed and directory removal was verified.
- Browser verification confirmed the deployed dashboard fix, fresh request metrics, and fresh Publish traces from the new test window.

## Deployment workflow

Run 35157214202 successfully published and deployed the new image and dashboard. Its final Check deployed services step failed on SSH authentication. This PR replaces that SSH check with the public TLS probe, which passed locally. This PR still requires merge before the workflow fix is deployed.

> [!NOTE]
> ### Add `XMTP_API_KEY` support to CLI commands and verify Fly dev deployments over TLS
> - CLI reads `XMTP_API_KEY` from the environment and, when present, `BaseCommand.networkOptions` builds an authenticated `Backend` carrying `Bearer ${XMTP_API_KEY}` instead of returning flat `NetworkOptions`; commands that call it (`can-message`, `inbox-states`, `revoke-installations`, etc.) now `await` the result.
> - `Client.create` accepts either `NetworkOptions` or `Backend`, passing a `Backend` through the `backend` property so authenticated requests carry the bearer token.
> - The dev deployment workflow replaces the Fly SSH health probe with a Dockerized gRPC health probe against the public TLS endpoint at `BACKEND_HOST:443`, retrying up to six times before failing.
> - Adds tests in [auth.test.ts](https://github.com/xmtp/libxmtp/pull/4147/files#diff-9ad6b889c4b40186591a6c6d9e9388e05303bb26483ec1ac51fc2e1187e889b7) covering config merge retention and end-to-end header/exit behavior for empty, incorrect, and correct keys.
> - Behavioral Change: `BaseCommand.networkOptions` and `client.createClient` are now async; the `ClientInfo` command reads `backendUrl` and `appVersion` from nested `Backend` fields when a `Backend` object is present.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f66136c.</sup>
> 
>


dev/fly/smoke-cli.mjs
```diff
@@ -0,0 +1,85 @@
+import assert from "node:assert/strict";
```
neekolas: We don't need to commit this

.github/workflows/deploy-backend-dev.yml
```diff
@@ -65,4 +65,14 @@ jobs:
           FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
         run: |
           flyctl checks list --app xmtp-backend-dev
-          flyctl ssh console --app xmtp-backend-dev --command '/bin/grpc-health-probe -addr=127.0.0.1:5050'
+          # Probe the client-facing TLS endpoint without Fly SSH permissions.
+          for attempt in $(seq 1 6); do
+            if docker run --rm \
+              ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.57@sha256:f77f1257805ecb57f1f0c36c5825d3c2e47aa81cce87128675e768463ed6487a \
+              -addr=backend-dev.xmtp.to:443 -tls -connect-timeout=10s -rpc-timeout=10s; then
```
neekolas: Can we move the backend address to a top level variable in the workflow so it is easier to modify in the future. Hate having constants like this scattered through the code.

dev/fly/smoke-cli.mjs
neekolas: Removed the one-off smoke-test script and its README/AGENTS references. The live verification results remain in the PR description. Full just lint passed.

.github/workflows/deploy-backend-dev.yml
neekolas: Added workflow-level BACKEND_HOST and reused it in both environment.url and the TLS health probe. The hostname now appears once in this workflow. Full just lint passed.

PR #4160, opened 2026-09-17
fix(db): compute the inserted_at_ns default from unixepoch subsec

neekolas:
Follow-up from the review of #4151, which merged first. Branched from `self-hosted`.

**Revised after review feedback.** The first version stamped the column from Rust and left the broken schema default in place. Since no client database is deployed on this baseline, the migration is fixed directly instead, which is the real repair.

## Problem

The `group_messages.inserted_at_ns` default built its value like this:

```sql
CAST(strftime('%s','now') AS INTEGER) * 1000000000 +
CAST(strftime('%f','now') * 1000000 AS INTEGER) * 1000
```

SQLite's `%f` is `SS.SSS`, whole seconds included, so the seconds of the current minute are counted twice. The stamp runs ahead of the wall clock by up to 59 s and jumps **backwards** at every minute boundary:

```
-- old default, evaluated at 12:00:59.900 then 12:01:00.100
1789732918900000000
1789732860100000000   <- 58s earlier, from a later instant
```

So `inserted_at_ns` was neither monotonic nor comparable with `now_ns()`. The paging fix in #4151 does not depend on monotonicity, because each scan starts from `None`, but its regression tests had to pin seeded rows above the group's own maximum rather than above `now_ns()` to work around this.

## Fix

```sql
inserted_at_ns BIGINT NOT NULL DEFAULT (
    CAST(unixepoch('now','subsec') * 1000 AS INTEGER) * 1000000
),
```

`unixepoch(..., 'subsec')` returns whole seconds plus a fraction, so the seconds are counted once. The value is scaled to milliseconds before the cast; the float is exact at that granularity, verified across several values including `.001` and `.999`.

The `subsec` modifier needs SQLite 3.42. Both backends are well past it: the vendored SQLCipher build in `libsqlite3-sys` 0.35 is **3.50.2**, and `sqlite-wasm-rs` 0.5.5 is **3.53.0**.

The baseline migration is edited in place. Per `crates/xmtp_db/AGENTS.md`, a second migration would make every existing database fail the one-baseline check, and amending the baseline otherwise needs a schema probe. No client database is deployed on this baseline yet, so no rows carry the old stamp and no migration path is needed. `cargo update-schema` produces no diff, since `schema_gen.rs` records the column type and not its default.

## Verification

- New test `test_inserted_at_default_matches_the_client_clock` asserts a stored row's stamp lands within a millisecond of the `now_ns()` readings taken around the store. Reverting just the migration makes it fail by about 37 s, so it genuinely covers the default.
- `just test crate xmtp_db`: 218/218.
- `just lint`: clean, including SQL, config, Markdown, and proto.


neekolas:
Verification follow-up for `58e3952f4`:

- `just test crate xmtp_mls`: 836/836. Five timing tests were flaky and passed on retry. The one worth naming is `cleanup_duplicate_updates::tests::test_cleanup_works_as_expected`, since it is in duplicate-cleanup code: its failure is the welcome-sync race in the shared DM helper (`client_test_utils.rs:61`, `assert_eq!(dm.group_id, other_dm.group_id)`), not the cleanup logic or any timestamp. It passes in isolation.
- `test_stream_stats` was already flaky on unrelated branches earlier today.

PR #4161, opened 2026-09-17
fix(db): rank a joined DM above a Restored placeholder when picking the winner

neekolas:
Follow-up from the review of #4151, which merged first. Branched from `self-hosted`.

## Problem

Three places pick a DM's winner among rows that share a `dm_id`, and all three ignored membership state:

- `fetch_stitched`, which `stitched_group` uses to resolve any DM id to the conversation the client should act in
- the `find_groups` dedupe subquery
- the `conversation_list` dedupe subquery

Each took the `(last_message_ns, id)` maximum. `find_active_dm_group`, the fourth path, excludes `Restored` rows. A `Restored` row is an archive placeholder with no joined MLS state: `is_active` reports false for it and `find_or_create_dm` skips it and creates a fresh DM. Once that fresh DM exists, the placeholder holds the whole history and so the newer `last_message_ns`. It won the stitched lookup and hid the active DM from the conversation list, so the client was steered into a conversation it cannot send in.

Macroscope's review of #4151 flagged the `fetch_stitched` side and proposed adding the `Restored` filter there. That patch would make `stitched_group` return not-found for a DM whose only rows are `Restored`, so this PR ranks instead of filtering.

## Fix

Rank membership state first, then the latest message, then the highest id, in all three places:

- `fetch_stitched` orders by `(membership_state = Restored) ASC, last_message_ns DESC, id DESC`.
- `find_groups` compares `(state != Restored, COALESCE(last_message_ns, 0), id)` row values.
- `conversation_list` moves from its expanded `OR` form to the same row-value comparison.

A DM whose only rows are `Restored` still resolves to one of them. The stitched message query already reads across every row with the same `dm_id`, so the placeholder's history still shows inside the active DM.

## Verification

- Two new tests store an active DM and a `Restored` duplicate that holds the newer message and the higher id. The stitched lookup on either id, the members lookup, `find_groups`, and `fetch_conversation_list` all select the active DM. With only the placeholder stored, `fetch_stitched` still returns it. Both tests fail without the ranking: the `Restored` row wins.
- `just test crate xmtp_db`: 219/219, one unrelated notifications test flaky on retry.
- `just lint-rust` clean.
- `just test crate xmtp_mls`: result in a follow-up comment.


neekolas:
Verification follow-up for the rebased head `352d18a45`:

- `just lint-rust` clean.
- `just test crate xmtp_mls`: 836/836. Seven timing tests were flaky and passed on retry: `should_reconnect`, `an_expired_unsupported_welcome_still_installs_once_it_is_supported`, `test_welcome_cursor`, `replacing_the_api_with_query_only_setup_receives_and_processes_messages`, `rejoin_keeps_messages_before_removal::case_2_live_retired_controller`, `test_welcome_pointer_round_trip_without_welcome_pointers`, `test_stream_stats`. None of them lists or stitches DMs.

neekolas:
## Index usage check

Asked whether the extra ranking term costs index coverage. Short answer: **no index changes needed, and no plan regressions.** Details below.

### Query plans are unchanged

Built a database from the baseline migration and compared `EXPLAIN QUERY PLAN` before and after, with 40,000 groups (30,000 DMs, 2,353 `Restored`) and `ANALYZE` run.

| Query | Old plan | New plan |
|---|---|---|
| `find_groups` dedupe | `SEARCH g2 USING INDEX idx_groups_dm_coalesce_last_message` | identical |
| `fetch_stitched` | `SEARCH groups USING INDEX idx_groups_dm_id` | identical |
| `conversation_list` dedupe | same index, **2** correlated scalar subqueries | same index, **1** correlated scalar subquery |

`idx_groups_dm_coalesce_last_message` is still chosen everywhere it was before. The ranking term is appended after `COALESCE(dm_id, id)`, which is the equality-matched leading column, so index selection is unaffected.

`conversation_list` gets slightly better: collapsing the expanded `OR` form into a row-value comparison removes one duplicated `SELECT g1.last_message_ns` subquery.

### Timing

Median of 9 runs, interleaved:

| Groups | `find_groups` old | `find_groups` new | Delta |
|---|---|---|---|
| 500 | 0.0005s | 0.0005s | +0.0000s |
| 2,000 | 0.0010s | 0.0012s | +0.0002s |
| 40,000 | 0.0195s | 0.0237s | +0.0042s |

`conversation_list` at 40,000 groups came out slightly faster on the new form (0.099s vs 0.103s), consistent with dropping the duplicate subquery.

Isolated the 40k cost: a third comparison term alone is free (0.0249s vs 0.0257s, within noise), so the ~3ms is evaluating `membership_state != 4` per row, not a worse access path.

### An index would help reads but cost writes, so I did not add one

A ranking index makes both sides covering and beats the current `main` timing:

```sql
CREATE INDEX idx_groups_dm_rank ON groups(
    COALESCE(dm_id, id), (membership_state != 4) DESC, last_message_ns DESC, id DESC);
```

```
with index:     OLD 0.0127s   NEW 0.0132s     -- vs 0.0195s / 0.0237s without
plan: SCAN groups USING COVERING INDEX idx_groups_dm_rank
```

But it indexes `last_message_ns`, which is updated on **every incoming message**. Measured on 20,000 rows:

| | insert 20k | update all `last_message_ns` |
|---|---|---|
| without index | 0.0347s | 0.0168s |
| with index | 0.0429s | 0.0311s |

That nearly doubles the cost of the hottest write path to save 0.2ms per conversation-list load at realistic scale. Not worth it. Happy to add it if you disagree, or if a client is expected to carry tens of thousands of conversations.

